PR #2, opened 2016-05-26
Update README

ghodss:
Clarify the difference between Heapster and kube-state-metrics.


ghodss:
@bgrant0607 @vishh Let me know if you think this rationale makes sense. I'm also open to reconsidering the rationale and considering the possibility of merging the two projects.

README.md
```diff
@@ -26,11 +26,31 @@ additional metrics!
...
-| nodes | ready=&lt;true|false&gt; |
+| nodes | ready=&lt;true\|false&gt; |
 | deployment_replicas | name=&lt;deployment-name&gt; namespace=&lt;deployment-namespace&gt; |
 | deployment_replicas_available | name=&lt;deployment-name&gt; namespace=&lt;deployment-namespace&gt; |
 | container_restarts | name=&lt;container-name&gt; namespace=&lt;pod-namespace&gt; pod_name=&lt;pod-name&gt; |
 
+## kube-state-metrics vs. Heapster
+
+[Heapster](https://github.com/kubernetes/heapster) is a project which fetches
+metrics (such as CPU and memory utilization) from a Kubernetes API server and
```
vishh: nit: It collects from nodes and not the API server.

README.md
```diff
@@ -26,11 +26,31 @@ additional metrics!
...
+Kubernetes, kube-state-metrics is focused on generating completely new metrics
+from Kubernetes' object state (e.g. metrics based on deployments, replica sets,
+etc.). The reason not to extend Heapster with kube-state-metrics' abilities is
+because the concerns are fundamentally different - while Heapster only needs to
+fetch, format and forward metrics that already exist, kube-state-metrics holds
+an entire snapshot of Kubernetes state in memory and continuously generates new
+metrics based off of it but has no responsibility for exporting its metrics
+anywhere.
+
+In other words, kube-state-metrics itself is designed to be another source for
```
vishh: Another use case is that of being compatible with monitoring systems that do not depend on heapster. For example, prometheus does not collect metrics via heapster today.

vishh:
Added a couple of comments. LGTM overall!

PR #8, opened 2016-09-02
Node metrics

fabxc:
@ghodss @grobie 

This reworks node metrics in the same manner as deployment metrics.
The node states are broken out into one time series for each possible
conditional state. One constant metric holds textual meta information
about cluster nodes.

The 3 metrics/node state are a bit unusual. Discussed this with @brian-brazil. There are different solutions with their pros and cons. The one chosen here appears to be the most robust against future changes to the possible conditions and confusing results when trying to aggregate.

There are actually more fields in the API's node information block. But I think obscure "system", "boot", and "machine" IDs should be only added after a valid use case has been discovered.


fabxc:
We might have to discuss whether "node" is too generic as a label too and "node_name" may be more explicit. Then we should standardize it to to deployment's etc. as well.

grobie:
The node label is pretty common across Prometheus exporters and similar to
instance. I vote to keep it.

node.go
```diff
@@ -0,0 +1,117 @@
...
+			"container_runtime_version",
+			"kubelet_version",
+			"kubeproxy_version",
+		}, nil,
+	)
+
+	descNodeStatusReady = prometheus.NewDesc(
+		"node_status_ready",
+		"The ready status of a cluster node.",
+		[]string{"node", "condition"}, nil,
```
brian-brazil: Is this per-kubelet information coming from the kubelet, or something else?

node_test.go
```diff
@@ -0,0 +1,130 @@
...
+					},
+					Status: api.NodeStatus{
+						Conditions: []api.NodeCondition{
+							{Type: api.NodeReady, Status: api.ConditionFalse},
+						},
+					},
+				},
+			},
+			want: metadata + `
+				node_status_ready{node="127.0.0.1",condition="true"} 1
```
brian-brazil: `node_` is already used by the node exporter. This should have a k8 related prefix.

brian-brazil:
> The node label is pretty common across Prometheus exporters and similar to
instance. I vote to keep it.

The node label is the instance label, this smells a bit.

node_test.go
fabxc: Was about to file a separate issue to discuss that.
You still haven't made the point how this is not addressed by your constant emphasis on using the `job` selector everywhere.

node.go
fabxc: This is the known cluster state on cluster nodes.

fabxc:
Things in Kubernetes have unique names. Before we were just using a `name` label here, now it's referencing the thing more specifically. The next stage would be `node_name`, `deployment_name`, etc. – but node seems to be as "bad" as it gets. I generally feel like keeping it, too.

This can all be changed in the near future. Let's get some generally sane metrics out and then dig into the nits.

node_test.go
brian-brazil: > You still haven't made the point how this is not addressed by your constant emphasis on using the job selector everywhere.

It isn't addressed (though this is a big reason why you want to specify some ofrm of job selector), so I didn't bring it up. This is a metric namespacing question. If someone e.g. combined k8 and node exporter metrics in once /mcsetri (and k8 has form in this regard) that'd be confusing.

These are not metrics about the machine, they've metrics about the beliefs of the k8 apiserver about the machine (if I'm understanding correctly).

brian-brazil: > You still haven't made the point how this is not addressed by your constant emphasis on using the job selector everywhere.

It isn't fully addressed (though mitigating this is a big reason why you want to specify some form of job selector), so I didn't bring it up. This is a metric namespacing question. If someone e.g. combined k8 and node exporter metrics in once /mcsetri (and k8 has form in this regard) that'd be confusing.

These are not metrics from the machine, they've metrics about the opinions of the k8 apiserver about the machine (if I'm understanding correctly).

PR #9, opened 2016-09-05
Container restart metric

brancz:
This reworks the container restart metric in the same manner as
deployment metrics. The format remains unchanged to the previous format
as in one metric per container via labels.

@fabxc @ghodss


brancz:
I signed it!

pod.go
```diff
@@ -0,0 +1,75 @@
...
+			ch <- m
+		}
+	}
+}
+
+func (pc *podCollector) collectPod(p *api.Pod) (res []prometheus.Metric) {
+	for _, cs := range p.Status.ContainerStatuses {
+		res = append(res, prometheus.MustNewConstMetric(
+			descContainerRestarts,
+			prometheus.GaugeValue,
```
brancz: Took this from the previous implementation, the semantic of this metric behaves like a counter though, wdyt @fabxc

pod.go
```diff
@@ -0,0 +1,75 @@
...
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+var (
+	descContainerRestarts = prometheus.NewDesc(
+		"container_restarts",
+		"The number of container restarts per container.",
+		[]string{"namespace", "pod_name", "container_name"}, nil,
```
fabxc: We ran into the discussion of naming here previously. If we stay consistent with the last two PRs, we strip the "_name" suffixes here and just have it as `pod` and `container`. 

The question is basically whether this will collide with labels added further up in the monitoring stack. My hunch is that with "_name" there's about equal likelihood. So as long as we don't make it "kubernetes_pod_name" etc. it probably doesn't matter.

Maybe opt for the simpler version of no suffix and stay consistent for now?

pod.go
```diff
@@ -0,0 +1,75 @@
...
+import (
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/kubernetes/pkg/api"
+	"k8s.io/kubernetes/pkg/labels"
+)
+
+var (
+	descContainerRestarts = prometheus.NewDesc(
+		"container_restarts",
```
fabxc: This should probably be `pod_container_restarts`, no?

pod.go
brancz: No strong opinion, so no suffix sounds good to me. Fixing..

pod.go
brancz: Yes makes sense.

pod.go
```diff
@@ -0,0 +1,75 @@
...
+func (pc *podCollector) collectPod(p *api.Pod) (res []prometheus.Metric) {
+	for _, cs := range p.Status.ContainerStatuses {
+		res = append(res, prometheus.MustNewConstMetric(
+			descContainerRestarts,
+			prometheus.GaugeValue,
+			float64(cs.RestartCount),
+			p.Namespace,
+			p.Name,
+			cs.Name,
+		))
```
fabxc: In the last PRs I went with having desc, type, and value in the first line together and labels together in the second (broken out in one/line if they became a lot).

This is more of a taste thing, but I expect there to be 10+ metrics and they'll almost all have the same standard labels. Might make for an easier code read if it's more compact and all the same anyway.

pod_test.go
```diff
@@ -0,0 +1,96 @@
...
+							},
+						},
+					},
+				},
+			},
+			want: metadata + `
+				container_restarts{container_name="container1",namespace="ns1",pod_name="pod1"} 0
+				container_restarts{container_name="container2",namespace="ns2",pod_name="pod2"} 0
+				container_restarts{container_name="container3",namespace="ns2",pod_name="pod2"} 1
+            `,
```
fabxc: Whitespaces seem to be spaces rather than tabs.

fabxc:
Cool, with that we'd be done with the base refactor and can add more metrics.

LGTM after comments.

pod.go
fabxc: Yes, that seems fine.

brancz: Alright, changed to counter.

PR #12, opened 2016-09-09
Add more node metrics

fabxc:
@brancz

### DIFF
--- a/node.go
+++ b/node.go
@@ -41,6 +41,48 @@ var (
 		"The ready status of a cluster node.",
 		[]string{"node", "condition"}, nil,
 	)
+	descNodeStatusOutOfDisk = prometheus.NewDesc(
+		"node_status_out_of_disk",
+		"Whether the node is out of disk space",
+		[]string{"node", "condition"}, nil,
+	)
+	descNodeStatusPhase = prometheus.NewDesc(
+		"node_status_phase",
+		"The phase the node is currently in.",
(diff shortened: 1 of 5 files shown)


node.go
```diff
@@ -71,30 +121,55 @@ func (nc *nodeCollector) Collect(ch chan<- prometheus.Metric) {
...
+	addGauge(descNodeInfo, 1,
 		n.Status.NodeInfo.KernelVersion,
 		n.Status.NodeInfo.OSImage,
 		n.Status.NodeInfo.ContainerRuntimeVersion,
 		n.Status.NodeInfo.KubeletVersion,
 		n.Status.NodeInfo.KubeProxyVersion,
 	)
+
+	// Collect node conditions and while default to false.
+	// TODO(fabxc): add remaining conditions: NodeMemoryPressure,  NodeDiskPressure, NodeNetworkUnavailable
```
brancz: What's holding us back from doing these?

About `NodeNetworkUnavailable` ... how are we gonna get this value if the network is unavailable? 😁

brancz:
LGTM 👍

node.go
fabxc: Was wondering the same – that's probably set if heartbeats to the node fail at a connection level.

What's holding us back is that the vendored deps do not yet contain the constants for these conditions and godep refuses to update the vendoring on my machine without any indication of reason :)

PR #13, opened 2016-09-12
add more pod metrics

brancz:
@fabxc

### DIFF
--- a/node.go
+++ b/node.go
@@ -140,9 +140,9 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n api.Node) {
 	for _, c := range n.Status.Conditions {
 		switch c.Type {
 		case api.NodeReady:
-			nodeConditionMetrics(ch, descNodeStatusReady, n.Name, c.Status)
+			addConditionMetrics(ch, descNodeStatusReady, c.Status, n.Name)
 		case api.NodeOutOfDisk:
-			nodeConditionMetrics(ch, descNodeStatusOutOfDisk, n.Name, c.Status)
+			addConditionMetrics(ch, descNodeStatusOutOfDisk, c.Status, n.Name)
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


brancz:
@fabxc in a follow up PR I would add the `kube` prefix to all metrics, as well as incorporate the missing `status` within the deployment metric names

node.go
```diff
@@ -168,19 +168,19 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n api.Node) {
 	addResource(descNodeStatusAllocateablePods, n.Status.Allocatable, api.ResourcePods)
 }
 
-// nodeConditionMetrics generates one metric for each possible node condition status.
-func nodeConditionMetrics(ch chan<- prometheus.Metric, desc *prometheus.Desc, name string, cs api.ConditionStatus) {
+// addConditionMetrics generates one metric for each possible node condition status.
```
fabxc: Doc comment that this requires the condition metric description to have "condition" as the last label.
That kind of collides with the suggestion to have label names alphabetically sorted. But that's fine for now.

pod.go
```diff
@@ -24,8 +24,48 @@ import (
...
+		[]string{"namespace", "pod", "phase"}, nil,
+	)
+	descPodStatusReady = prometheus.NewDesc(
+		"pod_status_ready",
+		"Describes whether the pod is ready to serve requests.",
+		[]string{"namespace", "pod", "condition"}, nil,
+	)
+	descPodContainerInfo = prometheus.NewDesc(
+		"pod_container_info",
+		"Information about container.",
```
fabxc: .. about a container in a pod... maybe?

pod.go
```diff
@@ -53,19 +101,44 @@ func (pc *podCollector) Collect(ch chan<- prometheus.Metric) {
...
+		addConstMetric(desc, prometheus.GaugeValue, v, lv...)
+	}
+	addCounter := func(desc *prometheus.Desc, v float64, lv ...string) {
+		addConstMetric(desc, prometheus.CounterValue, v, lv...)
+	}
+
+	addGauge(descPodInfo, 1,
+		p.Status.HostIP,
+		p.Status.PodIP,
+	)
```
fabxc: Short enough to keep in a single line I think.

pod_test.go
```diff
@@ -63,23 +75,89 @@ func TestPodCollector(t *testing.T) {
...
+				pod_status_phase{namespace="ns1",phase="Running",pod="pod1"} 1
+				pod_status_phase{namespace="ns2",phase="Pending",pod="pod2"} 1
+				# HELP pod_status_ready Describes whether the pod is ready to serve requests.
+				# TYPE pod_status_ready gauge
+				pod_status_ready{condition="false",namespace="ns1",pod="pod1"} 0
+				pod_status_ready{condition="false",namespace="ns2",pod="pod2"} 1
+				pod_status_ready{condition="true",namespace="ns1",pod="pod1"} 1
+				pod_status_ready{condition="true",namespace="ns2",pod="pod2"} 0
+				pod_status_ready{condition="unknown",namespace="ns1",pod="pod1"} 0
+				pod_status_ready{condition="unknown",namespace="ns2",pod="pod2"} 0
```
fabxc: Did you see my addition in the node metrics that allows you to only test a specific set of metrics per test case? Might be advisable here as well so you can catch all cases around condition and phase without blowing up the test case.

brancz: aha yes I was thinking of something similar, but didn't remember that this was already implemented

brancz:
fixed @fabxc

pod_test.go
```diff
@@ -35,52 +110,103 @@ func TestPodCollector(t *testing.T) {
...
-							},
-							api.ContainerStatus{
-								Name:         "container3",
-								RestartCount: 1,
-							},
-						},
-					},
-				},
-			},
+			pods: pods,
```
fabxc: Replicating the specific data that's tested here would provide a smaller scope so one doesn't always have to scroll up. The tradeoff would be more verbosity of course.

brancz: Fixed @fabxc . I prefer the decomposed version, as it shows the dependency of which information is required for which metric.

fabxc:
👍 cool, I think we can release tomorrow then.

PR #15, opened 2016-09-13
the `--dry-run` parameter was recently removed

brancz:
@fabxc


fabxc:
We might reconsider bringing this back as a development feature – probably rather named `--dump` then. However, haven't seen it coming up as a desirable feature in any Prometheus exporter so far.

PR #16, opened 2016-09-13
cleanup unused code

brancz:
these were artifacts from the previous approach where the metrics were collected periodically

@fabxc


fabxc:
👍

PR #21, opened 2016-09-19
Remove non kube metrics

ahakanbaba:
In this change we are removing the uninteresting metrics from the output related to the running kube-state-metrics process. They were prefixed with go_* 

We use a separate registry distinct from the defaultRegistry in the prometheus go client. 
Also use the recent promhttp implementation. 

Since the kube-state-metrics now depends on the prometheus/promhttp package, we also  committed a compatible version of the prometheus/promhttp into the vendor directory. 

Needless to say, all tests in this package still pass with this change. 

@brancz @fabxc

### DIFF
--- a/Godeps/Godeps.json
+++ b/Godeps/Godeps.json
@@ -96,6 +96,11 @@
 			"Rev": "8aae34f3ffc9a93c75efeaad87bb192aa7750900"
 		},
 		{
+			"ImportPath": "github.com/prometheus/client_golang/prometheus/promhttp",
+			"Comment": "v0.8.0-5-g8aae34f",
+			"Rev": "8aae34f3ffc9a93c75efeaad87bb192aa7750900"
+		},
+		{
 			"ImportPath": "github.com/prometheus/client_model/go",
 			"Comment": "model-0.0.2-12-gfa8ad6f",
 			"Rev": "fa8ad6fec33561be4280a8f0514318c79d7f6cb6"
--- a/main.go
+++ b/main.go
@@ -57,8 +57,6 @@ var (
 	kubeconfig = flags.String("kubeconfig", "./config", "absolute path to the kubeconfig file")
 
 	port = flags.Int("port", 80, `Port to expose metrics on.`)
-
-	prefix = flag.String("prefix", "kube_", "A prefix appended to the start of all kubernetes metric names")
 )
 
 func main() {
--- a/main.go
+++ b/main.go
@@ -37,6 +37,7 @@ import (
 	"k8s.io/client-go/1.4/tools/clientcmd"
 
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/prometheus/client_golang/prometheus/promhttp"
 )
 
 const (
@@ -56,6 +57,8 @@ var (
 	kubeconfig = flags.String("kubeconfig", "./config", "absolute path to the kubeconfig file")
 
 	port = flags.Int("port", 80, `Port to expose metrics on.`)
+
+	prefix = flag.String("prefix", "kube_", "A prefix appended to the start of all kubernetes metric names")
 )
 
 func main() {
@@ -226,6 +229,11 @@ func initializeMetrics(kubeClient clientset.Interface) {
 		}
 		return machines, nil
 	})
+
+	_ = promhttp.Handler()
+	// FixMe: This change line is only there for the go compiler not to compile.
+	// We first wanted to set the godeps correctly to include the promhttp in the
+	// vendor directory. In the following changes I will actually use the promhttp
 
 	prometheus.MustRegister(&deploymentCollector{store: dplLister})
 	prometheus.MustRegister(&podCollector{store: podLister})
--- a/main.go
+++ b/main.go
@@ -76,8 +76,11 @@ func main() {
 		glog.Fatalf("Failed to create client: %v", err)
 	}
 
-	initializeMetrics(kubeClient)
-	metricsServer()
+	// The speficic prometheus registry for the metrics
+	reg := prometheus.NewRegistry()
+
+	initializeMetrics(kubeClient, reg)
+	metricsServer(reg)
 }
 
 func createKubeClient() (kubeClient clientset.Interface, err error) {
@@ -131,13 +134,15 @@ func createKubeClient() (kubeClient clientset.Interface, err error) {
 	return kubeClient, nil
 }
 
-func metricsServer() {
+func metricsServer(reg prometheus.Gatherer) {
 	// Address to listen on for web interface and telemetry
 	listenAddress := fmt.Sprintf(":%d", *port)
 
 	glog.Infof("Starting metrics server: %s", listenAddress)
+
 	// Add metricsPath
-	http.Handle(metricsPath, prometheus.UninstrumentedHandler())
+	http.Handle(metricsPath, promhttp.HandlerFor(reg, promhttp.HandlerOpts{}))
+
 	// Add healthzPath
 	http.HandleFunc(healthzPath, func(w http.ResponseWriter, r *http.Request) {
 		w.WriteHeader(200)
@@ -178,7 +183,7 @@ func (l NodeLister) List() (v1.NodeList, error) {
 }
 
 // initializeMetrics creates a new controller from the given config.
-func initializeMetrics(kubeClient clientset.Interface) {
+func initializeMetrics(kubeClient clientset.Interface, reg prometheus.Registerer) {
 	dplStore, dplController := cache.NewNamespaceKeyedIndexerAndReflector(
 		&cache.ListWatch{
 			ListFunc: func(options api.ListOptions) (runtime.Object, error) {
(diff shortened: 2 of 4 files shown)


ahakanbaba:
I have signed it. 
I am part of the Box CLA. 
image

fabxc:
Thanks!

I'm generally a bit torn on this. These metrics are reporting health of the kube-state-metrics application itself, which we generally want to do in exporters.  
In the monitoring use case, these go_ metrics are basically negligible overhead compared to the total metrics and can provide critical information during an incident. 
I see how this is not too desirable for non-monitoring clients.

So far, in Prometheus land, we expected the clients to filter out metrics they do not care about.
However, recently the discussion about a standardized way for clients to filter the returned metrics based on request parameters.

I think an early-stage tool such as kube-state-metrics would be a good way to implement an initial proposal of this and see how it goes. This could look like

- `/metrics` returns everything
- `/metrics?filter=kube_.*` returns metrics prefixed with `kube_`
- `/metrics?filter=go_.*` returns metrics prefixed with `go_`

Would that sound like a sane alternative to you?
Another option would be to opt in/out by collector:

- `/metrics?collectors=deployment,pod,node`
- `/metrics?collectors=deployment,go_runtime`
- ...

@brian-brazil @beorn7

brian-brazil:
Based on discussions when we were in Berlin, the API will look something like the former and not the latter. If you need the latter then you're likely doing something so custom that you should handle it yourself via your own logic and a custom registry, rather than trying to plumb it through the prometheus client.

beorn7:
I can confirm. We essentially want a positive and a negative regexp match URL parameter. In the ideal case, the registry knows (via the Desc) what Collectors to call. In the case of a "wildcard" Desc (to be introduced in the next version of client_golang), the Collector is called in any case, but the registry still filters out any results not matching the parameters. (The same has to be done with Collectors that collect both matching and non-matching metrics.)

fabxc:
>  In the ideal case, the registry knows (via the Desc) what Collectors to call.

That's my central concern – if we can provide this, I'm fine with the regexp approach. However, I understand in some cases it's not really possible, e.g. MySQL exporter. One could probably provide some plumbing around this so developers can specify a prefix per collector?

beorn7:
I was pondering "prefix collectors" for a while, but came to the conclusion that they will add a lot of complexities with not much return.

What I am planning to add is a "describe by collect" helper, i.e. in cases where the metrics cannot be predicted at coding time but are constant over the runtime, you can implement describe by executing a collection and then returning the Desc's of the returned metrics.

There is still a case that isn't covered properly if all of the following is true:
- Collection is expensive (otherwise calling Collect and then filter before sending to the network would be fine).
- Metrics returned by a collector are not predictable at coding time (→ cannot use static Desc).
- Metrics returned by a collector change during runtime (→ cannot "describe by collect").

My preliminary conclusion is that this special case is not really solvable in general.

brian-brazil:
My thinking is in line with Beorn's. The special cases (e.g. JMX, and it turns out MySQL) are sufficiently special that we can't do much about them and have to leave it up to custom code by the exporter implementors

In the context of this PR, it doesn't appear to be special. I would raise the point that if one user is requesting the removal of the standard exports, then there's a chance they shouldn't be there for all users.

@ahakanbaba Would you mind explaining why you don't want these?

ahakanbaba:
Fabian thank you very much for the feedback.

On Mon, Sep 19, 2016 at 2:04 AM, Fabian Reinartz <notifications@github.com>
wrote:

> Thanks!
>
> I'm generally a bit torn on this. These metrics are reporting health of
> the kube-state-metrics application itself, which we generally want to do in
> exporters.
>
> In the monitoring use case, these go_ metrics are basically negligible
> overhead compared to the total metrics and can provide critical information
> during an incident.
> I see how this is not too desirable for non-monitoring clients.
>
Exactly that is not too desirable.

When you say ",which  we generally want to do in exporters" it suggest that
you have some existing or future use cases in mind that could use the go_*
metrics.

We did not have a use case and thought of them as really uninteresting.
The lack of tests for the go_* metrics also suggested that nobody cares
about them.

> So far, in Prometheus land, we expected the clients to filter out metrics
> they do not care about.
> However, recently the discussion about a standardized way for clients to
> filter the returned metrics based on request parameters.
>
Sorry, I did not quite understand this comment. Was there a discussion
about a grand scheme of filtering metrics ?

> I think an early-stage tool such as kube-state-metrics would be a good way
> to implement an initial proposal of this and see how it goes. This could
> look like
>
>    - /metrics returns everything
>    - /metrics?filter=kube_.* returns metrics prefixed with kube_
>
> We have some further questions about this recently introduced kube_
prefix. We plan to file an issue about that, just a heads up.

>
>    -
>    - /metrics?filter=go_.* returns metrics prefixed with go_
>
>
Do you foresee other types of prefixes here ? Is there a possibility of
adding new prefixes to the output ?

>
>    -
>
> Would that sound like a sane alternative to you?
> Another option would be to opt in/out by collector:
>
>    - /metrics?collectors=deployment,pod,node
>    - /metrics?collectors=deployment,go_runtime
>    - ...
>
>
I would also be interested in @ghodss 's input

ahakanbaba:
On Mon, Sep 19, 2016 at 9:18 AM, Brian Brazil <notifications@github.com>
wrote:

> My thinking is in line with Beorn's. The special cases (e.g. JMX, and it
> turns out MySQL) are sufficiently special that we can't do much about them
> and have to leave it up to custom code by the exporter implementors
>
> In the context of this PR, it doesn't appear to be special. I would raise
> the point that if one user is requesting the removal of the standard
> exports, then there's a chance they shouldn't be there for all users.
>
> @ahakanbaba <https://github.com/ahakanbaba> Would you mind explaining why
> you don't want these?
>

Hi Brian,
We really appreciate the feedback.

Please allow me some time to circle back in Box and give a definitive
answer.
Thanks

fabxc:
> We did not have a use case and thought of them as really uninteresting.
The lack of tests for the go_* metrics also suggested that nobody cares
about them.

There are no tests because they are created by the Prometheus client library. For the monitoring use case, they provide information about the health and state of kube_state_metrics. For example, if you see wrong or stale `kube_metrics_`, the `go_` metrics can indicate to you that `kube-state-metrics` itself might be having issues.

> Sorry, I did not quite understand this comment. Was there a discussion
about a grand scheme of filtering metrics ?

This is a discussion happening against the Prometheus client library. The comments above are talking about some of the aspects.

> Do you foresee other types of prefixes here ? Is there a possibility of
adding new prefixes to the output ?

Prefixes are simply a way to be explicit about what a metric is. If you are collecting metrics from hundreds of systems, you don't want to have metrics of the same name with different meanings. It also makes it easier to find them (e.g. auto completion).

ahakanbaba:
On Mon, Sep 19, 2016 at 11:13 AM, Fabian Reinartz <notifications@github.com>
wrote:

> We did not have a use case and thought of them as really uninteresting.
> The lack of tests for the go_* metrics also suggested that nobody cares
> about them.
>
> There are no tests because they are created by the Prometheus client
> library. For the monitoring use case, they provide information about the
> health and state of kube_state_metrics. For example, if you see wrong or
> stale kube_metrics_, the go_ metrics can indicate to you that
> kube-state-metrics itself might be having issues.
>
> Sorry, I did not quite understand this comment. Was there a discussion
> about a grand scheme of filtering metrics ?
>
> This is a discussion happening against the Prometheus client library. The
> comments above are talking about some of the aspects.
>
> Do you foresee other types of prefixes here ? Is there a possibility of
> adding new prefixes to the output ?
>
> Prefixes are simply a way to be explicit about what a metric is. If you
> are collecting metrics from hundreds of systems, you don't want to have
> metrics of the same name with different meanings. It also makes it easier
> to find them (e.g. auto completion).
>

Fabian,
thanks for the responses.

ahakanbaba:
Hi all for the comments, suggestions and insights. 
We have implemented a filter agent in front of the kube-state-metrics that purges the go_* metrics. 
So at this point, for us this pull request it not crucial. 

We do not want these go_* metrics because 
- We do not need them.
- Once they are in the metrics store database, they require some cognitive activity to discard and remember what they are. We have seen that they make browsing the saved metrics more complex.
- They also consume small but non-zero space. 

Since we actively remove the go_* metrics ourselves, we are abandoning this pull request. 
We still believe that retaining the go_* metrics in the kube-state-metrics defeats the Keep It Simple and You Aren't Gonna Need It principles. Preferably in the future the prometheus golang client would not add these metrics by default. 

Thanks everyone for their time.

fabxc:
Thanks for the update.
Indeed, an option to disable them should be there (@beorn7). Even with it, I'd still argue for keeping them. As said above, the kube-state-metrics exporter gives you metrics about the state of your cluster, but how would you debug the exporter if it starts responding really slowly etc.?
You would need metrics and they have to come from the exporter itself. It's yet another application in your cluster that you do want to monitor.

beorn7:
The option to disable the "automatic" metrics is already there: Use a custom registry and don't register the Go collector and process collector. If you want it configurable, make the registration of those collectors configurable.

fabxc:
Fair point, I keep forgetting that a large number of problems are solved with non-global registries now,

PR #23, opened 2016-09-19
Add kube_pod_status_scheduled metric

gdvalle:

PR #24, opened 2016-09-19
Add test for kube_pod_status_scheduled

gdvalle:
Forgot to include the test in #23.


brancz:
That was my bad. Thanks a lot! 👍

PR #25, opened 2016-09-20
Add Makefile

fabxc:
@brancz

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,26 @@
+all: build
+
+FLAGS =
+ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
+REGISTRY = gcr.io/google_containers
+TAG = v0.2.0
+
+deps:
+	go get github.com/tools/godep
+
+build: clean deps
+	$(ENVVAR) godep go build -o kube-state-metrics 
+
+test-unit: clean deps build
+	$(ENVVAR) godep go test --race . $(FLAGS)
+
+container: build
+	docker build -t ${REGISTRY}/kube-state-metrics:$(TAG) .
+
+push: container
+	gcloud docker push ${REGISTRY}/kube-state-metrics:$(TAG)
+
+clean:
+	rm -f kube-state-metrics
+
+.PHONY: all deps build test-unit container push clean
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main // import "github.com/kubernetes/kube-state-metrics"
+package main // `import "k8s.io/kube-statemetric"
```
brancz: ``` ` ``` -> `"`

Makefile
```diff
@@ -0,0 +1,26 @@
...
+deps:
+	go get github.com/tools/godep
+
+build: clean deps
+	$(ENVVAR) godep go build -o kube-state-metrics 
+
+test-unit: clean deps build
+	$(ENVVAR) godep go test --race . $(FLAGS)
+
+container: build
```
brancz: Maybe `container-image`? The make target `container` rather sounds like creating a running container.

Makefile
```diff
@@ -0,0 +1,26 @@
+all: build
+
+FLAGS =
+ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
+REGISTRY = gcr.io/google_containers
+TAG = v0.2.0
```
brancz: Wdyt about moving this into a file? I can see this easily being forgotten when creating a new release...

main.go
brancz: And typo `kube-statemetric` -> `kube-state-metrics`

Makefile
fabxc: The file was largely taken from the rescheduler repository. Figured we just keep it consistent.

Makefile
fabxc: Same as above. Let's move it into a file later if it turns out to be okay.

fabxc:
Fixed import path.

Makefile
brancz: sgtm

w0mbat:
I had used the `centurylink/golang-builder` container to build the binary and package everything with into a Docker image - do we want to update `Building the Docker container` section and describe it with the `make` command?

brancz:
@w0mbat that would be good I guess /cc @fabxc

PR #26, opened 2016-09-20
Update container build instructions

w0mbat:
I removed the docker run command to build the binary and docker image and replaced it with the appropriate make command

### DIFF
--- a/README.md
+++ b/README.md
@@ -95,11 +95,7 @@ those monitoring systems.
 Simple run the following command in this root folder, which will create a
 self-contained, statically-linked binary and build a Docker image:
 ```
-docker run --rm \
-  -v "$(pwd):/src" \
-  -v /var/run/docker.sock:/var/run/docker.sock \
-  centurylink/golang-builder \
-  kube-state-metrics
+make container
 ```
 
 # Usage


w0mbat:
One thing I noticed is that the resulting container image is now 43.15 MB - it was 32.95 MB when I was building it with the build container

brancz:
As the current Dockerfile builds off of scratch so it is mainly the binary that creates this size. As far as I can tell the golang-builder doesn't do [anything magic](https://github.com/CenturyLinkLabs/golang-builder/blob/63a09d9168c94ba80bff7914465b057a57ee937a/builder/build.sh#L20-L29), so we can look into it. However, we were instructed to use the Makefile as we have it right now. Once we have everything in place and working we can think of optimizing the whole process of building the binary and image.

brancz:
Just checked and what I described seems to be the case.

Before (compiled with: `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build -o kube-state-metrics`):

```
-rwxr-xr-x  1 fredericbranczyk staff 53343352 Sep 20 13:43 kube-state-metrics
```

After (compiled with: `GOOS=linux GOARCH=amd64 CGO_ENABLED=0 go build --installsuffix cgo --ldflags="-s" -o kube-state-metrics`): 

```
-rwxr-xr-x  1 fredericbranczyk staff 32983648 Sep 20 13:57 kube-state-metrics
```

As always though, it does increase build time. I'll merge this, but wanted to comment this as a reference to get back to when we think about optimizing.

PR #30, opened 2016-09-27
use kubernetes informer framework

### LINKED ISSUE #29: upgrade to new version of client-go
client-go has released a new client version, under 1.5

https://github.com/kubernetes/client-go

brancz:
Fixes #29 .

@fabxc @ghodss


brancz:
I signed it!

fabxc:
Do we see an immediate use case for the informers that should finally land soon? (https://github.com/kubernetes/client-go/issues/4)

If so, we might want to hold off so we don't constantly revendor.

brancz:
That makes sense. [This](https://github.com/kubernetes/client-go/issues/4#issuecomment-249627687) says the informer framework should be in the client once [this](https://github.com/kubernetes/kubernetes/pull/33334#event-803521757) is merged, which it is already, just waiting for submit queue. So I will hold back on this once the informer framework is included. Then we vendor it in this PR and then switch to it in an additional PR.

brancz:
Although it's merged now and available, it seems that there is still [a problem with it](https://github.com/kubernetes/client-go/issues/4#issuecomment-250061886), however, it looks like once kubernetes/kubernetes#33632 is fixed it will work. So holding off just a bit longer until then.

brancz:
As kubernetes/kubernetes#33632 is merged we're only waiting for submit queue again. 🎉

brancz:
Alright, 🍏 CI. Re-review @fabxc @ghodss ?

brancz:
Alright now implemented it with the k8s informer framework. @fabxc

deployment.go
```diff
@@ -70,7 +69,7 @@ var (
 )
 
 type deploymentStore interface {
-	List() (deployments []v1beta1.Deployment, err error)
+	List() []interface{}
 }
```
fabxc: `List() []interface{}` is super generic now. Not really necessary to keep one type per resource type anymore.
Maybe just:

```go
type lister interface {
    List() []interface{}
}
```

main.go
```diff
@@ -156,78 +155,24 @@ func metricsServer() {
...
+	ninf := cache.NewSharedInformer(nlw, &v1.Node{}, resyncPeriod)
+
+	prometheus.MustRegister(&deploymentCollector{store: dinf.GetStore()})
+	prometheus.MustRegister(&podCollector{store: pinf.GetStore()})
+	prometheus.MustRegister(&nodeCollector{store: ninf.GetStore()})
 
-	prometheus.MustRegister(&deploymentCollector{store: dplLister})
-	prometheus.MustRegister(&podCollector{store: podLister})
-	prometheus.MustRegister(&nodeCollector{store: nodeLister})
+	go dinf.Run(context.TODO().Done())
```
fabxc: We can just use background here as this is basically top-level. From my understanding, TODO is intended for when you want to connect a context chain later on but can't yet because the call chain doesn't fully support it yet.

deployment.go
fabxc: Or keep them as they were and add a wrapper around the cache's `List` method doing the typecast there. Saves us from dragging `interface{}` through the actual application logic layers and we can keep working with more idiomatic go.

brancz:
@fabxc Undid the interface weirdness and applied the other comments.

fabxc:
👍

Issue #33, opened 2016-10-06
Feature Request: Add a new metric for currently used resources in nodes. 


ahakanbaba:
Hi All,
We would like to propose a new metric addition that describes the used resources by pods in every node. 
Its name could be 
kube_node_current_pod_cpu_requests,
kube_node_requested_cpu_resources,
kube_node_allocated_cpu_resources. 

The names extend naturally for _memory_ 

Or something else suggested by you we are open to suggestions for names. 

As the desired semantics, we propose this metric to contain the sum of requested resources of all the currently running pods in that node.

The same computation is done every time one executes `kubectl describe node`. 
The resources table is computed every time from scratch. 
For example

```
 Namespace          Name                        CPU Requests    CPU Limits  Memory Requests Memory Limits
  ---------         ----                        ------------    ----------  --------------- -------------
  default           k8s-master-127.0.0.1                0 (0%)      0 (0%)      0 (0%)      0 (0%)
  default           k8s-proxy-127.0.0.1             0 (0%)      0 (0%)      0 (0%)      0 (0%)
  hello-world-hbaba-build   hello-world-hbaba-989929512-6jiaf       800m (20%)  20 (500%)   100Mi (2%)  100Mi (2%)
  hello-world-hbaba-build   hello-world-hbaba-989929512-8hq9v       800m (20%)  20 (500%)   100Mi (2%)  100Mi (2%)
  hello-world-hbaba-build   hello-world-hbaba-989929512-jjd5n       800m (20%)  20 (500%)   100Mi (2%)  100Mi (2%)
  hello-world-hbaba-build   hello-world-hbaba-989929512-n00t2       800m (20%)  20 (500%)   100Mi (2%)  100Mi (2%)
  kube-system           kube-dns-v10-z4ics              310m (7%)   310m (7%)   170Mi (4%)  170Mi (4%)
Allocated resources:
  (Total limits may be over 100%, i.e., overcommitted. More info: http://releases.k8s.io/HEAD/docs/user-guide/compute-resources.md)
  CPU Requests  CPU Limits  Memory Requests Memory Limits
  ------------  ----------  --------------- -------------
  3510m (87%)   80310m (2007%)  570Mi (14%) 570Mi (14%)
````

The code for that table is in `kubernetes/pkg/kubectl/describe.go` file in `describeNodeResource` function. It iterates over a collection that is accessed like this : 
[...]

PR #34, opened 2016-10-07
Addition of per node requested cpu and memory resources.

ahakanbaba:
This change relates to issue #33.
We added a new Collector that uses both the nodeList and podList.
For each node, the collector finds the relevant pods and sums their
requested resources.
This is very similar to the `kubectl describe node` implementation.

In this change automated tests are missing. I will follow up with
another commit for the tests.

I wanted to create this commit early to get responses and feedback as
soon as possible.

### DIFF
--- a/node_pod.go
+++ b/node_pod.go
@@ -1,0 +1,115 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+// In this package, the mentioned metrics need both the node list as well as the
+// pod list.
+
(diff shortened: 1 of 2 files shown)


node_pod.go
```diff
@@ -0,0 +1,115 @@
...
+
+	reqCpu := resource.Quantity{}
+	reqMem := resource.Quantity{}
+	nn := n.Name
+	for _, p := range pods {
+		// TODO: There should be a faster way to get the reduced set of
+		// relevant pods with List(ap.ListOptions{FieldSelector})
+		// Here we are iterating over all nodes every time and skipping
+		// uninteresting ones.
+		if isRelevantPod(p, nn) {
```
ahakanbaba: I am grad that you are also concerned about the performance. This iteration over all pods for every node would not scale. 
I have looked through, investigated through the api as much as I could but could not find a simple way to call List( options api.ListOptions) function easily from inside the Collect function. 
I am very open to suggestions and would change this accordingly.

brancz: Maybe this metric makes most sense to be directly in the kubelet, then there is no need to filter, which is purely produced by the indirection over the apiserver.

fabxc: Please see my comment [here](https://github.com/kubernetes/kube-state-metrics/issues/33#issuecomment-252182854).
I'd suggest to focus on exposing better base metrics and getting the labeling straight.

Once we have something stable, we should move certain metrics into the kubelet as well.

ahakanbaba:
I am closing this pull request in favor of #35 
#35 supersedes this.

PR #35, opened 2016-10-08
Added metrics describing the cpu and mem requests per container .

ahakanbaba:
These metrics are retrieved using the v1 api directly.
No additional acucmulation or computation is done.

The state is maintained per container in a pod in kubernetes.
This change exposes the state in kubernetes via  kube-state-metrics

The metrics are also tagged by the node,where  the containers are residing.
Our use case for these metrics are as follows:

We would like to accumulate these per node and understand how much
remaining un-requested resourced are available in each node.

This would give an indication of the "utilization" of the cluster to us.

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -76,13 +76,13 @@ var (
 
 	descPodContainerRequestedCpuMilliCores = prometheus.NewDesc(
 		"kube_pod_container_requested_cpu_millicores",
-		"The number of requested cpu millicores by a container",
+		"The number of requested cpu millicores by a container.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
 
 	descPodContainerRequestedMemoryBytes = prometheus.NewDesc(
 		"kube_pod_container_requested_memory_bytes",
-		"The number of requested memory bytes  by a container",
+		"The number of requested memory bytes  by a container.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
 )
@@ -162,12 +162,14 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 	nodeName := p.Spec.NodeName
 	for _, c := range p.Spec.Containers {
 		req := c.Resources.Requests
-		cpu := req[v1.ResourceCPU]
-		mem := req[v1.ResourceMemory]
+		if cpu, ok := req[v1.ResourceCPU]; ok {
+			addGauge(descPodContainerRequestedCpuMilliCores, float64(cpu.MilliValue()),
+				c.Name, nodeName)
+		}
+		if mem, ok := req[v1.ResourceMemory]; ok {
+			addGauge(descPodContainerRequestedMemoryBytes, float64(mem.Value()),
+				c.Name, nodeName)
+		}
 
-		addGauge(descPodContainerRequestedCpuMilliCores, float64(cpu.MilliValue()),
-			c.Name, nodeName)
-		addGauge(descPodContainerRequestedMemoryBytes, float64(mem.Value()),
-			c.Name, nodeName)
 	}
 }
--- a/pod.go
+++ b/pod.go
@@ -73,6 +73,18 @@ var (
 		"The number of container restarts per container.",
 		[]string{"namespace", "pod", "container"}, nil,
 	)
+
+	descPodContainerRequestedCpuMilliCores = prometheus.NewDesc(
+		"kube_pod_container_requested_cpu_millicores",
+		"The number of requested cpu millicores by a container",
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
+
+	descPodContainerRequestedMemoryBytes = prometheus.NewDesc(
+		"kube_pod_container_requested_memory_bytes",
+		"The number of requested memory bytes  by a container",
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
 )
 
 type podStore interface {
@@ -96,6 +108,8 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerStatusTerminated
 	ch <- descPodContainerStatusReady
 	ch <- descPodContainerStatusRestarts
+	ch <- descPodContainerRequestedCpuMilliCores
+	ch <- descPodContainerRequestedMemoryBytes
 }
 
 // Collect implements the prometheus.Collector interface.
@@ -143,5 +157,17 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 		addGauge(descPodContainerStatusTerminated, boolFloat64(cs.State.Terminated != nil), cs.Name)
 		addGauge(descPodContainerStatusReady, boolFloat64(cs.Ready), cs.Name)
 		addCounter(descPodContainerStatusRestarts, float64(cs.RestartCount), cs.Name)
+	}
+
+	nodeName := p.Spec.NodeName
+	for _, c := range p.Spec.Containers {
+		req := c.Resources.Requests
+		cpu := req[v1.ResourceCPU]
+		mem := req[v1.ResourceMemory]
+
+		addGauge(descPodContainerRequestedCpuMilliCores, float64(cpu.MilliValue()),
+			c.Name, nodeName)
+		addGauge(descPodContainerRequestedMemoryBytes, float64(mem.Value()),
+			c.Name, nodeName)
 	}
 }
--- a/pod_test.go
+++ b/pod_test.go
@@ -19,6 +19,7 @@ package main
 import (
 	"testing"
 
+	"k8s.io/client-go/1.4/pkg/api/resource"
 	"k8s.io/client-go/1.4/pkg/api/v1"
 )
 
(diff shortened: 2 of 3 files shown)


ahakanbaba:
The discussion in #33 is relevant to this pull request.

ahakanbaba:
Please note, in this pull request there are no tests. 
I wanted to get an initial reaction to these new metrics before investing time in terms of tests. 

Once I get a favorable response I will add tests too.

pod.go
```diff
@@ -144,4 +158,16 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 		addGauge(descPodContainerStatusReady, boolFloat64(cs.Ready), cs.Name)
 		addCounter(descPodContainerStatusRestarts, float64(cs.RestartCount), cs.Name)
 	}
+
+	nodeName := p.Spec.NodeName
```
brancz: I wonder at which point `NodeName` is set on the PodSpec when unspecified on creation. The concern is that this might be race-y with Prometheus starting to scrape this being unset and then set on the next scrape, which would make it a new time series.

pod.go
```diff
@@ -144,4 +158,16 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
...
+	for _, c := range p.Spec.Containers {
+		req := c.Resources.Requests
```
brancz: As far as I understand resource requests cannot change over a pods lifetime.

```
pod updates may not change fields other than `containers[*].image` or `spec.activeDeadlineSeconds`
```

Therefore it might make sense to put that information into the [PodContainerInfo](https://github.com/ahakanbaba/kube-state-metrics/blob/79761d06617809f78c16a5f129e8267e1e8fe7cd/pod.go#L152). Similar to the PodInfo or NodeInfo metrics.

ahakanbaba: Looking at the api definition, the comment mentions that the resources cannot change. 
The Container type  in `client-go/1.4/pkg/api/v1/types.go` has this.

```
 // Compute Resources required by this container.
 // Cannot be updated.
 // More info: http://releases.k8s.io/HEAD/docs/user-guide/persistent-volumes.md#resources
 Resources ResourceRequirements `json:"resources,omitempty" protobuf:"bytes,8,opt,name=resources"` 
```

Are you sure the resources really belongs in the descContainerInfo ? 
One argument is about the composition of the descContainerInfo. The current tags of descContainerInfo are all from PodStatus.ContainerStatus. This new addition of ResourceRequirements is part of the PodSpec not  PodStatus. 
If we add the per container resources to descPodContainerInfo, the metric will have tags both from the PodStatus and PodSpec. I do not argue that this is a fundamental problem but may indicate a problem in the encapsulation and scope of the metrics emitted from kube-state-metrics. I wanted to double check that this is really the direction we want to go. 

A second more important concern of mine is that the labelValues in [MustNewConstMetric](https://godoc.org/github.com/prometheus/client_golang/prometheus#MustNewConstMetric)  are strings. The resources that we collect are of millicpus and bytes types. Stringifying them may be problematic for the consuming system. 

As mentioned [before](https://github.com/kubernetes/kube-state-metrics/issues/33#issue-181472679) our use case is to accumulate these request resources for all containers, in all pods running in one node. The consuming system needs to handle an atoi like conversion and then accumulation.  @fabxc gave an example code snippet [here](https://github.com/kubernetes/kube-state-metrics/issues/33#issuecomment-252182854). I am not sure whether a similar accumulation is possible if the requestedResources are stringified lables to some metric. I have taken a brief look at [prometheus operators](https://prometheus.io/docs/querying/operators/) and it was not immediately obvious to me.

ahakanbaba:
Any other comments ? 
We have been blocked on this for couple of days now. It would be great to have some swift conclusion as soon as possible.
Thanks

pod.go
ahakanbaba: Does prometheus create a new time series if the tag of some metric changes over time ? 

Could you please give more details of the problem you foresee ? 
Would there be a very short lived time series with an empty value in the node tag? 
Is that your concern ?

pod.go
brancz: > A second more important concern of mine is that the labelValues in MustNewConstMetric are strings. The resources that we collect are of millicpus and bytes types. Stringifying them may be problematic for the consuming system.

My bad, I didn't think of that.

> As mentioned before our use case is to accumulate these request resources for all containers, in all pods running in one node. The consuming system needs to handle an atoi like conversion and then accumulation. @fabxc gave an example code snippet here. I am not sure whether a similar accumulation is possible if the requestedResources are stringified lables to some metric. I have taken a brief look at prometheus operators and it was not immediately obvious to me.

Agreed. Then let's do it this way.

pod.go
```diff
@@ -144,4 +158,16 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
...
+		cpu := req[v1.ResourceCPU]
+		mem := req[v1.ResourceMemory]
+
+		addGauge(descPodContainerRequestedCpuMilliCores, float64(cpu.MilliValue()),
+			c.Name, nodeName)
```
brancz: Above you declared the labels
```
"namespace", "pod", "container", "node"
```
but you are setting `namespace` to `c.Name` and `pod` to `nodeName`. Same below.

pod.go
brancz: Yes, a new combination of labels is a new time series in Prometheus. There could be is that whenever a Pod is created and the `NodeName` is _not yet_ set, then there will be two time-series once it is set, and the time-series where the `NodeName` is set will be missing the first scrape(s) (depending on when it is set). I would say this is fine for now, we should just keep it in mind.

brancz:
Unfortunately cannot edit the comment I wrote when submitting the review, so if you just add some tests, this should be ready to 🚢

pod.go
ahakanbaba: Hi Frederic, thanks for the detailed explanation. I am not very familiar with the sequence NodeName goes through at this point :(

ahakanbaba:
Great. Thanks @brancz. I will add tests today as soon as possible. I will absolutely need help to merge this and also cut a new release (if possible). So that we can immediately incorporate the new metrics on our side.

brancz:
I'll do my best, unfortunately @fabxc is on vacation, so there is not really anyone who can review a release. If it's not too much trouble for you maybe in the mean time you can build from master until we get the next release out there?

ahakanbaba:
Thank you so much for your response. 
I have pushed the changes that contains the test. Please let me know if you see any problem with that. 
I completely understand the delay caused by the timezones. 
In the meantime, I will investigate the "compiling from master" option internally. 
Thanks

pod.go
```diff
@@ -144,4 +158,16 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
...
 		addCounter(descPodContainerStatusRestarts, float64(cs.RestartCount), cs.Name)
 	}
+
+	nodeName := p.Spec.NodeName
+	for _, c := range p.Spec.Containers {
+		req := c.Resources.Requests
+		cpu := req[v1.ResourceCPU]
+		mem := req[v1.ResourceMemory]
+
+		addGauge(descPodContainerRequestedCpuMilliCores, float64(cpu.MilliValue()),
```
brancz: I'm wondering whether we should not add the metric when `cpu`/`mem` is `nil`, wdyt? Right now this produces metrics with a 0 value when no resource requests are set. Either way we need a test where no resource requests are set.

brancz:
Other than the requested change, could you please add documentation on the `README.md`?

pod.go
```diff
@@ -73,6 +73,18 @@ var (
 		"The number of container restarts per container.",
 		[]string{"namespace", "pod", "container"}, nil,
 	)
+
+	descPodContainerRequestedCpuMilliCores = prometheus.NewDesc(
+		"kube_pod_container_requested_cpu_millicores",
+		"The number of requested cpu millicores by a container",
```
brancz: Nitpick, but missing colon on both of these descriptions.

ahakanbaba:
Thanks for the comments. @brancz. 
I have addressed them. 

:) I forgot about the readme file entirely. Maybe once this project get larger, we could work on auto generating the readme file from the prometheus.NewDesc constructs.

brancz:
Thanks for bearing with me :) . LGTM 👍

brancz:
As mentioned above, we'll probably do a release within the next week. In the mean time it's probably best to build from master.

ahakanbaba:
Frederick,
it was my pleasure.
Thanks
Hakan

On Thu, Oct 13, 2016 at 2:03 AM, Frederic Branczyk <notifications@github.com

pod_test.go
```diff
@@ -54,6 +55,10 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_status_ready gauge
 		# HELP kube_pod_status_scheduled Describes the status of the scheduling process for the pod.
 		# TYPE kube_pod_status_scheduled gauge
+		# HELP kube_pod_container_requested_cpu_millicores The number of requested cpu millicores by a container.
+		# TYPE kube_pod_container_requested_cpu_millicores gauge
```
fabxc: Those should be base units, i.e. "cores", "bytes", "seconds", etc.
As the exposition is floating-point, no particular is better. But sticking with the simplest one everywhere, avoids having to account for it in queries.

PR #39, opened 2016-10-17
fix allocateable -> allocatable

### LINKED ISSUE #38: Typo in the metric name "allocateable"
As far as I know this should be spelled as allocatable. Without an "e" between the "t" and "a".

I apologize if there are alternate spellings of this word. I am by no means an expert in spelling. 

Nevertheless, I see value to be consistent with the [kubernetes repo](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/types.go#L2161) in terms of spelling. 

It also seems like google suggests the allocatable spelling 
https://www.google.com/webhp?sourceid=chrome-instant&ion=1&espv=2&ie=UTF-8#q=allocateable

brancz:
fixes #38 

@fabxc @ahakanbaba


fabxc:
👍

PR #40, opened 2016-10-17
normalize cpu cores unit

brancz:
As @fabxc pointed out in [this comment](https://github.com/kubernetes/kube-state-metrics/pull/35#discussion_r83591789) it is best practice to use base units.

@ahakanbaba


brancz:
Nope. That's why we currently have this in the `README.md`

> WARNING: THESE METRIC/TAG NAMES ARE UNSTABLE AND MAY CHANGE IN A FUTURE RELEASE.

The metric has also not been included in a release so far.

fabxc:
👍

PR #41, opened 2016-10-17
Fix resource fractions

fabxc:
This fixes the generation of resource metrics to respect milli
precision.

@brancz I realized I introduced this mistake when reviewing your last PR.


brancz:
Huh .. it's weird because I tried the 

```
resource.MustParse("100m")
```

And that kept giving me 1 cpu where it should have given me 0.1. Do you have an explanation to that?

Other than that LGTM 👍

fabxc:
Guess it just always applies `ceil()` internally when dropping the fraction. Same behavior for sub-milli fractions and `MilliValue`.

PR #42, opened 2016-10-17
don't panic when requesting help

### LINKED ISSUE #28: Panic when showing flags
I find it unexpected that a simple `kube-state-metrics -h` will cause a stack trace to be printed.

brancz:
fixes #28

@fabxc @grobie


fabxc:
👍

PR #45, opened 2016-10-21
healthcheck metrics

sebv:
This does the following:
- adds event watcher (with pod events filtering)
- adds 2 healthcheck related metrics:
  - total number of healthcheck failures
  - seconds since last healthcheck failure

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/45)

### DIFF
--- a/event.go
+++ b/event.go
@@ -1,0 +1,82 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package main
+
+import (
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+)
+
+var (
+	descPodNumOfHealthcheckFailures = prometheus.NewDesc(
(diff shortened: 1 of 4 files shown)


sebv:
I signed it!

sebv:
retest this please

brancz:
Sorry very busy currently, I have this on my todo list and will hopefully get to it soon.

brancz:
Sorry for only getting back to this now, just out of curiosity, what would alerts look like, that you would write with having these metrics available.

Also wdyt of having these metrics @fabxc @matthiasr ?

event.go
```diff
@@ -0,0 +1,82 @@
...
+	"time"
+
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/client-go/1.5/pkg/api/v1"
+)
+
+var (
+	descPodNumOfHealthcheckFailures = prometheus.NewDesc(
+		"kube_pod_healthcheck_num_of_failures",
```
fabxc: `kube_pod_healthcheck_failures_total` would be the best practice naming.

event.go
```diff
@@ -0,0 +1,82 @@
...
+		"Number of healthcheck failures for a given pod.",
+		[]string{"namespace", "pod"}, nil,
+	)
+	descSecondsSinceLastHealthcheckFailure = prometheus.NewDesc(
+		"kube_pod_healthcheck_seconds_since_last_failure",
```
fabxc: `kube_pod_healthcheck_last_failure` would be best practice here and have it return a unix timestamp (seconds, sub-seconds in fractions). Exporters should avoid doing any pre-calculation and aim to export raw data.

A metric processing system generally easily allows doing the calculations on the fly. Having the raw data allows more different calculations.
Especially as the value here is tied to the timestamp it is associated with and looking at the last value of the series has little meaning without its timestamp.

Whereas with exposing a timestamp, I can query my metric system like 

```time_now() - kube_pod_healthcheck_last_failure```

 and it will always give me the expected value w.r.t the time I did the query.

Does that reasoning make sense?

PR #49, opened 2016-11-17
Export per container limits

### LINKED ISSUE #48: Report CPU and Memory requests and limits
We are trying to keep an eye on the overall CPU and Memory requests and limits across the cluster.

`kubectl describe nodes` reports this information per node, but this exporter does not seem to expose it.

Example output:

```
Allocated resources:
  (Total limits may be over 100 percent, i.e., overcommitted. More info: http://releases.k8s.io/HEAD/docs/user-guide/compute-resources.md)
  CPU Requests	CPU Limits	Memory Requests	Memory Limits
  ------------	----------	---------------	-------------
 1095m (54%)	5480m (274%)	786Mi (10%)	3418Mi (45%)
```

As kubectl is reporting this information it is certainly available to this exporter as well.

Is there any way to get these numbers from the already exported metrics or could someone look into exporting them?

IMHO these "Cluster Usage" metrics are the most important metrics K8s has, and right now it's a little hard (or at least unclear) how to get them.

dominikschulz:
Fixes #48

### DIFF
--- a/README.md
+++ b/README.md
@@ -69,8 +69,8 @@ additional metrics!
 | kube_pod_container_status_restarts | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; |
 | kube_pod_container_requested_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
 | kube_pod_container_requested_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
-| kube_pod_container_limited_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
-| kube_pod_container_limited_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
+| kube_pod_container_limits_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
+| kube_pod_container_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
 
 ## kube-state-metrics vs. Heapster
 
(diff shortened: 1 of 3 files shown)


dominikschulz:
Could test this much, so far. Would be nice to get some feedback on the implementation and I will test it in my cluster ASAP.

pod.go
```diff
@@ -85,6 +85,18 @@ var (
 		"The number of requested memory bytes  by a container.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
+
+	descPodContainerLimitedCpuCores = prometheus.NewDesc(
+		"kube_pod_container_limited_cpu_cores",
+		"The number of limited cpu cores by a container.",
```
matthiasr: This is not your fault (it matches the request metrics) but I don't like the "limited X" / "requested X" wording here. @brancz WDYT about changing this before the next release? I would prefer `kube_pod_container_{limits,requests}_{cpu_cores,memory_bytes}` to match the hierarchy in the pod spec.

dominikschulz:
CLA is already signed. Don't know why the check still fails.

matthiasr:
I think the CLA check needs a comment (or time) to pick up a state change. It's green now.

pod.go
```diff
@@ -85,6 +85,18 @@ var (
...
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
+
+	descPodContainerLimitedMemoryBytes = prometheus.NewDesc(
+		"kube_pod_container_limited_memory_bytes",
+		"The number of limited memory bytes  by a container.",
```
brancz: two spaces between "bytes" and "by"

pod.go
brancz: Absolutely for it. I would even say to avoid further confusion and to stay as close as possible to the API to do `kube_pod_container_resources_{limits,requests}_{cpu_cores,memory_bytes}`. Let's start that naming right away and I'll fix the existing naming, so we have consistency for the next release.

README.md
```diff
@@ -69,6 +69,8 @@ additional metrics!
 | kube_pod_container_status_restarts | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; |
 | kube_pod_container_requested_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
 | kube_pod_container_requested_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
+| kube_pod_container_limited_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
```
brancz: This also needs updating.

pod.go
```diff
@@ -85,6 +85,18 @@ var (
 		"The number of requested memory bytes  by a container.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
+
+	descPodContainerLimitsCpuCores = prometheus.NewDesc(
+		"kube_pod_container_limits_cpu_cores",
+		"The number of limits on cpu cores by a container.",
```
brancz: Maybe rather "The limit on cpu cores to be used by a container". Wdyt?

dominikschulz: Sounds much better.

README.md
dominikschulz: Should be OK now.

brancz:
LGTM 👍

PR #52, opened 2016-11-26
Pod Metric Labels

Nalum:
I want to add the Node Name to the Pod metric labels as I think it make more sense to have that in there rather than or as well as the Node IP.

I also think given that the Node IP is used elsewhere it would be worth adding it to the NodeInfo metric. I'm willing to do this but haven't started on it yet as I'd like input as to which IP to use. I'll open an issue for discussion around that.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/52)

### DIFF
--- a/README.md
+++ b/README.md
@@ -57,20 +57,20 @@ additional metrics!
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_pod_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `start_time`=&lt;date-time since kubelet acknowledged pod&gt; |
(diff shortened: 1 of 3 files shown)


Nalum:
I signed it

brancz:
I feel like we really only need this information on the `kube_pod_info` metric, and all other metrics can be joined onto it. Wdyt @fabxc ?

fabxc:
Yes, this is one arbitrary piece of information added to every single metric of an already uniquely identifiable object. The next person wants another bit and adds another label to it and so on...

It gets particularly interesting when think that `node` is not actually a stable attribute if pods can be live-migrated at some point. Thus all time series of a pod would churn at once even though the pod is still there.

This belongs indeed into `kube_pod_info`.

Nalum:
I'll reduce the change to just having the `NodeName` on `kube_pod_info` if that is preferred.

Nalum:
I also just want to note that I'm running `go version go1.7.3 linux/amd64` and `make test-unit` requires `CGO_ENABLED=1` because of the `-race` flag.

brancz:
@Nalum the Makefile was contributed by google and they require us to leave it as is for now.

Other than that lgtm 👍, thanks for the contribution and sorry that it took so long!

PR #54, opened 2016-12-04
Add DaemonSet metrics

dominikschulz:
This PR adds DaemonSet state metrics.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/54)

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,15 @@ additional metrics!
 | kube_node_status_allocatable_memory_bytes | Gauge | `node`=&lt;node-address&gt;|
 | kube_node_status_allocatable_pods | Gauge | `node`=&lt;node-address&gt;|
 
+### DaemonSet Metrics
+
+| Metric name| Metric type | Labels/tags |
+| ---------- | ----------- | ----------- |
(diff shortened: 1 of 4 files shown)


brancz:
There is some work on upstream kubernetes/kubernetes regarding `DaemonSet` rolling updates. I haven't followed too much, can you confirm that it won't have the same problem as described in https://github.com/kubernetes/kube-state-metrics/pull/55#issuecomment-264807664.

dominikschulz:
No, sorry. Not sure about the upstream work. We're using DaemonSets mostly for running prometheus-node-exporter and this change only adds very few timeseries. Maybe the rolling updates you mention affect this, but I can't comment on that.

brancz:
Alright I'll have a look myself, but can't promise when I'll get to it. But greatly appreciate your work as it's something many people are looking for!

brancz:
So the upstream rolling updates implementation does not implement anything on top of daemonsets and will not be in the kubernetes 1.5 release anyways, so there is absolutely no reason to worry about the time series churn for this, if anything equivalent to the `Deployment` gets implemented for the `DaemonSet` we can still make this optional.

brancz:
lgtm 👍  thanks!

PR #55, opened 2016-12-04
Add ReplicaSet metrics.

dominikschulz:
This PR adds ReplicaSet metrics.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/55)

### DIFF
--- a/README.md
+++ b/README.md
@@ -72,6 +72,19 @@ additional metrics!
 | kube_pod_container_limits_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
 | kube_pod_container_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-address&gt; |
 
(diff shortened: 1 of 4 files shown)


brancz:
I'm a bit concerned about time-series churn with this one as `ReplicaSet`s are most likely used by `Deployment`s. Just wondering, do you have that many `ReplicaSet`s independent of `Deployment`s that this will make a difference?

brancz:
wdyt @fabxc ?

dominikschulz:
You're probably right about time series churn. Even in a rather small cluster we have a few hundred timeseries for these metrics alone.

wdyt if I make this feature optional?

brancz:
I'm not sure what the benefit of this is as I don't see it being used outside of Deployments, but I'm fine with it as long as it is optional and by default turned off. Let's take this opportunity then to introduce a model to enable or disable certain collectors, can you include that in this PR? I suggest to have all of the collectors enabled by default that are in master right now and allow them to be disabled/enabled in a similar fashion as it is done in the [node-exporter](https://github.com/prometheus/node_exporter). Can you do that @dominikschulz ?

brancz:
After some discussion this doesn't turn out to be such a big problem, as we are already exposing metrics based on `Pod`s, which naturally have a much higher churn than `ReplicaSet`s, and there seem to be no reported problems with `Pod`s. So if you can rebase this, we can work on getting this merged soon.

brancz:
continued with #78

PR #60, opened 2016-12-20
rename resource_quota to resourcequota

brancz:
The convention of using an object kind in the metric name is to use the
same that would be used when using `kubectl`. So in this case `kubectl
get resourcequota`.

As discussed in the metric convention document.

@fabxc @matthiasr

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/60)


fabxc:
👍

PR #65, opened 2016-12-30
Added node name to container status queries and adapted tests

OlivierBoucher:
Hi, this is simply a useful addition to the container status gauge metrics. 

Background: I'm using Google Cloud's Stackdriver Monitoring and the entries require a bit more context.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/65)

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -51,27 +51,27 @@ var (
 	descPodContainerStatusWaiting = prometheus.NewDesc(
 		"kube_pod_container_status_waiting",
 		"Describes whether the container is currently in waiting state.",
-		[]string{"namespace", "pod", "container"}, nil,
+		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
 	descPodContainerStatusRunning = prometheus.NewDesc(
 		"kube_pod_container_status_running",
 		"Describes whether the container is currently in running state.",
-		[]string{"namespace", "pod", "container"}, nil,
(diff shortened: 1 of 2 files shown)


OlivierBoucher:
I signed it!

brancz:
I wasn't aware that Google Cloud Stackdriver understands the Prometheus format. This can potentially be a problem for Prometheus as every new combination of key/value pairs create a new time series. In Prometheus you can simply join the labels from the `kube_pod_info` metric onto the metrics you proposed to add them to. Would that be a possibility for you?

OlivierBoucher:
@brancz It does not, I'm running a server that does manual conversion / posting to Stackdriver.

```go
                timeseries.Resource = &monitoring.MonitoredResource{
			Labels: map[string]string{
				"project_id":     PROJECT_ID,
				"container_name": metric.Labels["container"],
				"pod_id":         metric.Labels["pod"],
				"namespace_id":   metric.Labels["namespace"],
				"instance_id":    metric.Labels["node"],
				"cluster_name":   clusterName,
				"zone":           zone,
			},
			Type: "gke_container",
		}
 ```

brancz:
You should be able to "join" the labels from the info metric with your `map[string]string`, right?

brancz:
Closing this as this should be handled by the querying system not by de-normalizing the output.

PR #66, opened 2017-01-06
Add flag to override resync period

dhawal55:
The hard-coded cache period of 5 mins is too long for us as we want to setup alerts using kube-state-metrics. Adding flag to override the cache/resync period.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/66)

### DIFF
--- a/main.go
+++ b/main.go
@@ -39,9 +39,8 @@ import (
 )
 
 const (
-	resyncPeriod = 5 * time.Minute
-	metricsPath  = "/metrics"
-	healthzPath  = "/healthz"
+	metricsPath = "/metrics"
+	healthzPath = "/healthz"
 )
 
 var (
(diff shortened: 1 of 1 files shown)


fabxc:
The resync periods after how long the entire cache state is refreshed.
However, in between all updates are retrieved via the watch API. So in the
general case the data should always be up to date.

Did you run into any issues with metrics actually not adding up with the
known cluster state?

dhawal55:
@fabxc Yes, I'm creating a pod, running a quick test and deleting it. However, kube_state_metrics reports the pod phase as 'pending' for 5 mins.

matthiasr:
Could this be the staleness handling in Prometheus itself? Could you try
manually inspecting the kube-state-metrics /metrics endpoint while doing
your test?

strantalis:
Just to let you know I was experiencing the same thing and took a look. It looks like it has to do with the Prometheus staleness handling. The kube-state-metric updates quickly but is stale in Prometheus for a few minutes.

dhawal55:
Thank you guys. Yes, it is indeed stale data in Prometheus. Closing this issue as I don't see any need to override the resyncPeriod

strantalis:
Also just a thought we have run into issues where we have had date/time issues on our prometheus servers that have caused the metrics to show up incorrectly.

Issue #74, opened 2017-01-26
[feature-request] Deployment Conditions


vdavidoff:
That'd be great if kube-state-metrics exposed Deployment Conditions. My use case is to easily know if a Deployment is in progress or not, which I plan to use along with Prometheus and an inhibit rule in AlertManager.

I see Deployment Conditions were added to the go k8s client around git commit ae6775eeec5cc9e96cc5cec848f589158acf7d92, and I have tried to add this support in myself by modifying kube-state-metrics, but I'm running into issues building against the newer go client. This is at least partly because I'm hacking at it right now - I don't really know golang or the process of golang development.

I'm happy to keep trying on this, but it's possible one of you will figure it out before I do.

vdavidoff:
Well, I read some stuff and got it to build with my changes and HEAD from go client. Tomorrow I'll see if stuff actually works as I think it should.

brancz:
I wanted to update to latest version of client-go as well, so if you can already do that in the process of implementing this feature, that would be great!

vdavidoff:
Well, I think my changes worked. But in the process of confirming they did what I wanted, I think I learned that the "Progressing" Deployment Condition doesn't work how I thought it would. I suspected it'd be removed from the Deployment Conditions once a rollout was complete, but it appears to exist in perpetuity. I think the following is how Kubernetes determines if a rollout is complete/successful (without ever removing the "Progressing" condition):

pkg/controller/deployment/util/deployment_util.go: 1009 (v1.5.1)
```
// DeploymentComplete considers a deployment to be complete once its desired replicas equals its
// updatedReplicas and it doesn't violate minimum availability.
func DeploymentComplete(deployment *extensions.Deployment, newStatus *extensions.DeploymentStatus) bool {
        return newStatus.UpdatedReplicas == deployment.Spec.Replicas &&
                newStatus.AvailableReplicas >= deployment.Spec.Replicas-MaxUnavailable(*deployment)
}
```

If I'm not wrong, then I think I just need to expose max unavailable replicas (for now at least, as `kube_deployment_spec_strategy_rollingupdate_max_unavailable`) from the Deployment spec. Once I have that, I can use it in conjunction with the following metrics that are already available via kube-state-metrics:
* kube_deployment_spec_replicas
* kube_deployment_status_replicas
* kube_deployment_status_replicas_available
* kube_deployment_status_replicas_unavailable
* kube_deployment_status_replicas_updated

[...]

PR #75, opened 2017-01-27
client-go update and new Deployment metric

vdavidoff:
* Update to the newest tagged client-go release, v2.0.0-alpha.1
* Expose Deployment RollingUpdate Strategy MaxUnavailable

@brancz showed interest in pulling in a new client-go version as mentioned in [issue 74](https://github.com/kubernetes/kube-state-metrics/issues/74). I chose the newest tag there, even though it's alpha, because the newest non-alpha tag (v1.5.0) is effectively a NOOP between the client-go ref committed to master, and is earlier than the major changes that restructure the directory hierarchy in client-go (the 1.4 and 1.5 subdirectories went away, and 1.5 effectively moved up a level). In other words, it made no sense to update to v1.5.0, and I preferred to pull in something that was at least tagged instead of master.

The new metric exposed is also discussed in [issue 74](https://github.com/kubernetes/kube-state-metrics/issues/74).

I think I got this right but if there are any questions, comments, or criticisms, I'm all ears.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/75)


vdavidoff:
I signed the CLA.

brancz:
lgtm 👍  thanks a lot!

PR #77, opened 2017-02-02
allow choosing which collectors to run

brancz:
Extract collector registration into each collector file and improve flag handling and configure glog.

This is backward compatible, but will allow us to potentially disable new collectors by default.

@alexsomesan @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/77)

### DIFF
--- a/daemonset.go
+++ b/daemonset.go
@@ -19,7 +19,11 @@ package main
 import (
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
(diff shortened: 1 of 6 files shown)


brancz:
@fabxc comments adapted and collectors are now validated when parsing rather than after communication with the apiserver is already established.

fabxc:
:+1:

PR #81, opened 2017-02-08
fix make test units with cgo

andyxning:
`go test -race` can not be run with cgo disabled. This will result `make test-unit` failed.

FYI:
* https://blog.golang.org/race-detector
* https://github.com/golang/go/issues/14481

/cc @brancz @vdavidoff

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/81)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,9 @@
 all: build
 
 FLAGS =
-ENVVAR = GOOS=linux GOARCH=amd64 CGO_ENABLED=0
+COMMONENVVAR = GOOS=linux GOARCH=amd64
+BUILDENVVAR = CGO_ENABLED=0
+TESTENVVAR = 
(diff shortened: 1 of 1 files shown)


brancz:
Yes I am aware of this, the only reason why we have this `Makefile` like this is because we were instructed to use exactly that by Google employees.

I entirely agree with this change, and have always run tests myself. Maybe @piosz can comment on the parts of the `Makefile` that are important for internals at Google for the release process.

andyxning:
@brancz @fabxc Currently this PR will make `make test-unit` work properly. Seems good. :)

piosz:
LGTM

andyxning:
friendly ping @brancz @vdavidoff

PR #92, opened 2017-02-22
health check and resource limits?

qrpike:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/92)

### DIFF
--- a/kubernetes/deployment.yml
+++ b/kubernetes/deployment.yml
@@ -2,6 +2,7 @@ apiVersion: extensions/v1beta1
 kind: Deployment
 metadata:
   name: kube-state-metrics-deployment
+  namespace: kube-system
 spec:
   replicas: 1
   template:
@@ -15,3 +16,14 @@ spec:
         image: gcr.io/google_containers/kube-state-metrics:v0.4.1
         ports:
         - containerPort: 8080
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 8
+          timeoutSeconds: 4
+        resources:
+          limits:
+            cpu: .1
+            memory: 75M


qrpike:
I signed it!

kubernetes/deployment.yml
```diff
@@ -15,3 +16,14 @@ spec:
...
+          httpGet:
+            path: /healthz
+            port: 8080
+            scheme: HTTP
+          initialDelaySeconds: 8
+          timeoutSeconds: 4
+        resources:
+          limits:
+            cpu: .1
+            memory: 75M
```
brancz: While CPU resource seems fine, as `kube-state-metrics` quite literally just lists/watches, I'm wondering how you came up this as an upper bound for memory, as this highly depends on number of objects in your cluster. I'm fine with giving a suggestion of memory limits, just wondering what kind of cluster size/object count we're talking about here.

qrpike: I just doubled/tripled what it was using in my cluster. Which is ~16nodes with ~200pods running. We can make it 100M to be safe? But it just needs some kind of limit to be safe.

brancz:
Other than my comment something with the CLA check still doesn't seem right. Make sure that if you signed it as an individual to use the same email as you did in your git config when you created the commit.

qrpike:
@brancz I signed it as an org, but I registered using this github account.

kubernetes/deployment.yml
brancz: `100M` limit sounds good to me then

qrpike:
@brancz I'm curious how much data is duplicated between this and heapster? Not apparent if this is meant to work along side, or replace/be replaced by the other?

brancz:
Heapster and kube-state-metrics are trying to solve different things. The sole purpose of `kube-state-metrics` is converting Kubernetes Objects into the Prometheus exposition format, while staying as close as possible to the data accessible when debugging things through `kubectl`. This data can then be scraped by Prometheus or Prometheus compatible systems to explore/aggregate/alert on that data.

Heapster is slowly but surely abstracted away by the Core/Custom Metrics API. Those will eventually allow pluggable solutions, so the idea is that there may be a drop in replacement for Heapster using Prometheus in the background. Regarding their differences Heapster as far as I know and can read from the [docs](https://github.com/kubernetes/heapster/blob/master/docs/overview.md) it only collects data about nodes, pods and containers, while kube-state-metrics generally collects cluster state metrics.

So to answer your question, for the time being you'll want to run both along side each other and they're all part of the puzzle.

Regarding the CLA I'll need to figure out what's happening as I can't merge if the check doesn't pass.

kubernetes/deployment.yml
andyxning: How about we change `cpu: .1` to `cpu: 100m`. This makes it more clear. :)

andyxning: BTW, maybe we should set resource configuration like what is done for [dashboard](https://github.com/kubernetes/kubernetes/blob/master/cluster/addons/dashboard/dashboard-controller.yaml) by setting both `limits` and `requests` to make its QoS guranteed.

brancz: I agree with both of @andyxning's comments.

brancz:
Have you been able to figure out what the problem with the cla is?

andyxning:
@qrpike Could you please take some time to resolve the CLA and update and rebase the code. :) We need this change.

andyxning:
close this pr via #238 . @qrpike Feel free  to reopen it if you have any doubts.

PR #94, opened 2017-02-28
fix service label to match deployment pod label

padolan:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/94)

### DIFF
--- a/kubernetes/service.yaml
+++ b/kubernetes/service.yaml
@@ -5,11 +5,11 @@ metadata:
     prometheus.io/scrape: 'true'
   name: kube-state-metrics
   labels:
-    k8s-app: kube-state-metrics
+    app: kube-state-metrics
 spec:
   ports:
   - name: http-metrics
     port: 8080
     protocol: TCP
   selector:
-    k8s-app: kube-state-metrics
+    app: kube-state-metrics


fabxc:
Thanks. That's a correct fix of course.
Now the only question is whether it should be fixed like this or by making the deployment have the label `k8s-app`. I like to think that kube-state-metrics should be part of every production cluster and thus is an `k8s-app`, but more opinions on this would help.

@brancz 

@piosz do you know more on whether there's a clear definition on what falls under "k8s-app"?

piosz:
I don't know. We had similar discussion here https://github.com/kubernetes/contrib/pull/2420#discussion_r103153127.

@bgrant0607 what is the semantic/policy behind `k8s-app`?

brancz:
I agree with rather changing the label in the deployment manifest, as @fabxc mentioned kube-state-metrics is a cluster component.

@fabxc @piosz if it turns out there is no semantic/policy in place I think this one is worth formalizing.

bgrant0607:
@piosz Add-ons in the kube-system namespace use the k8s-app label key.

https://github.com/kubernetes/kubernetes/search?p=2&q=k8s-app&utf8=%E2%9C%93

piosz:
This is the current state. The question is whether it's the standard?

brancz:
#98 instead fixed the `Pod`'s label from `app` to `k8s-app` so it matches the `Service` object. As discussed, kube-state-metrics is a cluster add-on that is aimed to reach a stable state for the 1.7 release. Being a cluster add-on justifies the label to be `k8s-app` instead of `app`. Therefore closing this in favor of #98.

PR #96, opened 2017-03-03
Fix the deployment definition for v0.4.1

dblackdblack:
The deployment .yaml for v0.4.1 called out the docker image with tag 0.3.0. You'll probably have to bump version to 0.4.2 in order to fix.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/96)

### DIFF
--- a/kubernetes/deployment.yml
+++ b/kubernetes/deployment.yml
@@ -8,10 +8,10 @@ spec:
     metadata:
       labels:
         app: kube-state-metrics
-        version: "v0.3.0"
+        version: "v0.4.1"
     spec:
       containers:
       - name: kube-state-metrics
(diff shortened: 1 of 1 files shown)


dblackdblack:
I'm not signing this thing for such a trivial fix.  Feel free to reject this PR but please implement the fix.

andyxning:
@dblackdblack In case you will further contribute to other cncf projects, it would be great to sign the cla. :)

brancz:
I realize that this was a mistake on our side, but the deployment is up to date on master, what is the problem with grabbing the manifest from there?

What I'm saying is that I don't think this change justifies a release.

PR #100, opened 2017-03-07
Clarified that prometheus can use heapster as scrape target.

### LINKED ISSUE #99: The explanation for `kube-state-metrics-vs-heapster` is not accurate
In https://github.com/kubernetes/kube-state-metrics#kube-state-metrics-vs-heapster , it  claims that `some monitoring systems such as Prometheus do not use Heapster for metric collection at all and instead implement their own`.

I think that this may be not accurate, as actually, I can always use Heapster as a prometheus scrape targets, right?

gyliu513:
Fix #99

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/100)


README.md
```diff
@@ -137,8 +137,8 @@ anywhere.
 In other words, kube-state-metrics itself is designed to be another source for
 Heapster (although this is not currently the case).
 
-Additionally, some monitoring systems such as Prometheus do not use Heapster
-for metric collection at all and instead implement their own. Having
+Additionally, some monitoring systems such as Prometheus can use both Heapster
+and kube-state-metrics as scrape target for metric collection. Having
```
brancz: As far as I know heapster (as of today) does not expose a metrics endpoint for Prometheus to scrape so this is also not 100% correct.

What @fabxc was talking about on the mailing list was, that there are reasons to run both Prometheus and Heapster in your cluster. Heapster is required for autoscaling with the horizontal pod autoscaler. Heapster mainly collects lower level container metrics. Prometheus is for all kinds of metrics and it also has alerting features.

So you want to actually run both in your cluster today:

* Heapster for metric collection and autoscaling
* Prometheus for metric collection, alerting and dashboarding

Eventually as the metrics API becomes available in Kubernetes, we will be able to get rid of Heapster and replace it with a Prometheus based proxy that will be drop-in.

/cc @fabxc

gyliu513: @brancz but here the heapster does mention that it can support prometheus https://github.com/kubernetes/heapster/blob/master/docs/debugging.md#debuging , comments?

```
`/metrics` contains lots of metrics in Prometheus format that can indicate the root cause of Heapster problems. Example:
```

brancz: Yes, but as far as I understand that does not expose the state of the metrics that heapster itself collects, but rather only it's own state, that Prometheus could alert on.

gyliu513: Thanks @brancz , clear now ;-)

README.md
```diff
@@ -122,25 +122,29 @@ additional metrics!
...
 
 In other words, kube-state-metrics itself is designed to be another source for
 Heapster (although this is not currently the case).
 
-Additionally, some monitoring systems such as Prometheus do not use Heapster
-for metric collection at all and instead implement their own. Having
-kube-state-metrics as a separate project enables access to these metrics from
-those monitoring systems.
+Additionally, some monitoring systems such as Prometheus do not use Heapster for
+metric collection at all and instead implement their own, but the Prometheus can
```
brancz: but Prometheus can scrape metrics from heapster itself to alert on heapster's health

gyliu513: Done, thanks @brancz

brancz:
Nice! lgtm :+1: Thanks!

PR #105, opened 2017-03-14
kube_pod_info add controller info 

zouyee:
It's necessary for kube_pod_info to keep controller info
/cc @fabxc 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/105)


pod.go
```diff
@@ -152,6 +155,18 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerResourceLimitsMemoryBytes
 }
 
+func printControllers(annotation map[string]string) string {
+	value, ok := annotation[api.CreatedByAnnotation]
+	if ok {
+		var r api.SerializedReference
+		err := json.Unmarshal([]byte(value), &r)
```
mxinden: Maybe the Kubernetes Decoder would be better: "k8s.io/apimachinery/pkg/util/yaml" with function "NewYAMLOrJSONDecoder" as this JSON was marshaled by the Kubernetes encoder.

pod.go
```diff
@@ -152,6 +155,18 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerResourceLimitsMemoryBytes
 }
 
+func printControllers(annotation map[string]string) string {
```
mxinden: Could we rename this to `extractCreatedBy`?

pod_test.go
```diff
@@ -280,8 +280,8 @@ func TestPodCollector(t *testing.T) {
 				},
 			},
 			want: metadata + `
-				kube_pod_info{host_ip="1.1.1.1",namespace="ns1",pod="pod1",node="node1",pod_ip="1.2.3.4"} 1
-				kube_pod_info{host_ip="1.1.1.1",namespace="ns2",pod="pod2",node="node2",pod_ip="2.3.4.5"} 1
+				kube_pod_info{controllers="<none>",host_ip="1.1.1.1",namespace="ns1",pod="pod1",node="node1",pod_ip="1.2.3.4"} 1
```
mxinden: As this can only be a single controller we should rename this to `controller`.

mxinden:
@zouyee Thank you very much for the contribution.

pod_test.go
fabxc: Why do we name it controller to begin with? The annotation says `created_by` and not `controller`. Just for people to intuitively understand what it was populated by, I'd avoid introducing name mappings.

Anyone knows whether that annotation is scheduled to move into the main YAML and if, whether it will keep the name?

brancz: Agreed with `created_by`. I'm unsure but I felt like the `ownerReference` has similar if not the same information. Although by my understanding the `ownerReference` is used by the GC so if that is not the case for the `created_by` annotation then that gives the chance of writing things like the actual controller in there as in the statefulset controller or the replicaset controller created these, not a direct relation to an object in Kubernetes.

pod.go
zouyee: Though CreatedByAnnotation represents the key used to store the spec(json)
used to create the [resource](https://github.com/kubernetes/client-go/blob/65ac4e638a79d49b7dbd5e1cef4e7785af7cbd39/pkg/api/helpers.go#L440) ,controllers named come from `kubeclt describe po xxx`

pod.go
zouyee: @fabxc thank you for your response! Initially,I also confused with `controllers`, but in the returned result of `kubectl describe po pod-name` [describe code](https://github.com/kubernetes/kubernetes/blob/c38717b73ae5c35f91d2e30bf0470575c09ad9d0/pkg/printers/internalversion/describe.go#L541),according to [code](https://github.com/kubernetes/kubernetes/blob/1d97472361e61106a1da1ce33da97f243e4d0441/pkg/client/listers/core/internalversion/replicationcontroller_expansion.go#L37),controllers refer to which one actually manage the pod

pod_test.go
zouyee: @brancz  Though CreatedByAnnotation represents the key used to store the spec(json)
used to create the resource ,controllers named come from` kubeclt describe po xxx`,maybe we need to think about the consistency with other projects

zouyee:
@mxinden

brancz:
lgtm

brancz:
@zouyee can you remove the limit range commits from this PR?

zouyee:
👌

zouyee:
hi @brancz ,I have revert my commit, so sorry for my mistake

PR #104, opened 2017-03-09
replicationcontrollers are part of core not extentions

brancz:
Fixes #101 .

@fabxc 

/cc @tmegow

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/104)

PR #106, opened 2017-03-16
fix rc api namespace

andyxning:
`replicationcontrollers` resources is under `api/v1` namespace instead of `apis/extensions/v1beta1`. Thus we need to use `CoreV1()`. :)

/ping @dominikschulz @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/106)


brancz:
Dupe of #104, so closing there as this was merged.

PR #107, opened 2017-03-20
add limit range

zouyee:
add limit range metrics
@brancz 
cc/ @fabxc 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/107)

### DIFF
--- a/README.md
+++ b/README.md
@@ -71,7 +71,7 @@ additional metrics!
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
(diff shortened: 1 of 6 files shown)


brancz:
Can you add the readme changes from #105 in here? I think you may have accidentally committed them there.

zouyee:
I have added limit range metrics info in readme,thank you @brancz

brancz:
You need to add the collector to the available collectors, and possibly the default collectors. Otherwise it won't be able to be enabled.

https://github.com/kubernetes/kube-state-metrics/blob/1751d86b25463c60234a37753024c7122d39cc26/main.go#L53-L61

PR #110, opened 2017-03-24
kubernetes: add complete set of manifests

brancz:
With kubernetes 1.6 RBAC is enabled by default, so we should make sure that the manifests provided here follow the best practice. Therefore I added a full set of manifests from ServiceAccount, over Roles and Deployment so a user can really just do: `kubectl apply -f kubernetes/` to deploy on 1.6.

This uses the v1alpha1 objects of the RBAC API, which is still present in 1.6 but it likely subject for removal in following releases. But for now they still work, so once there is a stable release of 1.6 we can change these to the v1beta1 objects. For the time being these objects are compatible with 1.5 clusters as well.

@piosz @fabxc @mxinden

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/110)


kubernetes/kube-state-metrics-service.yaml
```diff
@@ -1,15 +1,16 @@
 apiVersion: v1
 kind: Service
 metadata:
-  annotations:
-    prometheus.io/scrape: 'true'
```
deads2k: The readme says this exists.  I don't know why, but its there.

kubernetes/kube-state-metrics-cluster-role.yaml
```diff
@@ -8,6 +8,7 @@ rules:
   - nodes
   - pods
   - resourcequotas
+  - replicationcontrollers
   verbs: ["list", "watch"]
 - apiGroups: ["extensions"]
```
deads2k: Deployments also exists in `apps` now. No need to be this pull, but you're likely to transition at some point.  For some period (if you support both), you can list both in your role

kubernetes/kube-state-metrics-cluster-role-binding.yaml
```diff
@@ -0,0 +1,12 @@
...
+metadata:
+  name: kube-state-metrics
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: ClusterRole
+  name: kube-state-metrics
+subjects:
+- kind: ServiceAccount
+  name: kube-state-metrics
+  namespace: kube-system
```
deads2k: This makes the resources it's provisioning for only valid in the `kube-system` namespace.  Including the namespace in the other resources may save you questions later.

deads2k:
The role looks good and using a specific service account is great.

kubernetes/kube-state-metrics-cluster-role.yaml
brancz: Currently we [explicitly use](https://github.com/kubernetes/kube-state-metrics/blob/042a59eb73005b30f506f312f01da7a88988c315/deployment.go#L90) the extensions API. But I wasn't aware of this so thanks for the heads up!

kubernetes/kube-state-metrics-cluster-role-binding.yaml
brancz: Good point! Will adapt.

kubernetes/kube-state-metrics-service.yaml
brancz: True, I'll add this back. It seems that the community has moved away from this use, but if not present it may confuse users who do expect it.

brancz: readded

kubernetes/kube-state-metrics-cluster-role-binding.yaml
brancz: adapted

brancz:
Addressed your comments @deads2k also added a readiness check.

deads2k:
permissions bits lgtm.

piosz:
Thanks @deads2k for the review!

fabxc:
👍

PR #117, opened 2017-03-29
fix start_time in kube_pod_info

zouyee:
@brancz hi,I found start_time label of kube_pod_info in readme that is not be implemented.

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/117)

### DIFF
--- a/Documentation/node-metrics.md
+++ b/Documentation/node-metrics.md
@@ -3,6 +3,7 @@
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
 | kube_node_info | Gauge | `node`=&lt;node-address&gt; <br> `kernel_version`=&lt;kernel-version&gt; <br> `os_image`=&lt;os-image-name&gt; <br> `container_runtime_version`=&lt;container-runtime-and-version-combination&gt; <br> `kubelet_version`=&lt;kubelet-version&gt; <br> `kubeproxy_version`=&lt;kubeproxy-version&gt; |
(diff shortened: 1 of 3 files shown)


zouyee:
gentle ping @brancz

brancz:
I'm wondering what kind of a query/alert you would want to do with this as it should be equivalent to the first samples through scraping.

pod.go
```diff
@@ -191,8 +191,12 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 	addCounter := func(desc *prometheus.Desc, v float64, lv ...string) {
 		addConstMetric(desc, prometheus.CounterValue, v, lv...)
 	}
+	for _, containerStatus := range p.Status.Conditions {
+		if containerStatus.Type == "Initialized" {
+			addGauge(descPodInfo, 1, p.Status.HostIP, p.Status.PodIP, containerStatus.LastTransitionTime.String(), nodeName, extractCreatedBy(p.Annotations))
```
andyxning: @zouyee 

* Does the `Initialized` status indicates that the pod has started?
* IIUC, there should only be one `Initialized` condition, so we can `break` once we find it.

andyxning: According to [client-go](https://github.com/kubernetes/client-go/blob/master/pkg/api/types.go#L1684-L1688),
`Initialized` indicates that `all init containers in the pod have started successfully`. `Ready` represents that `the pod is able to service requests and should be added to the load balancing pools of all matching services'.

PR #126, opened 2017-04-22
Expose daemonset status number ready gauge

### LINKED ISSUE #125: Add support for ready metrics for daemonset
Currently, the metrics exported for daemonset doesn't allow to detect that a damonset pod is not ready.
There is not equivalent of kube_replicaset_status_ready_replicas.

Would it be possible to add support for a similar metric for daemonset ?

Thanks in advance.

Yann

nvartolomei:
Closes #125 

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/126)

### DIFF
--- a/daemonset.go
+++ b/daemonset.go
@@ -42,6 +42,11 @@ var (
 		"The number of nodes that should be running the daemon pod.",
 		[]string{"namespace", "daemonset"}, nil,
 	)
+	descDaemonSetNumberReady = prometheus.NewDesc(
+		"kube_daemonset_status_number_ready",
+		"The number of nodes that should be running the daemon pod and have one or more of the daemon pod running and ready.",
+		[]string{"namespace", "daemonset"}, nil,
+	)
 	descDaemonSetMetadataGeneration = prometheus.NewDesc(
 		"kube_daemonset_metadata_generation",
 		"Sequence number representing a specific generation of the desired state.",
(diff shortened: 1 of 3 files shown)


daemonset.go
```diff
@@ -42,6 +42,11 @@ var (
 		"The number of nodes that should be running the daemon pod.",
 		[]string{"namespace", "daemonset"}, nil,
 	)
+	descDaemonSetNumberReady = prometheus.NewDesc(
+		"kube_daemonset_status_number_ready",
+		"The number of nodes that should be running the daemon pod and have one or more of the daemon pod running and ready.",
```
andyxning: How about `The number of nodes running at least on daemon pod and are ready`

daemonset_test.go
```diff
@@ -43,6 +43,8 @@ func TestDaemonSetCollector(t *testing.T) {
 		# TYPE kube_daemonset_status_number_misscheduled gauge
 		# HELP kube_daemonset_status_desired_number_scheduled The number of nodes that should be running the daemon pod.
 		# TYPE kube_daemonset_status_desired_number_scheduled gauge
+		# HELP kube_daemonset_status_number_ready The number of nodes that should be running the daemon pod and have one or more of the daemon pod running and ready.
```
andyxning: ditto

daemonset.go
brancz: Agreed

brancz:
other than the help text lgtm

nvartolomei:
The HELP text is a copy of [code doc from k8s](https://github.com/kubernetes/kube-state-metrics/blob/7b6eb5ab819d271886daae83d5ea784724c01a48/vendor/k8s.io/client-go/pkg/apis/extensions/v1beta1/types.go#L564-L566).

Also from the code doc, I drew a conclusion that number ready does not include misscheduled pods. This is why I left _The number of nodes that should be running the daemon pod_ part.

Let me know your final decision.

brancz:
Understood. Lgtm

andyxning:
Sounds reasonable.

PR #130, opened 2017-04-28
Add Pod/Node Labels metric

### LINKED ISSUE #108: kube_node_info should add node labels
We run several node pools and use labels to differentiate them (as well as taints). One node pool in particular is built to be a canary for future versions of our underlying OS. I'd like to be able to build an alert the fires if that node goes away (assumably because some change in the lower env causes it's kubelet to fail). 

Being able to to say sum(kube_node_info{mynodelabel="canary"}) == 0 would be awesome.

brancz:
Closes #108

@fabxc @mxinden

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/130)

### DIFF
--- a/pod.go
+++ b/pod.go
@@ -18,8 +18,9 @@ package main
 
 import (
 	"encoding/json"
- 	"fmt"
-	
+	"fmt"
+	"regexp"
+
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/net/context"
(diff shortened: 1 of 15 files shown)


pod_test.go
```diff
@@ -468,29 +470,48 @@ func TestPodCollector(t *testing.T) {
...
+						Namespace: "ns1",
+						Labels: map[string]string{
+							"app": "example",
+						},
+					},
+					Spec: v1.PodSpec{},
+				},
+			},
+			want: metadata + `
+				kube_pod_labels{kube_label_app="example",namespace="ns1",pod="pod1"} 1
```
fabxc: Not prefixing the label names is unfortunately going to cause problems down the road – but maybe we can cut it down to just be `label_`. The metric itself is already prefixed with `kube_`.

brancz:
@fabxc adapted according to your comment and added an info and label metric for `Service`s.

fabxc:
👍

pod.go
```diff
@@ -179,6 +201,31 @@ func (pc *podCollector) Collect(ch chan<- prometheus.Metric) {
...
+	}
+	return labelKeys, labelValues
+}
+
+func sanitizeLabelName(s string) string {
+	return invalidLabelCharRE.ReplaceAllString(s, "_")
+}
+
+func podLabelsDesc(labelKeys []string) *prometheus.Desc {
+	return prometheus.NewDesc(
```
andyxning: @brancz It seems that prometheus will use `fqName`(`descPodLabelsName`) and `help`(`descPodLabels`) to uniquely define an metric. Once we first register it, the labels can be modified on the fly. Am i right?

andyxning: Any docs about this behavior?

brancz: This is actually just a validation that the go client library performs that we turned off. Metrics/time-series are _only_ identified by their unique label-set. The description registering was a premature validation attempt that is too hard of a validation in cases like this, although typically a correctly validation. So tl;dr yes you're right.

Issue #72, opened 2017-01-22
[feature-request] Add metrics from CronJobs


mikekap:
Most important for monitoring: last successful run time. Last run time, duration, overrun indication, etc would be awesome too :).

CronJobs are still in alpha though, so it might be good to wait for them to at least make it to beta.

fcantournet:
Might also be worth having the PodStatuses for Jobs ( for the number of failures )

jhorwit2:
Cronjobs are being promoted to beta in 1.6 https://github.com/kubernetes/features/issues/19

this would be super nice to have

igniting:
Is there any progress on this?

brancz:
I am currently not aware of anyone working on this, if anyone wants to grab it, go head! :slightly_smiling_face:

vdavidoff:
I have started working on exposing Job metrics. I have a dev build working, I think, though it hasn't been thoroughly tested to ensure there are no edge cases (or cases simply unlike the state of my test clusters) where the code blows up. I haven't started on the unit tests, either. I'm about to start looking at CronJobs to see what can/should be exposed there.

Quite frankly I'm sort of hacking at this. It's going well but if anyone is interested in reviewing code and/or helping write the unit tests, that'd be swell. Otherwise I'll try to get it all done. I'm working here: https://github.com/vdavidoff/kube-state-metrics

vdavidoff:
I have Job and CronJob metrics working. I think the metrics I'm exposing, and how I'm exposing them, makes sense, but feedback on that as well as coding choices etc. is most welcome. In fact I highly recommend it.

I can assure you that someone who is more familiar with the tests than I am will be much quicker at getting the _test files done. My plan is to do them as I have time, but I don't know how long that'll be, and I realize that nothing I have done will get merged until those are complete. Or, I assume that's the case, at least.

brancz:
@vdavidoff feel free to open a PR already, that exactly the situation what they're for :slightly_smiling_face:. Reviews aren't there to make anyone feel bad or unqualified, we can all learn from them and we can track progress a lot better. There is no need for it to be completely done, it's perfectly fine to be work in progress.

If you can't manage to finish the unit tests, but the rest of the implementation is more or less done, we may just pick it up from where you left off - no work is wasted work. Thanks a lot for your effort!

vdavidoff:
https://github.com/kubernetes/kube-state-metrics/pull/131

brancz:
Closed by #131

PR #131, opened 2017-05-02
Expose Job and CronJob metrics

vdavidoff:
This is meant to address https://github.com/kubernetes/kube-state-metrics/issues/72

These changes aren't as baked as I'd normally want them to be before submitting a PR, but I was encouraged to go ahead and submit by @brancz, so here you go. I can confirm that I am able to build and run this code, apparently without any issues.

I am not a golang programmer and admittedly sort of reverse engineered what I have done here. I based both of these new modules (perhaps the wrong term) on deployment.go. Aside from the missing test files, the things I know might need to change are:
* For metics that are statuses like "failed", or "succeeded", I am just exposing one metric with the value of 0 or 1 (~boolean) rather than exposing each condition twice with a "condition" field of "true" or "false" and a value that indicates if the condition is true. I saw the use of a "condition" field in some other metrics, but it just seems unnecessary. I realize there may be a good reason why I should have used a condition field instead.
* I am not exporting "info" metrics (e.g. kube_pod_info) because I don't think they're necessary, but I'm not opposed to adding those if I should.
* I used metric type Counter for values that are seconds since the epoch, because that seemed to make more sense to me than gauge, since the value will only ever increase.
* When cycling through Job conditions, the code assumes that it will only ever see a Failed or Complete condition, and if it does, it can safely make assumptions (i.e. if the first condition is Failed, it can assume the succeeded metric should be 0 and failed should be 1). I think this is safe, but I may not understand why it's not.
* I'm using the extractCreatedBy function that's defined in pod.go in my job.go. This feels kind of weird to me since one file's code is reaching into another's without any explicit pointer to the other file, but maybe I am just not used to golang.
* Following existing examples, the name of the Job would be exposed in the "job" field, but Prometheus exports its own "job" field related to the job that scraped the metrics, which would result in our Job field being named "exported_job". To avoid this I named our field "job_name".

Please let me know what you'd like to see change here, and I'll happily make those changes.

Thanks.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/131)

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -14,7 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
-package main
+package collectors
 
 import (
 	"strconv"
@@ -125,6 +125,9 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 		string(j.Spec.ConcurrencyPolicy))
 
 	addGauge(descCronJobStatusActive, float64(len(j.Status.Active)))
-	addCounter(descCronJobStatusLastScheduleTime, float64(j.Status.LastScheduleTime.Unix()))
 	addGauge(descCronJobSpecSuspend, boolFloat64(*j.Spec.Suspend))
+
+	if j.Status.LastScheduleTime != nil {
+		addCounter(descCronJobStatusLastScheduleTime, float64(j.Status.LastScheduleTime.Unix()))
+	}
 }
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -17,8 +17,6 @@ limitations under the License.
 package collectors
 
 import (
-	"strconv"
-
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
 	"golang.org/x/net/context"
@@ -32,7 +30,7 @@ var (
 	descCronJobInfo = prometheus.NewDesc(
 		"kube_cronjob_info",
 		"Info about cronjob.",
-		[]string{"namespace", "cronjob", "schedule", "starting_deadline_seconds", "concurrency_policy"}, nil,
+		[]string{"namespace", "cronjob", "schedule", "concurrency_policy"}, nil,
 	)
 	descCronJobStatusActive = prometheus.NewDesc(
 		"kube_cronjob_status_active",
@@ -47,6 +45,11 @@ var (
 	descCronJobSpecSuspend = prometheus.NewDesc(
 		"kube_cronjob_spec_suspend",
 		"Suspend flag tells the controller to suspend subsequent executions.",
+		[]string{"namespace", "cronjob"}, nil,
+	)
+	descCronJobSpecStartingDeadlineSeconds = prometheus.NewDesc(
+		"kube_cronjob_spec_starting_deadline_seconds",
+		"Deadline in seconds for starting the job if it misses scheduled time for any reason.",
 		[]string{"namespace", "cronjob"}, nil,
 	)
 )
@@ -88,6 +91,7 @@ func (dc *cronJobCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descCronJobStatusActive
 	ch <- descCronJobStatusLastScheduleTime
 	ch <- descCronJobSpecSuspend
+	ch <- descCronJobSpecStartingDeadlineSeconds
 }
 
 // Collect implements the prometheus.Collector interface.
@@ -112,17 +116,11 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 		ch <- prometheus.MustNewConstMetric(desc, prometheus.CounterValue, v, lv...)
 	}
 
-	var startingDeadlineSeconds string
-	if j.Spec.StartingDeadlineSeconds == nil {
-		startingDeadlineSeconds = ""
-	} else {
-		startingDeadlineSeconds = strconv.FormatInt(*j.Spec.StartingDeadlineSeconds, 10)
+	if j.Spec.StartingDeadlineSeconds != nil {
+		addGauge(descCronJobSpecStartingDeadlineSeconds, float64(*j.Spec.StartingDeadlineSeconds))
 	}
 
-	addGauge(descCronJobInfo, 1,
-		j.Spec.Schedule,
-		startingDeadlineSeconds,
-		string(j.Spec.ConcurrencyPolicy))
+	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
 
 	addGauge(descCronJobStatusActive, float64(len(j.Status.Active)))
 	addGauge(descCronJobSpecSuspend, boolFloat64(*j.Spec.Suspend))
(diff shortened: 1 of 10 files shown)


brancz:
> For metics that are statuses like "failed", or "succeeded", I am just exposing one metric with the value of 0 or 1 (~boolean) rather than exposing each condition twice with a "condition" field of "true" or "false" and a value that indicates if the condition is true. I saw the use of a "condition" field in some other metrics, but it just seems unnecessary. I realize there may be a good reason why I should have used a condition field instead.

As the Jobs status fields for failed, succeeded and active are actually different from the other condition metrics, they actually don't represent a true/false condition, but rather how many jobs of this kind are active or have failed/succeeded. So the way you currently expose them is perfect.

> I am not exporting "info" metrics (e.g. kube_pod_info) because I don't think they're necessary, but I'm not opposed to adding those if I should.

This is something you should add and remove the duplicate `created_by` label on the other metrics. This metric can then be joined onto the other metrics therefore adding the label when necessary, but not unnecessarily having the extra label on all metrics.

> I used metric type Counter for values that are seconds since the epoch, because that seemed to make more sense to me than gauge, since the value will only ever increase.

Sounds reasonable. Other than that metrics types are currently not handled in any special way in Prometheus (although this may change at some point).

> Following existing examples, the name of the Job would be exposed in the "job" field, but Prometheus exports its own "job" field related to the job that scraped the metrics, which would result in our Job field being named "exported_job". To avoid this I named our field "job_name".

> When cycling through Job conditions, the code assumes that it will only ever see a Failed or Complete condition, and if it does, it can safely make assumptions (i.e. if the first condition is Failed, it can assume the succeeded metric should be 0 and failed should be 1). I think this is safe, but I may not understand why it's not.

The goal of kube-state-metrics is to 100% reflect the state of the objects received from the Kubernetes API, therefore it would be better to leave out any additional logic that makes assumptions about what certain fields mean. Have a look at how node conditions are handled for example and you can pretty much copy that 1 to 1 for the possible conditions for Jobs.

> I'm using the extractCreatedBy function that's defined in pod.go in my job.go. This feels kind of weird to me since one file's code is reaching into another's without any explicit pointer to the other file, but maybe I am just not used to golang.

That is perfectly fine.

> Following existing examples, the name of the Job would be exposed in the "job" field, but Prometheus exports its own "job" field related to the job that scraped the metrics, which would result in our Job field being named "exported_job". To avoid this I named our field "job_name".

kube-state-metrics in many places overlaps with labels that people likely are adding to their targets, which is why we recommend running it with the `honor_labels` configuration in Prometheus, which does not override those, therefore this should actually be `job` instead of `job_name`.

---

Only a few comments, overall not a whole lot. Thanks a lot for your effort! Once you apply those comments and add some tests I think this should be good to merge. Thanks a again!

vdavidoff:
Thank you for the detailed feedback. I have pushed a commit that I think addresses everything you hit on, plus fixes a few bugs I found. Details are in the commit message.

At this point I think the new/remaining questions I have are:
* For values which may be missing (CronJob startingDeadlineSeconds, Job completionTime, Job activeDeadlineSeconds) I'm setting the label value to the string literal "<none>" if the value is missing, which I saw in some preexisting metrics. I don't know if that's always right.
* I folded CronJob completions, startTime and completionTime into kube_cronjob_info, but I wonder if there's an argument for having any of those (namely the times) be their own metrics.
* I have explicitly protected some potential nil pointer dereferences I discovered, but I wonder if there are any other references I'm making when populating metics that might be empty or missing and cause similar issues.

brancz:
> For values which may be missing (CronJob startingDeadlineSeconds, Job completionTime, Job activeDeadlineSeconds) I'm setting the label value to the string literal "" if the value is missing, which I saw in some preexisting metrics. I don't know if that's always right.

That's fine as Prometheus throws out empty labels.

> I folded CronJob completions, startTime and completionTime into kube_cronjob_info, but I wonder if there's an argument for having any of those (namely the times) be their own metrics.

Interesting idea, I'd ask @fabxc what he thinks here. Do you think it's worth having these as metrics to make them queryable?

> I have explicitly protected some potential nil pointer dereferences I discovered, but I wonder if there are any other references I'm making when populating metics that might be empty or missing and cause similar issues.

I'll give it a thorough code review once we solve all other points :slightly_smiling_face: .

> I wonder if CronJob lasteScheduleTime (kube_cronjob_status_last_schedule_time) should be folded into kube_cronjob_info

Either way we should be consistent with the other time values, so let's see what @fabxc has to say about the second point.

fabxc:
Timestamp metrics can make a lot of sense and do here too, I think. One could is it for alerts like

```
ALERT CronJobNotRunning
  IF time() - kube_cronjob_status_last_schedule_time > 3600 * 24
  ANNOTATIONS {
    description = "cron job {{ $labels.cronjob }} did not run in the last 24 hours"
  }
```

brancz:
So putting all timestamps for these in metric values is the way to go here. This will also need a rebase @vdavidoff due to recent restructuring, sorry about that. But generally looks sane!

vdavidoff:
Thanks. I have pushed another set of changes rebased against master.

vdavidoff:
..and 4realz. This PR should now be rebased and all up to date.

collectors/job.go
```diff
@@ -0,0 +1,169 @@
...
+	"k8s.io/client-go/pkg/api"
+	v1batch "k8s.io/client-go/pkg/apis/batch/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descJobInfo = prometheus.NewDesc(
+		"kube_job_info",
+		"Information about job.",
+		[]string{"namespace", "job", "parallelism", "completions", "active_deadline_seconds"}, nil,
```
brancz: parallelism, completions and active_deadline_seconds should be their own metrics

collectors/cronjob.go
```diff
@@ -0,0 +1,133 @@
...
+	"k8s.io/client-go/pkg/api"
+	v2batch "k8s.io/client-go/pkg/apis/batch/v2alpha1"
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descCronJobInfo = prometheus.NewDesc(
+		"kube_cronjob_info",
+		"Info about cronjob.",
+		[]string{"namespace", "cronjob", "schedule", "starting_deadline_seconds", "concurrency_policy"}, nil,
```
brancz: starting_deadline_seconds should be its own metric

vdavidoff:
There you go. Thanks.

brancz:
Alright, that all looks good! Now just some tests and it's good to go! All the other collectors are thoroughly tested, you can guide yourself with those. It should be pretty straight forward, but if not let me know and I can point you in the right direction.

vdavidoff:
FYI I have not dropped this, I have just been too busy to work on it.

brancz:
No worries @vdavidoff!

vdavidoff:
I have committed a first shot at the tests. To continue the frankness, let me state that I started with `deployment_test.go` and hacked my way forward. I may have inadvertently cargo-culted some stuff from `deployment_test.go`. Also, I think my tests are testing sufficient cases, but maybe you think I missed some.

brancz:
lgtm, huge thansk for taking care of this and thanks for baring with me!

vdavidoff:
Thanks!

PR #132, opened 2017-05-03
*: cut v0.5.0

brancz:
It's been almost 3 months since the last release and we have a bunch of new features that were added since the last release, so we should do another release.

@fabxc @mxinden @andyxning 

/cc @piosz (to push the new image once tagged)

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/132)

PR #133, opened 2017-05-03
fix service and limitrange rbac roles and fix order and naming

brancz:
I noticed an inconsistency of the name of the LimitRange collector. In contrast to all the other collectors the LimitRange collector was singular. In addition I ordered the collectors in lexicographical order for ease to sync the list of available and default collectors.

In addition to that the default provided RBAC roles did not contain the recently added Services as well as LimitRange resources.

@fabxc @mxinden @andyxning 

This should be merged before #132 so we don't introduce a collector breaking change as the LimitRange collector has not been released yet.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/133)

### DIFF
--- a/kubernetes/kube-state-metrics-cluster-role.yaml
+++ b/kubernetes/kube-state-metrics-cluster-role.yaml
@@ -7,8 +7,10 @@ rules:
   resources:
   - nodes
   - pods
+  - services
   - resourcequotas
   - replicationcontrollers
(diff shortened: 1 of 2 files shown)


fabxc:
👍

andyxning:
👏 

/lgtm

PR #134, opened 2017-05-03
update image tags

brancz:
We'll primarily distribute the kube-state-metrics container image via quay from now on as none of the kube-state-metrics maintainers have the ability to update the image on gcr (but we do for quay) and therefore it creates an unnecessary hassle when releasing.

@fabxc @mxinden @andyxning

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/134)

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,7 +18,7 @@ the raw metrics.
 
 ## Container Image
 
-The latest container image can be found at `gcr.io/google_containers/kube-state-metrics:v0.4.1`.
+The latest container image can be found at `gcr.io/google_containers/kube-state-metrics:v0.5.0`.
 
 ## Metrics
 
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -13,7 +13,7 @@ spec:
       serviceAccountName: kube-state-metrics
       containers:
       - name: kube-state-metrics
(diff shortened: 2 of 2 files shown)


andyxning:
I am upvoting that we need to find a better and easy way to distribute kube-state-metrics images. However, changing from gcr to quay seems to be a breaking change for user who has an assumption that all kubernetes main components should be available on gcr. 

If we need to do the change, i suggest we make a statement about the transition in readme clearly. :)

zouyee:
👍

brancz:
Google is planning to open up the release process so we'll keep the primary source to be gcr for now, so I just updated the tags now and will ping Google people to take care of publishing the image.

fabxc:
👍

cemo:
It is still at not there. By the way why docker hub is not considered in the first place?

brancz:
We've always distributed the image via gcr, it was the default when the project started as it is the natural choice for a somewhat google scoped project. FWIW we (CoreOS) are also maintaining images of kube-state-metrics hosted on quay.io, that already has the v0.5.0 image. https://quay.io/repository/coreos/kube-state-metrics?tag=latest&tab=tags

Other than that Google has a goal of opening up the release process for the Kubernetes 1.7 release, which is supposedly going to allow us to get the images up more quickly. The current process is to release and wait for Googlers to upload the image.

PR #141, opened 2017-05-12
correct client version

zouyee:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/141)


brancz:
Does that actually make a difference? As far as I can tell both methods return the same thing.

andyxning:
@brancz According to the comment for [`Extensions`](https://github.com/kubernetes/client-go/blob/master/kubernetes/clientset.go#L71-L73), it is deprecated, we should explicitly choose a versioned one.

brancz:
Gotcha. lgtm

PR #146, opened 2017-05-16
add node summary grafana

### LINKED ISSUE #123: Grafana dashboard
I think that it might be really cool to have a dashboard using this project.

andyxning:
This PR fix #123 .

Currently, i suggest we contains cluster and single dashboards. 
* cluster: Aggregated dashboards for a cluster.
* single: single dashboards for a single resource(e.g. `node` or `pod`).

@brancz @andrewhowdencom @cemo

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/146)

### DIFF
--- a/grafana/README.md
+++ b/grafana/README.md
@@ -1,0 +1,22 @@
+# Grafana
+
+This directory contains useful out-of-box dashboards.
+
+# Constraints
+
+Dashboards contain [cluster](../grafana/dashboards/cluster) and [single](../grafana/dashboards/single) dashboards.
+* cluster: aggregated dashboards for a cluster.
+* single: dashboards for a single resource(e.g. `node` or `pod`).
+
+**Notice**
+* cluster dashboards use a `cluster` template variable. It can be set at target level as a custom label or set as an `external_labels` label in case Prometheus federation is used.
+
+# Dashboards
+
+* cluster
+  * [nodes](../grafana/dashboards/cluster/nodes.json)
+    * nodes unschedulable: nodes whose node `Unschedulable` status is `true`.
+    * nodes notready: nodes whose node `NodeReady` status is not `true`.
+    * ready node number: number of nodes whose node `NodeReady` status is `true`.
+* single
+
--- a/grafana/dashboards/cluster/nodes.json
+++ b/grafana/dashboards/cluster/nodes.json
@@ -1,0 +1,366 @@
+{
+  "__inputs": [
+    {
+      "name": "DS_K8S_MONITOR",
+      "label": "k8s_monitor",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__requires": [
+    {
+      "type": "grafana",
+      "id": "grafana",
+      "name": "Grafana",
+      "version": "4.2.0"
+    },
+    {
+      "type": "panel",
+      "id": "graph",
+      "name": "Graph",
+      "version": ""
+    },
+    {
+      "type": "datasource",
+      "id": "prometheus",
+      "name": "Prometheus",
+      "version": "1.0.0"
+    }
+  ],
+  "annotations": {
+    "list": []
+  },
+  "description": "kube-state-metrics cluster nodes",
+  "editable": true,
+  "gnetId": null,
+  "graphTooltip": 0,
+  "hideControls": false,
+  "id": null,
+  "links": [],
+  "rows": [
+    {
+      "collapse": false,
+      "height": "250px",
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "${DS_K8S_MONITOR}",
+          "fill": 1,
+          "id": 1,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "percentage": false,
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_spec_unschedulable{cluster=\"$cluster\"}) > 0",
+              "intervalFactor": 2,
+              "legendFormat": "{{node}}",
+              "refId": "A",
+              "step": 30
+            }
+          ],
+          "thresholds": [],
+          "timeFrom": null,
+          "timeShift": null,
+          "title": "Nodes Unschedulable",
+          "tooltip": {
+            "shared": true,
+            "sort": 0,
+            "value_type": "individual"
+          },
+          "type": "graph",
+          "xaxis": {
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            },
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            }
+          ]
+        }
+      ],
+      "repeat": null,
+      "repeatIteration": null,
+      "repeatRowId": null,
+      "showTitle": false,
+      "title": "Dashboard Row",
+      "titleSize": "h6"
+    },
+    {
+      "collapse": false,
+      "height": 250,
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "${DS_K8S_MONITOR}",
+          "fill": 1,
+          "id": 2,
+          "legend": {
+            "avg": false,
+            "current": false,
+            "max": false,
+            "min": false,
+            "show": true,
+            "total": false,
+            "values": false
+          },
+          "lines": true,
+          "linewidth": 1,
+          "links": [],
+          "nullPointMode": "null",
+          "percentage": false,
+          "pointradius": 5,
+          "points": false,
+          "renderer": "flot",
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_status_ready{condition!=\"true\",cluster=\"$cluster\"}) > 0",
+              "intervalFactor": 2,
+              "legendFormat": "{{node}}",
+              "metric": "",
+              "refId": "A",
+              "step": 30
+            }
+          ],
+          "thresholds": [],
+          "timeFrom": null,
+          "timeShift": null,
+          "title": "Nodes NotReady",
+          "tooltip": {
+            "shared": true,
+            "sort": 0,
+            "value_type": "individual"
+          },
+          "type": "graph",
+          "xaxis": {
+            "mode": "time",
+            "name": null,
+            "show": true,
+            "values": []
+          },
+          "yaxes": [
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            },
+            {
+              "format": "short",
+              "label": null,
+              "logBase": 1,
+              "max": null,
+              "min": null,
+              "show": true
+            }
+          ]
+        }
+      ],
+      "repeat": null,
+      "repeatIteration": null,
+      "repeatRowId": null,
+      "showTitle": false,
+      "title": "Dashboard Row",
+      "titleSize": "h6"
+    },
+    {
+      "collapse": false,
+      "height": 250,
+      "panels": [
+        {
+          "aliasColors": {},
+          "bars": false,
+          "datasource": "${DS_K8S_MONITOR}",
+          "fill": 1,
+          "id": 3,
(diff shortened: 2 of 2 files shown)


brancz:
I have mixed feelings about the usefulness about dashboards that can __only__ use metrics coming from kube-state-metrics whereas I'd imagine the bulk of metrics would want to be displayed with a number of other metrics together.

andyxning:
> I'd imagine the bulk of metrics would want to be displayed with a number of other metrics together

IIUC, @brancz means that we may combine kube-state-metrics with some other metrics about k8s, such as heapster, to give users more info. That's a good idea, but i don't think it's the current mission for kube-state-metrics:
* We have no idea whether users also deploy some other component, such as heapster.
* Dashboards with mix kube-state-metrics metrics with metrics collected by other component may may should be placed in other dependent repo. This makes it clear for users.

I am not opposed to adding mixed dashboards with metrics from more than kube-state-metrics. Just feels that it is not suitable to be included in any one of the components. :)

brancz:
I agree with everything that you're saying, I'm just questioning how useful the dashboards can be that are built solely from kube-state-metrics metrics.

andrewhowdencom:
I agree that perhaps they're not useful in and of themselves, but still having them is useful -- I can (for example) use graphs from this dashboard to construct other dashboards. It also allows those without a super understanding of k8s to get an insight into the cluster, and allows for conversations around state vs health etc.

brancz:
@andrewhowdencom fair enough! Makes sense.

andyxning:
@brancz An example i suggest in [#123 comment](https://github.com/kubernetes/kube-state-metrics/issues/123#issuecomment-294392145) may be reason for starting this. :)

@andrewhowdencom I still prefer use `grafana` folder just like heapster. `contrib` is mainly used for non core code instead grafana dashboards.

brancz:
Yeah I don't have a preference of the directory .. I think using a grafana dir is fine for now, we can restructure if it gets too much.

andyxning:
@brancz Could you please review this. :)

grafana/README.md
```diff
@@ -0,0 +1,25 @@
...
+* summary
+  * [nodes](../grafana/dashboards/summary/nodes.json)
+    * nodes unschedulable: nodes whose node `Unschedulable` status is `true`.
+    * nodes notready: nodes whose node `NodeReady` status is not `true`.
+    * ready node number: number of nodes whose node `NodeReady` status is `true`.
+* single
+
+# Caution
+
+After importing all necessary dashboards, users should adapt data source to the right one.
```
brancz: You just need to export the dashboard via the export feature rather than copying the json from the "View JSON" feature, that way when importing the user will be asked which datasource to use it with, then this is not an issue anymore.

grafana/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This directory contains useful out-of-box dashboards.
+
+# Constraints
+
+Dashboards contain [summary](../grafana/dashboards/summary) and [single](../grafana/dashboards/single) dashboards.
+* summary: Aggregated dashboards for a cluster(`alias`).
+* single: single dashboards for a single resource(e.g. `node` or `pod`).
+
+summary dashboards use `alias` as a template variable. `alias` can be set as a cluster name
```
brancz: I think having something like this is a sensible thing to do, but `alias` is too generic, so I think `cluster` would be much more clear on what this selection is supposed to represent.

grafana/dashboards/summary/nodes.json
```diff
@@ -0,0 +1,339 @@
...
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_spec_unschedulable{alias=\"$alias\"}) > 0",
```
brancz: I actually just discussed the multi replica usage with @fabxc and it turns out that scraping multiple replicas actually makes little sense especially when you want to summarize data then it shows up multiple times. So we should actually be scraping kube-state-metrics through a Service.

grafana/README.md
andyxning: @brancz Thanks for your explanation about Grafana export feature. 👍

grafana/README.md
andyxning: Will do.

andyxning: Readme has been updated. @brancz PTAL.

grafana/dashboards/summary/nodes.json
andyxning: IIUC, this is related to the multi cluster support for kube-state-metrics. For now, i use multiple kube-state-metrics instances for this, i.e., on kube-state-metric instance for one cluster. Thus i have used `alias` as the cluster name.

andyxning: Disclose: I run kube-state-metrics on bare metal instead of in a k8s cluster.

brancz: Yes sorry that makes sense, I thought you were using the `avg` aggregation for something else .. that was my own confusion.

grafana/README.md
```diff
@@ -0,0 +1,22 @@
...
+This directory contains useful out-of-box dashboards.
+
+# Constraints
+
+Dashboards contain [summary](../grafana/dashboards/summary) and [single](../grafana/dashboards/single) dashboards.
+* summary: Aggregated dashboards for a cluster.
+* single: dashboards for a single resource(e.g. `node` or `pod`).
+
+**Notice**
+* summary dashboards use `alias` as the `cluster` template variable.
```
brancz: I'm more confused by this then anything. Why do we not rename `alias` to `cluster` entirely and get rid of naming anything `alias`?

andyxning: @brancz You mean that we can use relabel feature in Prometheus to relabel `alias` to `cluster`?

andyxning: Example:

```yaml
scrape_configs:
  - job_name: 'kube_state_metrics'
    static_configs:
    - labels: {alias: 'dev', instance: 'dev', job: 'dev'}
      targets: ['10.0.0.1:8080']
    relabel_configs:
      - source_labels: [alias]
        action: replace
        target_label: cluster
        regex: (.*)
        replacement: $1
```

brancz: No I'm just wondering where this wondrous `alias` label is coming from. I've never seen it before, so I thought that it is something that you have in your Prometheus configuration like an external label or something. Because it's not part of the labels exposed by those metrics.

andyxning: Ah. I got you. `alias` is just my customized label. My fault.

So, @brancz what about we make a constraint: 
* summary dashboards use a `cluster` template variable which is set for each cluster in Promethues configuration.

brancz: From what I've seen people put their cluster name in the `external_labels` in Prometheus. So generally yes I think it's sensible to have a cluster selection, but also have the dashboard not care if there is only one cluster even if the metrics don't have that label.

andyxning: @brancz I totally agree with you. However, I do not know how to implement `if single cluster, do not show up cluster template var; otherwise, do it` in Grafana. 

If there is no such an easy way to implement this, how about we document `cluster` template var clearly and make it all the time.

brancz: `external_labels` is just a common pattern, your usage is also valid, the label then just comes from a different source, but ends up being similar at query time.

brancz: Either way I think alias should still be renamed to cluster.

andyxning: @brancz Yep. It is the user's choice of whether `external_labels` or target level custom `cluster` label is fine to them. 

Finally, we make a conclusion that we should use a `cluster` template var.

I have updated the `node.json` file and readme. Could you please take a look at this.

grafana/dashboards/summary/nodes.json
```diff
@@ -0,0 +1,366 @@
...
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_spec_unschedulable{cluster=\"$cluster\"}) > 0",
```
brancz: I think you're interested in the sum of unschedulable nodes no? In that case you would want to query more like this:

```
sum by(cluster) (kube_node_spec_unschedulable)
```

That way we get the sum of nodes that are in the unschedulable state.

The legend would show the `cluster`.

grafana/dashboards/summary/nodes.json
```diff
@@ -0,0 +1,366 @@
...
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_status_ready{condition!=\"true\",cluster=\"$cluster\"}) > 0",
```
brancz: Similar here, you are likely interested in the sum of nodes that are not in ready state:

```
sum by(cluster) (kube_node_status_ready{condition!="true"})
```

grafana/dashboards/summary/nodes.json
```diff
@@ -0,0 +1,366 @@
...
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "count(avg by(node)(kube_node_status_ready{cluster=\"$cluster\",condition=\"true\"}) > 0)",
```
brancz: Same here, the interesting thing to see is the number of nodes that are ready:

```
sum by(cluster) (kube_node_status_ready{condition="true"})
```

grafana/dashboards/summary/nodes.json
andyxning: The original intention is to list node name in notready or unschedulable status. My starting point is by most cases the nodes in notready or unschedulable status should be rarely small and we need their names to process them.

I am totally agree that we should also add a panel about the total number of nodes in notready or unschedulable status.

grafana/dashboards/cluster/nodes.json
```diff
@@ -0,0 +1,366 @@
...
+          "seriesOverrides": [],
+          "span": 12,
+          "stack": false,
+          "steppedLine": false,
+          "targets": [
+            {
+              "aggregator": "sum",
+              "downsampleAggregator": "avg",
+              "downsampleFillPolicy": "none",
+              "expr": "avg by(node)(kube_node_spec_unschedulable{cluster=\"$cluster\"}) > 0",
```
brancz: I fail to see why this should be `avg > 0`

andyxning: we only have `kube_node_spec_unschedulable` being `1` when it is unschedulable,  otherwise it is `0`. By setting higher than 0, i want to only match those which is really in unschedulable status. :)

As for `avg`, just an aggregate function.

brancz: `avg` does some heuristics IIRC so this may show different results from what a user would expect, so `sum` is definitely the right choice here.

Additionally we should aggregate by cluster instead of select by cluster, I think that makes this dashboard more usable as most users won't start out with having a cluster label on the metrics. That way this dashboard can be a quick success.

r4j4h:
Just wanted to mention in regards to aggregating by cluster or allowing to select that there is another approach that provides the usefulness (and the complication) of filtering by cluster while also allowing aggregation and that is use of the "Multi-Select" and "Select All" features of Grafana on the Template Variables.

The other change that would be needed is the queries would need `=` turned to RegExp `=~`. By default All with OR each option with a pipe, so `cluster=~"cluster-1|cluster-2|etc"` but with Custom All Value `.+` it will do `cluster=~".+"`. The Custom All Value is useful when the cardinality is so high that the query errors out due to length, but is unlikely to be needed here.

brancz:
Agreed @r4j4h!

Sorry @andyxning, I've been traveling the past week and a half, I'll try to get back to this as soon as I can.

grafana/dashboards/cluster/nodes.json
brancz: as a start I think we can merge this once this change is done

andyxning: @brancz

Sorry for the delay. Hard work. Will do this in recent two days. :)

brancz: No worries I didn't get back to it for a while as well. There is no hurry.

Starefossen:
Any progress on this?

andyxning:
@Starefossen Sorry for the half-way work. Too busy this week. 😭  Maybe I will continue to work on this. 

BTW, any suggestion or thoughts about how to implement this?

brancz:
Since we last discussed on this PR there were some efforts in the Prometheus community to start including "bundles" of alerting and dashboard definitions with exporters. The results of this was that even for the components maintained by the Prometheus Team (this was specifically tried out on the node-exporter), not maintainable or even applicable to all users. The result is that documenting useful examples of queries and alerting rules are better suitable for this purpose, that way everyone can take and adapt and even mix metrics of different sources, instead of having one dashboard per exporter.

andrewhowdencom:
@brancz Is an "either or" approach the required for this? At the minute, I just pick up pages from the Grafana sharing system, and start messing around with them.

The goal with having a dashboard (at least to me) isn't to provide a default to take this information and make it useful, but to take this information and give it meaning. I use them to explain how things work -- in the case of the node exporter, 

- "What is CPU? Why are there like 6 of them?"
- "What is entropy and why do I need it?"
- "What do we expect to see when the network adapter is the rate limiting step on the node?"

and so on. Actual problems use runbooks or whatever, so that meaning can be conveyed with the metrics (if this is above x, it likely means that y is true. see #there-is-something-wrong-with-bork)

andyxning:
close this PR via [#123 comment](https://github.com/kubernetes/kube-state-metrics/issues/123#issuecomment-328962797).

PR #153, opened 2017-06-23
Multi arch binaries

nlamirault:
Signed-off-by: Nicolas Lamirault <nicolas.lamirault@gmail.com>

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/153)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,6 +2,7 @@ all: build
 
 FLAGS =
 COMMONENVVAR = GOOS=linux GOARCH=amd64
+ARCHS = amd64 386 arm arm64
 BUILDENVVAR = CGO_ENABLED=0
 TESTENVVAR = 
 REGISTRY = gcr.io/google_containers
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -14,6 +15,9 @@ deps:
 build: clean deps
 	$(COMMONENVVAR) $(BUILDENVVAR) godep go build -o kube-state-metrics 
 
+crossbuild:
```
andyxning: @nlamirault Should also consider update the `clean` directive.

andyxning: @brancz should this be added to 1.0 milestone to support build multi arches binary releases?

brancz: I don't think we will start releasing in any other form than the container images for now. At least until google opens up the release process further. So for now this will just be a helper to build it yourself.

andyxning:
close this via #252

PR #154, opened 2017-06-27
collectors/pod: flatten created by label and extract owners

brancz:
The created by label previously was in the format "<kind>/<name>", this
is now normalized in to a respective label for each.

Additionally it is possible that a Pod has multiple OwnerReferences.
Previously multiple `kube_pod_info` metrics were generated for each
OwnerReference, unnecessarily duplicating all other labels of the
`kube_pod_info` metric.

The owner part has not been released yet, so it's non breaking,
however, the `created_by` label is breaking. This will be noted in
the changelog.

@fabxc @andyxning @mxinden @gouthamve

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/154)

### DIFF
--- a/Documentation/pod-metrics.md
+++ b/Documentation/pod-metrics.md
@@ -2,7 +2,8 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_pod_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `start_time`=&lt;date-time since kubelet acknowledged pod&gt; <br> `node`=&lt;node-name&gt;<br> `created_by`=&lt;created_by&gt;<br> `owner_kind`=&lt;owner kind&gt; <br> `owner_name`=&lt;owner name&gt; <br> `owner_is_controller`=&lt;whether owner is controller&gt;  |
+| kube_pod_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `start_time`=&lt;date-time since kubelet acknowledged pod&gt; <br> `node`=&lt;node-name&gt;<br> `created_by_kind`=&lt;created_by_kind&gt;<br> `created_by_name`=&lt;created_by_name&gt;<br> |
+| kube_pod_owner | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `owner_kind`=&lt;owner kind&gt; <br> `owner_name`=&lt;owner name&gt; <br> `owner_is_controller`=&lt;whether owner is controller&gt;  |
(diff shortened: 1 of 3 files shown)


fabxc:
Pulling owners into their own metric is a pretty intense normalization for a metric. I wonder how common that really is and whether just having multiple `kube_pod_info` metrics still wouldn't be the simpler choice in that case.   
One risk of course is that people generally don't expect it due to its rareness and then write queries that fail when actually confronted with a n-to-m match on joins.

andyxning:
I prefer that we split owner info into different metrics, this makes `kube_pod_info` more simplified and clear. By making kind and name separation, it will make queries about kind and name solely available and more easy.

>One risk of course is that people generally don't expect it due to its rareness and then write queries that fail when actually confronted with a n-to-m match on joins.

@brancz @fabxc Prometheus is good enough to support n-to-m match join. 👍  However, imho, the docs about `left join` and `right join` are not so clear and lack of clear examples. Maybe this can be resolved by adding some join query example in kube-state-metrics docs.

Documentation/pod-metrics.md
```diff
@@ -2,7 +2,8 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_pod_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `start_time`=&lt;date-time since kubelet acknowledged pod&gt; <br> `node`=&lt;node-name&gt;<br> `created_by`=&lt;created_by&gt;<br> `owner_kind`=&lt;owner kind&gt; <br> `owner_name`=&lt;owner name&gt; <br> `owner_is_controller`=&lt;whether owner is controller&gt;  |
+| kube_pod_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `start_time`=&lt;date-time since kubelet acknowledged pod&gt; <br> `node`=&lt;node-name&gt;<br> `created_by_kind`=&lt;created_by_kind&gt;<br> `created_by_name`=&lt;created_by_name&gt;<br> |
```
andyxning: @brancz How about we use `owner_kind` and `owner_name`. This make join query more easy. Just like we use the same label name for `namespace` and `pod`.

fabxc: What exactly do you mean? The owner metric uses `owner_kind` and `owner_name` already.

andyxning:
Imean the "created_by_kind" and "created_by_name" tag key in kube_pod_info

Documentation/pod-metrics.md
brancz: The point is that created by and owners can be different and are extracted from different sources, they are essentially unrelated data. For to be able to join on these I would recommend to use the [`label_replace`](https://prometheus.io/docs/querying/functions/#label_replace()) function.

andyxning: @brancz Seems good to me. 

BTW, having not known that Prometheus supports `label_replace` function which is awesome to this situation. :)

andyxning: SGTM

fabxc:
👍

PR #155, opened 2017-06-29
Add metric for cronjob scheduling delay

julia-stripe:
We're scheduling cronjobs! When they fail to get scheduled onto a node (for instance because the cluster is overloaded), we want to know how far behind they are.

This adds a delay metric for every job that

* isn't suspended
* has a `lastScheduleTime`

& uses the same cronjob library that kube-cron uses to calculate cron job schedule times.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/155)

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -17,8 +17,11 @@ limitations under the License.
 package collectors
 
 import (
+	"time"
+
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
+	"github.com/robfig/cron"
 	"golang.org/x/net/context"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api"
@@ -26,7 +29,10 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+type nowFuncT func() time.Time
+
 var (
+	nowFunc nowFuncT // so that we can mock out `time.Now()` in tests
 	descCronJobInfo = prometheus.NewDesc(
 		"kube_cronjob_info",
 		"Info about cronjob.",
@@ -52,7 +58,18 @@ var (
 		"Deadline in seconds for starting the job if it misses scheduled time for any reason.",
 		[]string{"namespace", "cronjob"}, nil,
 	)
+	descCronJobSchedulingDelay = prometheus.NewDesc(
+		"kube_cronjob_scheduling_delay",
+		"Number of seconds the cron job is delayed scheduling",
+		[]string{"namespace", "cronjob"}, nil,
+	)
 )
+
+func init() {
+	nowFunc = func() time.Time {
+		return time.Now()
+	}
+}
 
 type CronJobLister func() ([]v2batch.CronJob, error)
 
@@ -92,6 +109,7 @@ func (dc *cronJobCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descCronJobStatusLastScheduleTime
 	ch <- descCronJobSpecSuspend
 	ch <- descCronJobSpecStartingDeadlineSeconds
+	ch <- descCronJobSchedulingDelay
 }
 
 // Collect implements the prometheus.Collector interface.
@@ -103,6 +121,21 @@ func (cjc *cronJobCollector) Collect(ch chan<- prometheus.Metric) {
 	}
 	for _, cj := range cronjobs {
 		cjc.collectCronJob(ch, cj)
+	}
+}
+
+func getSchedulingDelaySeconds(schedule string, lastScheduleTime time.Time) float64 {
+	sched, err := cron.ParseStandard(schedule)
+	if err != nil {
+		glog.Errorf("Failed to parse cron job schedule '%s': %s", schedule, err)
+		return 0
+	}
+	nextScheduleTime := sched.Next(lastScheduleTime)
+	delay := nowFunc().Sub(nextScheduleTime).Seconds()
+	if delay < 0 {
+		return 0
+	} else {
+		return delay
 	}
 }
 
@@ -120,8 +153,12 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 		addGauge(descCronJobSpecStartingDeadlineSeconds, float64(*j.Spec.StartingDeadlineSeconds))
 	}
 
-	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
+	// If the cron job is suspended, don't track delays
+	if j.Status.LastScheduleTime != nil && !*j.Spec.Suspend {
+		addGauge(descCronJobSchedulingDelay, getSchedulingDelaySeconds(j.Spec.Schedule, j.Status.LastScheduleTime.Time))
+	}
 
+	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
 	addGauge(descCronJobStatusActive, float64(len(j.Status.Active)))
 	addGauge(descCronJobSpecSuspend, boolFloat64(*j.Spec.Suspend))
 
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -132,9 +132,8 @@ func getSchedulingDelaySeconds(schedule string, lastScheduleTime time.Time) floa
 	delay := timeNow().Sub(nextScheduleTime).Seconds()
 	if delay < 0 {
 		return 0
-	} else {
-		return delay
 	}
+	return delay
 }
 
 func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batch.CronJob) {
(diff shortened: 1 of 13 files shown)


collectors/cronjob.go
```diff
@@ -17,16 +17,22 @@ limitations under the License.
...
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api"
 	v2batch "k8s.io/client-go/pkg/apis/batch/v2alpha1"
 	"k8s.io/client-go/tools/cache"
 )
 
+type nowFuncT func() time.Time
+
 var (
+	nowFunc nowFuncT // so that we can mock out `time.Now()` in tests
```
jmillikin-stripe: This would usually be a plain assignment, `timeNow = time.Now` -- no need for the extra type or the wrapper in `init()`.

Alternatively, `now` could be passed to `getSchedulingDelaySeconds()` as a parameter and then we wouldn't need to stub it out at all.

(heh, you can take the reviewer out of go-readability-approvers@, but ...)

collectors/cronjob.go
```diff
@@ -106,6 +124,21 @@ func (cjc *cronJobCollector) Collect(ch chan<- prometheus.Metric) {
...
+	sched, err := cron.ParseStandard(schedule)
+	if err != nil {
+		glog.Errorf("Failed to parse cron job schedule '%s': %s", schedule, err)
+		return 0
+	}
+	nextScheduleTime := sched.Next(lastScheduleTime)
+	delay := nowFunc().Sub(nextScheduleTime).Seconds()
+	if delay < 0 {
+		return 0
+	} else {
```
jmillikin-stripe: Go style discourages `} else {` after a return:

```go
if delay < 0 {
	return 0
}
return delay
```

collectors/cronjob.go
```diff
@@ -120,8 +153,12 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 		addGauge(descCronJobSpecStartingDeadlineSeconds, float64(*j.Spec.StartingDeadlineSeconds))
 	}
 
-	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
+	// If the cron job is suspended, don't track delays
+	if j.Status.LastScheduleTime != nil && !*j.Spec.Suspend {
```
jmillikin-stripe: What happens to the metric if a CronJob has never been successfully scheduled, for example if it's too big for any machine in the cluster? I think we want to alert on that also.

julia-stripe: agreed! `CronJob` [only has](https://github.com/kubernetes/kubernetes/blob/60604f8/staging/src/k8s.io/api/batch/v2alpha1/types.go#L54-L148) a `LastScheduleTime` timestamp right now though (there's no `creationTime` to say when the job was created) so I don't see how to accomplish that with the current API

brancz:

kube-state-metrics maps values of Kubernetes API object fields one to one to metrics. Introducing this would break the consistency with the other metrics, as it is a computed value. Adding the `scheduling_delay` might make more sense to be in the upstream `CronJob` object and once available there it can be implemented to be exposed as a metric here. The implementation details would have to be worked out and might not end up being the scheduled delay. Maybe the final aggregation might also be done at query time meaning we would just need to add `next_schedule_time` and compute the delay with PromQL functions.

As already mentioned in the comment, that would also give us the chance to add `creation_time` to be able to alert on never having been scheduled, however, equivalent today would be `absent(cronjob_status_last_schedule_time)`, which evaluates to triggering if the time-series does not exist.

I'll be happy to review upstream proposals and implementations.

What do you think?

/cc @fabxc 

As a side note: for us to be able to accept contributions, you or your organization will need to sign the CLA.

julia-stripe:
Thanks so much for the feedback!

I think a `nextScheduleTime` metric is ultimately what we want here, like you say -- calculated as 

```
if lastScheduleTime:
    sched.Next(lastScheduleTime)
else:
    sched.Next(creationTime)
```

one reason I'm hesitant to propose adding that field to the upstream `CronJob` is that it's only really useful for monitoring -- the scheduler itself wouldn't have any reason to use it (if 8 jobs were meant to be scheduled between `lastScheduleTime` and now, then the expected `nextScheduleTime` is irrelevant except for monitoring purposes). Do you think it's reasonable to add something to a Kubernetes API object that is redundant / could be calculated from other properties in that object but is useful for monitoring? I'm a bit worried it'll get out of sync.

brancz:
I think I'm ok with just exposing `next_scheduled_time` in form of a unix timestamp as a metric instead of the delay, mainly as the delay introduces the necessity to inject time. Which we shouldn't do with the init function, but via dependency injection into the collector struct. However, since that necessity would disappear if we opt for `next_scheduled_time` instead of the delay, we wouldn't have to in the first place.

Then the [`time()`](https://prometheus.io/docs/querying/functions/#time()) function of PromQL can be used to calculate the actual delay between the expected. As everything is unix timestamps, we don't have to worry about timezones.

What do you think @fabxc ? (I'll be on vacation for the next two weeks so I'll leave it to @fabxc to handle this, otherwise I'll pick this up once I'm back)

jmillikin-stripe:
> As a side note: for us to be able to accept contributions, you or your organization will need to sign the CLA.

Acknowledged. We're currently working through the internal CLA request process and expect there won't be any issues (CLAs for e.g. Bazel went through fine).

julia-stripe:
Updated this with an implementation of a `kube_cronjob_next_schedule_time` metric. It turns out that there actually *is* a creation timestamp on cron jobs through the `ObjectMeta` metadata that all kube resources have (!!) so it uses that.

collectors/cronjob.go
```diff
@@ -106,6 +116,21 @@ func (cjc *cronJobCollector) Collect(ch chan<- prometheus.Metric) {
 	}
 }
 
+func getNextScheduledTime(schedule string, lastScheduleTime *unversioned.Time, createdTime unversioned.Time) time.Time {
+	sched, err := cron.ParseStandard(schedule)
+	if err != nil {
+		glog.Errorf("Failed to parse cron job schedule '%s': %s", schedule, err)
```
fabxc: I'd prefer keeping logging out of helper functions and return an error here.

brian-brazil:
NaN should only be used where it makes sense mathematically, so not exposing it would be a better option.

julia-stripe:
@k8s-bot test this

julia-stripe:
Review status: 0 of 13 files reviewed at latest revision, 4 unresolved discussions.

---

*[collectors/cronjob.go, line 35 at r2](https://reviewable.kubernetes.io:443/reviews/kubernetes/kube-state-metrics/155#-KnopkUGSjsgadnWs-l9-r2-35:-KoOj-1Bfu3aQ-xIj_CF:b-896fix) ([raw file](https://github.com/kubernetes/kube-state-metrics/blob/cc026a87fe9dd6704b98d513efda45a87879b251/collectors/cronjob.go#L35)):*
<details><summary><i>Previously, jmillikin-stripe wrote…</i></summary><blockquote>

This would usually be a plain assignment, `timeNow = time.Now` -- no need for the extra type or the wrapper in `init()`.

Alternatively, `now` could be passed to `getSchedulingDelaySeconds()` as a parameter and then we wouldn't need to stub it out at all.

(heh, you can take the reviewer out of go-readability-approvers@, but ...)
</blockquote></details>

Done.

---

*[collectors/cronjob.go, line 137 at r2](https://reviewable.kubernetes.io:443/reviews/kubernetes/kube-state-metrics/155#-KnopkUGSjsgadnWs-l9-r2-137:-KoOpFYVYpMJgkL0jiu1:b-896fix) ([raw file](https://github.com/kubernetes/kube-state-metrics/blob/cc026a87fe9dd6704b98d513efda45a87879b251/collectors/cronjob.go#L137)):*
<details><summary><i>Previously, jmillikin-stripe wrote…</i></summary><blockquote>

Go style discourages `} else {` after a return:

```go
if delay < 0 {
	return 0
}
return delay
```
</blockquote></details>

Done.

---

*[collectors/cronjob.go, line 122 at r5](https://reviewable.kubernetes.io:443/reviews/kubernetes/kube-state-metrics/155#-KnopkUGSjsgadnWs-l9-r5-122:-KoOpCnYqfDFu77Ma8Bb:b-896fja) ([raw file](https://github.com/kubernetes/kube-state-metrics/blob/3f0a013e496b239e57d2df5b779a5f04eacece12/collectors/cronjob.go#L122)):*
<details><summary><i>Previously, fabxc (Fabian Reinartz) wrote…</i></summary><blockquote>

I'd prefer keeping logging out of helper functions and return an error here.
</blockquote></details>

Done!

---


*Comments from [Reviewable](https://reviewable.kubernetes.io:443/reviews/kubernetes/kube-state-metrics/155)*

julia-stripe:
@fabxc moved the logging out of the helper function and signed the CLA!

fabxc:
Thanks 👍

fabxc:
@julia-stripe one thing I forgot: can you document the new metric in `Documentation/`?

PR #160, opened 2017-07-04
fix namespace test failure

zouyee:
@brancz Sorry for having replied to you so late, but things have really been hectic around here.

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/160)

### DIFF
--- a/collectors/namespace.go
+++ b/collectors/namespace.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collectors
+
+import (
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/net/context"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api"
+	"k8s.io/client-go/pkg/api/v1"
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descNamespacePhase = prometheus.NewDesc(
+		"kube_namespace_status_phase",
+		"kubernetes namespace status phase.",
+		[]string{
+			"name",
+			"phase",
+		}, nil,
+	)
+)
+
+// NamespaceLister define NamespaceLister type
+type NamespaceLister func() ([]v1.Namespace, error)
+
+// List return namespace list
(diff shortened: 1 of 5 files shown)


collectors/namespace.go
```diff
@@ -0,0 +1,102 @@
...
+		"kubernetes namespace status phase.",
+		[]string{
+			"name",
+			"create_time",
+			"status",
+		}, nil,
+	)
+)
+
+// NamespaceLister ...
```
fabxc: Needs to be completed.

collectors/namespace.go
```diff
@@ -0,0 +1,102 @@
...
+type NamespaceLister func() ([]v1.Namespace, error)
+
+// List ...
```
fabxc: Needs to be completed.

collectors/namespace.go
```diff
@@ -0,0 +1,102 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descNamespace = prometheus.NewDesc(
+		"kube_namespace_status_phase",
+		"kubernetes namespace status phase.",
+		[]string{
+			"name",
+			"create_time",
```
fabxc: The value for this label is not taken from `.status.phase` and thus doesn't belong into this metric.
It could be its own metric or part of an info metric. I would just leave it out for now though as we aren't using that piece of metadata elsewhere. If there's a valid use case we can still add it back.

fabxc:
Looks good aside from comments. Docs have to be added in `Documentation/`.
I noticed that we added phase metrics for other objects as well but are not following the enumeration semantics which we are using for conditions, i.e. having a series for each possible value and setting them to 0 or 1 respectively. I'll open a separate issue for that. Since this would need changes elsewhere too, it's not a blocker for this PR.

collectors/namespace.go
zouyee: I got it. Thanks you.

zouyee:
@brancz  PTAL

collectors/namespace.go
zouyee: now I only  reserved `name` and `status` fields, thank you.

collectors/namespace.go
zouyee: done!

collectors/namespace.go
```diff
@@ -0,0 +1,101 @@
...
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descNamespace = prometheus.NewDesc(
+		"kube_namespace_status_phase",
+		"kubernetes namespace status phase.",
+		[]string{
+			"name",
+			"status",
```
brancz: This would be called `phase`.

zouyee: done!

zouyee: It has beed fixed.

zouyee: fixed as requested

andyxning:
@zouyee 

Could you please rebase code to current master branch as we have introduce a new check in travisCI.

chlunde:
This was merged as part of #234, so I believe it can be closed, unless I missed something

andyxning:
@chlunde Thanks for pointing this out. I have checked and the metrics proposed in this PR has been implemented in #234 . So, closing this. 

@zouyee Thanks for you contributions! Feel free to reopen this PR if you have some thoughts.

PR #172, opened 2017-07-11
Change kube_cronjob_status_last_schedule_time from a counter to a gauge

julia-stripe:
If I understand correctly `kube_cronjob_status_last_schedule_time` should actually be a gauge (we shouldn't be adding unix timestamps together)

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/172)


fabxc:
Good catch, thank you! 👍

PR #175, opened 2017-07-24
Collect labels for deployments and statefulsets.

alamaison:
It builds but is untested because `go test` just says `?   	k8s.io/kube-state-metrics	[no test files]`.

I'm not a go programmer, so if this is not the correct way to run tests please let me know. 

I'm not sure I understand the significance of the default labels.  Copying how the service collector defines them, I've added `namespace` and the type of object being labelled.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/175)

### DIFF
--- a/collectors/deployment.go
+++ b/collectors/deployment.go
@@ -161,6 +161,9 @@ func (dc *deploymentCollector) collectDeployment(ch chan<- prometheus.Metric, d 
 		lv = append([]string{d.Namespace, d.Name}, lv...)
 		ch <- prometheus.MustNewConstMetric(desc, prometheus.GaugeValue, v, lv...)
 	}
+
+	labelKeys, labelValues := kubeLabelsToPrometheusLabels(d.Labels)
+	addGauge(deploymentLabelsDesc(labelKeys), 1, labelValues...)
 	addGauge(descDeploymentStatusReplicas, float64(d.Status.Replicas))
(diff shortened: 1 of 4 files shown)


brancz:

You can just run `make test-unit` to execute the tests locally.

We'll also need to you or your organization to sign the CLA for us to be able to accept these changes.

alamaison:
Thanks. I've added tests and fixed the errors.  Except one.  The stateful set label collection works as expected, one entry for each statefulset.  However the identical setup for deployments only gives labels for the first deployment.  Perhaps you could have a look to see if I'm doing something obviously wrong?

collectors/deployment.go
```diff
@@ -160,4 +180,7 @@ func (dc *deploymentCollector) collectDeployment(ch chan<- prometheus.Metric, d
 	} else {
 		addGauge(descDeploymentStrategyRollingUpdateMaxUnavailable, float64(maxUnavailable))
 	}
+
+	labelKeys, labelValues := kubeLabelsToPrometheusLabels(d.Labels)
+	addGauge(deploymentLabelsDesc(labelKeys), 1, labelValues...)
```
brancz: Just move these two lines to be the first metrics to be added (line 164). The reason these metrics are not showing up is the guard clause on line 173-175.

alamaison:
Derp.  Thanks for the help.  I've updated with a fix.

brancz:

PR #178, opened 2017-07-27
add readme for supported kubernetes version

### LINKED ISSUE #129: Documentation of the metrics
I can be considered pretty much new to monitoring land. It would be really cool to document metrics with some example usages. I know that some metrics are pretty much obvious from its name but its usage with some other metrics would be beneficial for all. I can also try my best to provide necessary information with my limited knowledge.

Another important thing is that compatibility of metrics with k8s version. Maybe something like `since` might be a good idea.

andyxning:
Fix #129 partially.

Add readme about the supported Kubernetes version for kube-state-metrics.   This makes the supported Kubernetes version info more clear. :)

/cc @brancz @cemo

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/178)

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,7 +14,14 @@ either by Prometheus itself or by a scraper that is compatible with scraping
 a Prometheus client endpoint. You can also open `/metrics` in a browser to see
 the raw metrics.
 
-*Requires Kubernetes 1.2+*
+## Kubernetes Version
+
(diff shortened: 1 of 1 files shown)


brancz:
Great! Thanks for taking care of this! Could you just add an additional sentence in there, something like: "all additional compatibility is only best effort, or happens to still/already be supported"

andyxning:
@brancz Done. PTAL.

PR #179, opened 2017-07-28
Add PersistentVolumeClaim metrics

cofyc:

As a admin operator of  multi-tenant clusters, I need to monitor PVC bound status. If a PVC failed (for example, in case RBD Provisioner is not working or Ceph cluster is full), I can know it as soon as possible.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/179)


cofyc:
cc @brancz

brancz:
lgtm :+1:

PR #181, opened 2017-07-28
collectors: generate a metric per possible phase per pod

### LINKED ISSUE #180: Handle pod status phase correctly
The [status phase metric](https://github.com/kubernetes/kube-state-metrics/blob/e8e5e144678064d3f4d3ba5bd8e1bbaeea9daf96/collectors/pod.go#L273) for pod's is currently set to 1 for the phase it is in. No series are exposed for for the phases it is not in. That's not the correct approach for enumerations. 

We've to export a series for each possible phase. Phases the pod is not in must be set to 0.
A correct example of this is the [phase metric for nodes](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/node.go#L232-L236).

brancz:
Fixes #180 

@fabxc @mxinden @gouthamve

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/181)

### DIFF
--- a/collectors/pod.go
+++ b/collectors/pod.go
@@ -270,7 +270,14 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 
 	labelKeys, labelValues := kubeLabelsToPrometheusLabels(p.Labels)
 	addGauge(podLabelsDesc(labelKeys), 1, labelValues...)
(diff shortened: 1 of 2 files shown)


andyxning:
SGTM

fabxc:
👍

PR #182, opened 2017-07-28
Add creation time metrics for all collectors

julia-stripe:
We want to track when pods are created so that we can monitor various pod-age-related metrics, like "are there pods that are > 30 minutes old and stuck in the 'Pending' state".

While adding a 'pod created time' metric I figured it made sense to add a created time metric for all the other collectors as well since they all have a `CreationTimestamp` that's easy to expose.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/182)

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -37,6 +37,11 @@ var (
 		"Info about cronjob.",
 		[]string{"namespace", "cronjob", "schedule", "concurrency_policy"}, nil,
 	)
+	descCronJobCreated = prometheus.NewDesc(
+		"kube_cronjob_created",
+		"Unix creation timestamp",
+		[]string{"namespace", "cronjob"}, nil,
+	)
 	descCronJobStatusActive = prometheus.NewDesc(
 		"kube_cronjob_status_active",
 		"Active holds pointers to currently running jobs.",
@@ -98,6 +103,7 @@ type cronJobCollector struct {
 // Describe implements the prometheus.Collector interface.
 func (dc *cronJobCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descCronJobInfo
+	ch <- descCronJobCreated
 	ch <- descCronJobStatusActive
 	ch <- descCronJobStatusLastScheduleTime
 	ch <- descCronJobSpecSuspend
@@ -150,6 +156,9 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 	}
 
 	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
+	if !j.CreationTimestamp.IsZero() {
+		addGauge(descCronJobCreated, float64(j.CreationTimestamp.Unix()))
+	}
 	addGauge(descCronJobStatusActive, float64(len(j.Status.Active)))
 	if j.Spec.Suspend != nil {
 		addGauge(descCronJobSpecSuspend, boolFloat64(*j.Spec.Suspend))
--- a/collectors/cronjob_test.go
+++ b/collectors/cronjob_test.go
@@ -30,9 +30,9 @@ var (
 	SuspendFalse               bool  = false
 	StartingDeadlineSeconds300 int64 = 300
 
-	ActiveRunningCronJob1LastScheduleTime, _          = time.Parse(time.RFC3339, "2017-05-26T12:00:07Z")
-	SuspendedCronJob1LastScheduleTime, _              = time.Parse(time.RFC3339, "2017-05-26T17:30:00Z")
-	ActiveCronJob1NoLastScheduledCreationTimestamp, _ = time.Parse(time.RFC3339, "2017-05-26T18:30:00Z")
+	ActiveRunningCronJob1LastScheduleTime          = time.Unix(1500000000, 0)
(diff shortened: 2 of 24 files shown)


collectors/cronjob_test.go
```diff
@@ -118,6 +120,10 @@ func TestCronJobCollector(t *testing.T) {
 				},
 			},
 			want: metadata + `
+				kube_cronjob_created{cronjob="ActiveCronJob1NoLastScheduled",namespace="ns1"} 1.4958234e+09
+				kube_cronjob_created{cronjob="ActiveRunningCronJob1",namespace="ns1"} -6.21355968e+10
```
andyxning: @julia-stripe

Why the value for kube_cronjob_created is a negative value.

collectors/cronjob.go
```diff
@@ -150,6 +156,7 @@ func (jc *cronJobCollector) collectCronJob(ch chan<- prometheus.Metric, j v2batc
 	}
 
 	addGauge(descCronJobInfo, 1, j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy))
+	addGauge(descCronJobCreated, float64(j.CreationTimestamp.Unix()))
```
brancz: The `CreationTimestamp` [is not necessarily set](https://github.com/julia-stripe/kube-state-metrics/blob/02ec0772506b4d95ca06de7f3fa3d4587785be9a/vendor/k8s.io/client-go/pkg/api/v1/types.go#L141-L150). So we should skip this metric whenever `CreationTimestamp == nil`.

collectors/cronjob_test.go
brancz: Curious about this as well.

julia-stripe:
Fixed those two things! The `-6.21355968e+10` number was just because that's what `time.Time{}.Unix()` evaluates to.

collectors/cronjob.go
brancz: Argh, sorry this was incorrect from my side. I read from the code comment that this `nil`'able, but in fact it's not a pointer actually, so what you did at first, was actually correct (* hides in shame *).

brancz:
Thanks for following up, I actually mistook something there on my last review, sorry about that! I'm still curious as to why we got that time, but it seem to be fine now.

Could you also make sure to add tests for each collector?

andyxning:
Agree with @brancz . Also, could you please choose another positive unix timestamp which is more intuitively.

julia-stripe:
done! switched everything to the timestamp 1500000000 which is easier to work with.

collectors/cronjob_test.go
```diff
@@ -49,6 +49,8 @@ func TestCronJobCollector(t *testing.T) {
 	const metadata = `
 		# HELP kube_cronjob_info Info about cronjob.
 		# TYPE kube_cronjob_info gauge
+		# HELP kube_cronjob_created Unix creation timestamp
```
andyxning: @julia-stripe should add a `kube_cronjob_created` metric here just like what you added for [`daemonset_test.go`](https://github.com/kubernetes/kube-state-metrics/pull/182/files#diff-edd133937b9a4e7fe14e47765a166c71R73)

andyxning: I have saw it. Sorry for the wrong advice.

collectors/persistentvolumeclaim_test.go
```diff
@@ -49,6 +50,7 @@ func TestPersistentVolumeClaimCollector(t *testing.T) {
 				{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "mysql-data",
+						CreationTimestamp: metav1.Time{Time: time.Unix(1500000000, 0)},
```
andyxning: @julia-stripe there is no test for the added `CreationTimestamp`.

collectors/pod_test.go
```diff
@@ -86,6 +87,7 @@ func TestPodCollector(t *testing.T) {
 				{
 					ObjectMeta: metav1.ObjectMeta{
 						Name:      "pod1",
+						CreationTimestamp: metav1.Time{Time: time.Unix(1500000000, 0)},
```
andyxning: ditto.

andyxning:
Overall lgtm except two nits.

brancz:
Thanks for bearing with us @julia-stripe and sorry again for my mistake in the review. Once @andyxning's last comments are addressed this lgtm as well.

As we've released the 1.0-rc.1 yesterday I'd like to hold off on merging until after we've released the final 1.0, this will land in the next minor release then.

collectors/pod_test.go
julia-stripe: fixed!

brancz:
lgtm as well, thanks a lot @julia-stripe ! Also that's what reviews are for :slightly_smiling_face:. As I mentioned before, I'll hold off on merging this until after we do the 1.0 final release, mainly for myself to avoid having to cherripick actual fixes.

julia-stripe:
looks like the 1.0 release is out now? (no hurry though)

andyxning:
@brancz Maybe it is time to merge postponed PRs before 1.0 GA.

brancz:
Absolutely @andyxning! I was sick and then at a PromCon all week so I apologize if I haven't been keeping up with everything in that time :slightly_smiling_face: . I hope to catch up with everything next week.

brancz:
Thanks @julia-stripe !

andyxning:
@brancz The body is the capital of revolution. Healthy comes first. :)

PR #184, opened 2017-07-29
add kube_pod_start_time metric

### LINKED ISSUE #163: kube_pod_info{start_time=...} documented, but unimplemented
See https://stackoverflow.com/a/44862179/717998.

andyxning:
Fix #163

This PR add `kube_pod_start_time` metric with the value of pod start time in unix timestamp.

/cc @errordeveloper @fabxc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/184)

### DIFF
--- a/collectors/pod_test.go
+++ b/collectors/pod_test.go
@@ -36,6 +36,10 @@ func TestPodCollector(t *testing.T) {
 	// Fixed metadata on type and help text. We prepend this to every expected
 	// output so we only have to modify a single place when doing adjustments.
 	var test = true
+
+	startTime := 1501569018
+	metav1StartTime := metav1.Unix(int64(startTime), 0)
+
 	const metadata = `
 		# HELP kube_pod_container_info Information about a container in a pod.
 		# TYPE kube_pod_container_info gauge
@@ -53,6 +57,8 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_container_status_waiting gauge
 		# HELP kube_pod_info Information about pod.
 		# TYPE kube_pod_info gauge
+		# HELP kube_pod_start_time Start time in unix timestamp for a pod.
+		# TYPE kube_pod_start_time gauge
 		# HELP kube_pod_owner Information about the Pod's owner.
 		# TYPE kube_pod_owner gauge
 		# HELP kube_pod_status_phase The pods current phase.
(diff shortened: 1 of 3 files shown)


matthiasr:
Shouldn't this be a separate metric? How would I use this label? As a metric, together with #182, I could for example monitor for kubelets that are slow to start pods.

andyxning:
@matthiasr Yes, you're right. Having not thought about that. :) 

I have added another `kube_pod_start_time` metric with the start time as the value. PTAL.

andyxning:
@matthiasr 

> Could you also add a test for the not-started-yet case
> Is not having this metric at all for not-started pods what we want? or should we initialise it to some well-known value (0? there wasn't any kubernetes in 1970).

I guess the value for `StartTime` is `nil` when the pod has not been started. So i add a [check about the value of `StartTime`](https://github.com/kubernetes/kube-state-metrics/pull/184/files#diff-f415922d1cb0da031b2ee1d0850f21a1R271). We will add the `kube_pod_start_time` metric only when the value of `StartTime` is not nil. 

As for the default value when `StartTime` is `nil`, i am open to give it a default value. But, i think keep `kube_pod_start_time` not recorded will be more clearer.

> remove the mention of the un-implemented label on kube_pod_info?

Good catch! Will do.

collectors/pod_test.go
```diff
@@ -267,8 +277,9 @@ func TestPodCollector(t *testing.T) {
 						NodeName: "node1",
 					},
 					Status: v1.PodStatus{
-						HostIP: "1.1.1.1",
-						PodIP:  "1.2.3.4",
+						HostIP:    "1.1.1.1",
+						PodIP:     "1.2.3.4",
+						StartTime: &unversionedStartTime,
```
brancz: for the sake of safety, could you add a test that does not have this set?

brancz:
Talked to @fabxc and it's correct to not initialize the metric with 0. We can use the `unless` operator to detect those pods: https://prometheus.io/docs/querying/operators/#logical-set-binary-operators

matthiasr:
great, thanks

brancz:
There seems to be a conflict @andyxning, could you resolve those and add a test for the nil time case? Then this lgtm.

collectors/pod_test.go
andyxning: @brancz There are two pods for test `kube_pod_info` and `kube_pod_start_time`. I have only added non `nil` `StartTime` to `pod1` and leaving `StartTime` for `pod2` uses the default value for a pointer `nil`. Thus, we have tested both cases. :)

andyxning:
@brancz Done. PTAL.

collectors/pod_test.go
```diff
@@ -36,6 +38,9 @@ func TestPodCollector(t *testing.T) {
 	// Fixed metadata on type and help text. We prepend this to every expected
 	// output so we only have to modify a single place when doing adjustments.
 	var test = true
+	startTime := time.Now()
```
brancz: Actually we're checking whether our expectation is our expectation here, could you do a constant time here and then write out the expected output?

andyxning: @brancz Done.

brancz:
lgtm thanks!

PR #185, opened 2017-07-30
bump client go to v4.0.0-beta.0 to support kubernetes 1.7

### LINKED ISSUE #183: Update client-go to support Kubernetes 1.7
I think we should update client-go to `4.0.beta.0` to support Kubernetes 1.7 since it has been released.

This should be done before 1.0 GA.

/cc @brancz @fabxc

andyxning:
Fix #183

This PR will bump client-go to `v4.0.0-beta.0` to support Kubernetes 1.7.

/cc @brancz @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/185)


brancz:
This generally seems good to me. 

Maybe @loburm can quickly comment on whether it's sane to update now, after we've run the scalability tests. As I mentioned in the issue respective to this PR I think we should do this and think it should be safe as the Kubernetes components use all of the functionality we use here and are more regularly load/scale tested.

andyxning:
@loburm can you please help re-bench kube-state-metrics against the new branch of upgrading cilent-go against Kubernetes 1.7.

brancz:
@andyxning I'm hoping that @loburm's answer will be that that's not really necessary. I imagine that the scalability tests are a time intensive task, so if not necessary I would like to not defer the release a lot longer.

andyxning:
Yep. Just like what you said in [comment](https://github.com/kubernetes/kube-state-metrics/pull/185#issuecomment-318998743). It should be ok to upgrade as kube-state-metrics only uses normal and long lived apis in client-go. Those apis should only be enhanced. :)

brancz:
I think we should move forward on this, regardless of whether we may need to re-run scalability tests this is definitely something we should do.

brancz:
lgtm thanks @andyxning !

PR #190, opened 2017-07-31
add check for nil before deference

andyxning:
This PR will check for pointer values before dereference to avoid panic dereferencing a nil pointer.

/cc @brancz @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/190)


brancz:
I wonder whether we can put some linting in place for this?

Other than that this lgtm on green.

andyxning:
Yes, we should auto check for this instead of manually. I will search for a solution for this later.

brancz:
Awesome! Great stuff! :+1:

Issue #186, opened 2017-07-31
Want to add a new node condition metric(KernelDeadlock)


WIZARD-CXY:
Hi~
     I am using node-problem-detector (https://github.com/kubernetes/node-problem-detector) for detecting any kernel and docker problems on my node. I notice there already has some node metrics(https://github.com/kubernetes/kube-state-metrics/blob/master/Documentation/node-metrics.md). I want to add more conditions like KernelDeadlock node condition into kube-state-metrics. After  a quick code searching, I found it is easy to add it to the existing system. I can mimic node ready condition https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/node.go#L60 to finish the job.For the future, I think it is nice to have more node condition type. Because it is not a default node condition in k8s, I wonder is it ok for me to add this to the kube-state-metrics?

andyxning:
@WIZARD-CXY We have discussed about adding `KernelDeadlock` condition to Kubernetes api and it is closed for the reason that npd is a side-car component for Kubernetes and npd has not any enforcement about the condition name which you ref as `KernelDeadlock`.

For more info about this discussion, [go to Kubernetes #44099](https://github.com/kubernetes/kubernetes/pull/44099)

WIZARD-CXY:
@andyxning thx

PR #193, opened 2017-08-01
Add a flexible node condition metric `kube_node_status_condition`

cofyc:

This metric use `type` label to enumerate all conditions contained in
node Status.Conditions array.

Condition is enumerable, and finite. Adding it in label, makes it
extensible. Especially, Kuberentes community already have a project
[node-problem-detector](https://github.com/kubernetes/node-problem-detector) which reports customized condition `KernelDeadlock` for cluster node.

**Special notes for your reviewer**:

If `kube-state-metrics` can expose all types of conditions, it will be very convenient to monitor all possible node conditions, whether they are reported by Kubernetes core or by third-party addons, like`node-problem-detector`.

And if new node conditions are added in [kubernetes](https://github.com/kubernetes/kubernetes) in future, they will be exposed automatically.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/193)

### DIFF
--- a/Documentation/node-metrics.md
+++ b/Documentation/node-metrics.md
@@ -17,3 +17,4 @@
 | kube_node_status_memory_pressure | Gauge | `node`=&lt;node-address&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
 | kube_node_status_disk_pressure | Gauge | `node`=&lt;node-address&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
 | kube_node_status_network_unavailable | Gauge | `node`=&lt;node-address&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
+| kube_node_status_condition | Gauge | `node`=&lt;node-address&gt; <br> `type`=&lt;node-condition-type&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
--- a/Documentation/node-metrics.md
+++ b/Documentation/node-metrics.md
@@ -5,8 +5,6 @@
 | kube_node_info | Gauge | `node`=&lt;node-address&gt; <br> `kernel_version`=&lt;kernel-version&gt; <br> `os_image`=&lt;os-image-name&gt; <br> `container_runtime_version`=&lt;container-runtime-and-version-combination&gt; <br> `kubelet_version`=&lt;kubelet-version&gt; <br> `kubeproxy_version`=&lt;kubeproxy-version&gt; |
 | kube_node_labels | Gauge | `node`=&lt;node-address&gt; <br> `label_NODE_LABEL`=&lt;NODE_LABEL&gt;  |
 | kube_node_spec_unschedulable | Gauge | `node`=&lt;node-address&gt;|
-| kube_node_status_ready| Gauge | `node`=&lt;node-address&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
-| kube_node_status_out_of_disk | Gauge | `node`=&lt;node-address&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
 | kube_node_status_phase| Gauge | `node`=&lt;node-address&gt; <br> `phase`=&lt;Pending\|Running\|Terminated&gt; |
 | kube_node_status_capacity_cpu_cores | Gauge | `node`=&lt;node-address&gt;|
 | kube_node_status_capacity_memory_bytes | Gauge | `node`=&lt;node-address&gt;|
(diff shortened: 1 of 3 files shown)


andyxning:
@cofyc Please take a look at this #186

cofyc:
hi, @andyxning 

I do not propose`kubernetes` or `kube-state-metrics` to support `KernelDeadlock` condition builtin. I suggest to expose all node conditions (node `Status.Condtions` array) in a flexible way. `kube-state-metrics` do not need to know `KernelDeadlock` condition, just expose all conditions directly.

cofyc:
@matthiasr 

If we agree on this feature. I prefer deprecating the old metrics (`kube_node_status_ready`, `kube_node_status_out_of_disk` etc), as I think metric name is better to be constant. Even we can remove the old metrics in a future release which can introduces breaking changes to reduce extra resource consumption (network traffic, etc).

brancz:
@fabxc wdyt of removing the old metrics before we do the 1.0 release? Once we do the 1.0 release just removing metrics won't be an option, so this is realistically our only chance to change for a long time. Wdyt?

fabxc:
There are certainly arguments for generating a unique name vs a type label alike. The former plays better with the naive auto-completion dahsboard and monitoring systems have today. The conditions are semantically very different – more so if custom ones are added.

I'd be interested in your opinion @brian-brazil

brian-brazil:
I think a label makes more sense, as this is basically a table as it's not really practical for users to know all potential label values in advance as there will be many that change across versions.

`type` is a very generic label name though, I'd suggest `status`given the metric name.

I'd also suggest trying to make this a simple boolean rather than true/false/file not found.

fabxc:
I'm pretty sure I discussed the very case of true/false/unknown with you back then and we concluded that `unknown` is sufficiently different from `false` to make the whole thing a tri-state enumeration – especially given that `unknown` does not indicate absence of the condition but is set to this state by the k8s API.

cofyc:
hi, @brian-brazil 

Kubernetes `Status.Conditions` array contains a list of [NodeCondition](https://kubernetes.io/docs/api-reference/v1.6/#nodecondition-v1-core) objects.

Each NodeCondition object has following two important fields:

- status: string, Status of the condition, one of True, False, Unknown.
- type: string, Type of node condition.

We are trying to expose all conditions in metrics. Currently, `kube-state-metrics` exposes builtin conditions in this format: `kube_node_status_<type_of_condition_in_snake_case>{node="...", condition="status field of condition"} [1|0]`.

My proposal in this PR is to expose all conditions in one format: `kube_node_status_condition{node="...", type="type field of condition", condition="status field of condition} [1|0]`. 

If `type` is a very generic label name, what about this format: `kube_node_status_condition{node="...", condition="type field of condition", status="status field of condition} [1|0]`? As I think it is more consistent with naming in k8s API.

/cc @brancz @fabxc

andyxning:
As for third-party components like node-problem-detector, users may change the condition name for kernel deadlock. We can not pre enumerate all the conditions with metrics.

Instead, using condition label will be more flexible. 

I am fine with what @cofyc proposed in [comment above](https://github.com/kubernetes/kube-state-metrics/pull/193#issuecomment-319335660).

brancz:
@cofyc let's remove the old metrics and just use this "generic" metric to create then extensibly for all conditions in the array.

cofyc:
@brancz @andyxning 
A new https://github.com/kubernetes/kube-state-metrics/pull/193/commits/f8becf4543217c145c400b31b043c505a0ea5c6b commit is pushed to remove old node status condition metrics.

Now, what's your opinion about label names of this new metric, which one is better?

- kube_node_status_condition{node="...", type="type field of condition", condition="status field of condition} [1|0]
- kube_node_status_condition{node="...", condition="type field of condition", status="status field of condition} [1|0]
- others?

brancz:
> kube_node_status_condition{node="...", condition="type field of condition", status="status field of condition} [1|0]

Sounds like the appropriate naming.

cofyc:
@brancz 
Done in commit https://github.com/kubernetes/kube-state-metrics/pull/193/commits/dda7a4f1e25df4d6e3a6a6f018a51b0808fdea91.

Example `/metrics` output:

```
# curl -s http://localhost:8081/metrics | grep ^kube_node_status
...
kube_node_status_condition{condition="DiskPressure",node="127.0.0.1",status="false"} 1
kube_node_status_condition{condition="DiskPressure",node="127.0.0.1",status="true"} 0
kube_node_status_condition{condition="DiskPressure",node="127.0.0.1",status="unknown"} 0
kube_node_status_condition{condition="KernelDeadlock",node="127.0.0.1",status="false"} 1
kube_node_status_condition{condition="KernelDeadlock",node="127.0.0.1",status="true"} 0
kube_node_status_condition{condition="KernelDeadlock",node="127.0.0.1",status="unknown"} 0
kube_node_status_condition{condition="MemoryPressure",node="127.0.0.1",status="false"} 1
kube_node_status_condition{condition="MemoryPressure",node="127.0.0.1",status="true"} 0
kube_node_status_condition{condition="MemoryPressure",node="127.0.0.1",status="unknown"} 0
kube_node_status_condition{condition="OutOfDisk",node="127.0.0.1",status="false"} 0
kube_node_status_condition{condition="OutOfDisk",node="127.0.0.1",status="true"} 0
kube_node_status_condition{condition="OutOfDisk",node="127.0.0.1",status="unknown"} 1
kube_node_status_condition{condition="Ready",node="127.0.0.1",status="false"} 0
kube_node_status_condition{condition="Ready",node="127.0.0.1",status="true"} 1
kube_node_status_condition{condition="Ready",node="127.0.0.1",status="unknown"} 0
...
```

brancz:
Great! Thanks for the quick review <-> code cycle :slightly_smiling_face: ! lgtm :+1:

WIZARD-CXY:
cool

Issue #124, opened 2017-04-18
1.0 stabilization


fabxc:
As discussed in the last SIG instrumentation meeting, we plan to do a first stable release of kube-state-metrics.  
As we have been mostly adding functionality for a while, rather than changing existing one, there's nothing fundamental to change here.

* [x] double-check all existing metrics for compliance with our [guidelines](https://github.com/kubernetes/community/blob/master/contributors/devel/instrumentation.md)
* [x] double-check current functionality does not conflict with future plans of fetching partial metrics, i.e. only for pod metrics for certrain deployments
* [x] load test kube-state-metrics to ensure it scales with large clusters and derive resource requirements (@piosz, can you help with that?)
* [x] provide deployment manifest that scales with cluster size using [pod nanny](https://github.com/kubernetes/contrib/tree/master/addon-resizer)

piosz:
@kubernetes/api-reviewers are there any guidelines/requirements for graduating a feature to GA which doesn't Kubernetes API?

cc @wojtek-t @gmarek re: scalability testing

smarterclayton:
I think we'd probably recommend documenting what the compatibility expectations of that feature are going forward (in a doc in that repo), make sure there is a process for API changes reasonably consistent with the use of the goals, and then make sure the feature repo contains an issue to the graduated feature.

piosz:
@loburm will help with scalability tests

brancz:
I'm now back from vacation and can help coordinate. @loburm and @piosz let me know how/when you want to tackle this.

loburm:
@brancz for the scalability testing we need to have some scenario to test against. I think we should concentrate only on testing metric related to nodes and pods mostly (I assume that other parts should consume significantly smaller amount of resources). How many nodes and pod should be present in the test scenario?

brancz:
@loburm I'm completely new to the load tests, so I suggest to start with whatever seems reasonable to you. My thoughts are the same as yours, the number of pods metrics are expected to increase linearly with the number of other objects, so focusing on those and nodes sounds perfect for our load scenarios.

Testing with the recommended upper bound of recommended pods/nodes in a single cluster would be best to see if we can actually handle this, but I'm not sure that's reasonable given that we have never performed load tests before.

piosz:
We had a chat offline and we will try to test the following scenarios:
1. 100 nodes cluster, 30pods/node
2. O(1000) nodes cluster, 30 pods/node

@loburm will verify:
[...]

PR #196, opened 2017-08-01
README: add resource recommendation

brancz:
Based on @matthiasr's [comment](https://github.com/kubernetes/kube-state-metrics/issues/124#issuecomment-318727127).

@fabxc @andyxning

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/196)

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,6 +37,20 @@ additional metrics!
 
 See the [`Documentation`](Documentation) directory for documentation of the exposed metrics.
 
+## Resource recommendation
+
+Resource usage changes with the size of the cluster. As a general rule, you should allocate
+
+* 200MiB memory
+* 0.1 cores
+
+For clusters of more than 100 nodes, allocate at least
+
+* 2MiB memory per node
+* 0.001 cores per node
+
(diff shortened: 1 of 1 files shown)


brancz:
Agreed. What I wanted to avoid is people giving their kube-state-metric 6mb request and limit. 200mb will get you very far from our findings.

matthiasr:
hmm but in the way you phrased it, I would expect that I need 206MiB for a 3 node cluster, so I'd probably add even more, and then it does get a bit excessive. What do you think about

```
Resource usage changes with the size of the cluster. As a general rule, you should allocate

* 200MiB memory
* 0.1 cores

For clusters of more than 100 nodes, allocate at least

* 2MiB memory per node
* 0.001 cores per node

These numbers are based on [scalability tests](https://github.com/kubernetes/kube-state-metrics/issues/124#issuecomment-318394185) at 30 pods per node.
```

This makes it easier for people with "small" clusters, and the math checks out the same. The minimum CPU request is higher just to make things match up nicely, if someone really cares about reclaiming those 0.09 cores they will want to see what _exactly_ they need.

brancz:
Fair enough, I like your version better :slightly_smiling_face:. Adapted.

PR #200, opened 2017-08-03
add autoscaler for kube-state-metrics

WIZARD-CXY:
For https://github.com/kubernetes/kube-state-metrics/issues/124
(provide deployment manifest that scales with cluster size using pod nanny)
This setup is tested on my gke cluster, now kube-state-metrics pod has two container, kube-state-metrics and pod-nanny
```
➜  kube-state-metrics git:(addautoscaler) ✗ kubectl get po -n kube-system
NAME                                                  READY     STATUS    RESTARTS   AGE
event-exporter-v0.1.4-4272745813-m1sqp                2/2       Running   0          6d
fluentd-gcp-v2.0-290kq                                2/2       Running   0          6d
fluentd-gcp-v2.0-7wmrd                                2/2       Running   0          6d
fluentd-gcp-v2.0-qtr9r                                2/2       Running   0          6d
heapster-v1.4.0-2764992688-s3q6v                      2/2       Running   0          6d
kube-dns-1413379277-387r0                             3/3       Running   0          6d
kube-dns-1413379277-z057j                             3/3       Running   0          6d
kube-dns-autoscaler-3880103346-73r5g                  1/1       Running   0          6d
kube-proxy-gke-cluster-1-default-pool-d10ff02d-cft7   1/1       Running   0          6d
kube-proxy-gke-cluster-1-default-pool-d10ff02d-mmfk   1/1       Running   0          6d
kube-proxy-gke-cluster-1-default-pool-d10ff02d-x4pg   1/1       Running   0          6d
kube-state-metrics-1585740284-w6lnr                   2/2       Running   0          1m
kubernetes-dashboard-1962351010-km5wg                 1/1       Running   0          6d
l7-default-backend-2954409777-rrwkq                   1/1       Running   0          6d
```
The cpu and memory parameters can be adjusted too.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/200)

### DIFF
--- a/kubernetes/kube-state-metrics-cluster-role.yaml
+++ b/kubernetes/kube-state-metrics-cluster-role.yaml
@@ -5,11 +5,8 @@ metadata:
 rules:
 - apiGroups: [""]
   resources:
-  - pods
-  verbs: ["list", "watch", "get"]
-- apiGroups: [""]
-  resources:
   - nodes
+  - pods
   - services
   - resourcequotas
   - replicationcontrollers
@@ -18,11 +15,8 @@ rules:
   verbs: ["list", "watch"]
 - apiGroups: ["extensions"]
   resources:
-  - deployments
-  verbs: ["list", "watch", "get", "update"]
-- apiGroups: ["extensions"]
-  resources:
   - daemonsets
+  - deployments
   - replicasets
   verbs: ["list", "watch"]
 - apiGroups: ["apps"]
@@ -34,4 +28,3 @@ rules:
   - cronjobs
   - jobs
   verbs: ["list", "watch"]
-
--- a/kubernetes/kube-state-metrics-cluster-role.yaml
+++ b/kubernetes/kube-state-metrics-cluster-role.yaml
@@ -5,8 +5,11 @@ metadata:
 rules:
 - apiGroups: [""]
   resources:
-  - nodes
   - pods
+  verbs: ["list", "watch", "get"]
+- apiGroups: [""]
+  resources:
+  - nodes
   - services
   - resourcequotas
   - replicationcontrollers
@@ -15,8 +18,11 @@ rules:
   verbs: ["list", "watch"]
 - apiGroups: ["extensions"]
   resources:
-  - daemonsets
   - deployments
+  verbs: ["list", "watch", "get", "update"]
+- apiGroups: ["extensions"]
+  resources:
+  - daemonsets
   - replicasets
   verbs: ["list", "watch"]
 - apiGroups: ["apps"]
--- a/kubernetes/kube-state-metrics-role.yaml
+++ b/kubernetes/kube-state-metrics-role.yaml
@@ -1,0 +1,16 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: Role
+metadata:
+  namespace: kube-system
+  name: kube-state-metrics-resizer
+rules:
+- apiGroups: [""]
+  resources:
+  - pods
+  verbs: ["get"]
+- apiGroups: ["extensions"]
+  resources:
+  - deployments
+  resourceNames: ["kube-state-metrics"]
+  verbs: ["get", "update"]
+
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -4,7 +4,7 @@ metadata:
   name: kube-state-metrics
   namespace: kube-system
 spec:
-  replicas: 2
+  replicas: 1
   template:
     metadata:
       labels:
@@ -30,4 +30,30 @@ spec:
           limits:
             memory: 50Mi
             cpu: 200m
-
+      - name: pod-nanny
+        image: gcr.io/google_containers/addon-resizer:1.0
+        resources:
+          limits:
+            cpu: 100m
+            memory: 30Mi
+          requests:
+            cpu: 100m
+            memory: 30Mi
+        env:
+          - name: MY_POD_NAME
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.name
+          - name: MY_POD_NAMESPACE
+            valueFrom:
+              fieldRef:
+                fieldPath: metadata.namespace
+        command:
+          - /pod_nanny
+          - --container=kube-state-metrics
+          - --cpu=100m
+          - --extra-cpu=1m
+          - --memory=30Mi
+          - --extra-memory=2Mi
+          - --threshold=5
+          - --deployment=kube-state-metrics
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -30,7 +30,7 @@ spec:
           limits:
             memory: 50Mi
             cpu: 200m
-      - name: pod-nanny
+      - name: addon-resizer
         image: gcr.io/google_containers/addon-resizer:1.0
         resources:
           limits:
--- a/kubernetes/kube-state-metrics-role-binding.yaml
+++ b/kubernetes/kube-state-metrics-role-binding.yaml
@@ -1,0 +1,14 @@
+apiVersion: rbac.authorization.k8s.io/v1beta1
+kind: RoleBinding
+metadata:
+  name: kube-state-metrics
+  namespace: kube-system
+roleRef:
+  apiGroup: rbac.authorization.k8s.io
+  kind: Role
+  name: kube-state-metrics-resizer
+subjects:
+- kind: ServiceAccount
+  name: kube-state-metrics
+  namespace: kube-system
+


WIZARD-CXY:
I think since the nanny scales resources linearly with the number of nodes in the cluster, I can use small starting resource for kube-state-metrics container. Here I use request resource for resource base requirement in nanny.

brancz:
Having never used the nanny pod, this might be a incorrect, but I'd expect it to change the deployment manifest, meaning it needs the RBAC roles for it, no?

brancz:
I looked at the addon-resizer code and from what I can tell it needs:

* nodes: list, watch
* deployments: get, update
* pods: get

What that would basically mean is that we need to add the last two points to our RBAC roles.

WIZARD-CXY:
Good point, but I don't add the last two, it works... and the example in https://github.com/kubernetes/autoscaler/tree/master/addon-resizer
is a simple deployment without any other RBAC config.Maybe it exists some tricks

brancz:
Depending on the `ServiceAccount` you're using, you might have given your pod cluster admin roles as it's in the kube-system namsepace.

@piosz @loburm are you aware of the RBAC roles the addon-resizer needs? From looking at the code it doesn't seem to do anything special so I'd expect it to require it.

WIZARD-CXY:
@brancz I started a new namespace and deployed pod-nanny, it complains it can't get pod and update  deployment due to limited permission.So maybe is it due to kube-system(a special ns)?

brancz:
Yes that was my suspicion. Would you care to adapt? :slightly_smiling_face:

WIZARD-CXY:
you mean add 
```
deployment: update
pods: get

`` `
in this file https://github.com/kubernetes/kube-state-metrics/blob/master/kubernetes/kube-state-metrics-cluster-role.yaml?

**Comment by brancz** [comment]
Yes, but also get for deployments, as currently it's just list/watch.

**Comment by WIZARD-CXY** [comment]
yeah, exactly what i want to say due to this func
https://github.com/kubernetes/autoscaler/blob/master/addon-resizer/nanny/kubernetes_client.go#L69

**Comment by WIZARD-CXY** [comment]
Changed according to your suggestion,PTAL @brancz.Tested ok in  my cluster.

File: kubernetes/kube-state-metrics-cluster-role.yaml [hunk 1]
```diff
@@ -15,6 +18,10 @@ rules:
   verbs: ["list", "watch"]
 - apiGroups: ["extensions"]
   resources:
+  - deployments
+  verbs: ["list", "watch", "get", "update"]
+- apiGroups: ["extensions"]
+  resources:
   - daemonsets
   - deployments
```
brancz: this can be removed as it's redundant with the above rule

WIZARD-CXY: sorry will do

WIZARD-CXY:
I read the doc https://kubernetes.io/docs/admin/authorization/rbac/, and find that "Default RBAC policies grant scoped permissions to control-plane components, nodes, and controllers, but grant no permissions to service accounts outside the “kube-system” namespace (beyond discovery permissions given to all authenticated users)." 
maybe service account in kube-system namespace is privileged just as our kube-state-metrics sa in kube-system.

WIZARD-CXY:
maybe @piosz @loburm  can give an answer?

brancz:
We actually run kube-state-metrics not in the kube-system namespace, so we don't really have that problem, or rather therefore we do have this very problem. Yes `kube-system` is a special namespace, and that's sort of why we chose it to be the example namespace for this component as we can always assume it's there. If you want to we can change the example manifests to deploy to the default namespace instead, maybe that makes the need for RBAC more clear, wdyt?

WIZARD-CXY:
I think it should be in kube-system, no need to change IMHO

brancz:
Agreed, but it's still up to the user, so RBAC roles should be properly documented.

piosz:
@kubernetes/sig-api-machinery-pr-reviews for rbac stuff

piosz:
Also we use pod-nanny with Heapster already. I'm not sure where we can find rbac rules for it.

FYI @kubernetes/sig-autoscaling-misc

kubernetes/kube-state-metrics-cluster-role.yaml
```diff
@@ -15,8 +18,11 @@ rules:
   verbs: ["list", "watch"]
 - apiGroups: ["extensions"]
   resources:
-  - daemonsets
   - deployments
+  verbs: ["list", "watch", "get", "update"]
```
liggitt: Letting this update any replicaset and daemonset in the system doesn't seem appropriate. I'd expect those permissions limited to the namespace and specific replicaset/daemonset it is responsible for. Also, the update permission belongs in a separate role specific to the nanny, since someone who is not using the nanny should not need to give write permission to their metrics pod

WIZARD-CXY: replicaset and daemonset are just given list and watch permission, no update.

liggitt: Ah, sorry, misread the diff. Same comment applies to deployments re update permission just on this particular deployment in this namespace

brancz: In that case we need to separate the list/watch from the get and update as list/watch needs to be on all namespaces for the purpose of collecting the objects to produce the metrics of those. Thus get and update only need to be on that specific namespace.

liggitt: > In that case we need to separate the list/watch from the get and update as list/watch needs to be on all namespaces for the purpose of collecting the objects to produce the metrics of those. Thus get and update only need to be on that specific namespace.

exactly, in a Role in the `kube-system` namespace, bound with a `RoleBinding`. the rule should also specify `resourceNames:["kube-state-metrics"]` to only allow modifying the relevant deployment.

WIZARD-CXY: good point

WIZARD-CXY: But I think maybe it's a little too tight, from the code we can see that pod-nanny only get and update the relevant deployment we specified. I think we can trust it.

brancz: That's less about not trusting the code/binary of the addon-resizer, than an attacker being able to modify arbitrary deployments.

WIZARD-CXY:
So I will submit another commit to fix the rbac issue

WIZARD-CXY:
@liggitt @brancz. I add Role and Rolebinding according to your suggestion.PTAL

kubernetes/kube-state-metrics-role.yaml
```diff
@@ -0,0 +1,17 @@
...
+- apiGroups: [""]
+  resources:
+  - pods
+  resourceNames: ["kube-state-metrics"]  
+  verbs: ["get"]
+- apiGroups: ["extensions"]
+  resources:
+  - deployments
+  resourceNames: ["kube-state-metrics"]
+  verbs: ["get", "update"]
```
liggitt: Might want to allow patch as well… not sure how the nanny updates

WIZARD-CXY: the nanny updates is using https://github.com/kubernetes/autoscaler/blob/master/addon-resizer/nanny/kubernetes_client.go#L69

WIZARD-CXY: just using get and update

WIZARD-CXY:
@liggitt another question here: If I already have a role to give "list pod" permission, is it right that I already have "get a specific pod" permission?

WIZARD-CXY:
@brancz I found that we actually don't need to add "get pods" role even if we deploy kube-state-metrics all in another fresh-new namespace.Like below, I deploy it in monitor namspace
```

➜  kube-state-metrics git:(addautoscaler) ✗ kubectl get po -n monitor
NAME                                  READY     STATUS    RESTARTS   AGE
kube-state-metrics-2057856429-k0t4d   2/2       Running   0          1m
```

brancz:
I would expect this to cause errors at runtime, did you have a look at the logs of the addon-resizer? The actions should be distinct.

WIZARD-CXY:
I looked at the log of addon-resizer, it didn't complain. Below is its log.

The addon-resizer actually changed the kube-state-metrics resource configuration from cpu 100 to 103 and mem 30 to 36Mi



You can test my pr in your environment and see if anything change. My env is gke k8s 1.7.2

brancz:
@ericchiang @liggit is this intended, that get/list RBAC actions are both allowed when only one of them is specified?

liggitt:
> @ericchiang @liggitt is this intended, that get/list RBAC actions are both allowed when only one of them is specified?

That is not the case. The clusterrole already grants list pods to the service account, so anything running with that service account would have list permission. If the resizer is doing a get of a pod, either RBAC (or another configured authorizer) is allowing an explicit get. Did you set up your gke cluster with legacy abac permissions disabled?

WIZARD-CXY:
@liggitt no extra config, I just use default setting

WIZARD-CXY:
Do you know how to get apiserver start configuration on gke? I can't find it anywhere, maybe google hide it...

liggitt:
> Do you know how to get apiserver start configuration on gke? I can't find it anywhere, maybe google hide it...

I don't think it's accessible. Did you start your cluster passing `--no-enable-legacy-authorization` to gcloud? If not, I think you still have legacy ABAC policy in place that makes every service account a cluster admin

WIZARD-CXY:
@liggit, I checked the configuration is "Legacy authorization	
Enabled", the default value is enabled as u can see below

WIZARD-CXY:
Finally got these authorization thing, thanks to @brancz @liggitt. add "get pod" rule in pod-nanny role.PTAL

brancz:
Happy we could figure it out, lgtm now :+1: . I'll hold off on a final lgtm from @liggitt though.

liggitt:
seems fine as a starting point. as a test, I'd recommend deploying the addon-resizer in its own pod with its own service account, and seeing if the role is sufficient.

WIZARD-CXY:
@liggitt If we deploy addon-resizer in its own pod with its own sa, we also need a cluster role for list nodes permission due to it must get the cluster node number to determine whether to scale or not. Tested on my cluster with this. Now it is using the kube-state-metrics clusterrole to work.

liggitt:
> we also need a cluster role for list nodes permission due to it must get the cluster node number to determine whether to scale or not. Tested on my cluster with this. Now it is using the kube-state-metrics clusterrole to work.

that's worth noting, but given the Role for the addon-resizer is pretty specific to kube-state-metrics, granting the kube-state-metrics clusterrole to the resizer wouldn't be terrible. the important thing is to keep the deployment mutation permissions separate so that someone running without the addon-resizer wouldn't be required to give their metrics pod permission to modify the deployment. this PR does that, so I think it's fine as-is.

WIZARD-CXY:
fixed @brancz ptal

WIZARD-CXY:
tested ok on my cluster

brancz:
Great thanks! I think we can move forward and get 1.0 out!

PR #201, opened 2017-08-03
add namespace cmd arg

### LINKED ISSUE #152: Get metrics only for one namespace
Dear all,

I try to deploy monitoring on a CaaS. I'm limited to one namespace on a Kubernetes cluster and I try to monitor deployments, replicas, respawn in this namespace.
When I try to deploy kube-state-metrics I have a lot of error because it list all cluster objects : 
_User *** cannot list all pods in the cluster_

Is it possible to limit listing activity only for one or some namespace depit of all cluster?

Regards

andyxning:
Fix #152

This PR add `namespaces` command line argument to kube-state-metrics. It's value is in the same format as `collectors`. If specified with delimited namespaces, kube-state-metrics will only collect resources in those namespaces.

/cc @brancz @matthiasr

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/201)

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -26,7 +26,7 @@ import (
 	"golang.org/x/net/context"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/pkg/api"
+
 	v2batch "k8s.io/client-go/pkg/apis/batch/v2alpha1"
 	"k8s.io/client-go/tools/cache"
 )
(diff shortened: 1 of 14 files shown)


brancz:
There are loads of things going on here, let's first look at the problem again.

Issue #152 describes that the problem is that the caches are actually trying to request objects from all namespaces, however, this implementation still requests objects from all namespaces, but then filters those out that are not in the given namespace.

The solution is actually much more simple than what you proposed here. The caches already take the namespace they should run in as an argument, so all that has to be done is pass the namespace from a flag, and then pass it to the collector.

Aside from that I appreciate the attempt to refactor some things, but I'd prefer to separate refactoring from feature development (at least by commits). In terms of this feature though I'd prefer to keep it as much on track as possible and avoid the refactoring here as all of this is very core to kube-state-metrics working reliably.

andyxning:
Yep. @brancz you are right. I have implemented it wrongly. :( 

Will correct it later.

andyxning:
@brancz I have updated the implementation. PTAL.

BTW, in the newest implementation, i have updated all namespace argument list and watch except `node` which is not binding to any namespace.

andyxning:
Maybe this pr will also be accept after 1.0 GA. @brancz

brancz:
lgtm now :+1: nice! Yes we'll hold off until the 1.0. Nice job though.

zouyee:
sounds good, something like that [`Namespace to watch for scrapy. Default is to watch all namespaces` ](https://github.com/kubernetes/ingress/tree/master/controllers/nginx#command-line-arguments)

andyxning:
@brancz I have rebased the code. PTAL.

brancz:
Turns out the wonderful interface for the register funcitons is not as good as I thought :slightly_smiling_face: that's totally not your fault but if it becomes confusing we might want to think about functional arguments or something like that to make this more clean.

But these changes lgtm here :+1:

PR #203, opened 2017-08-03
Split the UT and Build in different Stages

asifdxtreme:
Split the UT and Build Stages in different jobs of travis CI so that it's easy for user to trace what went wrong in CI

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/203)


brancz:
Nice! Thanks! lgtm :+1:

PR #206, opened 2017-08-04
gofmt

asifdxtreme:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/206)


brancz:
gofmt should not be run on the vendor directory, aside from that, do you maybe want to setup running gofmt on every run on travis? That way we don't have to make sure ourselves.

andyxning:
Maybe we should add another stage `Fmt code` in travis CI.

asifdxtreme:
@brancz @andyxning  yes adding this check in travis CI will be better so that non-formatted code does not pass the check, Thanks for your suggestion, I will add this check in travis conf

asifdxtreme:
Raised a new PR with the suggested changes https://github.com/kubernetes/kube-state-metrics/pull/210

PR #211, opened 2017-08-07
Add godoc reference

asifdxtreme:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/211)


matthiasr:
This isn't a library, is Godoc useful for that?

asifdxtreme:
@matthiasr At present it might not be much useful but when the code size grows then it will easy to browse the interfaces or functions

andyxning:
I am not apposed with this PR. IMO, i agree with @mattfarina . Godoc is useful for a golang lib instead of a binary. :)

brancz:
Godoc still exists even if we don't have the badge, so I'd argue that this would indeed confuse people and possibly make it look like we encourage that type of usage.

I'd say we don't move forward with this. Feel free to comment if you strongly disagree. I'm happy to be shown otherwise, especially regarding:

> when the code size grows then it will easy to browse the interfaces or functions

Currently this codebase is quite small and uncomplicated, and I'm hoping to have the same opinion in the future :slightly_smiling_face: .

PR #214, opened 2017-08-09
Update usage docs with cli args

unguiculus:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/214)

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,6 +103,26 @@ make container
 Simply build and run kube-state-metrics inside a Kubernetes pod which has a
 service account token that has read-only access to the Kubernetes cluster.
 
+The kube-state-metrics binary takes a number of arguments.
+
+```
+$ kube-state-metrics --help
+Usage of kube-state-metrics:
+      --apiserver string                 The URL of the apiserver to use as a master
+      --collectors map[string]struct{}   Collectors to be enabled
+  -h, --help                             Print help text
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -103,6 +103,26 @@ make container
...
+Usage of kube-state-metrics:
+      --apiserver string                 The URL of the apiserver to use as a master
+      --collectors map[string]struct{}   Collectors to be enabled
+  -h, --help                             Print help text
+      --in-cluster                       If true, use the built in kubernetes cluster for creating the client (default true)
+      --kubeconfig string                Absolute path to the kubeconfig file
+      --port int                         Port to expose metrics on. (default 80)
+```
+
+In most cases, it is fine to run it without any arguments. By default, [these collectors](https://github.com/kubernetes/kube-state-metrics/blob/master/main.go#L45-L59) are enabled. If you don't need all of them or
```
zouyee: something like `--feature-gates` is needed.

brancz: @zouyee I think what you're describing already exists with the `--collectors=...` flag described in this PR.

andyxning: How about we only document and detail `collectors` functionality. List and maintain all the command line arguments is a heavy job.

brancz: I remember there was a tool to run go snippets and include it in docs, I can't find it right now though.

PR #158, opened 2017-07-03
split deploy configuration

zouyee:
mkdir deployWithRBAC and deployWithoutRBAC

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/158)

PR #217, opened 2017-08-10
add deploy file with no rbac config

zouyee:
merge files and add deploy file with no rbac config

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/217)


zouyee:
@brancz PTAL

brancz:
I still stand by [my comment on this](https://github.com/kubernetes/kube-state-metrics/pull/158#issuecomment-315775653) PR. All production Kubernetes offerings I am aware of have RBAC on by default, this gives the wrong incentive to officially support non RBAC environments in my opinion. Why do you think it's important to have this in the repo? I'd expect everyone even without RBAC enabled to be able to run kube-state-metrics just as well with the already provided manifests.

zouyee:
ok, i miss sth.

PR #218, opened 2017-08-10
add gpu resource monitoring

### LINKED ISSUE #216: [feature] add pod gpu resource limit and request monitoring
Hi~
   Our team use gpu for ml task and we think it is nice to have gpu resource monitoring in kube-state-metrics, Now I find the doc, only to find it doesn't have yet。I'm happy to submit a pr to add this feature

WIZARD-CXY:
fix #216

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/218)

### DIFF
--- a/Documentation/pod-metrics.md
+++ b/Documentation/pod-metrics.md
@@ -19,3 +19,6 @@
 | kube_pod_container_resource_requests_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_container_resource_limits_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_container_resource_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
+| kube_pod_container_resource_requests_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
+| kube_pod_container_resource_limits_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
+
--- a/collectors/pod.go
+++ b/collectors/pod.go
@@ -143,6 +143,18 @@ var (
 		"The limit on memory to be used by a container in bytes.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
+
+	descPodContainerResourceRequestsNvidiaGPUDevices = prometheus.NewDesc(
+		"kube_pod_container_resource_requests_nvidia_gpu_devices",
+		"The number of requested gpu devices by a container.",
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
+
+	descPodContainerResourceLimitsNvidiaGPUDevices = prometheus.NewDesc(
+		"kube_pod_container_resource_limits_nvidia_gpu_devices",
+		"The limit on gpu devices to be used by a container.",
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
 )
 
 type PodLister func() ([]v1.Pod, error)
@@ -196,6 +208,8 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerResourceRequestsMemoryBytes
 	ch <- descPodContainerResourceLimitsCpuCores
 	ch <- descPodContainerResourceLimitsMemoryBytes
+	ch <- descPodContainerResourceRequestsNvidiaGPUDevices
+	ch <- descPodContainerResourceLimitsNvidiaGPUDevices
 }
 
 func extractCreatedBy(annotation map[string]string) *api.ObjectReference {
@@ -335,13 +349,22 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 				c.Name, nodeName)
 		}
 
+		if gpu, ok := req[v1.ResourceNvidiaGPU]; ok {
+			addGauge(descPodContainerResourceRequestsNvidiaGPUDevices, float64(gpu.Value()), c.Name, nodeName)
+		}
+
 		if cpu, ok := lim[v1.ResourceCPU]; ok {
 			addGauge(descPodContainerResourceLimitsCpuCores, float64(cpu.MilliValue())/1000,
 				c.Name, nodeName)
 		}
+
 		if mem, ok := lim[v1.ResourceMemory]; ok {
 			addGauge(descPodContainerResourceLimitsMemoryBytes, float64(mem.Value()),
 				c.Name, nodeName)
+		}
+
+		if gpu, ok := lim[v1.ResourceNvidiaGPU]; ok {
+			addGauge(descPodContainerResourceLimitsNvidiaGPUDevices, float64(gpu.Value()), c.Name, nodeName)
 		}
 	}
 }
--- a/collectors/pod_test.go
+++ b/collectors/pod_test.go
@@ -78,6 +78,10 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_container_resource_limits_cpu_cores gauge
 		# HELP kube_pod_container_resource_limits_memory_bytes The limit on memory to be used by a container in bytes.
 		# TYPE kube_pod_container_resource_limits_memory_bytes gauge
+		# HELP kube_pod_container_resource_requests_nvidia_gpu_devices The number of requested gpu devices by a container.
+		# TYPE kube_pod_container_resource_requests_nvidia_gpu_devices gauge
+		# HELP kube_pod_container_resource_limits_nvidia_gpu_devices The limit on gpu devices to be used by a container.
+		# TYPE kube_pod_container_resource_limits_nvidia_gpu_devices gauge
 	`
 	cases := []struct {
 		pods    []v1.Pod
@@ -437,12 +441,14 @@ func TestPodCollector(t *testing.T) {
 								Name: "pod1_con1",
 								Resources: v1.ResourceRequirements{
 									Requests: map[v1.ResourceName]resource.Quantity{
-										v1.ResourceCPU:    resource.MustParse("200m"),
-										v1.ResourceMemory: resource.MustParse("100M"),
+										v1.ResourceCPU:       resource.MustParse("200m"),
+										v1.ResourceMemory:    resource.MustParse("100M"),
+										v1.ResourceNvidiaGPU: resource.MustParse("3"),
 									},
 									Limits: map[v1.ResourceName]resource.Quantity{
-										v1.ResourceCPU:    resource.MustParse("200m"),
-										v1.ResourceMemory: resource.MustParse("100M"),
+										v1.ResourceCPU:       resource.MustParse("200m"),
+										v1.ResourceMemory:    resource.MustParse("100M"),
+										v1.ResourceNvidiaGPU: resource.MustParse("3"),
 									},
 								},
 							},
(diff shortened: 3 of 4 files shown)


WIZARD-CXY:
@brancz PTAL

WIZARD-CXY:
will this feature be added to 1.0?

collectors/pod_test.go
```diff
@@ -468,25 +476,29 @@ func TestPodCollector(t *testing.T) {
...
+										v1.ResourceCPU:       resource.MustParse("500m"),
+										v1.ResourceMemory:    resource.MustParse("400M"),
+										v1.ResourceNvidiaGPU: resource.MustParse("5"),
 									},
 									Limits: map[v1.ResourceName]resource.Quantity{
-										v1.ResourceCPU:    resource.MustParse("500m"),
-										v1.ResourceMemory: resource.MustParse("400M"),
+										v1.ResourceCPU:       resource.MustParse("500m"),
+										v1.ResourceMemory:    resource.MustParse("400M"),
+										v1.ResourceNvidiaGPU: resource.MustParse("5"),
```
matthiasr: could/should this also be non-integer values? if so, can you mix things up a bit so the tests also cover that?

collectors/pod.go
```diff
@@ -137,6 +137,18 @@ var (
 		"The limit on memory to be used by a container in bytes.",
 		[]string{"namespace", "pod", "container", "node"}, nil,
 	)
+
+	descPodContainerResourceRequestsGPU = prometheus.NewDesc(
+		"kube_pod_container_resource_requests_gpu",
```
matthiasr: since the resources in Kubernetes are specifically _NVidia_ GPUs, that should be reflected here.

collectors/pod_test.go
WIZARD-CXY: gpu is unlike cpu,"it is not possible to request a portion of a GPU " from k8s gpu scheduling doc https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/

collectors/pod.go
WIZARD-CXY: I think it is ok to just use gpu, maybe we can add a label to specify the gpu vendor.Adding a vendor name to metric name may be kinda of wierd IMHO.

andyxning: As for recording gpu vendor, a `vendor` label should be fine.

collectors/pod_test.go
andyxning: As for now, integer should be fine enough. :)

collectors/pod.go
andyxning: @WIZARD-CXY Be consistent with cpu cpu and memory metric name is preferred. 

`kube_pod_container_resource_requests_gpu` -> `kube_pod_container_resource_requests_gpu_cores`
`kube_pod_container_resource_limits_gpu ` -> `kube_pod_container_resource_limits_gpu_cores`

collectors/pod.go
```diff
@@ -137,6 +137,18 @@ var (
...
+		"The number of requested gpu by a container.",
```
andyxning: ditto.

`The number of requested gpu by a container.` -> `The number of requested gpu cores by a container.`

collectors/pod.go
```diff
@@ -137,6 +137,18 @@ var (
...
+		[]string{"namespace", "pod", "container", "node"}, nil,
+	)
+
+	descPodContainerResourceLimitsGPU = prometheus.NewDesc(
+		"kube_pod_container_resource_limits_gpu",
+		"The limit on gpu to be used by a container.",
```
andyxning: ditto.

`The limit on gpu to be used by a container.` -> `The limit on gpu cores to be used by a container.`

collectors/pod.go
```diff
@@ -324,13 +338,22 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 				c.Name, nodeName)
 		}
 
+		if gpu, ok := req[v1.ResourceNvidiaGPU]; ok {
```
andyxning: According to the [doc](https://kubernetes.io/docs/tasks/manage-gpus/scheduling-gpus/), 
```
GPUs can be specified in the limits section only
```
Request nvidia cpu will be none exist for now, request nvidia cpu metric seems useless. @WIZARD-CXY

collectors/pod.go
WIZARD-CXY: @andyxning gpu is not like cpu, one gpu  has many cores, so we should not use cores here, it actually denotes one gpu device afaik.

collectors/pod.go
WIZARD-CXY: ditto

collectors/pod.go
WIZARD-CXY: @andyxning I have same question with you before. After some digging, I got resp when you get the pod from the apiserver, it will have request as below.

So for consistency I add gpu resource request record here

WIZARD-CXY: It is ok for me to remove "gpu resource request metric", it is not a big deal.
Actually u can see from the code, the doc maybe some kinda wrong. It is ok to have request as long as it is same with limit.
https://github.com/kubernetes/kubernetes/blob/master/pkg/api/validation/validation.go#L3717

andyxning: Seems like apiserver copies the value setted for limits to requests?

WIZARD-CXY: "Seems like apiserver copies the value setted for limits to requests?"
not seeing the code doing this work from a quick search.
 But I remember long time ago (maybe k8s can use gpu feature just come up)the "GPUs can be specified in the limits section only" hasn't be included in the doc.maybe it is added recently.

WIZARD-CXY:
@andyxning any other problems?

collectors/pod.go
andyxning: According to https://github.com/kubernetes/kubernetes/blob/master/pkg/api/validation/validation_test.go#L3295-L3329, it seems that apiserver allows both request for gpu not setted and setted.

collectors/pod.go
andyxning: @WIZARD-CXY Thanks for providing correct info!

> it actually denotes one gpu device afaik.

Can this be renamed to `kube_pod_container_resource_limits_gpu_devices`? This is only my thought and i am opening to other thoughts. The intention behind this is that we need to make the unit for a metric clear. :)

andyxning:
@WIZARD-CXY Comments updated.

collectors/pod.go
WIZARD-CXY: get your intention,I am ok with kube_pod_container_resource_limits_gpu_devices

WIZARD-CXY: @others if other people do not have second opinion on this, I will change to kube_pod_container_resource_limits_gpu_devices today

brancz: If I understand correctly this should end with the suffix `_total`, as this is the count of devices available, correct? That suffix is Prometheus metric naming convention. https://prometheus.io/docs/practices/naming/

WIZARD-CXY: u mean kube_pod_container_resource_limits_gpu_total?

brancz: Actually I made a mistake here, the limit can be changed in all directions so the `_total` suffix is actually not appropriate to be used here.

brancz: `kube_pod_container_resource_limits_gpu_devices` sounds good to me

WIZARD-CXY: ok

WIZARD-CXY:
@andyxning I don't add vendor label in this pr, because I think it is enough for now(just nvidia here), If we have more gpu vendor used in k8s, we can easily add it in the future.

andyxning:
@WIZARD-CXY I am fine with not adding gpu vendor as a label and not testing the gpu value with float. As for now, we are compatible with the latest Kubernetes release(correct me if i am wrong), right? :) 

As Kubernetes evolves, we should and will make kube-state-metrics keep up with it.

brancz:
Are you familiar with what the future around gpu limits and requests are? If it's unclear at this point I'd actually to have the nvidia part in here somewhere, and if we go by the existing metrics that would mean it needs to be included in the name. Given the alpha stage of this feature having a very specific metric rather than one would be more appropriate to deprecate at a later point instead of changing the meaning of a general "gpu" metric (as we don't know at this point what this feature is going to look like in the future in Kubernetes).

WIZARD-CXY:
Maybe @vishh can answer brancz's question

vishh:
The Resource Type for Nvidia GPUs will be changing in the next release. It's an alpha feature. I'd recommend holding off on integration with the rest of the ecosystem tools until we graduate GPU support out of alpha, which is expected to happen in by kubernetes v1.9 (~November '17)

WIZARD-CXY:
According to vishh said above, we should wait a bit longer(about 3 months) to get this gpu monitoring to work. I think it is a long wait and I can't wait that long so I will get it work in my company. I can't agree more  as @andyxning suggested 
"As Kubernetes evolves, we should and will make kube-state-metrics keep up with it." 
So my suggestion is adding nvidia as a gpu vendor in the label for now and  change it as the gpu support graduated in the k8s 1.9. I'm open to other opinions.:)

andyxning:
> According to vishh said above, we should wait a bit longer(about 3 months) to get this gpu monitoring to work.

Agree with @WIZARD-CXY, it is too long. AFAIK, many users has adopted this GPU feature event it is still in alpha phase.

 > So my suggestion is adding nvidia as a gpu vendor in the label for now and change it as the gpu support graduated in the k8s 1.9. 

@WIZARD-CXY I think we should accept the advise from @brancz and @matthiasr, i.e., adding "nvidia" to the metric name. Using about `kube_pod_container_resource_requests_nvidia_gpu_devices` and `kube_pod_container_resource_limits_nvidia_gpu_devices`. As for now this is more compitable with current gpu resource name `alpha.kubernetes.io/nvidia-gpu`.

If later Kubernetes changes the gpu resource name from `ResourceNvidiaGPU` to something like `ResourceGPU`(just in my assumption), we can then use `kube_pod_container_resource_limits_gpu_devices` and `kube_pod_container_resource_requests_gpu_devices` with gpu vendor labeled if supported. And, we can deprecate alpha support metrics with the new generic ones. Deprecating metrics is less fragile than changing an existing ones.

WIZARD-CXY:
@andyxning agree

WIZARD-CXY:
@andyxning @brancz @matthiasr  change metric name to kube_pod_container_resource_requests_nvidia_gpu_devices and kube_pod_container_resource_limits_nvidia_gpu_devices. PTAL

andyxning:
nit: squash commits. :)

Others sgtm.

WIZARD-CXY:
I want to hear @brancz's or @matthiasr's opinion on whether waiting or just using this gpu metrics now?

WIZARD-CXY:
@andyxning after all is set, I will squash the commits

brancz:
I'm fine with introducing it now, but we need documentation and we need do explicitly say that this is a metric that is using alpha stage data and is already deprecated as it's planned to be removed once the "real" GPU resources are here. This needs to be clearly stated in the documentation.

andyxning:
Agreed with @brancz . @WIZARD-CXY Could you please add doc about these metrics.

WIZARD-CXY:
ok

WIZARD-CXY:
@andyxning @brancz updated.I add a second commit to fix gofmt issue in some other code to make gofmtcheck happy, otherwise I won't pass the travis ci.

brancz:
I merged the PR that was blocking this. Please rebase again, then this should be good to go.

Documentation/pod-metrics.md
```diff
@@ -19,3 +19,6 @@
 | kube_pod_container_resource_requests_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_container_resource_limits_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_container_resource_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
+| kube_pod_container_resource_requests_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
+| kube_pod_container_resource_requests_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
```
brancz: small typo, the second entry should be limits instead of requests

WIZARD-CXY:
@brancz fixed typo, PTAL

brancz:
as mentioned in my previous comment, this PR needs a rebase to ensure [these changes](https://github.com/kubernetes/kube-state-metrics/pull/229) are included

WIZARD-CXY:
@brancz rebased ptal

andyxning:
lgtm

brancz:
Thanks a lot for baring with us :slightly_smiling_face: ! lgtm :+1:

WIZARD-CXY:
Glad we can make kube-state-metrics better and better!

thomasjungblut:
sorry to warm this up, I was just about to use this metric to compute how many GPUs are used within the cluster. I can select by namespace, but not by pod status- which means all finished pods that still remain in the cluster are returned by that metric too. 

How hard would it be to add this as another label on top of this?

brancz:
You need to perform a join with the condition metric that describes the Pod's state, then you can select GPU metrics of running pods.

Issue #212, opened 2017-08-08
Alpha features should not be used by default


unguiculus:
v1.0.0-rc.1 uses alpha features which may not be available for every one. At least, they are not in our case on GKE. It doesn't seem to be do harm, but the log is flooded with this:

```
E0808 12:43:47.637347       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:48.739642       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:49.742757       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:50.745885       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:51.749028       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:52.752262       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:53.755323       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:54.839768       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:55.939709       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
E0808 12:43:56.943320       1 reflector.go:201] k8s.io/kube-state-metrics/collectors/cronjob.go:86: Failed to list *v2alpha1.CronJob: the server could not find the requested resource
```

Please make the use of alpha feature configurable. Ideally they should be turned off by default.

brancz:
[...]

PR #220, opened 2017-08-14
checkout kubernetes cluster supported feature

zouyee:
https://github.com/kubernetes/kube-state-metrics/issues/212

This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/220)

### DIFF
--- a/main.go
+++ b/main.go
@@ -29,6 +29,8 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/spf13/pflag"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	utilruntime "k8s.io/apimachinery/pkg/util/runtime"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/pkg/api"
 	"k8s.io/client-go/rest"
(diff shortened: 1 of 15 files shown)


zouyee:
@brancz

brancz:
Nice! For backward compatibility we will have to put this behind a flag and only use it when explicitly enabled. Furthermore I'd prefer to introduce a map that maps the api group + version to the collector identifier, that way we can keep using the available collectors list for the `--collectors=...` flag as used previously, and only have to change a single string in case an API is promoted/deprecated.

WIZARD-CXY:
hmm agreed with @brancz

main.go
```diff
@@ -43,34 +45,34 @@ const (
...
-		"nodes":                  collectors.RegisterNodeCollector,
-		"pods":                   collectors.RegisterPodCollector,
-		"replicasets":            collectors.RegisterReplicaSetCollector,
-		"replicationcontrollers": collectors.RegisterReplicationControllerCollector,
-		"resourcequotas":         collectors.RegisterResourceQuotaCollector,
-		"services":               collectors.RegisterServiceCollector,
-		"statefulsets":           collectors.RegisterStatefulSetCollector,
-		"persistentvolumeclaims": collectors.RegisterPersistentVolumeClaimCollector,
+		"cronjobs/v2alpha1":         collectors.RegisterCronJobCollector,
+		"daemonsets/v1beta1":        collectors.RegisterDaemonSetCollector,
```
andyxning: Hard code the version for each group here is fragile and maybe diverse from the actually used version in each register function. Maybe we should find another way to correlate the group version used in register function and the keys used in `availableCollectors`.

zouyee: now I add `init function` in each collector.

brancz: Could you please find another solution for this @zouyee ? So far we've been able to keep global things to a minimum, but 14 additional init functions just makes this overly complex to maintain.

zouyee: I have restored to the original state, used `variable` to shown the GroupVersionResource of collector.

zouyee:
@brancz  PTAL

andyxning:
@zouyee Could you please resolve the conflict. We need this feature and after conflict is resolved we can continue to review.

zouyee:
I will fix it!

Issue #221, opened 2017-08-15
kube-state-metrics should not use CreatedByAnnotation


crimsonfaith91:
Based on this [announcement](https://groups.google.com/forum/#!msg/kubernetes-dev/juMOsdaCml0/FwVNJA9uBAAJ), `CreatedByAnnotation` will be deprecated in 1.8 in favor of `ControllerRef`. However, the annotation is still used in this repo [here](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go#L195).

As `CreatedByAnnotation` will be deprecated, we may have to remove this [gauge](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go#L272) or set it to `<none>`. I wish to ask for feedback regarding removal of the gauge.

zouyee:
[according to the doc, Kubernetes automatically sets the value of ownerReference for objects created or adopted by ReplicationController, ReplicaSet, StatefulSet, DaemonSet, and Deployment.](https://github.com/kubernetes/kubernetes.github.io/blob/490fb99a307d808ead693123d0fbccdb6286f247/docs/concepts/workloads/controllers/garbage-collection.md), maybe it is time to migrate from Annotation. @brancz

brancz:
ControllerRef is already implemented, so the labels that we have will just be populated with `<none>`, which is essentially free in Prometheus (2.0) as strings are symbolized. This is an easier case, as we can just keep the label for compatibility, but I can imagine some day a field will disappear that will have a broader effect. We should be prepared for those situations. As we so heavily depend on kubernetes/kubernetes changes for this I'd suggest our compatibility guarantees go along with the client-go compatibility. So once we upgrade client-go 1.8 we remove these labels. 

Wdyt @fabxc ?

zouyee:
I think [`api`](https://github.com/kubernetes/api/blob/master/core/v1/meta.go#L65) will be used, `OwnerReferences` related function has been migrated.

brancz:
Yes, we're [already using that](https://github.com/kubernetes/kube-state-metrics/blob/fda17aa549fe96533b6cf09efee4581a8a06e451/collectors/pod.go#L274).

zouyee:
[kubernetes/api](https://github.com/kubernetes/api/blob/master/core/v1/meta.go#L65) will be updated, because `client-go` structure has beed refactored

andyxning:
[...]

PR #223, opened 2017-08-16
remove kube_pod_info gauge

crimsonfaith91:
Addressing https://github.com/kubernetes/kube-state-metrics/issues/221
`kube_pod_info` gauge is not needed anymore as `CreatedByAnnotation` will be deprecated and removed.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/223)

### DIFF
--- a/Documentation/pod-metrics.md
+++ b/Documentation/pod-metrics.md
@@ -2,7 +2,6 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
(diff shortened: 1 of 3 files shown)


zouyee:
It's not the right time  to remove that metric.

andyxning:
The conclusion of #221 is to set `created_by_kind` and `created_by_name` to `<none>` when kube-state-metrics updates its client-go dep to support Kubernetes 1.8. Sorry for the unclear conclusion.

Seems we should postpone this PR. @brancz

crimsonfaith91:
@zouyee @andyxning @brancz As it is not the right time to remove the `kube_pod_info` metric, there is no any change to the codebase needed. I will close this PR for now, and reopen it after 1.8 release. Thanks!

PR #226, opened 2017-08-17
Add PVC requested storage metric and storageclass label

cofyc:

With `kube_persistentvolumeclaim_resource_requests_storage` metric, we can monitor users requested capacity. This is like `kube_pod_container_resource_requests_(cpu_cores|memory_bytes)` metrics.

With new label `storageclass` in pvc metrics, we can group metrics by storage class. In practice, a cluster may have multiple storage class for different dynamic volume types. For example, in our multi-tenant cluster, we deployed rbd/nfs storage classes and corresponding provisioners. For monitor/alerting, it's better if we can distinguish metrics by storage class name.

**Special notes for your reviewer**:

Example output:

```
# curl -s http://localhost:80/metrics | grep '^kube_persistent'
kube_persistentvolumeclaim_resource_requests_storage{namespace="default",persistentvolumeclaim="mongo-persistent-storage-mongo-0",storageclass="rbd"} 1.073741824e+09
kube_persistentvolumeclaim_resource_requests_storage{namespace="default",persistentvolumeclaim="registry-storage",storageclass="rbd"} 2.147483648e+09
kube_persistentvolumeclaim_resource_requests_storage{namespace="kube-system",persistentvolumeclaim="alertmanager-kirk-monitor-db-alertmanager-kirk-monitor-0",storageclass="rbd"} 1.073741824e+09
kube_persistentvolumeclaim_resource_requests_storage{namespace="kube-system",persistentvolumeclaim="kirk-monitor-grafana",storageclass="rbd"} 1.073741824e+09
kube_persistentvolumeclaim_resource_requests_storage{namespace="kube-system",persistentvolumeclaim="prometheus-kirk-monitor-db-prometheus-kirk-monitor-0",storageclass="rbd"} 1.073741824e+09
```

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/226)

### DIFF
--- a/Documentation/persistentvolumeclaim-metrics.md
+++ b/Documentation/persistentvolumeclaim-metrics.md
@@ -2,4 +2,5 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
+| kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
+| kube_persistentvolumeclaim_resource_requests_storage| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
--- a/Documentation/persistentvolumeclaim-metrics.md
+++ b/Documentation/persistentvolumeclaim-metrics.md
@@ -2,9 +2,9 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_persistentvolumeclaim_info| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
-| kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
-| kube_persistentvolumeclaim_resource_requests_storage| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
+| kube_persistentvolumeclaim_info | Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
+| kube_persistentvolumeclaim_status_phase | Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
+| kube_persistentvolumeclaim_resource_requests_storage_bytes | Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; |
 
 Note:
 
--- a/Documentation/persistentvolumeclaim-metrics.md
+++ b/Documentation/persistentvolumeclaim-metrics.md
@@ -4,3 +4,7 @@
 | ---------- | ----------- | ----------- |
 | kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
 | kube_persistentvolumeclaim_resource_requests_storage| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
+
+Note:
+
+- A special `<none>` string will be used if PVC has no storage class.
--- a/Documentation/persistentvolumeclaim-metrics.md
+++ b/Documentation/persistentvolumeclaim-metrics.md
@@ -2,6 +2,7 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
+| kube_persistentvolumeclaim_info| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
 | kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
 | kube_persistentvolumeclaim_resource_requests_storage| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
 
--- a/collectors/persistentvolumeclaim.go
+++ b/collectors/persistentvolumeclaim.go
@@ -26,13 +26,22 @@ import (
 )
 
 var (
+	descPersistentVolumeClaimInfo = prometheus.NewDesc(
+		"kube_persistentvolumeclaim_info",
+		"Information about persistent volume claim.",
+		[]string{
+			"namespace",
+			"storageclass",
+			"persistentvolumeclaim",
+		}, nil,
+	)
 	descPersistentVolumeClaimStatusPhase = prometheus.NewDesc(
 		"kube_persistentvolumeclaim_status_phase",
 		"The phase the persistent volume claim is currently in.",
 		[]string{
 			"namespace",
-			"persistentvolumeclaim",
 			"storageclass",
+			"persistentvolumeclaim",
 			"phase",
 		}, nil,
 	)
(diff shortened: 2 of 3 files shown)


collectors/persistentvolumeclaim.go
```diff
@@ -29,13 +29,23 @@ import (
...
 		[]string{
 			"namespace",
 			"persistentvolumeclaim",
+			"storageclass",
 			"phase",
 		}, nil,
 	)
+	descPersistentVolumeClaimResourceRequestsStorage = prometheus.NewDesc(
+		"kube_persistentvolumeclaim_resource_requests_storage",
+		"The capacity of storage requested by the persistent volume claim.",
```
zouyee: the info of  StorageClass  required by the claim,not the resource of sc.

collectors/persistentvolumeclaim.go
```diff
@@ -87,9 +98,25 @@ func (collector *persistentVolumeClaimCollector) Collect(ch chan<- prometheus.Me
...
+	// Use beta annotation first
+	if class, found := claim.Annotations[v1.BetaStorageClassAnnotation]; found {
+		return class
+	}
+
+	if claim.Spec.StorageClassName != nil {
+		return *claim.Spec.StorageClassName
+	}
+
+	return ""
```
zouyee: I think `return "<none>" `will be better

cofyc: Updated. 
I also found "<none>" string is already used in other collectors, like [pod](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go). If it's common practice to use this non-empty special string to indicate absence of label value, maybe we need to document it somewhere, for developers and also for end-users.

collectors/persistentvolumeclaim.go
```diff
@@ -87,9 +98,26 @@ func (collector *persistentVolumeClaimCollector) Collect(ch chan<- prometheus.Me
...
 
+// getPersistentVolumeClaimClass returns StorageClassName. If no storage class was
+// requested, it returns "".
+func getPersistentVolumeClaimClass(claim *v1.PersistentVolumeClaim) string {
+	// Use beta annotation first
+	if class, found := claim.Annotations[v1.BetaStorageClassAnnotation]; found {
+		return class
+	}
+
+	if claim.Spec.StorageClassName != nil {
```
andyxning: +1 for nil pointer check. :)

collectors/persistentvolumeclaim.go
```diff
@@ -87,9 +98,26 @@ func (collector *persistentVolumeClaimCollector) Collect(ch chan<- prometheus.Me
 	}
 }
 
+// getPersistentVolumeClaimClass returns StorageClassName. If no storage class was
+// requested, it returns "".
+func getPersistentVolumeClaimClass(claim *v1.PersistentVolumeClaim) string {
+	// Use beta annotation first
+	if class, found := claim.Annotations[v1.BetaStorageClassAnnotation]; found {
```
andyxning: I am not so familiar with pvc actually. @cofyc Could you please spare some time explain the orders.

collectors/persistentvolumeclaim.go
andyxning: @cofyc According to https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go#L257-L266, pod uses `<none>` as the default value for none exist label values.

> maybe we need to document it somewhere, for developers and also for end-users.

👍 . Absolutely we should do this.

collectors/persistentvolumeclaim.go
andyxning: As stated in the [doc](https://kubernetes.io/docs/concepts/storage/persistent-volumes/#class-1)
> In the past, the annotation volume.beta.kubernetes.io/storage-class was used instead of storageClassName attribute. This annotation is still working, however it won’t be supported in a future Kubernetes release.

Seems the preference order should be:
* spec
* annotation

zouyee: I think the checkout order is right.

cofyc: hi, @andyxning 
I copied [the code from Kubernetes core repo](https://github.com/kubernetes/kubernetes/blob/4bfe9b1a568be137f72cf5dfc8bc2c1f13874548/pkg/api/helper/helpers.go#L561). This function is also used by [pv controller](https://github.com/kubernetes/kubernetes/blob/402e48b072b32ae79f59074d2fc67856f5a58849/pkg/controller/volume/persistentvolume/pv_controller.go).

andyxning: @zouyee @cofyc Thanks for the info!

Since [`GetPersistentVolumeClaimClass`](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/helper/helpers.go#L561) is an exportable function, reference that function is more preferable instead of copying in another place. 

This makes kube-state-metrics more maintainable. 

WDYT @cofyc .

cofyc: hi, @andyxning 

I had tried to use external package, but failed because:

- [PersistentVolumeClaim from core repo](https://github.com/kubernetes/kubernetes/blob/v1.7.4/pkg/api/v1/helper/helpers.go#L475) use api types under `k8s.io/kubernetes/pkg/api/v1`, not `k8s.io/client-go/pkg/api/v1`.
- k8s.io/client-go `v4.0.0-beta.0` branch only has a [GetPersistentVolumeClaimClass](https://github.com/kubernetes/client-go/blob/v4.0.0-beta.0/pkg/api/helper/helpers.go#L549) function which use `k8s.io/client-go/pkg/api.PersistentVolumeClaim` not `k8s.io/client-go/pkg/api/v1.PersistentVolumeClaim`

So, I found no way to reference an existing package directly. 

IMHO, it's acceptable (though not good) to copy [GetPersistentVolumeClaimClass](https://github.com/kubernetes/kubernetes/blob/master/pkg/api/helper/helpers.go#L561) here because `StorageClass` feature is stable now, and this function has beta annotation backward support. If future releases of `client-go` contains such helper utilities which we can use directly, we can switch to them at that time.

zouyee: It is better to use `client-go`v4.0.0-beta.0  than copy codes directly, which has existed in `vendor` dir.

cofyc: @zouyee 
client-go `v4.0.0-beta.0` branch only has [a function](https://github.com/kubernetes/client-go/blob/v4.0.0-beta.0/pkg/api/helper/helpers.go#L549) which use `k8s.io/client-go/pkg/api.PersistentVolumeClaim`. In kube-state-metrics collectors, we all use types under `k8s.io/client-go/pkg/api/v1` package, types are mismatch.

brancz:
As the external function is tiny, I think copying it is justified over the dependency complexity.

The only thing is that labels are used to uniquely identify a time-series. And since the storage class is not really identifying it it should go into an info metric, like most other resources have in kube-state-metrics. That way we can then join the metrics on those labels that actually uniquely identify them and build the data we want that way.

cofyc:
@brancz Like `kube_job_info`?

[Updated](https://github.com/kubernetes/kube-state-metrics/pull/226/files).

collectors/persistentvolumeclaim.go
andyxning: @cofyc In fact i am not sure the relation between client-go and `pkg/api/v1`. However, some functionalities are both available in these two packages.

In fact, there is a corresponding [`GetPersistentVolumeClaimClass`](https://github.com/kubernetes/client-go/blob/df46f7f13b3da19b90b8b4f0d18b8adc6fbf28dc/pkg/api/helper/helpers.go#L549) in client-go in `k8s.io/client-go/pkg/api/helper` package.

This will also needs to update Godeps.

collectors/persistentvolumeclaim.go
```diff
@@ -26,15 +26,34 @@ import (
...
+			"persistentvolumeclaim",
+		}, nil,
+	)
 	descPersistentVolumeClaimStatusPhase = prometheus.NewDesc(
 		"kube_persistentvolumeclaim_status_phase",
-		"The phase the claim is currently in.",
+		"The phase the persistent volume claim is currently in.",
 		[]string{
 			"namespace",
+			"storageclass",
```
brancz: My point was that only the info metric actually needs this label, then the storageclass can be joined onto this metric using the info metric. This metric only needs the labels that uniquely identify it, which is just namespace, persistentvolumeclaim and phase.

cofyc: Understood now, thanks for explaining!

PR is [updated](https://github.com/kubernetes/kube-state-metrics/pull/226/files).

Documentation/persistentvolumeclaim-metrics.md
```diff
@@ -2,4 +2,10 @@
 
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
-| kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
+| kube_persistentvolumeclaim_info| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; |
+| kube_persistentvolumeclaim_status_phase| Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; |
```
brancz: These docs need to be adapted as the `storageclass` label is not on them anymore.

cofyc:
@brancz Docs updated.

collectors/persistentvolumeclaim.go
```diff
@@ -26,15 +26,32 @@ import (
...
-		"The phase the claim is currently in.",
+		"The phase the persistent volume claim is currently in.",
 		[]string{
 			"namespace",
 			"persistentvolumeclaim",
 			"phase",
 		}, nil,
 	)
+	descPersistentVolumeClaimResourceRequestsStorage = prometheus.NewDesc(
+		"kube_persistentvolumeclaim_resource_requests_storage",
```
brancz: This needs a `_bytes` suffix.

collectors/persistentvolumeclaim.go
```diff
@@ -86,16 +105,39 @@ func (collector *persistentVolumeClaimCollector) Collect(ch chan<- prometheus.Me
...
+
 	// Set current phase to 1, others to 0 if it is set.
 	if p := pvc.Status.Phase; p != "" {
 		addGauge(descPersistentVolumeClaimStatusPhase, boolFloat64(p == v1.ClaimLost), string(v1.ClaimLost))
 		addGauge(descPersistentVolumeClaimStatusPhase, boolFloat64(p == v1.ClaimBound), string(v1.ClaimBound))
 		addGauge(descPersistentVolumeClaimStatusPhase, boolFloat64(p == v1.ClaimPending), string(v1.ClaimPending))
 	}
+
+	if storage, ok := pvc.Spec.Resources.Requests[v1.ResourceStorage]; ok {
+		addGauge(descPersistentVolumeClaimResourceRequestsStorage, float64(storage.Value()))
```
brancz: We need to make sure this is in base units, so we need to make sure it's in bytes.

collectors/persistentvolumeclaim.go
andyxning: @cofyc Friendly ping about import `GetPersistentVolumeClaimClass` from client-go instead of copy&&paste. This needs to update the Godep. 

@brancz WDYT.

brancz: If the types match then I'd also prefer vendoring.

cofyc:
@brancz 
`_bytes` suffix added.

collectors/persistentvolumeclaim.go
cofyc: It's correct way to gets capacity resource in bytes. See [example](https://github.com/kubernetes-incubator/external-storage/blob/master/ceph/rbd/pkg/provision/rbd_util.go#L44).

collectors/persistentvolumeclaim.go
cofyc: hi, @andyxning
[GetPersistentVolumeClaimClass](https://github.com/kubernetes/client-go/blob/v4.0.0-beta.0/pkg/api/helper/helpers.go#L549) in client-go use `k8s.io/client-go/pkg/api.PersistentVolumeClaim`, but we all use types under `k8s.io/client-go/pkg/api/v1` package. Types mismatch.

collectors/persistentvolumeclaim.go
brancz: Got it, it's absolutely correct the way you already had it, sorry for that.

brancz:
I'll give @andyxning another chance to comment on the vendoring part, I didn't look into it very far, but if the types mismatch (and there is no other function that does) I don't have an issue with having the function here as it's pretty small.

andyxning:
@cofyc Sorry for not making it clear before commenting. I am fine with copying the logic now.

Now, lgtm also.

PR #210, opened 2017-08-07
Add gofmt check in travis CI

asifdxtreme:
Add a new stage "go fmt" check in travis CI  as per suggestions in https://github.com/kubernetes/kube-state-metrics/pull/206

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/210)

PR #229, opened 2017-08-19
collectors: run go fmt on all collectors

brancz:
After merging #182 CI started failing as we recently introduced a check for that.

@andyxning let's make sure we rebase / ask to rebase all PRs opened after that change.

/cc @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/229)

### DIFF
--- a/collectors/cronjob_test.go
+++ b/collectors/cronjob_test.go
@@ -31,8 +31,8 @@ var (
 	StartingDeadlineSeconds300 int64 = 300
 
 	ActiveRunningCronJob1LastScheduleTime          = time.Unix(1500000000, 0)
-	SuspendedCronJob1LastScheduleTime              = time.Unix(1500000000 + 5.5 * 3600, 0) // 5.5 hours later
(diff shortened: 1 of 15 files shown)


andyxning:
As go fmt check is introduced in #210. So, prs after that are already checked with it.

PRs before #210 should conflict automatically if they change existing codes that are refactored later by `go fmt`. After #210, prs before #210 has not receive any conflicts it seems that all is fine according to `go fmt`.

The exceptions are adding new code. They should be rebased. The good news is that there are only 
five passed PRs needs to be be rebased. I think we can add a new comment requesting for rebase. I don't think we can do this automatically.

andyxning:
After check all five prs before #210, three of them are already have conflicts. So actually we just need to ping two user to rebase. One of the PR is mine. I have rebased it. The other is #160.

andyxning:
> As go fmt check is introduced in #210. So, prs after that are already checked with it.

I am curious why prs after #210 are not error on travisCI fmt stage.

WIZARD-CXY:
LGTM please merge it asap, my pr is hold off due to the failed fmtcheck test...

PR #71, opened 2017-01-22
Add HorizontaPodAutoscaler metrics.

bluecamel:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/71)

PR #230, opened 2017-08-20
implement horizontal pod autoscaler (HPA) metrics

joemiller:
This picks up the effort in PR #71 with the following changes:

- adds `.Spec.TargetCPU` and `.Status.CurrentCPU` metrics
- adds `kube_hpa_labels` metrics similar to other collectors
- refactored to fit the new code arrangement in ./collectors

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/230)


joemiller:
CLA signed

joemiller:
rebased master

brancz:
With HPA v2 coming up, are the objects and fields going to stay the same, or at least get a new type somehow? There is a question of which objects we want to be compatible with.

joemiller:
@brancz I believe you're probably correct. I am not very familiar with the details of HPA v2 or where it's heading. Do you (or anyone) already have some thoughts on how we might make this work for v1 and v2?

brancz:
@DirectXMan12 as @joemiller and I are clueless around HPA v2, could you comment on whether there are major API object differences we need to look out for or maybe even just treat them as completely separate objects and include a version in the metric name or something?

DirectXMan12:
So, basically, HPA v2 generalizes the concept of "target CPU" and "current CPU" to "target metrics" and "current metrics".  For each HPA, you have one or more targets.  Each target is of one of three types:

- resource metrics (like CPU and memory), which can either be percentage-of-request targets (like before) or raw targets.
- pod metrics, where the metric is describing each pod in the target scalable (raw target only)
- object metrics, where the metric describes a different object entirely (raw target only)

In general, the corresponding "current metrics" in the status look like their targets, except that for resource metrics, when using a percentage-of-request target, the current raw value will *also* be populated.

For example, if you have:

```yaml
spec:
  targetCPUUtilizationPercentage: 30
  ...
status:
  currentCPUUtilizationPercentage: 20
  ...
```

in `autoscaling/v1`, in `autoscaling/v2beta1` you'd get:

```yaml
spec:
  metrics:
  - type: Resource
    resource:
      name: cpu
      targetAverageUtilization: 20
  ...
status:
  metrics:
  - type: Resource
    resource:
      name: cpu
      currentAverageUtilization: 20
      currentAverageValue: 400m
  ...
```

brancz:
@DirectXMan12 sorry for not following entirely, but is the `v2beta1` autoscaling API already available?

DirectXMan12:
the move from autoscaling/v2alpha1 to v2beta1 just merged for 1.8 yesterday.

DirectXMan12:
P.S. I feel like exposing the metrics HPA collects is a bit strange, since those metrics are likely already in Prometheus.  I'm all for having metrics on the HPA, but re-exporting the metrics seems a little bit strange (although I guess seeing the difference between "CPU as the HPA sees it" and "CPU as Prometheus sees it" would be valuable).

joemiller:
Our use case is for our capacity planning process. When a squad deploys an
app we'd like to automatically incorporate its HPA into our capacity
planning processes. Being able to quickly/automatically put together
dashboards showing which apps/HPAs are spending much of their time at or
near 100% of their max-replicas, for example. We've also had some HPAs
where targetCPU/currentCPU were not well configured and we'd like to
surface those easily. I'm not sure if this HPA data is available in
prometheus. I'm only aware of individual pod and container CPU metrics in
prom.

brancz:
Yes I think especially those values configuring the HPA (rather than the "current" values) are valuable, but as you already mentioned, probably doesn't hurt if we can see if they diverge.

As the two objects are quite different I'm wondering how we should go about the implementation in kube-state-metrics, should we prefix the metrics originating from the v1alpha1 objects with that version? We haven't had any precedence of this so far so it'd probably be worth taking a moment or two to think about this.

/cc @fabxc

joemiller:
I am still interested in moving this PR forward. Apologies for the delay. I've had some time to think about a solution, and this is my proposal based largely on @DirectXMan12's description of hpa v2beta1 spec:

proposal:
1. two collectors - "hpav1" and "hpva2"
    1. for cli arg --collectors:  "horizontalpodautoscalersv1", "horizontalpodautoscalersv2"  ( or "horizontalpodautoscalers_v1 .. ?)
2. prometheus tag format same for both:
    1. {hpa="hpa1",namespace="ns1"}
3. hpav1 metric names:
    1. spec:
        1. kube_hpav1_spec_min_replicas
        2. kube_hpav1_spec_max_replicas
        3. kube_hpav1_spec_target_cpu
    2. status:
        1. kube_hpav1_status_desired_replicas
        2. kube_hpav1_status_current_replicas
        3. kube_hpav1_status_current_cpu
4. hpva2 metric naming scheme:
    1. spec:
        1. kube_hpav2_spec_min_replicas
        2. kube_hpav2_spec_max_replicas
        3. kube_hpav2_spec_metrics_resource_name_targetAverageUtilization
    2. status:
        1. kube_hpav2_status_desired_replicas
        1. kube_hpav2_status_current_replicas
        2. kube_hpav2_status_metrics_resource_cpu_currentAverageUtilization
        3. kube_hpav2_status_metrics_resource_cpu_currentAverageValue

joemiller:
Ok, having written my proposal above and _then_ doing the legwork of trying to implement I think the best way forward right now is:

- scale back this PR to only report the following for both v1 and v2 hpas. This would allow us to move forward with our capacity planning needs for reporting on HPAs and the delta between current and max replicas.
```
kube_hpa_spec_min_replicas
kube_hpa_spec_max_replicas
kube_hpa_status_desired_replicas
kube_hpa_status_current_replicas
```

joemiller:
ping @brancz @DirectXMan12  I pushed a new commit that scales this PR back to only focus on replica counts as rationalized in previous comment above ^^.   Thoughts on moving forward with this approach?

DirectXMan12:
the reduced-scope proposal seems reasonable to me.

brancz:
Thanks for revising! This lgtm :+1: , we're currently in RC for v1.1.0, so I would like to not merge new functionality and out of ease of maintenance, we'll hold off on merging until after GA of v1.1.0.

jakobant:
Hi, is there a plan to merge this soon?  If so when can I expect this to be in a new release.

brancz:
This lgtm, but it will be some time until this feature lands in a release, as we have just released `v1.1.0`.

PR #231, opened 2017-08-21
add job owner info in kube_job_info

zouyee:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/231)

### DIFF
--- a/Documentation/job-metrics.md
+++ b/Documentation/job-metrics.md
@@ -14,3 +14,4 @@
 | kube_job_complete | Gauge | `job`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; |
 | kube_job_failed | Gauge | `job`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; |
 | kube_job_created | Gauge | `job`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; |
+| kube_job_owner | Gauge | `pod`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `owner_kind`=&lt;owner kind&gt; <br> `owner_name`=&lt;owner name&gt; <br> `owner_is_controller`=&lt;whether owner is controller&gt;  |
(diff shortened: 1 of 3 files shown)


brancz:
Since we're deprecating the created by labels elsewhere, shouldn't we just add metrics for owner references?

zouyee:
I forget it...

andyxning:
@zouyee 

I have found nothing info about ownerReference for cronjob in the provided url. Could you please add more info for this.

andyxning:
According to [gc doc](https://kubernetes.io/docs/concepts/workloads/controllers/garbage-collection/#owners-and-dependents), it does not state that jobs created by cronjob will have `ownerreference` to cronjob. 

I have test this with minikube on Kubernetes 1.7.2 and it indeed does not have a `ownerreference` to the cronjob.

Maybe this should first be added in Kubernetes and then collect them in kube-state-metrics.

brancz:
As all of Kubernetes is moving away from the created-by annotations I don't think we should introduce them to begin with, it's just something people will get upset by when we deprecate it shortly after it was introduced.

andyxning:
Yes. That is why i suggest we should first add `ownerreference` to jobs in Kubernetes. After created by annotation is deprecated, we can still use the `ownerreference` to accomplish the same job as what created by annotation does. :)

brancz:
Sorry, I wasn't very clear in my previous comment, I meant to say that I agree with you @andyxning . :slightly_smiling_face:

zouyee:
In my opinion，according to [owenreference](https://github.com/kubernetes/kubernetes.github.io/blob/490fb99a307d808ead693123d0fbccdb6286f247/docs/concepts/workloads/controllers/garbage-collection.md),kubernetes would not add ownerreference for cronjob dynamically, it would be suitable to using annotation before deprecated

andyxning:
@zouyee Could you please detail the reason about why kubernetes will not add ownerreference for jobs created by cronjob.

zouyee:
according to this processing listhttps://github.com/kubernetes/community/blob/master/contributors/design-proposals/controller-ref.md#implementation

brancz:
With CronJobs being the only consumer/possible owner of Job objects, do you think it's worth generating 3+ extra metrics for every job?

PR #234, opened 2017-08-22
Add metrics for namespace label and created time

chlunde:
This allows prometheus reporting based on labels per namespace.

For example, if each namespace has a team label, you can report on the
memory use per team like this:

    sum(sum(container_memory_usage_bytes) by(namespace) * on(namespace)
    group_right() kube_namespace_labels) by (label_team)

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/234)

### DIFF
--- a/collectors/namespace.go
+++ b/collectors/namespace.go
@@ -1,0 +1,102 @@
+/*
+Copyright 2016 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collectors
+
+import (
+	"github.com/golang/glog"
+	"github.com/prometheus/client_golang/prometheus"
+	"golang.org/x/net/context"
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/pkg/api"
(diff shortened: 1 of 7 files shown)


chlunde:
CLA

brancz:
There is already an [open PR for namespace metrics](https://github.com/kubernetes/kube-state-metrics/pull/160). I'm fine with moving forward with this one as the other one needs a rebase, but please ensure to include the functionality that the other PR had as well.

Additionally we will need license headers in the beginning of the files.

chlunde:
Oops, I didn't see that PR.  I'll merge them.

I see statefulset is missing headers too (I happened to use that file when starting)

brancz:
Oh, thanks for the catch! Could you add the license headers to the statefulset file as well then (just add a separate commit for it)?

chlunde:
@brancz Sure.

One issue: technically, a namespace has a `name`, but using the label `namespace` (as I did) makes it easy to use without relabeling.  What would you like?

brancz:
`namespace` as the label-name is absolutely correct :slightly_smiling_face: as you already mentioned that makes joins very easy, and that's the intention

collectors/namespace.go
```diff
@@ -0,0 +1,138 @@
...
+// List return namespace list
+func (l NamespaceLister) List() ([]v1.Namespace, error) {
+	return l()
+}
+
+// RegisterNamespaceCollector registry namespace collector
+func RegisterNamespaceCollector(registry prometheus.Registerer, kubeClient kubernetes.Interface, namespace string) {
+	client := kubeClient.CoreV1().RESTClient()
+	var selector fields.Selector
+	if namespace != api.NamespaceAll {
```
brancz: it's safe to query for all namespaces here, the namespace restriction is only meant for limiting RBAC roles

chlunde:
I redid it with my commit on top of @zouyee's work, because git fixups across merge commits didn't work very well.

brancz:
lgtm thanks to you two!

PR #238, opened 2017-08-23
Match deployment resources to recommendation

### LINKED ISSUE #112: No logs available
I have kube-state-metrics running as a deployment via ansible on my clusters:
```
kind: Deployment
metadata:
  name: kube-state-metrics
spec:
  replicas: 4
  template:
    metadata:
      labels:
        app: kube-state-metrics
    spec:
      containers:
      - name: kube-state-metrics
        image: gcr.io/google_containers/kube-state-metrics:v0.3.0
        ports:
        - name: metrics
          containerPort: 8080
        resources:
          requests:
            memory: {{ kube_state_mem_req }}
            cpu: 100m
          limits:
            memory: {{ kube_state_mem_lim }}
            cpu: 200m
```
I've had to bump kube_state_mem_(req|lim) to 800Mi in order to get the pods to stay up; the pods have started OOMKilling/CrashLoopBackoff'ing.

I'd like to know why, but the containers are basically inscrutable. There's no way to shell in and `docker logs` is empty.

It'd be great if there was more information on what's going on, please and thanks!

matthiasr:
Adjust the deployment to match the recommended resources (#196). The baseline resources are set to the basic recommendation for a 100 node cluster.

Pod nanny (#200) does not support "the baseline includes the first 100 nodes", so instead set the threshold until it needs to adjust wide. This over-estimates resource needs for intermediate cluster sizes, but I'd rather be on the safe side. @WIZARD-CXY is this sensible?

Fixes #112.

/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/238)

### DIFF
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -25,10 +25,10 @@ spec:
           timeoutSeconds: 5
         resources:
           requests:
-            memory: 30Mi
+            memory: 200Mi
             cpu: 100m
           limits:
-            memory: 50Mi
+            memory: 200Mi
             cpu: 200m
       - name: addon-resizer
         image: gcr.io/google_containers/addon-resizer:1.0
@@ -53,7 +53,7 @@ spec:
           - --container=kube-state-metrics
           - --cpu=100m
           - --extra-cpu=1m
-          - --memory=30Mi
+          - --memory=200Mi
           - --extra-memory=2Mi
-          - --threshold=5
+          - --threshold=100
           - --deployment=kube-state-metrics
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -25,7 +25,7 @@ spec:
           timeoutSeconds: 5
         resources:
           requests:
-            memory: 200Mi
+            memory: 100Mi
             cpu: 100m
           limits:
             memory: 200Mi
@@ -53,7 +53,7 @@ spec:
           - --container=kube-state-metrics
           - --cpu=100m
           - --extra-cpu=1m
-          - --memory=200Mi
+          - --memory=100Mi
           - --extra-memory=2Mi
-          - --threshold=100
+          - --threshold=5
           - --deployment=kube-state-metrics


kubernetes/kube-state-metrics-deployment.yaml
```diff
@@ -53,7 +53,7 @@ spec:
           - --container=kube-state-metrics
           - --cpu=100m
           - --extra-cpu=1m
-          - --memory=30Mi
+          - --memory=200Mi
           - --extra-memory=2Mi
-          - --threshold=5
-          - --deployment=kube-state-metrics
\ No newline at end of file
+          - --threshold=100
```
brancz: The docs of the addon-resizer say this about the threshhold flag

> A number between 0-100. The dependent's resources are rewritten when they deviate from expected by more than threshold.

I'm not even sure what the numbers mean, but given it's 0-100 I'm thinking it's %, which would mean that only once the memory would accumulate to 400Mi it would trigger. Am I understanding this correctly, or does it mean when 100 x extra memory is reached start modifying?

WIZARD-CXY: it is percentage. I think 5 is ok,  100 is too large

WIZARD-CXY:
@matthiasr yes, you are right about "This over-estimates resource needs for intermediate cluster sizes" , I think it is too large to set to 200M, it may get kube-state-metrics unscheduled onto a dense cluster because it requests too much memory. 
I think it is no need to set this base large, just let pod-nanny work, If you insist, I think 100M as base is fine with me.

brancz:
I'm perfectly fine with 100Mi as well. Just giving users something that covers the 95% use case is enough I think. Tweaking purely on node count won't work all the way anyways, when people hit that limit then manual intervention is required either way. So if this still satisfies the use for the majority of users it sounds perfectly good to me.

WIZARD-CXY:
I think it is just like self-driving car. For me I do not want manual intervention. I like pod-nanny automatically do its work personally.

matthiasr:
Alright, changed the initial memory request to 100MiB and the threshold to 5%. PTAL

brancz:
@WIZARD-CXY agreed, but in that sense the addon-resizer is just too dumb and we'll just need a general vertical pod autoscaler (which is coming with the HPA v2) and then we can scale on any metric we want, which will make this experience much smoother and more refined.

Other than that this PR lgtm.

matthiasr:
> I think it is just like self-driving car. For me I do not want manual intervention.

Of course, but you can't hitch up a 40-tonne trailer and expect it to still work :)

brancz:
Thanks @matthiasr !

andyxning:
@brancz @matthiasr @WIZARD-CXY I think we should close #92 as that pr is a dup.

WIZARD-CXY:
Let me give a closer look at HPA v2

WIZARD-CXY:
I think we should try vertically scale based on metrics number ksm monitors, not on node number. It is more accurate and general. I will see what HPA v2 can do about self-defined metrics.

brancz:
As far as I know HPA v2 is still early stage (alpha or pre-alpha IIRC). It will one day be the better alternative, but we can't expect users to enable alpha/pre-alpha features right now, however an example if they already exist wouldn't hurt having in this repo.

WIZARD-CXY:
@brancz I think we use vpa here because memory resource doesn't easily suit for hpa. Anyone who is interesting can see more info here https://github.com/kubernetes/autoscaler, although it is still young.

brancz:
Yes if I remember correctly, vertical pod autoscaling is part of the HPA v2. And yes it's still very young so until it's the officially supported way of autoscaling kube-state-metrics it will be some time.

PR #241, opened 2017-08-23
Add check for missing documentation

### LINKED ISSUE #235: Missing metrics in documentation
A few metrics are missing or typoed in the docs.

```bash
grep -hoE '(kube_[^ ]+)' Documentation/* | sort -u > documented_metrics
grep -hoE '(kube_[^{]+){' collectors/*_test.go | sort -u | tr -d '{' > tested_metrics # looking for TYPE is probably better
diff -u0 tested_metrics documented_metrics
```

Gives the following (and two false positives):
```diff
--- tested_metrics      2017-08-22 21:09:53.715925313 +0200
+++ documented_metrics  2017-08-22 21:09:34.960378911 +0200
@@ -1 +0,0 @@
-kube_cronjob_created
@@ -8 +6,0 @@
-kube_daemonset_created
@@ -14,2 +11,0 @@
-kube_deployment_created
-kube_deployment_labels
@@ -26 +21,0 @@
-kube_job_created
@@ -29 +24 @@
-kube_job_spec_active_deadline_seconds
+kube_job_spec_active_dealine_seconds
@@ -38,2 +32,0 @@
-kube_limitrange_created
-kube_node_created
@@ -64 +56,0 @@
-kube_pod_created
@@ -72 +63,0 @@
-kube_replicaset_created
@@ -79 +69,0 @@
-kube_replicationcontroller_created
@@ -88,2 +77,0 @@
-kube_resourcequota_created
-kube_service_created
@@ -92,2 +79,0 @@
-kube_statefulset_created
-kube_statefulset_labels
```

chlunde:
Fixes #235

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/241)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,10 +12,15 @@ gofmtcheck:
 	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi     
 
 doccheck:
-	grep -hoE '(kube_[^ |]+)' Documentation/* | sort -u > documented_metrics
-	sed -n 's/.*# TYPE \(kube_[^ ]\+\).*/\1/p' collectors/*_test.go | sort -u > tested_metrics
-	diff -u0 tested_metrics documented_metrics
-	rm -f tested_metrics documented_metrics
+	@echo "- Checking if documentation is up to date..."
+	@grep -hoE '(kube_[^ |]+)' Documentation/* | sort -u > documented_metrics
+	@sed -n 's/.*# TYPE \(kube_[^ ]\+\).*/\1/p' collectors/*_test.go | sort -u > tested_metrics
+	@diff -u0 tested_metrics documented_metrics || (echo "ERROR: Metrics with - are present in tests but missing in documentation, metrics with + are documented but not tested."; exit 1)
+	@echo OK
+	@rm -f tested_metrics documented_metrics
+	@echo "- Checking for orphan documentation files"
+	@cd Documentation; for doc in *.md; do if [ "$$doc" != "README.md" ] && ! grep -q "$$doc" *.md; then echo "ERROR: No link to documentation file $${doc} detected"; exit 1; fi; done
+	@echo OK
 
 deps:
 	go get github.com/tools/godep
--- a/Makefile
+++ b/Makefile
@@ -11,6 +11,12 @@ PKGS = $(shell go list ./... | grep -v /vendor/)
 gofmtcheck:
 	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi     
 
+doccheck:
+	grep -hoE '(kube_[^ |]+)' Documentation/* | sort -u > documented_metrics
+	sed -n 's/.*# TYPE \(kube_[^ ]\+\).*/\1/p' collectors/*_test.go | sort -u > tested_metrics
+	diff -u0 tested_metrics documented_metrics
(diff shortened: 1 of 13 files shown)


chlunde:
Fails on purpose, will follow up with missing docs

WIZARD-CXY:
great

Makefile
```diff
@@ -11,6 +11,12 @@ PKGS = $(shell go list ./... | grep -v /vendor/)
 gofmtcheck:
 	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi     
 
+doccheck:
+	grep -hoE '(kube_[^ |]+)' Documentation/* | sort -u > documented_metrics
```
brancz: Let's drop a little comment in here so that we 3 months from now still know what this make target is doing :slightly_smiling_face:, maybe just something small like "This make target extracts the metrics from the documentation by taking all strings starting with `kube_` and compares it with the metrics it can find by taking all `kube_` strings it can find in the tests." I'm sure we can write this better, just an English description would be great for someone who gets a failing CI job.

andyxning: 👍  @chlunde . 

Agree with @brancz . More details about this target especially when it fails should be great for contributors to learn and make a quick correction.

matthiasr: To that end, maybe don't add this as a comment but as output before the check is run.

```
@echo "Checking if documentation is up to date..."
@echo "Metrics with - are present in tests but missing in documentation, metrics with + are documented but not tested."
```

(too verbose?)

matthiasr: Or just list the offending metrics on error …

brancz: Nice suggestion @matthiasr! No I think that will make it feel more like a legitimate test case, so I think your suggestion makes a lot of sense and is not too verbose.

andyxning:
readme.md under Documents dir should also pay attention to @chlunde. Actually we should also update it.

brancz:
@andyxning I'm not sure I understand what you're getting at, can you elaborate? Do you mean we should try to ensure we have a docs link for each collector?

andyxning:
@brancz Yes. We already have a list of those links in [Documents Readme.md](https://github.com/kubernetes/kube-state-metrics/blob/master/Documentation/README.md#documentation).

chlunde:
Something like this?

All OK:
```
$ make doccheck
- Checking if documentation is up to date...
OK
- Checking for orphan documentation files
/home/chlunde/go/src/k8s.io/kube-state-metrics/Documentation
OK
```

Orphan file:
```
$ touch Documentation/new.md
$ make doccheck
- Checking if documentation is up to date...
OK
- Checking for orphan documentation files
/home/chlunde/go/src/k8s.io/kube-state-metrics/Documentation
ERROR: No link to documentation file new.md detected
```

Docs/tests differ:
```
$ echo kube_foo_bar > Documentation/new.md 
$ make doccheck
- Checking if documentation is up to date...
--- tested_metrics	2017-08-24 19:48:24.963979635 +0200
+++ documented_metrics	2017-08-24 19:48:24.957979762 +0200
@@ -24,0 +25 @@
+kube_foo_bar
ERROR: Metrics with - are present in tests but missing in documentation, metrics with + are documented but not tested.
make: *** [Makefile:18: doccheck] Error 1
```

brancz:
Awesome! Thanks a lot for your effort! lgtm :+1:

PR #245, opened 2017-08-28
Fix gcloud push command

kawych:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/245)


brancz:
This change is already on master, and I don't think we'll actually still need it on this branch again, but just in case we will, I have no objections to this.

lgtm on green

brancz:
Thanks!

PR #244, opened 2017-08-25
merge v1.0.1 release back into master

brancz:
I tagged and published the release and pushed the image to `quay.io/coreos/kube-state-metrics:v1.0.1`, and asked @loburm to publish the gcr.io image.

@andyxning @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/244)

PR #224, opened 2017-08-17
add service type

zouyee:
https://github.com/kubernetes/kube-state-metrics/issues/122
`kube_service_spec_type{namespace="default",service="test-service",type="ClusterIP"}` and `kube_service_info{namespace="default",service="test-service",clusterIP="192.168.0.1"}`

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/224)

PR #88, opened 2017-02-12
bump deployment kube-state-metrics version to v0.4.1

andyxning:
Bump deployment.yml kube-state-metrics version to v0.4.1.

/cc @brancz @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/88)

Issue #85, opened 2017-02-10
Provide binary for releases 


rvrignaud:
Hello,

It would be great if community could provide Linux/amd64 binaries for releases.

Thoughts ?

brancz:
I am completely for this, however, it would have to be completely automated and as far as I understand this is part of the effort of opening up google's internal release process to non google maintainers.

/cc @saad-ali @grodrigues3 @piosz @david-mcmahon

andyxning:
@rvrignaud I am not sure whether you want release binaries or release images.

As for release binaries, maybe #153 will fully accomplish your requirement.

brancz:
As mentioned in the PR, I don't think we will start releasing binaries, until google opens up the release process to us. For the time being it will just be amd64 container images that are published on gcr by google people and on quay by me or @fabxc.

grodrigues3:
@kubernetes/sig-release-feature-requests

coffeepac:
@brancz is this the issue you were talking about during sig-inst today, re: opening the build process?

brancz:
@coffeepac yes this, [this](https://github.com/kubernetes/kube-state-metrics/pull/134) and [this](https://github.com/kubernetes/kube-state-metrics/pull/246). They all revolve around the same topic, essentially how can we as non-Googlers do releases independently.

PR #246, opened 2017-08-28
move default image registry to quay

andyxning:
This PR will move kube-state-metrics image registry from `gcr.io` to `quay.io`.

This is mainly because there are some times([#224 comment](https://github.com/kubernetes/kube-state-metrics/pull/244#issuecomment-325298896), [#88 comment](https://github.com/kubernetes/kube-state-metrics/pull/88#issuecomment-279228915)) the latest image can not be pushed to gcr.io timely. Users may be delayed to use the latest kube-state-metrics for this reason.  A better way was to use `quay.io` as the docker registry as @brancz should be able to push the latest image when a new kube-state-metrics is released.

/cc @brancz @loburm @fabxc @piosz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/246)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -4,7 +4,7 @@ FLAGS =
 COMMONENVVAR = GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(subst x86_64,amd64,$(patsubst i%86,386,$(shell uname -m)))
 BUILDENVVAR = CGO_ENABLED=0
 TESTENVVAR = 
-REGISTRY = gcr.io/google_containers
+REGISTRY = quay.io/coreos
 TAG = $(shell git describe --abbrev=0)
 PKGS = $(shell go list ./... | grep -v /vendor/)
 
--- a/README.md
+++ b/README.md
@@ -27,7 +27,13 @@ Currently, `client-go` is in version `v4.0.0-beta.0`.
 
 ## Container Image
 
-The latest container image can be found at `gcr.io/google_containers/kube-state-metrics:v0.5.0`.
+The latest container image can be found at:
+* `quay.io/coreos/kube-state-metrics:v1.0.1`
+* `gcr.io/google_containers/kube-state-metrics:v1.0.1`
+
+**Note**:
+The recommended docker registry for kube-state-metrics is `quay.io`. kube-state-metrics on
+`gcr.io` is only maintained on best effort as it requires external help from Google employees.
 
 ## Metrics Documentation
 
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -13,7 +13,7 @@ spec:
       serviceAccountName: kube-state-metrics
       containers:
       - name: kube-state-metrics
-        image: gcr.io/google_containers/kube-state-metrics:v1.0.1
+        image: quay.io/coreos/kube-state-metrics:v1.0.1
         ports:
         - name: http-metrics
           containerPort: 8080


Makefile
```diff
@@ -35,7 +35,7 @@ container: build
 	docker build -t ${REGISTRY}/kube-state-metrics:$(TAG) .
 
 push: container
-	gcloud docker -- push ${REGISTRY}/kube-state-metrics:$(TAG)
```
brancz: As I don't use this make target, we can leave this for the google people as a help, as the image should still be published on gcr.

README.md
```diff
@@ -27,7 +27,7 @@ Currently, `client-go` is in version `v4.0.0-beta.0`.
 
 ## Container Image
 
-The latest container image can be found at `gcr.io/google_containers/kube-state-metrics:v0.5.0`.
+The latest container image can be found at `quay.io/coreos/kube-state-metrics:v1.0.1`.
```
brancz: Let's enumerate both container image repositories here, but write a note, that the gcr image is only maintained on best effort as it requires external action from Google employees as no Google employee actively works on kube-state-metrics.

brancz:
I'm fine with doing this, but we should still try to maintain the gcr repository, as people have expressed their interest in the gcr image before.

@piosz last time I opened a PR like this we closed it, as we said this process will be opened up to non google employees, what is the state on this, as we have trouble regarding this on every release?

Makefile
andyxning: done.

README.md
andyxning: done. PTAL.

andyxning:
@piosz @loburm This situation also have been encountered in the maintenance of heapster.

piosz:
@kubernetes/sig-release-feature-requests could you please address the questions above?
@bgrant0607 is this ok to have official Kubernetes images in a different repo than `gcr.io/google_containers `?

piosz:
... and apologies for not doing great job here, but we only have a limited capacity. I agree that lack of access to the official Docker repository for non-googlers is painful.

andyxning:
Any updates?

andyxning:
Any updates?

bgrant0607:
@andyxning @brancz 

In one of the previously cited PRs I see a comment "I pinged some Googlers". Who and how? Was the release team notified? 

What is the release process for kube-state-metrics? At a glance, I didn't see this in the README and didn't see other obvious docs in the top level dir. Our [repo template](https://github.com/kubernetes/kubernetes-template-project) suggests RELEASE.md.

The general issue for solving this is:
https://github.com/kubernetes/release/issues/281

How large is the image?

Also, how do users know which Kubernetes releases a given release of kube-state-metrics will work with?

brancz:
Whenever we want to release a version of kube-state-metrics, I ping @piosz and @loburm as they are active members of sig-instrumentation and kube-state-metrics is owned by sig-instrumentation.

Previous attempts (that I could quickly find) around this include:

- https://github.com/kubernetes/kube-state-metrics/issues/73
- https://github.com/kubernetes/kube-state-metrics/issues/85

We picked up development from the previous maintainer and just continued as it was handled before: cut release, notify Googler to push image. I've previously been told that opening the release process is already in progress, so I for the time being decided to wait it out as the current approach wasn't giving us, as maintainers, too much pain.

> How large is the image?

I'm not sure I understand the relevance, but my docker host is telling me `43.9MB`.

> Also, how do users know which Kubernetes releases a given release of kube-state-metrics will work with?

The README has a section on [this](https://github.com/kubernetes/kube-state-metrics#kubernetes-version), basically whatever client-go support matrix is.

andyxning:
@bgrant0607 One more thing. heapster has the same release image not found problem in 
* [heapster  #1777](https://github.com/kubernetes/heapster/pull/1777)
* [heapster #1676](https://github.com/kubernetes/heapster/issues/1676)

> Was the release team notified?

@brancz I don't think we have a release team.

> What is the release process for kube-state-metrics? 

What is a reference implementation of a normalized release process?

piosz:
@andyxning There is a dedicated @kubernetes/sig-release-members team.

jdumars:
@kubernetes/sig-release-members is not the release team, it's the SIG that sources many (but not all) release team members.  https://github.com/orgs/kubernetes/teams/kubernetes-release-managers/members consists of the current release team.

piosz:
@jdumars thanks for the info. Sorry for the confusion then.

bgrant0607:
@brancz An image so small could be posted to github:

https://help.github.com/articles/creating-releases/
https://help.github.com/articles/distributing-large-binaries/

andyxning:
@bgrant0607 Github releases can be used to ship binaries. But, we are now discussing the image release. :) 

Seems github does not support image release. :)

bgrant0607:
@andyxning Bits are bits. It's true that Github doesn't support pulling images via a registry API.

Anyway, this isn't part of the official Kubernetes release and doesn't seem to be used in https://github.com/kubernetes/kubernetes/tree/master/cluster/addons. Other repos (e.g., kompose) also push their release artifacts to random places. We haven't standardized that yet.

So long as no tests run by prow pull this image (we've had a lot of failures caused by other registries in the past), I don't have a strong opinion right now.

piosz:
/lgtm
As we have the other Kubernetes projects not pushed to `google_containers` and this is not a part of [cluster/addons](https://github.com/kubernetes/kubernetes/tree/master/cluster/addons), I'm fine with changing the default registry, so that we are not the bottleneck here. The only concern is that now CoreOS is the bottleneck but let's see how this will work. Google will also build all versions and push them to `google_containers` registry, so please keep pinging us when the new version is ready.

brancz:
Yes I would suggest that we just document both image registries and users are free to choose which they want, and we'll continue as we do today.

andyxning:
@piosz @bgrant0607 Thanks for your suggestions on this. 

> please keep pinging us when the new version is ready.

Absolutely. we need to work together to make kube-state-metrics more friendly to users. :)

PR #252, opened 2017-09-06
Making targets for multi-arch

Pensu:
This PR ensures multi arch builds for kube-state-metrics.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/252)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,12 +1,15 @@
-all: build
-
 FLAGS =
-COMMONENVVAR = GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(subst x86_64,amd64,$(patsubst i%86,386,$(shell uname -m)))
 BUILDENVVAR = CGO_ENABLED=0
 TESTENVVAR = 
 REGISTRY = gcr.io/google_containers
 TAG = $(shell git describe --abbrev=0)
 PKGS = $(shell go list ./... | grep -v /vendor/)
+ARCH ?= $(shell go env GOARCH)
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
+
+
+IMAGE = $(REGISTRY)/kube-state-metrics
+MULTI_ARCH_IMG = $(IMAGE)-$(ARCH)
 
 gofmtcheck:
 	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi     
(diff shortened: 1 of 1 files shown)


Pensu:
@luxas Can you please take a look?

Makefile
```diff
@@ -22,22 +40,41 @@ doccheck:
...
-test-unit: clean deps build
-	$(COMMONENVVAR) $(TESTENVVAR) godep go test --race $(FLAGS) $(PKGS)
+all-push: $(addprefix sub-push-,$(ALL_ARCH))
 
-container: build
-	docker build -t ${REGISTRY}/kube-state-metrics:$(TAG) .
+container: .container-$(ARCH)
+.container-$(ARCH):
+	cp -r * $(TEMP_DIR)
+	OOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -o $(TEMP_DIR)/kube-state-metrics
```
luxas: GOOS?

Pensu: Typo, Updated!

Makefile
```diff
@@ -1,12 +1,30 @@
...
 TESTENVVAR = 
 REGISTRY = gcr.io/google_containers
 TAG = $(shell git describe --abbrev=0)
 PKGS = $(shell go list ./... | grep -v /vendor/)
+ARCH ?= $(shell go env GOARCH)
+ALL_ARCH = amd64 arm arm64 ppc64le s390x
+
+
+IMAGE = $(REGISTRY)/kube-state-metrics
+MULTI_ARCH_IMG = $(IMAGE)-$(ARCH)
```
brancz: Newer versions of the docker image/registry format actually allow having the same repository for multiple arches iirc, would it be possible to use that here instead, as this is otherwise a breaking change compared to the previously distributed image repository, also is the busybox base actually necessary? Seems like a significant change compared to previously used `scratch`.

Pensu: @brancz I guess you are talking about manifest list, I don't think that has been implemented for gcr repo yet! So, until then we need to have different images for different arch.
I have removed busybox and put the scratch back.

Makefile
```diff
@@ -22,22 +25,40 @@ doccheck:
...
+
+test-unit: clean build
+	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(TESTENVVAR) go test --race $(FLAGS) $(PKGS)
+
+TEMP_DIR := $(shell mktemp -d)
+
+all: all-container
+
+sub-container-%:
+	$(MAKE) ARCH=$* container
```
andyxning: Prefix `$(MAKE)` with `@`, i.e., `@$(MAKE)`. This way make will not print something like `make[1]: Leaving directory '/home/tiger/gopath/src/k8s.io/kube-state-metrics'`

Makefile
```diff
@@ -22,22 +25,40 @@ doccheck:
...
+
+sub-push-%:
+	$(MAKE) ARCH=$* push
```
andyxning: ditto

Makefile
andyxning: After some Googling, it turns out [the tools to create and push these manifest lists don't exist yet, but support for pull a manifest list has been aded to Docker 1.10](https://github.com/moby/moby/pull/18785). Also, there is a [tool](https://github.com/estesp/manifest-tool) to create and push manifest. But it is not the Docker official one. The tools for supporting manifest has not been supported well for now and @Pensu has added a compatibility to push `kubernetes/kube-state-metrics:v1.0.1` when arch is 'amd64'.

So, this should be fine. @brancz WDYT.

Makefile
```diff
@@ -22,22 +25,40 @@ doccheck:
...
-	$(COMMONENVVAR) $(TESTENVVAR) godep go test --race $(FLAGS) $(PKGS)
+all-push: $(addprefix sub-push-,$(ALL_ARCH))
 
-container: build
-	docker build -t ${REGISTRY}/kube-state-metrics:$(TAG) .
+container: .container-$(ARCH)
+.container-$(ARCH):
+	cp -r * $(TEMP_DIR)
+	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -o $(TEMP_DIR)/kube-state-metrics
+	docker build -t $(MULTI_ARCH_IMG):$(TAG) $(TEMP_DIR)
```
andyxning: @Pensu Should add  a check for amd64 arch and build `$(IMAGE):$(TAG)` image or otherwise an error will be fired when later pushing.

Makefile
```diff
@@ -22,22 +25,40 @@ doccheck:
...
+	gcloud docker -- push $(MULTI_ARCH_IMG):$(TAG)
+ifeq ($(ARCH), amd64)
+	gcloud docker -- push $(IMAGE):$(TAG)
+endif
 
 clean:
 	rm -f kube-state-metrics
 
-.PHONY: all deps build test-unit container push clean
+.PHONY: all build test-unit container push clean
```
andyxning: @Pensu Should `all-push`, `all-container` be added to `.PHONY`

Makefile
mkumatag: As mention by @andyxning, we need to add a fallback case for amd64 to tag the image to older format without any arch in it.

Pensu: Have added a check for amd64 arch.

Makefile
Pensu: Done!

Makefile
Pensu: Updated

Makefile
Pensu: Updated

Makefile
```diff
@@ -22,22 +25,46 @@ doccheck:
...
+container: .container-$(ARCH)
+.container-$(ARCH):
+	cp -r * $(TEMP_DIR)
+	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -o $(TEMP_DIR)/kube-state-metrics
+	docker build -t $(MULTI_ARCH_IMG):$(TAG) $(TEMP_DIR)
 
-test-unit: clean deps build
-	$(COMMONENVVAR) $(TESTENVVAR) godep go test --race $(FLAGS) $(PKGS)
+ifeq ($(ARCH), amd64)
+	# Adding check for amd 64
```
andyxning: `s/amd 64/amd64/g` @Pensu

andyxning:
@Pensu @mkumatag I want to postpone this PR until @brancz comment on this. :) This should not be long.

Makefile
andyxning: Others lgmt.

Pensu: Done

Pensu:
ping @brancz Can you please take a look?

brancz:
Sorry I was out sick, and am only slowly catching up with everything. I'm generally ok with these changes, but can't make any promises, on the release of these images. While I understand the need for these images, it's also a much higher burden of maintenance to now maintain different platforms as well and also release those. As the release process itself is not automated and will likely take more time and effort to get setup this work falls back on the maintainers of kube-state-metrics. That said, having these changes at least enables people to build these images easily for their own needs, which is already a win, but I can't currently make much promises on what the release of these images will look like.

Pensu:
@brancz Thanks a lot. Yeah, I understand the concerns with multi platfrom enablement but since we have been doing it for most of other projects in the same way, I don't think there would be much bigger issues. Would it be possible to just push the images to gcr repo once (for other archs, we can keep the amd64 intact) so that they can be used by everyone?

Pensu:
@brancz If you need any help with maintainance for multi-arch, please do let me know.

brancz:
The pain I described is primarily around the non-automated release process, which shouldn't be the case in the first place, but the circumstances we're in right now don't allow us to do it differently.

PR #253, opened 2017-09-09
add pprof

andyxning:
Add `pprof` support to kube-state-metrics. This will make debug more easily.

/cc @brancz @fabxc

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/253)

### DIFF
--- a/main.go
+++ b/main.go
@@ -34,6 +34,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
+	"net/http/pprof"
+
 	"k8s.io/kube-state-metrics/collectors"
 )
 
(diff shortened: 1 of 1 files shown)


brancz:
I'd prefer if we don't do this via init functions, I'd like to move away from the global DefaultHTTPServer (or whatever the name is), and instead instantiate our own mux and register the handlers on that. In the spirit of that and to allow that in the future let's "manually" register the handlers in our metrics server function for now, so we can visibly see that those handlers need to be added once we do move to our own mux. If you want to you can also do all of the above in this PR.

andyxning:
Curious about the difference about the default mux and customized mux. :) @brancz

brancz:
The thing is that we never know, what might be registering something onto a global http server. It's the same idea why basically all global and init style things are bad.

andyxning:
Sounds reasonable. Comments addressed. @brancz PTAL.

brancz:
lgtm, thanks for taking care of this!

Issue #122, opened 2017-04-12
[ingress]feature request


zouyee:
gentle ping  @brancz 
I will work on adding some metrics,here is the relevant information . would you have some suggeustions？
#### service
| Metric name| Metric type | Labels/tags |
| ---------- | ----------- | ----------- |
| kube_service_spec_servicetype | Gauge | `clusterIP `=&lt;cluster ip&gt; <br> `namespace`=&lt;service-namespace&gt; <br>`service`=&lt;service-name&gt;<br>`type`=&lt;ClusterIP\|NodePort\|None\&gt;
#### pv
| Metric name| Metric type | Labels/tags |
| ---------- | ----------- | ----------- |
| kube_persistentvolume_status | Gauge | `persistentvolume`=&lt;pv-name&gt; <br> `namespace`=&lt;pvc-namespace&gt; <br>`phase`=&lt;Bound\|Failed\|Pending\|Available\|Released&gt;<br>`volume`=&lt;pvc-namespace&gt;

 #### namespace
| Metric name| Metric type | Labels/tags |
| ---------- | ----------- | ----------- |
| kube_namespace_status_phase | Gauge | `name`=&lt;namespace-name&gt; <br> `namespace`=&lt;pvc-namespace&gt; <br>`phase`=&lt;Terminating\|Active&gt;<br>`create_time`=&lt;date-time the server time when this object was created&gt;  

#### ingress
| Metric name| Metric type | Labels/tags |
| ---------- | ----------- | ----------- |
| kube_ingress_info | Gauge | `name`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br>`create_time`=&lt;date-time the server time when this object was created&gt;
| kube_ingress_metadata-generation | Gauge | `name`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt;
|kube_ingress_loadbalancer|Gauge|`name`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt;<br>`IP`=&lt;IP set for loadbalancer&gt; <br>`hostname`=&lt;loadbalancer hostname based on dns&gt;

brancz:
These all sounds reasonable @zouyee ! :slightly_smiling_face:

zouyee:
thanks,sorry for my late reply

zouyee:
namespace DONE

zouyee:
[...]

PR #254, opened 2017-09-12
add collected resource objects length log

andyxning:
This PR will add the log about collected resource object number. This may be useful for debug as described in [#122 comment](https://github.com/kubernetes/kube-state-metrics/issues/112#issuecomment-328302552).

/cc @brancz @fabxc 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/254)


brancz:
good idea! lgtm :+1:

PR #256, opened 2017-09-14
Created kube_pod_container_status_waiting_reason

### LINKED ISSUE #255: Report waiting reason as well
It would be neat to report as well why a container is in `Waiting` state, for example, `ContainerCreating`, etc.

I'll try to pull this off but I thought it would be neat to have the issue as well in case I don't :D

caarlos0:
closes #255

1. not sure if this is the better approach to this
2. not sure if those are the only possible waiting reasons. Couldn't find docs on this, poke into the kubernetes source and added the ones I found :(

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/256)

### DIFF
--- a/Documentation/pod-metrics.md
+++ b/Documentation/pod-metrics.md
@@ -11,6 +11,7 @@
 | kube_pod_status_scheduled | Gauge |  `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `condition`=&lt;true\|false\|unknown&gt; |
 | kube_pod_container_info | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `image`=&lt;image-name&gt; <br> `image_id`=&lt;image-id&gt; <br> `container_id`=&lt;containerid&gt; |
 | kube_pod_container_status_waiting | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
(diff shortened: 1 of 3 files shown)


andyxning:
@caarlos0 

Thanks for your contributions!

After some search in Kubernetes, there are no enumerate values or predefined const state values for wait state. We absolutely can define some existing ones but we can not guarantee they will not change in the future Kuberenetes releases. This is feasible. So i  am not in favor of adding this.

smarterclayton:
Just these two are probably enough - ErrImagePull is the most common, and there are a few others.  While they won't change, knowing that you have large numbers of pods waiting for pulls vs anything else is probably a good signal.  I.e. if you suddenly have waits for anything else other than pulling, you're probably experiencing something very broken.

caarlos0:
@smarterclayton exactly, pods keep in `ContainerCreating` when docker is broken on the node, or if weave is out of IP addrs, etc...

would be nice to have metrics about this =)

brancz:
Agreed. We can white-list these reasons and prevent carnality explosion, but still have useful information.

As a side note: it more and more feels like we're building the infra-store rather than time-series, which is not necessarily a bad thing (if using Prometheus 2.0), just a thought. Actually kind of neat.

smarterclayton:
OomKill is huge too.  Right now that has to be inferred from discontinuities in the memory graph.

Yeah, agree on the infra-store side - I have that concern too.

caarlos0:
Hi, can we merge this? Anything else I need to do?

I'm having problems because I don't have this metric :(

brancz:
There is a job failing, you're missing documentation for this metric.

caarlos0:
@brancz oh, will fix that ASAP

caarlos0:
@brancz done, sorry for the delay

brancz:
lgtm :+1: thanks!

caarlos0:
thanks @brancz, 

any eta for a release?

brancz:
I can't give an exact ETA, but it should be happening in the upcoming weeks.

caarlos0:
Nice, thanks!

PR #259, opened 2017-09-15
add go flag to parse glog flags

drinktee:
It is need to parse glog flags. So we can use  glog flags such as --v=4, --logtostderr=true

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/259)

### DIFF
--- a/main.go
+++ b/main.go
@@ -130,6 +130,7 @@ func main() {
 	flags := pflag.NewFlagSet("", pflag.ExitOnError)
 	// add glog flags
 	flags.AddGoFlagSet(flag.CommandLine)
+	flags.Lookup("logtostderr").Value.Set("true")
 	flags.BoolVar(&options.inCluster, "in-cluster", true, `If true, use the built in kubernetes cluster for creating the client`)
 	flags.StringVar(&options.apiserver, "apiserver", "", `The URL of the apiserver to use as a master`)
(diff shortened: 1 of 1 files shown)


brancz:
For backward compatibility, we would need to still default to setting the `logtostderr` flag to true.

drinktee:
@brancz  I have set logtostderr to true. Please review this. Thx.

main.go
```diff
@@ -126,13 +126,11 @@ type options struct {
...
-	// configure glog
-	flag.CommandLine.Parse([]string{})
-	flag.Lookup("logtostderr").Value.Set("true")
-
 	options := &options{collectors: make(collectorSet)}
 	flags := pflag.NewFlagSet("", pflag.ExitOnError)
-
+	// add glog flags
+	flags.AddGoFlagSet(flag.CommandLine)
+	flags.Lookup("logtostderr").Value.Set("true")
```
andyxning: @drinktee @brancz 

Currently, `flags.Lookup("logtostderr").Value.Set("true")` will set `logtostderr` to `true` which will be conflict with the help output for `logtostderr`.

By adding  
```
flags.Lookup("logtostderr").DefValue = "true"
flags.Lookup("logtostderr").NoOptDefVal = "true"
```

On this way we can also set the default value for `logtostderr` to `true` and this will be displayed correctly in `help` output.

drinktee: Done

PR #261, opened 2017-09-18
add Compatibility matrix

zouyee:
Due to breaking changes between Kubernetes API versions, some features might not work in kube-state-metric ,using compatibility matrix  to mark  supported/unsupported version range

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/261)

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,24 @@ either by Prometheus itself or by a scraper that is compatible with scraping
 a Prometheus client endpoint. You can also open `/metrics` in a browser to see
 the raw metrics.
 
-## Kubernetes Version
+## Table of Contents
+
+- [Versioning](#versioning)
+  - [Kubernetes Version](#kubernetes-version)
+  - [Compatibility matrix](#compatibility-matrix)
+  - [Container Image](#container-image)
+- [Metrics Documentation](#metrics-documentation)
+- [Resource recommendation](#resource-recommendation)
+- [kube-state-metrics vs. Heaspter](#kube-state-metrics-vs-heapster)
+- [Setup](#setup)
+  - [Building the Docker container](#building-the-docker-container)
+- [Usage](#usage)
+  - [Kubernetes Deployment](#kubernetes-deployment)
+  - [Deployment](#deployment)
+
+### Versioning
+
+#### Kubernetes Version
 
 kube-state-metrics uses [`client-go`](https://github.com/kubernetes/client-go) to talk with
 Kubernetes clusters. The supported Kubernetes cluster version is determined by `client-go`.
@@ -25,7 +42,19 @@ The compatibility matrix for client-go and Kubernetes cluster can be found
 All additional compatibility is only best effort, or happens to still/already be supported.
 Currently, `client-go` is in version `v4.0.0-beta.0`.
 
-## Container Image
+#### Compatibility matrix
+
+| kube-state-metrics | client-go | **Kubernetes 1.4**  | **Kubernetes 1.5** | **Kubernetes 1.6** | **Kubernetes 1.7** |
+|--------------------|-----------|---------------------|--------------------|--------------------|--------------------|
+| **v0.4.0** |  v2.0.0-alpha.1   |          ✓          |         ✓          |        -           |         -          |
+| **v0.5.0** |  v2.0.0-alpha.1   |          ✓          |         ✓          |        -           |         -          |
+| **v1.0.x** |  4.0.0-beta.0     |          ✓          |         ✓          |        ✓           |         ✓          |
+| **master** |  4.0.0-beta.0     |          ✓          |         ✓          |        ✓           |         ✓          |
+
+- `✓` Fully supported version range.
+- `-` The Kubernetes cluster has features the client-go library can't use (additional API objects, etc).
+
+#### Container Image
 
 The latest container image can be found at:
 * `quay.io/coreos/kube-state-metrics:v1.0.1`
@@ -35,7 +64,7 @@ The latest container image can be found at:
 The recommended docker registry for kube-state-metrics is `quay.io`. kube-state-metrics on
 `gcr.io` is only maintained on best effort as it requires external help from Google employees.
 
-## Metrics Documentation
+### Metrics Documentation
 
 There are many more metrics we could report, but this first pass is focused on
 those that could be used for actionable alerts. Please contribute PR's for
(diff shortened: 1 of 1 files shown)


brancz:
I'm not sure we should be introducing a compatibility matrix here. We already have this information here: https://github.com/kubernetes/kube-state-metrics#kubernetes-version. Even if we do decide to introduce such a matrix, we will need to adapt it to match the client-go matrix.

andyxning:
I am against adding compatibility matrix for kube-state-metrics directly. 

As stated by @brancz , we have declared that kube-state-metrics depends on client-go to keep compatibility with Kubernetes. We should not state the compatibility matric again and it is difficult to keep in sync with client-go's compatibility matrix manually.

matthiasr:
As a user, how do I know which version of kube-state-metrics was built
against which version of client-go?

brancz:
I think @matthiasr has a point, the compatibility matrix purely by client-go version is not completely obvious from a user perspective. To be honest I don't know which version of kube-state-metrics was compiled with which version of client-go, without checking the code.

andyxning:
Sounds reasonable. Maybe adding a compatibility matrix for kube-state-metrics about client-go will meeet what @mattfarina proposed.

Also, by doing so, we do actually get a way from kube-state-metrics -> client-go -> Kubernetes which should be clearly enough for users for follow along and find the supported Kubernetes.

mattfarina:
@andyxning What I proposed? Did you mean @matthiasr?

Is there a place inside of client-go that you can get the version information? For client-go and for the k8s cluster it's pointed against? From a user experience point of view... if i were using something based against client-go 3.x and I was talking to a k8s 1.7 cluster I would want to know about compatibility issues from versions.

Just thinking out loud.

andyxning:
@mattfarina Sorry for the mistaken @. :) And, thanks for your participation in talking about this.

IIUC, @brancz you agree that we need to add a matrix about client-go and kube-state-metrics instead of kube-state-metrics and Kubernetes.

zouyee:
ordinary users: may be want to know the version of kube-state-metrcis against k8s versions.
developer : want to know the version of kube-state-metrics which is compatible with the version of client-go and kubernetes.

brancz:
I'm thinking of a direct mapping of kube-state-metrics to Kubernetes version, that's really what users care about. This should be relatively low maintenance, basically only a new line needs to be added when we do a new release.

andyxning:
How about we add client-go version for each kube-state-metrics next to the version of kube-state-metrics. I.e.,

|  kube-state-metrics version (client-go version)  | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
|---------------------|----------------|----------------|----------------|----------------|
| **v0.4.0**   | ✓              |✓               | ✓              | ✓              |
| **v0.5.0**  | ✓              | ✓              | ✓              | ✓              |
| **v1.0.x**(**4.0.0-beta.0**)   | ✓              | ✓              | ✓              | ✓              |
| **master**(**4.0.0-beta.0**)   | ✓              | ✓              | ✓              | ✓              |

BTW, Maybe merge of sections [Kubernetes Version](https://github.com/zouyee/kube-state-metrics/blob/7f7aa20ef31cdf3ba376cda793d72aa66ee04c61/README.md#kubernetes-version) and [Compatibility matrix](https://github.com/zouyee/kube-state-metrics/blob/7f7aa20ef31cdf3ba376cda793d72aa66ee04c61/README.md#compatibility-matrix) will make README more compact.

brancz:
I like the idea of at least mentioning the client-go version and merging Kubernetes Version and Compatibility Matrix sections. We just need to make sure to actually adapt that matrix as I don't think all the check marks are supposed to be there.

andyxning:
@brancz I have updated the example compatibility matrix in [#261 comment](https://github.com/kubernetes/kube-state-metrics/pull/261#issuecomment-330893931).

@zouyee Could you please update the PR according to [#261 comment1](https://github.com/kubernetes/kube-state-metrics/pull/261#issuecomment-330893931) and [#261 comment2](https://github.com/kubernetes/kube-state-metrics/pull/261#issuecomment-331080091).

@matthiasr @mattfarina WDYT?

matthiasr:
if you're adding the client-go version per kube-state-metrics version, why not make it another column? I think it would be easier to read than adding two pieces of information into one cell.

brancz:
Fair point @matthiasr. I think a separate column makes sense.

Thanks for updating @zouyee! I think once we have the client-go version in a separate column this should be good to merge. Just one more thing, since you're already at it, you can also remove the "Kubernetes Version" section we currently have, that mentions the compatibility guarantees - the sections seem duplicated.

brancz:
@zouyee I think there was a misunderstanding, the idea was just to have a separate column that also states the client-go version separate from the kube-state-metrics version where previously it was "kube-state-metrics-version/client-go-version".

Basically this:

|  kube-state-metrics version | client-go version | Kubernetes 1.4 | Kubernetes 1.5 | Kubernetes 1.6 | Kubernetes 1.7 |
|---------------------|---------|----------------|----------------|----------------|----------------|
| **v0.4.0** | **v2.0.0-alpha.1**  | ✓              |✓               | -              | -              |
| **v0.5.0** | **v2.0.0-alpha.1** | ✓              | ✓              | -              | -              |
| **v1.0.x** | **4.0.0-beta.0**   | ✓              | ✓              | ✓              | ✓              |
| **master** | **4.0.0-beta.0**   | ✓              | ✓              | ✓              | ✓              |

zouyee:
sorry for my mistake :( 😅😅

brancz:
I'd say the past year of releases is appropriate so roughly the past 4-5 minor releases. New versions we should only add explicitly if we have actually released a kube-state-metrics version that is compiled with an officially compatible client-go version, before that it's just a best effort support.

brancz:
This lgtm, sorry I haven't followed up until now.

andyxning:
> When do we remove old Kubernetes versions? If we don't the table will get wider and wider.

IIUC, kubernetes keeps latest two release maintainable. So as for kube-state-metrics, keep latest five release in the compatibility matrix should be fine. @matthiasr

> When do we add new Kubernetes versions? When they hit Beta, RC, release?

If this is needed to update kubernetes version, i ihink it should be fine to do this in the master(dev) branch. For example client-go uses master branch to keep up with kubernetes before a new compatible client-go version is released.

PR #262, opened 2017-09-20
Improve a usage message for collectors flag

superbrothers:
This PR Improves a usage message for collectors flag. It makes a sentence clear and shows a default value.

```
      --collectors string                Comma-separated list of collectors to be enabled. Defaults to "services,persistentvolumeclaims,nodes,replicationcontrollers,statefulsets,replicasets,resourcequotas,jobs,namespaces,daemonsets,deployments,limitranges,pods,cronjobs"
```

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/262)

### DIFF
--- a/main.go
+++ b/main.go
@@ -22,6 +22,7 @@ import (
 	"log"
 	"net/http"
 	"os"
+	"sort"
 	"strings"
 
 	"github.com/golang/glog"
@@ -83,7 +84,9 @@ type collectorSet map[string]struct{}
 
 func (c *collectorSet) String() string {
 	s := *c
-	return strings.Join(s.asSlice(), ",")
+	ss := s.asSlice()
+	sort.Strings(ss)
+	return strings.Join(ss, ",")
 }
 
 func (c *collectorSet) Set(value string) error {
--- a/main.go
+++ b/main.go
@@ -112,7 +112,7 @@ func (c collectorSet) isEmpty() bool {
 }
 
 func (c *collectorSet) Type() string {
-	return "map[string]struct{}"
+	return "string"
 }
 
 type options struct {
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -112,7 +112,7 @@ func (c collectorSet) isEmpty() bool {
 }
 
 func (c *collectorSet) Type() string {
-	return "map[string]struct{}"
```
andyxning: We should not change the type string here. 

With using `fmt.Sprintf("%v", &defaultCollectors)`, we can also get the default collectors `"services,persistentvolumeclaims,nodes,replicationcontrollers,statefulsets,replicasets,resourcequotas,jobs,namespaces,daemonsets,deployments,limitranges,pods,cronjobs"`.

superbrothers: I know, but I feel it is more understandable for me that Type method returns `string` than `map[string]struct{}`.

"string"
```
      --collectors string                Comma-separated list of collectors to be enabled. Defaults to "services,persistentvolumeclaims,nodes,replicationcontrollers,statefulsets,replicasets,resourcequotas,jobs,namespaces,daemonsets,deployments,limitranges,pods,cronjobs"
```

"map[string]struct{}"
```
      --collectors map[string]struct{}   Comma-separated list of collectors to be enabled. Defaults to "nodes,replicationcontrollers,cronjobs,persistentvolumeclaims,namespaces,daemonsets,jobs,limitranges,replicasets,resourcequotas,services,deployments,statefulsets,pods"
```

What do you think?

brancz: I think for non go developers it's hard to understand that `map[string]struct{}` is a set, so I think the method @andyxning suggested will be easier to understand for users.

andyxning: > I think for non go developers it's hard to understand that map[string]struct{} is a set

I think what @brancz means is that what @superbrothers suggests is more friendly to end users. :) So, 

/lgtm

andyxning:
@superbrothers Thanks for the contribution. 👏

main.go
brancz: Sorry yes, that was correct I mixed up who said what :slightly_smiling_face:, thanks for the contribution @superbrothers !

Issue #257, opened 2017-09-15
kube-state-metrics consuming too much memory


jac-stripe:
kube-state-metrics is using >400mb of RAM. It is also very slow when I query `/metrics`. The kubernetes cluster has 2700 job objects. It seems surprising that this would consume 400mb of RAM for metrics aggregation. Below is a pprof top trace. This is running the latest git revision (d316c013fae8965bfb75bafda9453ca2ef54c48f)
```
(pprof) top
Showing nodes accounting for 526.72MB, 86.90% of 606.14MB total
Dropped 148 nodes (cum <= 3.03MB)
Showing top 10 nodes out of 110
      flat  flat%   sum%        cum   cum%
  195.01MB 32.17% 32.17%   202.01MB 33.33%  github.com/prometheus/client_golang/prometheus.makeLabelPairs
  101.26MB 16.71% 48.88%   148.26MB 24.46%  github.com/prometheus/client_golang/prometheus.(*Registry).Gather
   74.28MB 12.26% 61.13%    74.81MB 12.34%  k8s.io/kube-state-metrics/collectors.RegisterJobCollector.func1                                                                 47MB  7.75% 68.89%       47MB  7.75%  github.com/prometheus/client_golang/prometheus.populateMetric
   27.60MB  4.55% 73.44%    30.60MB  5.05%  k8s.io/client-go/pkg/api/v1.codecSelfer1234.decSliceVolume
   23.01MB  3.80% 77.24%    23.01MB  3.80%  runtime.rawstringtmp
   18.97MB  3.13% 80.37%    19.55MB  3.22%  github.com/golang/protobuf/proto.(*Buffer).EncodeStringBytes
   15.50MB  2.56% 82.92%   217.51MB 35.88%  github.com/prometheus/client_golang/prometheus.NewConstMetric
   13.50MB  2.23% 85.15%    14.02MB  2.31%  runtime.mapassign
   10.58MB  1.74% 86.90%    12.71MB  2.10%  compress/flate.NewWriter
```

brancz:
What does "very" slow mean? Up to 10 seconds response time wouldn't be unusual for a huge request. 

[...]

PR #76, opened 2017-01-27
Add kube_deployment_spec_strategy_rollingupdate_max_unavailable

vdavidoff:
This documentation update got missed in [PR 75](https://github.com/kubernetes/kube-state-metrics/pull/75).

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/76)

PR #264, opened 2017-09-21
use protobuf instead of json

brancz:
client-go defaults to using json as the content type, as suggested in #257 it would be a good idea to use protobuf instead. As currently all objects used in kube-state-metrics support protobuf, this should be safe to do.

@smarterclayton @andyxning

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/264)

### DIFF
--- a/main.go
+++ b/main.go
@@ -201,8 +201,10 @@ func isNotExists(file string) bool {
 }
 
 func createKubeClient(inCluster bool, apiserver string, kubeconfig string) (kubeClient clientset.Interface, err error) {
+	var config *rest.Config
+
 	if inCluster {
-		config, err := rest.InClusterConfig()
+		config, err = rest.InClusterConfig()
 		if err != nil {
 			return nil, err
 		}
@@ -217,9 +219,6 @@ func createKubeClient(inCluster bool, apiserver string, kubeconfig string) (kube
 		}
 		glog.Infof("service account token present: %v", tokenPresent)
 		glog.Infof("service host: %s", config.Host)
-		if kubeClient, err = clientset.NewForConfig(config); err != nil {
-			return nil, err
-		}
 	} else {
 		loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 		// if you want to change the loading rules (which files in which order), you can do so here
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -217,26 +219,25 @@ func createKubeClient(inCluster bool, apiserver string, kubeconfig string) (kube
...
 		if err != nil {
 			return nil, err
 		}
-		kubeClient, err = clientset.NewForConfig(config)
-		if err != nil {
-			return nil, err
-		}
+	}
+
+	config.ContentType = "application/vnd.kubernetes.protobuf"
```
smarterclayton: Use `AcceptContentTypes = "application/vnd.kubernetes.protobuf,application/json` instead

brancz: Yes, of course. Now we're getting an error though in the node collector:

```
E0921 18:19:56.181928    1414 streamwatcher.go:109] Unable to decode an event from the watch stream: invalid character '\x00' looking for beginning of value
E0921 18:19:56.186126    1414 streamwatcher.go:109] Unable to decode an event from the watch stream: invalid character '\x00' looking for beginning of value
W0921 18:19:56.186202    1414 reflector.go:323] k8s.io/kube-state-metrics/collectors/node.go:140: watch of *v1.Node ended with: very short watch
```

Any idea?

andyxning: seems like encoder and decoder mismatch

andyxning: @smarterclayton What is the difference between `application/vnd.kubernetes.protobuf` and `application/vnd.kubernetes.protobuf,application/json`.

andyxning: According to [client-go #76](https://github.com/kubernetes/client-go/issues/76), protobuf is not the default data type and may change in the future.

@smarterclayton @brancz Should we make this optional through command line argument.

brancz: @andyxning My guess would be that `application/vnd.kubernetes.protobuf,application/json` is content negotiation, that when an endpoint doesn't support proto, then it responds in json and sets the content type header appropriately.

I don't think we should make it optional, I don't see a reason why people would want to choose.

andyxning: @brancz If apiserver supports content negotiation, it should be fine to use this and we do not need to make protobuf optional since client-go should fall back to json data type once protobuf is not supported.

andyxning: @brancz Thanks for pointing this out. It should be according to [rfc 2296](https://tools.ietf.org/html/rfc2296).

andyxning: I have tested this PR with Kubernetes 1.6.3. Everything is fine. However, in order to debug i suggest we get and log the actual content type used between apiserver and kube-state-metrics.

andyxning:
This is the most important improvement i want for v1.2.0 milestone. I will first learn client-go, and will continue work on this.

main.go
xigang: > @brancz If apiserver supports content negotiation, it should be fine to use this and we do not need to make protobuf optional since client-go should fall back to json data type once protobuf is not supported.

@andyxning I used crd typed client and  set `application/vnd.kubernetes.protobuf,application/json` is content negotiation. but the client is abnormal.

```
Failed to watch xxxxxx: failed to list *v1alpha1.DuplicateSet: object *unstructured.UnstructuredList does not implement the protobuf marshalling interface and cannot be encoded to a protobuf message
```

https://kubernetes.io/docs/reference/using-api/api-concepts/#alternate-representations-of-resources

Issue #265, opened 2017-09-25
multi-version resource metrics supported


zouyee:
Is it suitable to change the arch of collectors like [typed](https://github.com/kubernetes/client-go/tree/release-5.0/kubernetes/typed),according to the comments of the [PR](https://github.com/kubernetes/kube-state-metrics/pull/220/)
![2017-09-25 9 53 41](https://user-images.githubusercontent.com/5785605/30789475-619afe38-a1d8-11e7-8744-9acb70317597.png)
@andyxning @brancz

brancz:
Our compatibility guarantees revolve around the client-go version so as APIs get deprecated we move to the new APIs. I feel that is going to be a better experience for the user, then breaking all metrics available, as API changes are generally small throughout the evolution of an object.

andyxning:
This is a good question. What if there are two available group versions for a resource object. As for now, we hard code the specific group version for a resource version.

I am not quite sure that whether there are chances more than one group versions are available.

Aside from the above concerns, what if a reource graduate from beta to GA, which version should kube-state-metrics supports. 

BTW, log the currently used resource version for a resource should be user friendly for debugging.

andyxning:
This will definitely make kube-state-metrics more easy to avolve and maintain. I agree with this approach. I suggest we add a statement about the resource version behavior in readme to make this clear to end users.  @brancz WDYT.

andyxning:
Sorry for the press wrong button. :(

brancz:
No worries, clicking the wrong button happens :slightly_smiling_face: . A note in the readme sounds good. Probably most suitable in the upcoming compatibility matrix section.

andyxning:
yep. Kubernebtes resource group version compatibility. @zouyee since you are the contributor to the kunernetes compatibility pr, could you please update the pr by adding the new resource group version compatibility.

vincevaghalsen:
Only if the client generates programmatic IDs to exclude redundant metrics. Prometheus should always be initiated with a multilevel manifest in the cluster.

PR #266, opened 2017-09-27
[kubernetes 1.8] update client-go(release-5.0) and add k8s.io/api(rel…

zouyee:
- [x] update client-go(release-5.0) for kubernetes 1.8

- [x] add k8s.io/api for kubernetes 1.8

- [ ] migrate DaemonSet, Deployment, ReplicaSet, and StatefulSet from v1beta1 to v1beta2

- [ ] migrate cronjob from v1alpha1 to v1beta1

@andyxning  @brancz PTAL!

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/266)


brancz:
> migrate DaemonSet, Deployment, ReplicaSet, and StatefulSet from v1beta1 to v1beta2

When are the v1beta1 objects targeted for removal, if at all? I would prefer keeping the metrics as is as long as possible, if there are any there breaking changes for us, as it will keep backward compatibility for users of older clusters.

zouyee:
- [ ] keep v1beta1 metrics of  DaemonSet, Deployment, ReplicaSet, and StatefulSet 

- [ ] add v1beta2 metrics of  DaemonSet, Deployment, ReplicaSet, and StatefulSet 

- [ ] add version info in these metrics

- [ ] merge [feture collector](https://github.com/kubernetes/kube-state-metrics/pull/220)
@brancz WDYT?

andyxning:
@brancz According to #265, i think we should keep a single geoup version for a resource, right?

IIUC, we should update to v1beta2 for deployment since it is the latest stable support for deployment.

As a re think about this, i think what the compatibilty promise difference between client-go and kube-state-meteics is that client-go supports different group version support for any resource bit kube-state-metrics does not.

Currently used group version for a resource should also be considered as a compability promise since there are more than one group version available.

brancz:
Sorry if I wasn't clear enough, I meant we in kube-state-metrics support the oldest API available in the latest release. That would mean that we stick to `v1beta1` until those are actually permanently removed (if that actually happens).

zouyee:
Whether we have plans to support multiple versions,such as v1beta1(for backward compatibility) and v1beta2(for new changes) ?
For example, observedGeneration has been changed in apps/v1beta2 StatefulSet observedGeneration from a pointer to an int for consistency.

brancz:
> For example, observedGeneration has been changed in apps/v1beta2 StatefulSet observedGeneration from a pointer to an int for consistency.

Things like that don't make a difference for the generated metric. The point is that we at least need to try to support older APIs, but *must* support latest stable Kubernetes (latest as in latest stable at the point where we release kube-state-metrics).

zouyee:
>Things like that don't make a difference for the generated metric. The point is that we at least need to try to support older APIs, but must support latest stable Kubernetes (latest as in latest stable at the point where we release kube-state-metrics).
ok,I got it,but is it a more appropriate approach to update the versions of client-go and introduce api package which could support new features better?

brancz:
As part of our effort to support as many Kubernetes versions as possible we'll have to hold off on newest APIs and use what will support the most versions.

zouyee:
thank you for your response, I got it

andyxning:
@zouyee The first two tasks has been done in #269 . PTAL.

andyxning:
close this PR via #271 and #269 .

PR #267, opened 2017-09-28
log the group version for resources

andyxning:
This PR will log the currently used group version for each resource. This should be user friendly in debugging as discussed in [#256 comment](https://github.com/kubernetes/kube-state-metrics/issues/265#issuecomment-332525436).

/cc @brancz @matthiasr

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/267)


zouyee:
👍

PR #276, opened 2017-10-17
added termination reason metric

caarlos0:
very much like #256, but for containers that were terminated.

It should be useful to detect apps that are being constantly OOMKilled, for example.

I got the possible reasons from kubernetes codebase:

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_container.go#L328-L361

+cc @brancz 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/276)


brancz:
We need to be careful with the amount of metrics we add especially around pods and containers, as they grow linearly with the other objects. This seems like extremely useful information though so I think it's a valuable addition. We're currently in the last steps of releasing `v1.1.0` so we're gonna hold off on merging this until after `v1.1.0` has gone GA.

Code wise this lgtm :+1: thanks!

caarlos0:
Nice, once v1.1.0 is out I'll ping you again here then 🍻

caarlos0:
@brancz ping

brancz:
Thanks for the ping!

caarlos0:
Thanks :D

PR #280, opened 2017-10-22
update deploy config apiversion

zouyee:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/280)

### DIFF
--- a/kubernetes/kube-state-metrics-role-binding.yaml
+++ b/kubernetes/kube-state-metrics-role-binding.yaml
@@ -1,4 +1,4 @@
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1
 kind: RoleBinding
 metadata:
   name: kube-state-metrics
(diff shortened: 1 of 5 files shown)


kubernetes/kube-state-metrics-role-binding.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1beta2
```
andyxning: This should not be changed to `v1beta2` as `Role` and `RoleBinding` has moved to `v1` along side `ClusterRoleBinding` and `ClusterRole`.

@zouyee

kubernetes/kube-state-metrics-role.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: rbac.authorization.k8s.io/v1beta1
+apiVersion: rbac.authorization.k8s.io/v1beta2
```
andyxning: ditto

andyxning:
Others lgtm.

PR #282, opened 2017-10-22
Add simple e2e test

andyxning:
This PR add simple e2e test for kube-state-metrics.

This PR leverage minikube to setup an available Kubernetes cluster. This PR reference https://kinvolk.io/blog/2017/10/running-kubernetes-on-travis-ci-with-minikube/ doc.

e2e test logic
* set up an kubernetes v1.8.0 cluster with minikube
* apply kube-state-metrics manifests. Exit with errors
* build and run kube-state-metrics in standalone
* wait for kube-state-metrics to run by accessing `healthz` endpoint
* access `metrics` endpoint
* access `healthz` endpoint to check for that kube-state-metrics is still running after accessing `metrics` endpoint.

**Note**: 
After adding promtool as the metric validate tool, `kube_pod_container_status_restarts` as a Counter is not a valid metric name. This will result in `kube_pod_container_status_restarts: counter metrics should have "_total" suffix` error.

So, in order to make promtool work properly, A name from `kube_pod_container_status_restarts` to `kube_pod_container_status_restarts_total` will be made. This is a breaking change. Which should  be noted additionally.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/282)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,14 @@
+sudo: required
+
 language: go
+
 go:
 - 1.8
+
 install:
 - mkdir -p $HOME/gopath/src/k8s.io
 - mv $TRAVIS_BUILD_DIR $HOME/gopath/src/k8s.io/kube-state-metrics
+
 jobs:
   include:
     - stage: Go fmt
@@ -14,3 +19,5 @@ jobs:
       script: make build
     - stage: Check that all metrics are documented
       script: make doccheck
+    - stage: E2e
+      script: make e2e
--- a/collectors/pod_test.go
+++ b/collectors/pod_test.go
@@ -50,8 +50,8 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_labels gauge
 		# HELP kube_pod_container_status_ready Describes whether the containers readiness check succeeded.
 		# TYPE kube_pod_container_status_ready gauge
-		# HELP kube_pod_container_status_restarts The number of container restarts per container.
-		# TYPE kube_pod_container_status_restarts counter
+		# HELP kube_pod_container_status_restarts_total The number of container restarts per container.
+		# TYPE kube_pod_container_status_restarts_total counter
 		# HELP kube_pod_container_status_running Describes whether the container is currently in running state.
 		# TYPE kube_pod_container_status_running gauge
 		# HELP kube_pod_container_status_terminated Describes whether the container is currently in terminated state.
@@ -213,11 +213,11 @@ func TestPodCollector(t *testing.T) {
 				},
 			},
 			want: metadata + `
-				kube_pod_container_status_restarts{container="container1",namespace="ns1",pod="pod1"} 0
-				kube_pod_container_status_restarts{container="container2",namespace="ns2",pod="pod2"} 0
-				kube_pod_container_status_restarts{container="container3",namespace="ns2",pod="pod2"} 1
+				kube_pod_container_status_restarts_total{container="container1",namespace="ns1",pod="pod1"} 0
+				kube_pod_container_status_restarts_total{container="container2",namespace="ns2",pod="pod2"} 0
+				kube_pod_container_status_restarts_total{container="container3",namespace="ns2",pod="pod2"} 1
 				`,
-			metrics: []string{"kube_pod_container_status_restarts"},
+			metrics: []string{"kube_pod_container_status_restarts_total"},
 		}, {
 			pods: []v1.Pod{
 				{
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -1,0 +1,143 @@
+#!/bin/bash
+
+# Copyright 2017 The Kubernetes Authors All rights reserved.
+
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#    http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+set -o pipefail
+
+KUBERNETES_VERSION=v1.8.0
+KUBE_STATE_METRICS_LOG_DIR=./log
+KUBE_STATE_METRICS_IMAGE_NAME='quay.io/coreos/kube-state-metrics'
+KUBE_STATE_METRICS_IMAGE_NAME_PATTERN='quay.io\/coreos\/kube-state-metrics'
+PROMETHEUS_VERSION=1.8.1
+
+mkdir -p $KUBE_STATE_METRICS_LOG_DIR
+
+# setup a Kubernetes cluster
+curl -sLo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+
+minikube version
+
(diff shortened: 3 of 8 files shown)


scripts/e2e.sh
```diff
@@ -0,0 +1,59 @@
...
+
+export KUBECONFIG=$HOME/.kube/config
+sudo minikube start --vm-driver=none --kubernetes-version=v1.8.0
+
+# this for loop waits until kubectl can access the api server that Minikube has created
+for i in {1..150}; do # timeout for 5 minutes
+   echo "--------" 
+   cat $HOME/.kube/config
+   minikube logs
+   cat /Users/andy/.minikube/machines/minikube/config.json
```
zouyee: delete it.

.travis.yml
```diff
@@ -1,16 +1,15 @@
...
 go:
 - 1.8
+
 install:
 - mkdir -p $HOME/gopath/src/k8s.io
 - mv $TRAVIS_BUILD_DIR $HOME/gopath/src/k8s.io/kube-state-metrics
+
 jobs:
   include:
-    - stage: Go fmt
```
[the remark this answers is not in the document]
andyxning: @brancz Yep. Just WIP for now. When it is ready, we should only add an additional `make e2e`.

andyxning:
Ping @brancz 

This PR is ready to review. :)

scripts/e2e.sh
```diff
@@ -0,0 +1,127 @@
...
+   minikube logs
+   echo "Kubernetes does not start within 3 minutes"
+   exit 1
+fi
+
+set -e
+
+kubectl version
+
+# set up kube-state-metrics manifests
```
zouyee: could you combine these operations instead of using `kubectl apply -f ./kubernetes/`

andyxning: @zouyee I have tried about that. Seems `kubectl apply -f ./kubernetes/` could not report error with exit code non zero. We need this to make sure that kubernetes manifests are all good to use with test kuberentes cluster.

scripts/e2e.sh
```diff
@@ -0,0 +1,127 @@
...
+    echo "kube-state-metrics does not start within 1 minute"
+    exit 1
+fi
+
+set -e
+
+echo "kube-state-metrics is up and running"
+
+echo "access kube-state-metrics metrics endpoint"
+curl -s "127.0.0.1:$KUBE_STATE_METRICS_PORT/metrics" >$KUBE_STATE_METRICS_LOG_DIR/metrics
```
brancz: We should check this metrics output with `promtool`. It can check wether metrics output is valid etc. That way if we mess up something with the metrics endpoint and get some http error and promtool fails.

scripts/e2e.sh
```diff
@@ -0,0 +1,127 @@
...
+kubectl create -f ./kubernetes/kube-state-metrics-role-binding.yaml
+kubectl create -f ./kubernetes/kube-state-metrics-role.yaml
+
+kubectl create -f ./kubernetes/kube-state-metrics-deployment.yaml
+
+kubectl create -f ./kubernetes/kube-state-metrics-service.yaml
+
+# set up a kube-state-metrics instance
+make build
+./kube-state-metrics --in-cluster=false --kubeconfig=$KUBECONFIG --port=$KUBE_STATE_METRICS_PORT --log_dir=$KUBE_STATE_METRICS_LOG_DIR --logtostderr=false &
```
brancz: Shouldn't we build the binary and container image before we start it as containers in the cluster? We primarily support running kube-state-metrics inside of the cluster, so that's the case we should be testing.

scripts/e2e.sh
brancz: Also we should check that the return status code is 200.

zouyee:
```
go get github.com/prometheus/prometheus/cmd/promtool
valid=`cat ./log/metrics |$GOPATH/bin/prometool check metrics`
...
```

andyxning: @brancz Will do

@zouyee Thanks for the tip.

scripts/e2e.sh
andyxning: @brancz I have thought about that. The main problem is that kube-state-metrics pod is deployed with `ClusterIP` service type. This is impossible for external tools to access the metrics endpoint of kube-state-metrics(Correct me if i am wrong).

By running kube-state-metrics in standalone mode, we can do this more easily. :)

BTW, as for the kube-state-metrics manifests, currently we can only check the syntax error with `kubectl create` because it will return with an non-zero exit code when something wrong happens.

The more desirable e2e test is we run at least one resource instance for each resource and we then access `/metrics` endpoint to check for the possible nil pointer deference panic. For now we only have kube-state-metrics manifests resource instances. This should be enhanced by another PR later.

andyxning: @brancz Opened issue #290 to track for adding more resource to e2e test.

brancz: The proxy endpoint of the service resource through the apiserver allows us to access ClusterIP type services.

That looks something like this when you use it with `kubectl proxy`:

```
kubectl proxy &
curl http://localhost:8001/api/v1/proxy/namespaces/kube-system/services/kube-state-metrics:8080/metrics
```

Using kubectl proxy makes this easier as it takes care of AuthN/AuthZ.

scripts/e2e.sh
andyxning: @brancz @zouyee 

**Note**: 
After adding promtool as the metric validate tool, `kube_pod_container_status_restarts` as a Counter is not a valid metric name. This will result in `kube_pod_container_status_restarts: counter metrics should have "_total" suffix` error.

So, in order to make promtool work properly, A name from `kube_pod_container_status_restarts` to `kube_pod_container_status_restarts_total` will be made. This is a breaking change. Which should  be noted additionally.

scripts/e2e.sh
andyxning: @brancz Awesome for pointing the usage of kubectl proxy. 👍 

The other reason that i want to use standalone mode is that we can easily get the log output of kube-state-metrics. This is good in case we want to check the log output. This can also accomplished by `kubectl logs`.

With e2e test, we need to check the newly kube-state-metrics binary. But in kube-state-metrics-deployment.yml, we use v1.1.0 kube-state-metrics image. In case to use the latest kube-state-metrics image, we need to:
* build the latest image
* get the latest image name and replace v1.1.0 in kube-state-metric-deployment.yml
* `kubectl create -f kube-state-metric-deployment.yml`

Actually i have thought about this when doing this PR. I will make the test more complicated and we should make sure that the image used in deployment is the latest build one.

brancz: Small tip: kubectl also understands json, which is probably easier to modify in code than yaml (eg. yaml -> json -> modify with jq -> use with kubectl create)

andyxning: @brancz Comments addressed. PTAL.

collectors/pod_test.go
```diff
@@ -50,8 +50,8 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_labels gauge
 		# HELP kube_pod_container_status_ready Describes whether the containers readiness check succeeded.
 		# TYPE kube_pod_container_status_ready gauge
-		# HELP kube_pod_container_status_restarts The number of container restarts per container.
-		# TYPE kube_pod_container_status_restarts counter
+		# HELP kube_pod_container_status_restarts_total The number of container restarts per container.
+		# TYPE kube_pod_container_status_restarts_total counter
```
brancz: I'm fine with this change, we just have to explicitly declare it in the release notes.

brancz:
Nice job :+1: lgtm

collectors/pod_test.go
andyxning: @brancz Yep. This is a breaking change.

howardburgess: This should have been mentioned in the release notes https://github.com/kubernetes/kube-state-metrics/releases/tag/v1.2.0-rc.0

PR #284, opened 2017-10-24
fix travis ci build

andyxning:
As for now `make` does not equal to `make build`. We should make it explicitly to test with `make build`.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/284)


brancz:
lgtm, I also prefer explicitness

PR #285, opened 2017-10-24
Omit static compute resource requests and limits from deployment

svend:
The addon-resizer will dynamically set the pod compute resource values. If the
values are also set statically in the deployment configuration, reapplying the
configuration will result in the pods getting replaced. Without the static
resource, the deployment configuration can be reapplied, and the pods will not
be replaced.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/285)

### DIFF
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -26,13 +26,6 @@ spec:
             port: 8080
           initialDelaySeconds: 5
           timeoutSeconds: 5
-        resources:
-          requests:
-            memory: 100Mi
-            cpu: 100m
-          limits:
-            memory: 200Mi
-            cpu: 200m
       - name: addon-resizer
         image: gcr.io/google_containers/addon-resizer:1.0
         resources:


brancz:
Just to make sure I understand correctly, the resource request and limit will be applied to the running pods, without disruption?

svend:
By design, the addon-resizer will periodically update the resource settings for the deployment, which results in the pods getting restarted if the resource settings changed.

https://github.com/kubernetes/autoscaler/blob/74d20a6cc732be5058ef8fefb739240ca3d75e8c/addon-resizer/README.md#nanny-program-and-arguments

Having the resource limits in the deployment configuration means applying the config will overwrite the resource limits set by the addon-resizer. This results in the pods getting recreated with the statically defined resource limits. Then the addon-resizer runs, sets the resource limits back to what it thinks they should be, and the pods get recreated again.

andyxning:
@svend IIUC, you mean that without setting a limit and request for kube-state-metrics container, we can avoid container stop and recreate?

Does Kubelet will restart a container when the request or limit resource change? If this is the truth, by avoid setting limit and request for kube-state-metrics container can only stop kube-state-metrics from recreate at start. After that, once a new limit or request will be updated, it will be recreated also.

brancz:
That how I understood it at first as well @andyxning , but with the last comment I'm not sure anymore. If this avoids the pod being recreated at initial deployment I'm happy with accepting the change, otherwise I think it's nice to give people a base-line default recommendation.

svend:
@andyxning The addon-resizer will change the deployment's resource settings if the current resource settings are not within 5% of the calculated resource setttings (`--threshold=5`), which will depend on the number off nodes and the extra cpu and memory per node (`--extra-cpu=1m` and`--extra-memory=2Mi`). Any change the deployment's resource settings change the deployment's .spec.template which results in a rollout (and the running pods are replaced).

@brancz This change shouldn't change the initial deployment behavior. Currently, the initial deployment is created with the static resource settings, then the addon-resizer will update the resource settings, which will result in a rollout.

This PR is to avoid unnecessary rollouts on subsequent runs of `kubectk apply`.

Here is an example of the current behavior.

```
# Create a new deployment
$ kubectl -n ksm-test apply -f kube-state-metrics-deployment.yaml
deployment "kube-state-metrics" created

# The initial pod is getting replaced
$ kubectl -n ksm-test get po -l k8s-app=kube-state-metrics
NAME                                  READY     STATUS        RESTARTS   AGE
kube-state-metrics-2374292213-fkz5m   2/2       Running       0          14s
kube-state-metrics-2673708515-996dh   0/2       Terminating   0          16s

# New, resized pod is running
$ kubectl -n ksm-test get po -l k8s-app=kube-state-metrics
NAME                                  READY     STATUS    RESTARTS   AGE
kube-state-metrics-2374292213-fkz5m   2/2       Running   0          35s

# Run apply again
$ kubectl -n ksm-test apply -f kube-state-metrics-deployment.yaml
deployment "kube-state-metrics" configured

# Second run of apply triggered a rollout
$ kubectl -n ksm-test get po -l k8s-app=kube-state-metrics
NAME                                  READY     STATUS        RESTARTS   AGE
kube-state-metrics-2374292213-fkz5m   2/2       Running       0          52s
kube-state-metrics-2673708515-9124z   1/2       Terminating   0          3s
```

The rollout on the second run of `kubectl apply` is what I'm trying to avoid with this PR.

brancz:
Got it! Thanks for the explanation, this makes sense and lgtm :+1:!

PR #291, opened 2017-10-30
Update CronJobs to v1beta1

f0:
with kubernetes 1.8 cronjobs are at version v1beta1, and enabled per default

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/291)

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -27,7 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
-	v2batch "k8s.io/api/batch/v2alpha1"
+	v2batch "k8s.io/api/batch/v1beta1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/tools/cache"
 )
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -27,7 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
-	v2batch "k8s.io/api/batch/v1beta1"
+	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/tools/cache"
 )
@@ -70,9 +70,9 @@ var (
 	)
 )
 
-type CronJobLister func() ([]v2batch.CronJob, error)
+type CronJobLister func() ([]batchv1beta1.CronJob, error)
 
-func (l CronJobLister) List() ([]v2batch.CronJob, error) {
+func (l CronJobLister) List() ([]batchv1beta1.CronJob, error) {
 	return l()
 }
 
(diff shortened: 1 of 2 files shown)


f0:
cla singed

collectors/cronjob.go
```diff
@@ -27,7 +27,7 @@ import (
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
 
-	v2batch "k8s.io/api/batch/v2alpha1"
+	v2batch "k8s.io/api/batch/v1beta1"
```
andyxning: @f0 Seems that `batch/v2alpha1` is still available for k8s 1.8

andyxning: @brancz This is where we need to make resource group version configurable. I.e., for a single resource, users can specify which resource version to use aside from the default one which is for now we use the longest support one.

andyxning: I have no idea about whether it is possible to implement this. Because we need to change which package to import for the user specified group version. 

Automatically group version negotiation is also an option. But it is out of control when the auto detect group version is not the desired one. 

What i thought is to support one command option to let user to specify the desired resource group version.

brancz: I think making this generally configurable is going to create chaos. I believe we should just stick to the oldest API available in the latest Kubernetes release. That way we have the broadest compatibility matrix, while keeping it simple.

f0: @andyxning yes, but its recomended to move to v1beta1. v2alpha1 needs to be enabled explicit, v1beta1 is enabled per default (because it's beta now).

brancz: I would expect that once the beta API is automatically enabled the alpha API is still available and also autoenabled for backward compatibility.

f0: @brancz  yes its available, but if i read the changelog right, there are some downsides with v2alpha1
From the Changelog:
'''
Be aware that if you specify the deprecated version, you may encounter Resource Not Found errors. These errors occur because the new controllers look for the new version during a rolling update.
'''

brancz: I see. Then we should move to v1beta1 I think. Generally we can't give any guarantees around alpha objects I think. @andyxning we should note that in the compatibility guarantees.

collectors/cronjob_test.go
```diff
@@ -20,7 +20,7 @@ import (
 	"testing"
 	"time"
 
-	v2batch "k8s.io/api/batch/v2alpha1"
+	v2batch "k8s.io/api/batch/v1beta1"
```
brancz: Can you fix these import renames? v2 is now off.

f0: @brancz  hm do you mean v2beta1? Can not find this version in the kubernetes docs

brancz: I mean `v2batch` should be `batchv1beta1`.

f0:
@brancz   fixed

brancz:
Lgtm, thanks!

collectors/cronjob.go
```diff
@@ -70,21 +70,21 @@ var (
...
-type CronJobLister func() ([]v2batch.CronJob, error)
+type CronJobLister func() ([]batchv1beta1.CronJob, error)
 
-func (l CronJobLister) List() ([]v2batch.CronJob, error) {
+func (l CronJobLister) List() ([]batchv1beta1.CronJob, error) {
 	return l()
 }
 
 func RegisterCronJobCollector(registry prometheus.Registerer, kubeClient kubernetes.Interface, namespace string) {
 	client := kubeClient.BatchV2alpha1().RESTClient()
```
cpointner: Shouldn't that become `BatchV1beta1`?

brancz: @cpointner you are right! Do you mind creating a PR for this?

cpointner: #311 (CLA is in the works)

brancz: Thanks a lot! Feel free to ping us once you're ready with the CLA!

PR #301, opened 2017-11-08
add ksm version command line argument

zouyee:
add ksm buildinfo api and scrapy tagert info which log kubernetes cluster version
/cc @andyxning  @brancz 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/301)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,18 +1,20 @@
 FLAGS =
 BUILDENVVAR = CGO_ENABLED=0
-TESTENVVAR = 
+TESTENVVAR =
 REGISTRY = quay.io/coreos
 TAG = $(shell git describe --abbrev=0)
 PKGS = $(shell go list ./... | grep -v /vendor/)
 ARCH ?= $(shell go env GOARCH)
+BuildDate=$(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
+Commit = $(shell git rev-parse --short HEAD)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
-
+PKG=k8s.io/kube-state-metrics
 
 IMAGE = $(REGISTRY)/kube-state-metrics
 MULTI_ARCH_IMG = $(IMAGE)-$(ARCH)
 
 gofmtcheck:
-	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi     
+	@go fmt $(PKGS) | grep ".*\.go"; if [ "$$?" = "0" ]; then exit 1; fi
 
 doccheck:
 	@echo "- Checking if documentation is up to date..."
@@ -26,7 +28,8 @@ doccheck:
 	@echo OK
 
 build: clean
-	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -o kube-state-metrics
+	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -ldflags "-s -w -X ${PKG}/version.Release=${TAG} -X ${PKG}/version.Commit=${Commit} -X ${PKG}/version.BuildDate=${BuildDate}" \
+	-o kube-state-metrics
 
 test-unit: clean build
 	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(TESTENVVAR) go test --race $(FLAGS) $(PKGS)
--- a/main.go
+++ b/main.go
@@ -21,6 +21,7 @@ import (
 	"fmt"
 	"log"
 	"net/http"
+	"net/http/pprof"
 	"os"
 	"sort"
 	"strings"
@@ -35,9 +36,8 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
-	"net/http/pprof"
-
 	"k8s.io/kube-state-metrics/collectors"
+	"k8s.io/kube-state-metrics/version"
 )
 
 const (
@@ -145,6 +145,7 @@ func main() {
 	flags.IntVar(&options.port, "port", 80, `Port to expose metrics on.`)
 	flags.Var(&options.collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &defaultCollectors))
 	flags.StringVar(&options.namespace, "namespace", metav1.NamespaceAll, "namespace to be enabled for collecting resources")
+	versionFlag := flags.BoolP("version", "", false, "kube-state-metrics version information")
 
 	flags.Usage = func() {
 		fmt.Fprintf(os.Stderr, "Usage of %s:\n", os.Args[0])
@@ -154,6 +155,11 @@ func main() {
 	err := flags.Parse(os.Args)
 	if err != nil {
 		glog.Fatalf("Error: %s", err)
+	}
+
+	if *versionFlag {
+		fmt.Printf("%#v", version.GetVersion())
+		os.Exit(0)
 	}
 
 	if options.help {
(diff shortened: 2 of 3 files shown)


version/version.go
```diff
@@ -0,0 +1,36 @@
...
+	// REPO returns the git repository URL
+	REPO = "UNKNOWN"
+	// COMMIT returns the short sha from git
+	COMMIT = "UNKNOWN"
+)
+
+// String returns information about the release.
+func String() string {
+	return fmt.Sprintf(`-------------------------------------------------------------------------------
+Kube-state-metrics
```
brancz: lower case "k"

main.go
```diff
@@ -275,6 +279,12 @@ func metricsServer(registry prometheus.Gatherer, port int) {
 		w.WriteHeader(200)
 		w.Write([]byte("ok"))
 	})
+	// Add buildInfo
+	mux.HandleFunc("/build", func(w http.ResponseWriter, r *http.Request) {
```
brancz: This is generally an antipattern, we don't want to reveal the version that is running to any user. I think it's good enough to log the version on start up.

version/version.go
```diff
@@ -0,0 +1,36 @@
...
+	RELEASE = "UNKNOWN"
+	// REPO returns the git repository URL
+	REPO = "UNKNOWN"
+	// COMMIT returns the short sha from git
+	COMMIT = "UNKNOWN"
+)
+
+// String returns information about the release.
+func String() string {
+	return fmt.Sprintf(`-------------------------------------------------------------------------------
```
andyxning: Actually i do not like printing version info like this in un structure format which is very difficult to parse for users if needed.

Another good example to do this job is what kubernetes does for kubelet which is printing version info in json format.

brancz: I'd prefer a structured format as well

Makefile
```diff
@@ -26,7 +31,8 @@ doccheck:
 	@echo OK
 
 build: clean
-	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -o kube-state-metrics
+	GOOS=$(shell uname -s | tr A-Z a-z) GOARCH=$(ARCH) $(BUILDENVVAR) go build -ldflags "-s -w -X ${PKG}/version.RELEASE=${TAG} -X ${PKG}/version.COMMIT=${COMMIT} -X ${PKG}/version.REPO=${REPO_INFO}" \
```
andyxning: I prefer we should only log the kube-state-metrics version and git commit id for version info. Nothing we need else.

main.go
andyxning: Agreed with @brancz . We do not need this http endpoint for querying version info.

version/version.go
andyxning: A simple but useful way to achieve this goal is to define a struct and logout the json output of the version instance.

andyxning:
@zouyee Thanks for the contribution. @brancz and i have left some comments. Those comments need to be addressed before we merge it.

Makefile
```diff
@@ -1,18 +1,22 @@
...
 ARCH ?= $(shell go env GOARCH)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
-
+PKG=k8s.io/kube-state-metrics
 
 IMAGE = $(REGISTRY)/kube-state-metrics
 MULTI_ARCH_IMG = $(IMAGE)-$(ARCH)
 
+ifndef COMMIT
+  COMMIT := git-$(shell git rev-parse --short HEAD)
```
andyxning: I think the `git-` prefix is useless. We only need the commit id.

version/version.go
```diff
@@ -0,0 +1,23 @@
+/*
+Copyright 2015 The Kubernetes Authors.
```
andyxning: `s/2015/2017/g`

version/version.go
```diff
@@ -0,0 +1,23 @@
...
+limitations under the License.
+*/
+
+package version
+
+var (
+	// RELEASE returns the release version
+	RELEASE = "UNKNOWN"
+	// REPO returns the git repository URL
+	REPO = "UNKNOWN"
```
andyxning: `REPO` is useless.

main.go
```diff
@@ -245,10 +245,12 @@ func createKubeClient(inCluster bool, apiserver string, kubeconfig string) (kube
 	// can't reach the server, making debugging hard. This makes it easier to
 	// figure out if apiserver is configured incorrectly.
 	glog.Infof("Testing communication with server")
-	_, err = kubeClient.Discovery().ServerVersion()
+	v, err := kubeClient.Discovery().ServerVersion()
 	if err != nil {
 		return nil, fmt.Errorf("ERROR communicating with apiserver: %v", err)
 	}
+	glog.Infof("Running in Kubernetes Cluster version v%v.%v (%v) - git (%v) commit %v - platform %v",
```
andyxning: `s/Running in/Running with/g`

andyxning: `Running in Kubernetes Cluster version v%v.%v (%v) - git (%v) commit %v - platform %v` -> `Running with Kubernetes cluster version: v%s.%s. git version: %s. git tree state: %s. commit: %s. platform: %s`

andyxning: It would be good to add the version info about the kubernetes which kube-state-metrics running against.

andyxning:
@zouyee We need a `--version` command line argument for kube-state-metrics to print the version info about kube-state-metrics.

version/version.go
```diff
@@ -0,0 +1,51 @@
...
+type Version struct {
+	GitCommit string
+	BuildDate string
+	RELEASE   string
+	GoVersion string
+	Compiler  string
+	Platform  string
+}
+
+// GetVersion returns representing the version
```
andyxning: `GetVersion returns representing the version` -> `GetVersion returns kube-state-metrics version`

version/version.go
```diff
@@ -0,0 +1,51 @@
...
+
+package version
+
+import (
+	"fmt"
+	"runtime"
+)
+
+var (
+	// RELEASE returns the release version
```
andyxning: `s/RELEASE/Release/g`

version/version.go
```diff
@@ -0,0 +1,51 @@
...
+	RELEASE = "UNKNOWN"
```
andyxning: ditto

version/version.go
```diff
@@ -0,0 +1,51 @@
...
+	// COMMIT returns the short sha from git
```
andyxning: `s/COMMIT/Commit/g`

version/version.go
```diff
@@ -0,0 +1,51 @@
...
+	COMMIT = "UNKNOWN"
```
andyxning: ditto

main.go
```diff
@@ -145,6 +145,7 @@ func main() {
 	flags.IntVar(&options.port, "port", 80, `Port to expose metrics on.`)
 	flags.Var(&options.collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &defaultCollectors))
 	flags.StringVar(&options.namespace, "namespace", metav1.NamespaceAll, "namespace to be enabled for collecting resources")
+	versionFlag := flags.BoolP("version", "", false, "kube-state-metrics build information")
```
andyxning: `kube-state-metrics build information` -> `kube-state-metrics version information`

main.go
```diff
@@ -156,6 +157,11 @@ func main() {
 		glog.Fatalf("Error: %s", err)
 	}
 
+	if *versionFlag {
+		fmt.Printf("kube-state-metrics version: %#v", version.GetVersion())
```
andyxning: `kube-state-metrics version: %#v` -> `%#v`

I prefer we do not add any prefix and keep it simple.

andyxning:
/lgtm

Will merge this after CI is green.

PR #302, opened 2017-11-08
Allow using RBAC to protect metrics endpoint

### LINKED ISSUE #297: Allow using RBAC to protect metrics endpoint(s)
In order to prevent any Pod in a Kubernetes cluster to request the metrics endpoint of kube-state-metrics, it should allow enabling RBAC validation with a running apiserver. Since kube-state-metrics heavily relies on a running apiserver anyways, this is no additional dependency.

@andyxning

brancz:
Closes #297

@andyxning @smarterclayton

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/302)


andyxning:
Sorry for the late. Will review it recently. :)

main.go
```diff
@@ -268,8 +299,17 @@ func metricsServer(registry prometheus.Gatherer, port int) {
 	mux.Handle("/debug/pprof/symbol", http.HandlerFunc(pprof.Symbol))
 	mux.Handle("/debug/pprof/trace", http.HandlerFunc(pprof.Trace))
 
+	promhandler := promhttp.HandlerFor(registry, promhttp.HandlerOpts{})
 	// Add metricsPath
-	mux.Handle(metricsPath, promhttp.HandlerFor(registry, promhttp.HandlerOpts{}))
+	mux.Handle(metricsPath, http.HandlerFunc(func(w http.ResponseWriter, req *http.Request) {
+		ok := AuthRequest(auth, w, req)
+		if !ok {
```
andyxning: Maybe a 401 response status code and some description in the response body is more user friendly and directly.

andyxning: By just `return`, golang http server will return response status code with 200 and nothing response body which is quite confused.

scripts/e2e.sh
```diff
@@ -133,6 +133,12 @@ echo "check metrics format with promtool"
 wget -q -O /tmp/prometheus.tar.gz https://github.com/prometheus/prometheus/releases/download/v$PROMETHEUS_VERSION/prometheus-$PROMETHEUS_VERSION.linux-amd64.tar.gz
 tar zxfv /tmp/prometheus.tar.gz -C /tmp
 cat $KUBE_STATE_METRICS_LOG_DIR/metrics | /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool check-metrics
+RESULT=$?
+if [ $RESULT -eq 1 ]; then
+    echo
```
andyxning: `echo "metrics data format is error"` should be more user friendly when this actually happens and this will help user to

andyxning: I think we need to run `set +e` before `cat $KUBE_STATE_METRICS_LOG_DIR/metrics | /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool check-metrics`. If some command line returns non-zero exit code, the `e2e.sh` will exit immediately.

andyxning:
@brancz Have reviewed some of PR. Will complete review it tomorrow.

IMO, i thing adding authn and authz to kube-state-metrics seems to make kube-state-metrics complicated.

main.go
brancz: Added.

brancz:
> IMO, i thing adding authn and authz to kube-state-metrics seems to make kube-state-metrics complicated.

I agree but it's necessary, as otherwise all these metrics are readable by any pod in your network that has network access. That is not acceptable.

main.go
andyxning: lgtm

kubernetes/kube-state-metrics-cluster-role.yaml
```diff
@@ -14,6 +14,10 @@ rules:
   - persistentvolumeclaims
   - namespaces
   verbs: ["list", "watch"]
+- apiGroups: ["autoscaling"]
```
andyxning: lgtm

scripts/e2e.sh
```diff
@@ -146,7 +146,7 @@ tar zxfv /tmp/prometheus.tar.gz -C /tmp
 cat $KUBE_STATE_METRICS_LOG_DIR/metrics | /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool check-metrics
```
andyxning: add before this line
```shell
set +o pipefail
set +x
```

scripts/e2e.sh
```diff
@@ -146,7 +146,7 @@ tar zxfv /tmp/prometheus.tar.gz -C /tmp
...
 RESULT=$?
 if [ $RESULT -eq 1 ]; then
-    echo
+    echo "metrics data is invalid after checking metrics with promtool\n\n"
     cat $KUBE_STATE_METRICS_LOG_DIR/metrics
     exit 1
 fi
```
andyxning: add after this line
```shell
set -o pipefail
set -x
```

scripts/minikube-rbac.yaml
```diff
@@ -0,0 +1,78 @@
+# Wide open access to the cluster (mostly for kubelet)
```
andyxning: To be honest, i am not familiar with rbac. But i think @smarterclayton could take a look at this.

main.go
```diff
@@ -146,6 +166,12 @@ func main() {
 	flags.Var(&options.collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &defaultCollectors))
 	flags.StringVar(&options.namespace, "namespace", metav1.NamespaceAll, "namespace to be enabled for collecting resources")
 
+	// Auth flags.
+	flags.BoolVar(&options.auth.Authentication.Anonymous.Enabled, "anonymous-auth", true, "Enables anonymous requests to the Kubelet server. Requests that are not rejected by another authentication method are treated as anonymous requests. Anonymous requests have a username of system:anonymous, and a group name of system:unauthenticated.")
```
andyxning: Maybe `Kubelet server` should be changed to `kube-state-metrics`.

andyxning:
@brancz Just one comment needs to be addressed from me. TBH, i am not quite sure about the RBAC auth. Maybe we need to find some other kubernetes members to help review RBAC part.

andyxning:
@caesarxuchao @deads2k @liggitt @mtaufen Could you please help us to review RBAC related commits.

scripts/minikube-rbac.yaml
brancz: the `scripts/minikube-rbac.yaml` is just some rbac roles to give to kube-dns, etc. as minikube doesn't provision them (RE: https://github.com/kubernetes/minikube/issues/1734#issuecomment-323443032), probably a good idea to put the link into the file

main.go
brancz: yes absolutely :slightly_smiling_face:

liggitt:
@smarterclayton, is there a reason this wouldn't be better served as an authz proxy that used SAR requests?

scripts/minikube-rbac.yaml
andyxning: Agree. Add the link or the issue to the comment is quite useful. I am not getting the meaning of `sd` when first review this. :)

brancz:
@liggitt does such a proxy implementation already exist?

liggitt:
I was thinking something along the lines of https://github.com/openshift/oauth-proxy/#delegate-authentication-and-authorization-to-openshift-for-infrastructure

I'm not sure if @smarterclayton made that able to run against the authentication.k8s.io/authorization.k8s.io APIs as well, but there's no reason it couldn't

brancz:
Got it. I've been thinking about something like this as well actually.

simonpasquier:
@liggitt FWIW I've got an working [template](https://github.com/simonpasquier/openshift_templates/tree/master/kube-state-metrics) where I deploy kube-state-metrics behind OAuth proxy. One thing that would need fixing though is to make kube-state-metrics listening on localhost only. Right now it binds always on all interfaces, only the port number can be changed...

brancz:
I’ve been working on pulling this functionality into a standalone proxy that we can drop in as a sidecar. I should be able to release this within the upcoming week. Everything is working, it just needs a bit more documentation

brancz:
So I finally released a first version of what I mentioned above, what I call "kube-rbac-proxy". Check it out :slightly_smiling_face: ! https://github.com/brancz/kube-rbac-proxy

(using that as a sidecar would solve this PR and we can close it without merging :slightly_smiling_face: )

brancz:
With [brancz/kube-rbac-proxy](https://github.com/brancz/kube-rbac-proxy) and #318 merged, this is now possible, without this PR being necessary.

PR #219, opened 2017-08-14
add simple componentstatus metrics

zouyee:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/219)

PR #317, opened 2017-12-15
Revert "add componentstatus metrics"

andyxning:
This reverts commit 5452f063935185d32c2fffcdbcf1f3decf1db3da.

As for now `componentstatuses` does not support `watch` method, we can not watch on it.  This should fix [219 comment](https://github.com/kubernetes/kube-state-metrics/pull/219#issuecomment-351824814).

ping @zouyee 
/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/317)


zouyee:
https://github.com/kubernetes/client-go/blob/v5.0.0/informers/core/v1/componentstatus.go#L49

main.go
```diff
@@ -66,7 +65,6 @@ var (
 		"horizontalpodautoscalers": struct{}{},
 	}
 	availableCollectors = map[string]func(registry prometheus.Registerer, kubeClient clientset.Interface, namespace string){
-		"componentstatuses":        collectors.RegisterComponentStatusCollector,
```
zouyee: `componentstatuses` options need to be removed from `--collectors`

andyxning: @zouyee The default value for `--collectors` is of `defaultCollectors` keys which has `componentstatuses` removed.

brancz:
I don't think we have to revert, let's just fix the problem I described in the comment on the original PR.

brancz:
You were right, sorry, I didn't quite understand at first.

PR #318, opened 2017-12-17
Allow setting host/ip to bind HTTP server to

brancz:
If one wants to use a proxy to be the only path to the metrics endpoint that kube-state-metrics exposes, we need to be able to allow binding to a specific interface rather than all (which is what we are currently doing). When the `--host` flag is not set, the previous behavior is continued, meaning this is not a breaking change.

@andyxning

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/318)

### DIFF
--- a/main.go
+++ b/main.go
@@ -20,10 +20,12 @@ import (
 	"flag"
 	"fmt"
 	"log"
+	"net"
 	"net/http"
 	"net/http/pprof"
 	"os"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/golang/glog"
@@ -130,6 +132,7 @@ type options struct {
 	kubeconfig string
 	help       bool
 	port       int
+	host       string
 	collectors collectorSet
 	namespace  string
 	version    bool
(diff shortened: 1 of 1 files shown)


main.go
```diff
@@ -267,9 +269,9 @@ func createKubeClient(inCluster bool, apiserver string, kubeconfig string) (kube
 	return kubeClient, nil
 }
 
-func metricsServer(registry prometheus.Gatherer, port int) {
+func metricsServer(registry prometheus.Gatherer, host string, port int) {
 	// Address to listen on for web interface and telemetry
-	listenAddress := fmt.Sprintf(":%d", port)
+	listenAddress := fmt.Sprintf("%s:%d", host, port)
```
andyxning: One nit:

`net.JoinHostPort(host, strconv.Itoa(port))` is more preferred, imo.

andyxning:
Overall lgtm. Just one nit. @brancz :)

brancz:
@andyxning PTAL

PR #327, opened 2017-12-24
add rbac config for endpoint

andyxning:
Add rbac config for endpoints.

@jordigilh 
/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/327)

### DIFF
--- a/kubernetes/kube-state-metrics-cluster-role.yaml
+++ b/kubernetes/kube-state-metrics-cluster-role.yaml
@@ -13,6 +13,7 @@ rules:
   - limitranges
   - persistentvolumeclaims
   - namespaces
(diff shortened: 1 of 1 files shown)


jordigilh:
LGTM. Thanks for taking care of this part :)
/Jordi

brancz:
This has come up a couple times now, how can we test for this in CI, so we don't forget about it?

andyxning:
> This has come up a couple times now, how can we test for this in CI, so we don't forget about it?

We need to enhance the e2e tests to cover this. This may happen in next release v1.3.0. I will take care of this.

PR #328, opened 2017-12-24
bump client-go to v6.0.0

andyxning:
This PR will bump client-go dep to v6.0.0 to support Kubernetes 1.9.0.

Note: 
This PR removes `CreatedByAnnotation` annotation by using `GetControllerOf` from [`metav1`](https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/controller_ref.go#L32-L40) as `CreatedByAnnotation` has been deleted in Kubernetes 1.9.

/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/328)


brancz:
I’m wondering whether we should do a release with client go 5.0 before merging this. Did any of the API versions we use change?

Also did you make sure not to override the “manual” changes to the Prometheus lib?

andyxning:
> I’m wondering whether we should do a release with client go 5.0 before merging this. Did any of the API versions we use change?

So for i have only see that `CreatedByAnnotation` annotation has been removed.

>  did you make sure not to override the “manual” changes to the Prometheus lib?

Yep, i have remembered this since the first time i update `client-go`.

zouyee:
> I’m wondering whether we should do a release with client go 5.0 before merging this. Did any of the API versions we use change?
I agreed with @brancz  cut a new release before merging the PR.
And according to the 1.9 release ChangLog, old version API could be compatible which has been GA

brancz:
Actually it seems that for now the old APIs are still available in client-go, so I'm happy to update this.

@zouyee what do you mean by this?

> By the way, Whether we could migrate package management tool ?

brancz:
Anyhow, let's move forward with this.

andyxning:
> By the way, Whether we could migrate package management tool ?

@brancz I guess what @zouyee means is that we may need to replace `godep` with other golang dependency management tool, such as `govendor` or `glide`.

As for me, i suggest we suspend this transition for now. IIRC, there are also some issues and discussions about  replacing `godep` with other dependency management tools such as `glide`. But none of the repos under Kubernetes Org has do this transition(correct me if i am wrong :-) ). In order to keep the consistency and be more productive to kubernetes developers who has been familiar with `godep`, i suggest postpone this transition. 

@brancz @zouyee WDYT.

brancz:
I'm fine with sicking as is for now until the Kubernetes project converges to glide/dep. Aside from that the tinkering with the prometheus client as we do right now won't be possible then anymore, so we actually can't right now.

PR #330, opened 2017-12-27
add kube-state-metrics own metrics

### LINKED ISSUE #296: Expose metrics about kube-state-metrics itself
kube-state-metrics currently only exposes metrics about the Kubernetes objects, but kube-state-metrics itself should also be "monitorable". For this purpose I would suggest an additional metrics endpoint (maybe `/metrics/kube-state-metrics`, there we can expose metrics like the number of objects used to generate metrics and go runtime as well as process statistics.

@andyxning

andyxning:
close #296 

This PR:
* add `--ksm-metrics-port` command line argument to expose ksm own metrics
* add go and process metrics
* expose `ksm_resources_per_scrape ` and `ksm_scrape_error_total`
  * `scrape_error_total`: Number of resources returned per scrape
  * `ksm_scrape_error_total`: Total scrape errors encountered when scraping a resource

/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/330)

### DIFF
--- a/collectors/collectors.go
+++ b/collectors/collectors.go
@@ -18,8 +18,26 @@ package collectors
 
 import (
 	"time"
+
+	"github.com/prometheus/client_golang/prometheus"
 )
 
 var (
 	resyncPeriod = 5 * time.Minute
+
+	ScrapeErrorTotalMetric = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "ksm_scrape_error_total",
+			Help: "Total scrape errors encountered when scraping a resource",
+		},
+		[]string{"resource"},
+	)
+
+	ResourcesPerScrapeMetric = prometheus.NewSummaryVec(
+		prometheus.SummaryOpts{
+			Name: "ksm_resources_per_scrape",
+			Help: "Number of resources returned per scrape",
+		},
+		[]string{"resource"},
+	)
 )
--- a/main.go
+++ b/main.go
@@ -38,7 +38,7 @@ import (
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
-	"k8s.io/kube-state-metrics/collectors"
+	kcollectors "k8s.io/kube-state-metrics/collectors"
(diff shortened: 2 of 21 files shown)


collectors/collectors.go
```diff
@@ -18,8 +18,30 @@ package collectors
...
 )
 
 var (
 	resyncPeriod = 5 * time.Minute
+
+	kubeStateMetricsSubsystem = "ksm"
+
+	ScrapeErrorTotalMetric = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Subsystem: kubeStateMetricsSubsystem,
```
brancz: This is likely to disappear in the next version of the Prometheus client so I would prefer to just prepend the prefix everywhere.

main.go
```diff
@@ -153,6 +154,7 @@ func main() {
 	flags.StringVar(&options.kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
 	flags.BoolVarP(&options.help, "help", "h", false, "Print help text")
 	flags.IntVar(&options.port, "port", 80, `Port to expose metrics on.`)
+	flags.IntVar(&options.ksmMetricsPort, "ksm-metrics-port", 81, `Port to expose kube-state-metrics own metrics on.`)
```
brancz: Typically this is referred to as the "telemetry-port". And I would like to be able to specify the address/interface to bind to as well, as it may be different from `options.host`.

collectors/collectors.go
andyxning: Yep. Just leanred this change from https://github.com/prometheus/client_golang/issues/240, will do.

main.go
andyxning: Done.

andyxning: Added `--telemetry-host` and `--telemetry-host` for kube-state-metrics self metrics.

andyxning:
@brancz Comments addressed. PTAL.

brancz:
lgtm, thanks for adding this!

PR #333, opened 2018-01-05
kubernetes: Add correct componentstatuses to ClusterRole

brancz:
I believe this is all that needs to be fixed in regards to the errors previously printed.

@andyxning @zouyee

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/333)


brancz:
nvm I was too quick there

PR #335, opened 2018-01-05
Cut 1.2.0-rc.0

brancz:
Happy new year :tada: ! We have gathered *a lot* of features and due to the recent Kubernetes 1.9 release, we should also release to ensure compatibility. Assuming everything goes well, we will be aiming for Monday the 15th of January for the stable 1.2.0 release.

@andyxning @zouyee @fabxc 

@piosz @loburm when merged I'd be great if one of you can push the release image to k8s.gcr.io.

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/335)

PR #340, opened 2018-01-07
bump image version and apiVersion

zouyee:
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/340)

  /cc @brancz

### DIFF
--- a/README.md
+++ b/README.md
@@ -66,8 +66,8 @@ release.
 #### Container Image
 
 The latest container image can be found at:
-* `quay.io/coreos/kube-state-metrics:v1.1.0`
-* `k8s.gcr.io/kube-state-metrics:v1.1.0`
+* `quay.io/coreos/kube-state-metrics:v1.2.0-rc.0`
+* `k8s.gcr.io/kube-state-metrics:v1.2.0-rc.0`
 
 **Note**:
 The recommended docker registry for kube-state-metrics is `quay.io`. kube-state-metrics on
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -1,4 +1,5 @@
 apiVersion: apps/v1beta2
+# Kubernetes versions after 1.9.0 should use apps/v1
 kind: Deployment
 metadata:
   name: kube-state-metrics
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -21,7 +21,7 @@ KUBERNETES_VERSION=v1.8.0
 KUBE_STATE_METRICS_LOG_DIR=./log
 KUBE_STATE_METRICS_IMAGE_NAME='quay.io/coreos/kube-state-metrics'
 KUBE_STATE_METRICS_IMAGE_NAME_PATTERN='quay.io\/coreos\/kube-state-metrics'
-PROMETHEUS_VERSION=1.8.1
+PROMETHEUS_VERSION=2.0.0
 
 mkdir -p $KUBE_STATE_METRICS_LOG_DIR
 
@@ -40,7 +40,7 @@ mkdir $HOME/.kube || true
 touch $HOME/.kube/config
 
 export KUBECONFIG=$HOME/.kube/config
-sudo minikube start --vm-driver=none --kubernetes-version=$KUBERNETES_VERSION
+sudo minikube start --vm-driver=none --kubernetes-version=$KUBERNETES_VERSION --logtostderr
 
 minikube update-context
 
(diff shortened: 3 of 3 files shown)


README.md
```diff
@@ -66,7 +66,7 @@ release.
 #### Container Image
 
 The latest container image can be found at:
-* `quay.io/coreos/kube-state-metrics:v1.1.0`
+* `quay.io/coreos/kube-state-metrics:v1.2.0-rc.0`
 * `k8s.gcr.io/kube-state-metrics:v1.1.0`
```
andyxning: Actually this is just an example demonstrating that the image for kube-state-metrics is available both on `quay.io` and `k8s.gcr.io`. So, personally i think the tag version can be stale. However if we need to keep those tags in sync with the release, we should update both for `quay.io` and `k8s.gcr.io`. 

@zouyee

kubernetes/kube-state-metrics-deployment.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: apps/v1beta2
+apiVersion: apps/v1 # for Kubernetes versions before 1.9.0 use apps/v1beta2
```
andyxning: sgtm

andyxning:
> Invalid Kubernetes version.
The following Kubernetes versions are available when using the localkube bootstrapper: 
	- v1.8.0
	- v1.7.5
	- v1.7.4
	- v1.7.3
	- v1.7.2
	- v1.7.0
	- v1.7.0-rc.1
	- v1.7.0-alpha.2
	- v1.6.4
	- v1.6.3
	- v1.6.0
	- v1.6.0-rc.1
	- v1.6.0-beta.4
	- v1.6.0-beta.3
	- v1.6.0-beta.2
	- v1.6.0-alpha.1
	- v1.6.0-alpha.0
	- v1.5.3
	- v1.5.2
	- v1.5.1
	- v1.4.5
	- v1.4.3
	- v1.4.2
	- v1.4.1
	- v1.4.0
	- v1.3.7
	- v1.3.6
	- v1.3.5
	- v1.3.4
	- v1.3.3
	- v1.3.0
make: *** [e2e] Error 1

Seems minikube do not support Kubernetes v1.9.x. @zouyee For now, we can just leave the version to 1.8.0.

README.md
zouyee: For now, ksm image in k8s.gcr.io has not been updated.

andyxning: Yep. But it will be updated by @loburm later. I think we do not need another PR to just update this as @loburm has acknowledged the updating in https://github.com/kubernetes/kube-state-metrics/pull/335#issuecomment-355589236. :)

zouyee:
> Seems minikube do not support Kubernetes v1.9.x. @zouyee For now, we can just leave the version to 1.8.0.
We could fix it with  the flag `--bootstrapper=kubeadm`.

brancz:
Let's convert to the kubeadm bootstrapper in a separate PR and in this one just fix the versions in the readme.

I can see a lot of potential problems with the kubeadm bootstrapper coming up so I'd prefer it not to block the correct versions in the readme.

zouyee:
>I can see a lot of potential problems with the kubeadm bootstrapper coming up so I'd prefer it not to block the correct versions in the readme.

 I guess maybe  travis-ci  does not supported systemd, which I have passed in local cluster

scripts/e2e.sh
```diff
@@ -21,7 +21,7 @@ KUBERNETES_VERSION=v1.8.0
 KUBE_STATE_METRICS_LOG_DIR=./log
 KUBE_STATE_METRICS_IMAGE_NAME='quay.io/coreos/kube-state-metrics'
 KUBE_STATE_METRICS_IMAGE_NAME_PATTERN='quay.io\/coreos\/kube-state-metrics'
-PROMETHEUS_VERSION=1.8.1
+PROMETHEUS_VERSION=2.0.0
```
andyxning: With prometheus version bump to 2.0.0, the check metrics command for promtool should be `promtool check-metrics` instead of `promtool check metrics`.

kubernetes/kube-state-metrics-deployment.yaml
```diff
@@ -1,4 +1,5 @@
 apiVersion: apps/v1beta2
+# for Kubernetes versions after 1.9.0 could use apps/v1
```
andyxning: `for Kubernetes versions after 1.9.0 could use apps/v1` -> `Kubernetes versions after 1.9.0 should use apps/v1`

PR #341, opened 2018-01-08
add release prodceduer doc

andyxning:
Add release procedure doc in case we may forget to do some steps for releasing. :)

@brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/341)

### DIFF
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -1,0 +1,19 @@
+## Procedures
+* make a PR to update:
+  * kube-state-metrics image tag for both `quay.io` and `k8s.gcr.io`
+  * compatibility matrix
+  * changelog entry
+    * only include user relevant changes
+    * entries should follow in order below
+    ```
+    [CHANGE]
+    [FEATURE]
+    [ENHANCEMENT]
+    [BUGFIX]
+    ```
+    * all lines should be full sentences
+  * kube-state-metrics image tag used in Kubernetes deployment yaml config
+* cut the new release branch, i.e., `release-1.2`, or merge/cherry-pick changes onto the minor release branch you intend to tag the release on
+* cut the new release tag, i.e., `v1.2.0-rc.0`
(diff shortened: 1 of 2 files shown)


RELEASE.md
```diff
@@ -0,0 +1,10 @@
+## Procedures
+* make a PR to update:
+  * kube-state-metrics image tag for both `quay.io` and `k8s.gcr.io`
+  * compatibility matrix
+  * change log
```
brancz: change log -> "changelog entry" ; Also I like to write the changelog to only include user relevant changes. Also I like to keep the same order of changelog entries:

Entries in the CHANGELOG.md are meant to be in this order:

[CHANGE]
[FEATURE]
[ENHANCEMENT]
[BUGFIX]

Last but not least, all lines in the changelog should be full sentences.

RELEASE.md
```diff
@@ -0,0 +1,10 @@
...
+  * kube-state-metrics image tag used in Kubernetes deployment yaml config
+* cut the new release branch, i.e., `release-1.2`
```
brancz: or merge/cherry-pick changes onto the minor release branch you intend to tag the release on

RELEASE.md
andyxning: Done.

Btw, the types of change in [keepchangelog](http://keepachangelog.com/en/1.0.0/) are 
```
Added
Changed
Deprecated
Removed
Fixed
Security
```

In case we have sticked to use `[CHANGE]/[FEATURE]/[ENHANCEMENT]/[BUGFIX]`, i am ok we keep using those types.

RELEASE.md
andyxning: Will do.

RELEASE.md
brancz: change, feature, enhancement, bugfix is what we used so far, so let's stick with it.

brancz:
lgtm thanks for putting this together!

Issue #337, opened 2018-01-05
ksm v1.2.0-rc.0 on K8s 1.5.3 - Failed to list *v1beta1.CronJob


dannyk81:
Hey!

Was eager to try the new RC, however I'm getting the following errors when it tries to collect CronJobs:
```
E0105 16:13:29.620093       1 reflector.go:205] k8s.io/kube-state-metrics/collectors/cronjob.go:103: Failed to list *v1beta1.CronJob: the server could not find the requested resource
```

We are running Kubernetes 1.5.3, the following Api Versions are supported:
```
$ kubectl api-versions
apps/v1beta1
authentication.k8s.io/v1beta1
authorization.k8s.io/v1beta1
autoscaling/v1
batch/v1
batch/v2alpha1
certificates.k8s.io/v1alpha1
extensions/v1beta1
policy/v1beta1
rbac.authorization.k8s.io/v1alpha1
storage.k8s.io/v1beta1
v1
```

I believe the issue is that Batch API version in K8s 1.5.3 is still `batch/v2alpha1` and server doesn't support `batch/v1beta1`.

`batch/v1beta1` is available with Kubernetes 1.8 onwards (according to [this](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/))

WDYT?

dannyk81:
Ok, so I just saw https://github.com/kubernetes/kube-state-metrics/issues/295 🤦‍♂️ 

It seems that by switching the API version in 1.2.0 to `batch/v1beta1` it made it incompatible with K8s <1.8

Is there any way to support both? based on K8s api server version perhaps?

andyxning:
Actually we have talked about using different resource group version according to different Kubernetes versions in #256 and #220 . But we have not make an conclusion about how to do this. :(

Will think about this later and hope to make it before v1.2.0 GA release.

brancz:
> Will think about this later and hope to make it before v1.2.0 GA release.

I would prefer not to introduce any features between the v1.2.0 rc.0 and GA. We are not bound to the Kubernetes release cycle, so it's perfectly legit to do a v1.3.0 release once we figure out the discovery of availble APIs.

dannyk81:
Sounds resaonble, so anyone using K8s <1.8 should stick to v1.1.0 until this is resolved? unless the CronJob metrics are not relevant I guess.

andyxning:
@dannyk81 I am afraid that is true. We will try to fix it asap.

[...]

PR #343, opened 2018-01-12
[WIP] add support for resource group version

### LINKED ISSUE #357: Proposal for supporting different group versions for a single resource.
This Proposal is used to demonstrating the problem how to support different group version for a single resource.

docs: https://docs.google.com/document/d/1f7Q_FVODNNc5IytBYbLbLLIWSlefuk4d9YYbIsj99Ec/edit#

/cc @brancz 
/ping @zouyee  as you have proposed a similar PR for attempting to support this functionality.

We need to make a conclusion about this in this week and implement it later as this has been delayed for a long time and we definitely need this feature.

andyxning:
Close #357.

Add check for supported resource group version. This can help us:
* check whether the specified resource group version supports list and watch methods inside kube-state-metrics instead of relying on e2e tests. With e2e,  we can only test kube-state-metrics against one Kubernetes version. By enabling this function inside kube-state-metrics, we can fail fast when running with any Kubernetes with various versions and configurations. This can also help us from using resources which do not support both list and watch methods. Btw, this can help us avoid what we do for componentstatus in #219 . However, for now we have no 
* check whether the specified resource is supported by the server and fail fast. I prefer fail fast than log the resource on found on the server error. This makes kube-state-metrics's log more clean and make users to be informed earlier about the abnormal.
* Add supports for multi group version of a resource
  * add `--collectors-config` to specify the wanted group version of a collector in the format of `collector1=groupVersion,collector2=groupVersion`.
  * add `--list-registered-collectors` to list all supported collectors and the corresponding group versions
  * The logic for this functionality to work is like below:
    * first get all enabled collectors with the default group version
    * merge the enabled collectors  with the specified collectors config
    * check the finally enabled collectors against the server 
    * run the specified collectors with a group version 

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/343)

/cc @brancz


andyxning:
Before we make a conclusion about multi group version support for a resource, we can not make sure what the group version each resource will use before we register the collectors. So, we only check the resource with only one group version.

zouyee:
how to validate apps version?

andyxning:
> how to validate apps version?

For now, i have no idea about how to check and customize resource group version. So, i do not consider that in this PR.

andyxning:
Ping @brancz

andyxning:
@brancz @zouyee PTAL.

main.go
```diff
@@ -95,75 +158,128 @@ func (pl promLogger) Println(v ...interface{}) {
...
 	flags.StringVar(&options.apiserver, "apiserver", "", `The URL of the apiserver to use as a master`)
 	flags.StringVar(&options.kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
 	flags.BoolVarP(&options.help, "help", "h", false, "Print help text")
 	flags.IntVar(&options.port, "port", 80, `Port to expose metrics on.`)
 	flags.StringVar(&options.host, "host", "0.0.0.0", `Host to expose metrics on.`)
 	flags.IntVar(&options.telemetryPort, "telemetry-port", 81, `Port to expose kube-state-metrics self metrics on.`)
 	flags.StringVar(&options.telemetryHost, "telemetry-host", "0.0.0.0", `Host to expose kube-state-metrics self metrics on.`)
-	flags.Var(&options.collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &defaultCollectors))
+	flags.Var(&options.collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &supportedCollectors))
+	flags.Var(&options.collectorsConfig, "collectors-config", fmt.Sprintf("Comma-separated list of collectors group version to be used. Defaults to %q", &collectorsConfig))
```
brancz: I feel this gives the notion to the user, that multiple versions are actually supported, which we don't, so I would just limit this PR to automatically enabling the collectors that are available/discovered in the API.

main.go
```diff
@@ -215,6 +347,18 @@ func main() {
...
 	}
 
+	supportedResourceMetaInfo, err := getServerSupportedResourceMetaInfo(kubeClient)
+	if err != nil {
+		glog.Fatalf("Failed to list server supported resource meta info: %v", err)
+	}
+
+	glog.V(4).Info(supportedResourceMetaInfo)
+
+	err = validateCollectorsAgainstServer(supportedResourceMetaInfo, collectors)
```
brancz: while I agree this is a good behavior, it's not backward compatible, I think we need to hide this behind a flag

andyxning: > while I agree this is a good behavior, it's not backward compatible, 

IMHO, this should be done earlier and it is a prerequisite condition before kube-state-metrics start against any Kubernetes server. Fatal things should fail fast instead of just logging the error like what we do for now. 

There are two conditions that the check may not pass:
* the collector is not available
* the currently used collector group version is not available

For the first one, we already have `--collectors` command line argument to specify a collectors whitelist. For the second one, in this PR we have the ability to customize the collector group version for now. So, if the default collector group version does not supported against the server, we can use `--collectors-config` to customize the group version for the collector.

So, i think this may not be backward incompatible. :) 

@brancz WDYT.

main.go
andyxning: > so I would just limit this PR to automatically enabling the collectors that are available/discovered in the API

IIUC, you mean that for now, there is only one group version support for each collectors and we need to add all available/discovered collectors in the API to kube-state-metrics.

Yes. For now, there is only one group version for each collector. I prepare to add a second group version to cronjob as it has been discussed for a long time in https://github.com/kubernetes/kube-state-metrics/issues/337 and other issues. And after that we suggest we cut a release to let end users to give this feature a try and we can collect feedback earlier.

As for other additional group version support for different collectors, they should be added on-demand when only end users request for them.

collectors/collectors.go
```diff
@@ -41,3 +46,44 @@ var (
...
+}
+
+func BoolFloat64(b bool) float64 {
+	if b {
+		return 1
+	}
+	return 0
+}
+
+// AddConditionMetrics generates one metric for each possible node condition
```
zouyee: s/node condtion/resource condition/

collectors/collectors.go
```diff
@@ -19,11 +19,16 @@ package collectors
...
+
 	"github.com/prometheus/client_golang/prometheus"
+	"k8s.io/api/core/v1"
 )
 
 var (
-	resyncPeriod = 5 * time.Minute
+	invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
+
+	ResyncPeriod = 5 * time.Minute
```
zouyee: Could we make `ResyncPeriod ` to configurable?

collectors/test/test.go
```diff
@@ -0,0 +1,183 @@
...
+	}
+	sort.Strings(names)
+	result := make([]*dto.MetricFamily, 0, len(names))
+	for _, name := range names {
+		result = append(result, metricFamiliesByName[name])
+	}
+	return result
+}
+
+// The below sorting code is copied form the Prometheus client library modulo the added
```
zouyee: modulo?

main.go
```diff
@@ -95,75 +158,128 @@ func (pl promLogger) Println(v ...interface{}) {
...
 	flags.AddGoFlagSet(flag.CommandLine)
 	flags.Lookup("logtostderr").Value.Set("true")
 	flags.Lookup("logtostderr").DefValue = "true"
 	flags.Lookup("logtostderr").NoOptDefVal = "true"
 	flags.BoolVar(&options.inCluster, "in-cluster", true, `If true, use the built in kubernetes cluster for creating the client`)
+	flags.BoolVar(&options.listRegisteredCollectors, "list-registered-collectors", false, `If true, list registered collectors with corresponding supported group versions`)
 	flags.StringVar(&options.apiserver, "apiserver", "", `The URL of the apiserver to use as a master`)
 	flags.StringVar(&options.kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
 	flags.BoolVarP(&options.help, "help", "h", false, "Print help text")
 	flags.IntVar(&options.port, "port", 80, `Port to expose metrics on.`)
```
zouyee: 80 -> 8080？

zouyee: 80 -> 8080?

main.go
brancz: hmm ... shouldn't we just disable collectors that are not available and log that?

main.go
brancz: Yeah, I get that we could add new collectors for different versions, but that looks like a whole lot of code duplication right now. I'm not sure it's that big of a problem, as starting with 1.9 there is no collector that is still in alpha, feels like we're over-complicating it.

collectors/collectors.go
andyxning: If that is necessary, we can do this in another PR instead of this one. :)

collectors/test/test.go
andyxning: Will do.

collectors/collectors.go
andyxning: Will do.

main.go
andyxning: Agreed.

main.go
andyxning: Seems fine to me. There are two ways to inform end users about this error. Log and ignore or fail fast. I am in favor of fail fast. But log and ignore is also good to me. :)

Done. PTAL.

main.go
andyxning: If we choose to use this implementation, it will make code duplication indeed. Maybe we can do some abstract and refactor to extract common code and support for custom with arguments.

PR #346, opened 2018-01-15
Update README.md

zouyee:


README.md
```diff
@@ -90,11 +90,11 @@ additional metrics!
...
 >
 > **collectors**
-> * v2alpha1/cronjob
+> * v1beta1/cronjob
 >
-> If users want to enable this feature when kubernetes version larger than 1.7,
+> If users want to enable this feature when kubernetes version greater or equal to 1.8,
 > It must be configured, with the following parameter setting for apiserver.
-> `--runtime-config=batch/v2alpha1=true`
+> `--runtime-config=batch/v1beta1=true`
```
brancz: beta APIs are enabled by default, so I think we should remove this part all together

andyxning: This is actually related how we use a single resource against multiple group versions. Before we make any conclusion about this, i suggest we postpone updating this as what @brancz said for Kubernetes whose version is greater than v1.9 will have `batch/v1beta` enabled by default.

PR #371, opened 2018-02-17
Use Kubernete's own func that creates config by apiserver or kubeconfig

metalmatze:
Instead of implementing this logic ourselves we can simply leverage Kubernet's built-in function, which does exactly what we need.

We want to run kube-state-metrics outside of a cluster and simply give it an apiserver url. Right now this also needs a kubeconfig.

### DIFF
--- a/main.go
+++ b/main.go
@@ -35,7 +35,6 @@ import (
 	"github.com/spf13/pflag"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	clientset "k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
 
 	kcollectors "k8s.io/kube-state-metrics/collectors"
@@ -134,7 +133,6 @@ func (c *collectorSet) Type() string {
 }
 
 type options struct {
-	inCluster     bool
 	apiserver     string
 	kubeconfig    string
 	help          bool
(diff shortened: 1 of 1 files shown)


metalmatze:
Oh. I didn't see that I need to comment. :grin: 
CLA is signed.

brancz:
This seems fine to me, I slightly worry about backward compatibility if we remove the `--in-cluster` flag, but on the other hand I expect this to be a very small amount of users and it's a change that clearly breaks and not in a weird behavior at runtime. I think I'm ok with it, wdyt @andyxning ?

main.go
```diff
@@ -201,16 +198,13 @@ func main() {
 		glog.Infof("Using %s namespace", options.namespace)
 	}
 
-	if isNotExists(options.kubeconfig) && !(options.inCluster) {
-		glog.Fatalf("kubeconfig invalid and --in-cluster is false; kubeconfig must be set to a valid file(kubeconfig default file name: $HOME/.kube/config)")
-	}
-	if options.apiserver != "" {
-		glog.Infof("apiserver set to: %v", options.apiserver)
+	if options.apiserver == "" && options.kubeconfig == "" {
```
andyxning: These two lines are unnecessary as [`BuildConfigFromFlags`](https://github.com/kubernetes/client-go/blob/master/tools/clientcmd/client_config.go#L531) will check and log this.

andyxning:
@metalmatze I have added a comment. PTAL. 

BTW, as stated by @brancz , This is not a backward compatible change. Especially when `in-cluster` and `kubeconfig` or `apiserver` are both set or default kubeconfig file exists. Because after the change, kube-state-metrics will use kubeconfig config instead of `in-cluster` config for `BuildConfigFromFlags` will first use `kubeconfig` instead of `in-cluster`.

However, i also think that this should not be a big problem as we have already set `in-cluster` to `true` as the default behavior and do not specify `in-cluster` command line argument in kube-state-metrics deployment. Setting both `in-cluster` to `true` and `kubeconfig` or `apiserver` or there is an existing default kubeconfig file should be rarely. Overall this lgtm.

We should add a statement in the next release note to make this change widly known.

metalmatze:
I have remove the 3 unnecessary lines.

PR #375, opened 2018-02-27
fix port usage misunderstanding

### LINKED ISSUE #373: port confusion, document default port documentation in docker image
hi,
the default ports for the metrics in the binary is 80 and 81 while the readme says the default for the main metrics is 8080 in the first paragraph and the telemetry port default is not mentioned at all.
To make it worse the docker image sets one of them by entrypoint to 8080 but leaves the telemetry port at the privileged 81.

This is pretty confusing, the readme should not state that the default is 8080 when this it really is 80 in the binary, the documentation should probably mention that the docker image changes the default and the docker image should probably also change the default of the telemetry port so that it does not default to a privileged port (or not change the default at all).

andyxning:
Fix #373 

/cc @brancz 
ping @juliantaylor

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/375)

### DIFF
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -22,6 +22,8 @@ spec:
         ports:
         - name: http-metrics
           containerPort: 8080
+        - name: telemetry
+          containerPort: 8081
         readinessProbe:
           httpGet:
             path: /healthz
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,6 +3,7 @@ FROM scratch
 COPY kube-state-metrics /
 VOLUME /tmp
 
-ENTRYPOINT ["/kube-state-metrics", "--port=8080"]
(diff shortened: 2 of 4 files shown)


kubernetes/kube-state-metrics-deployment.yaml
```diff
@@ -22,6 +22,8 @@ spec:
         ports:
         - name: http-metrics
           containerPort: 8080
+        - name: telemetry-metrics
```
brancz: From the build failure:

> The Deployment "kube-state-metrics" is invalid: spec.template.spec.containers[0].ports[1].name: Invalid value: "telemetry-metrics": must be no more than 15 characters

I'd suggest `self-metrics`.

andyxning: I have changed the port name to `telemetry`. :)

brancz:
lgtm

juliantaylor:
thanks

PR #380, opened 2018-03-01
update go version to 1.10

andyxning:
This PR will update Golang version to 1.9 as 1.10 has been released. We should use the newer golang version. :)

/cc @brancz

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/380)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 language: go
 
 go:
-- 1.8
+- "1.10"
 
 install:
 - mkdir -p $HOME/gopath/src/k8s.io


brancz:
let's go with 1.10 I would say

andyxning:
@brancz Done. PTAL.

.travis.yml
```diff
@@ -3,7 +3,7 @@ sudo: required
 language: go
 
 go:
-- 1.8
+- 1.10
```
brancz: Because yaml, this gets parsed as a decimal and ends up being `1.1`, so we need to quote it.

andyxning: What a big hole it is! Having thought everything in yaml is a string first. After changing it, everything seems ok now.

PR #384, opened 2018-03-07
Fix: default ports

nlamirault:
The defaults arguments are not ``8080`` and ``8081`` (See [here](https://github.com/kubernetes/kube-state-metrics/blob/release-1.2/main.go#L155) and [here](https://github.com/kubernetes/kube-state-metrics/blob/release-1.2/main.go#L157))


andyxning:
@nlamirault The kube-state-metrics image sets `--port` and `--telemetry-port` to `8080` and `8081`. So the usage of `8080` and `8081` in pod spec is right. 

Close this as it is the desired settings. @nlamirault  Feel free to reopen it if you still have some questions.

nlamirault:
Sorry for noise.

PR #385, opened 2018-03-08
Add _labels for pv and pvc.

tcolgate:
This adds kube_persistentvolume_labels, and
kube_persistentvolumeclaim_lables. These are useful for associating the
pv and pvc with the app they were deployed for (via labels)

### DIFF
--- a/Documentation/persistentvolume-metrics.md
+++ b/Documentation/persistentvolume-metrics.md
@@ -3,5 +3,6 @@
 | Metric name| Metric type | Labels/tags |
 | ---------- | ----------- | ----------- |
 | kube_persistentvolume_status_phase | Gauge | `persistentvolume`=&lt;pv-name&gt; <br> `namespace`=&lt;pv-namespace&gt; <br>`phase`=&lt;Bound\|Failed\|Pending\|Available\|Released&gt;|
+| kube_persistentvolume_labels | Gauge | `persistentvolume`=&lt;persistentvolume-name&gt; <br> `namespace`=&lt;persistentvolume-namespace&gt; <br> `label_PERSISTENTVOLUME_LABEL`=&lt;PERSISTENTVOLUME_LABEL&gt;  |
 | kube_persistentvolume_info | Gauge | `persistentvolume`=&lt;pv-name&gt; <br> `namespace`=&lt;pv-namespace&gt;<br> `storageclass`=&lt;storageclass-name&gt; |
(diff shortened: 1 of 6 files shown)


andyxning:
@tcolgate 
> These are useful for associating the
pv and pvc with the app they were deployed for (via labels)

Any example about this?

tcolgate:
@andyxning as it's stands, there's no way (that I can work out), for graphing the volume useage of a pod. We'd need something like:

`kube_pod_spec_volumes{name="xxxx", persistentvolumeclaim="nnnn",....} = 1`

but a metrics on volumes is probably going to be messy (probably a good idea at some point though).
Exposing the labels for pvs and pvcs:
- makes it consistent with other collectors
- makes it possible to find all the PVCs for a given pod/deployment/....

e.g.
`kube_persistentvolumeclaim_labels{label_app="myapp"} * on (persistentvolumeclaim) kubelet_volume_stats_used_bytes`

andyxning:
@tcolgate 

I mean if you want to associate pv and pvc with pods by labels, does Kubernetes has some strict specification about that pv/pvc and pods will have some common labels when pods reference a pv/pvc or it is just be explicitly used in your use?

tcolgate:
@andyxning 
No, there is no strict requirement, but there is a convention to use an `app` label , so if I have deployed sentry, with a redis install, with a pv, then the objects created will likely have an `app=sentry-redis` label, ad the pvc template will add a similar label, so they can be casually associated by the convention.
Also, the labels are useful for other things (e.g. associating storage against the team that uses it). Is there a reason this wasn't included? it seems like an obvious, and useful addition to me.

I'm happy to look at pod_spec_volumes labels (which would let you actually tie a PV to a pod), but I think I'd still want this change anyway, and the two are orthogonal.

brancz:
I agree, I think this is a useful addition either way.

tcolgate:
@brancz I am curious as to why the pv metrics include a namespace. Forgive me if I'm wrong, but I don't think PVs are namespaced objects? I'm not sure if removing ti would be considered a breaking change, prometheus considers any label with a value of "" as being unset, so it never actually shows up. It's slightly confusing when reading the code though.

If you concur I can fix in another PR.

brancz:
Once again, you're absolutely right. I don't think it's a breaking change. Thanks for noticing! An additional PR would be highly appreciated.

PR #389, opened 2018-03-09
Add Pod persistentvolumeclaim binding information.

tcolgate:
This introduces two metrics that are useful when monitoring disk usage
of a pod. They allow you to determine the persistentvolumeclaims
associated with a specific pod, and how they are bound.

Without this pods can only be loosely associated with volumes through
labels.

I looked at providing metrics on other attached volume types, however
this is not useful at this time as the kubelet only exports statistics
for PVC mounted volumes. Most volumes can/should be mounted via a PVC
anyway so this is a relatively minor limitation.


tcolgate:
I think this rounds out the support for PVC monitoring. With this change included we can go from pod -> pvc -> kubelet pvc metrics.

Issue #388, opened 2018-03-09
Add taint support


derekwaynecarr:
A node whose volume is stuck attaching for more than X minutes is automatically tainted.

Ability to alert on this taint would be useful to couple automated remedy to reboot or repair the node or do other corrective action where appropriate.

Surfacing node status conditions helps for issues derived from node problem detector but more problems are likely to actually be surfaced via taints.

PR #391, opened 2018-03-10
Add metrics for node taints

derekwaynecarr:
Fixes https://github.com/kubernetes/kube-state-metrics/issues/388

### DIFF
--- a/collectors/node.go
+++ b/collectors/node.go
@@ -64,6 +64,12 @@ var (
 		[]string{"node"}, nil,
 	)
 
+	descNodeSpecTaint = prometheus.NewDesc(
+		"kube_node_spec_taint",
+		"The taint of a cluster node.",
+		[]string{"node", "key", "value", "effect"}, nil,
+	)
+
 	descNodeStatusCondition = prometheus.NewDesc(
 		"kube_node_status_condition",
 		"The condition of a cluster node.",
@@ -157,6 +163,7 @@ func (nc *nodeCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descNodeCreated
 	ch <- descNodeLabels
 	ch <- descNodeSpecUnschedulable
+	ch <- descNodeSpecTaint
 	ch <- descNodeStatusCondition
 	ch <- descNodeStatusPhase
 	ch <- descNodeStatusCapacityCPU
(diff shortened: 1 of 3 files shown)


andyxning:
@derekwaynecarr You need to add `kube_node_spec_taint` to the [node metrics documentation](https://github.com/kubernetes/kube-state-metrics/blob/master/Documentation/node-metrics.md).

derekwaynecarr:
@andyxning thanks for pointer, updated the doc.

collectors/node_test.go
```diff
@@ -263,6 +265,29 @@ func TestNodeCollector(t *testing.T) {
...
+						Taints: []v1.Taint{
+							{Key: "node.kubernetes.io/memory-pressure", Value: "true", Effect: v1.TaintEffectPreferNoSchedule},
+							{Key: "Accelerated", Value: "gpu", Effect: v1.TaintEffectPreferNoSchedule},
+							{Key: "Dedicated", Effect: v1.TaintEffectPreferNoSchedule},
+						},
+					},
+				},
+			},
+			want: metadata + `
+				kube_node_spec_taint{effect="PreferNoSchedule",key="Dedicated",node="127.0.0.1",value=""} 1
```
andyxning: @brancz Does Prometheus allows tag value to be an empty string?

derekwaynecarr: i thought this would work similar to labels whose value also could be the empty string.

https://github.com/kubernetes/kubernetes/blob/8d7d7a5e0d4d7e75f5a860574346944b8cc0fc43/staging/src/k8s.io/apimachinery/pkg/util/validation/validation.go#L88

it would be great to confirm as the taint value associated with the taint key is optional.

brancz: Labels with empty values are thrown away at ingestion time, so yes it's perfectly fine to have that.

collectors/node.go
```diff
@@ -219,6 +226,14 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n v1.Node) {
 
 	addGauge(descNodeSpecUnschedulable, boolFloat64(n.Spec.Unschedulable))
 
+	// Collect node taints
+	for _, taint := range n.Spec.Taints {
+		// Taints are applied to repel pods from nodes that do not have a corresponding
+		// toleration.  Many node conditions are optionally reflected as taints
+		// by the node controller in order to simplify scheduling constraints.
+		addGauge(descNodeSpecTaint, 1, taint.Key, taint.Value, string(taint.Effect))
```
brancz: I'm a bit worried about the cardinality here. Do you have some numbers of combinations that are typical in clusters for each node?

brancz:
Other than my worry on the cardinality, this lgtm

collectors/node.go
ironcladlou: Good point; given normal users likely don't have authorization to mutate taints on a node, it does seem unlikely that these would grow unbounded in practice, but as a defensive measure would it make any sense to introduce a hard coded upper bound to the amount of taints which we'll process (and produce an event or log if that limit is reached)?

brancz:
Considering few entities ever have the ability to edit these, I think it's reasonable to do "unbound".

collectors/node.go
derekwaynecarr: i anticipate an upper bound of num(taints) < 10 for the most tainted of nodes.

~5 node conditions -> mapped to node taints if you enable a current alpha flag which syncs them
~1 node taint identified now for volumes stuck
~1 node taint to steer workloads with matching toleration (i.e. accelerated, storage-optimized, highmem-lowcpu, etc.)

cluster-operators or sufficiently privileged actors will apply taints to nodes.

Issue #368, opened 2018-02-15
Add optional GET arguments to /metrics


dohnto:
# Current situation

Our company have on-premise kubernetes cluster for several teams. One team use several namespaces. We want to run single kube-state-metrics as a deployment inside kubernetes and we want to allow teams to run their own prometheus and scrape kube-state-metrics deployment.

However, every team has to have their special relabel configuration and drop other namespaces otherwise dataseries count would rise significantly. Also most of the time series are transferred through networked and they are just dropped by prometheus.

# Proposed solution

Would it be possible to add optional GET arguments to metricsPath? I could imagine something like:
```
/metrics?namespace=team1-staging&namespace=team1-production
/metrics?collectors=deployments&namespace=team1-staging
```

Although I am not sure whether this is compatible with gRPC. I am willing to do the implementation but first I would like to discuss it with maintainers.

Thank you for any ideas or tips how to solve our issue.

brancz:
Generally I like all that you're proposing and I think it would make sense to add these filters by parameters. However you are inevitably going to run into a scalability problem, where the sheer amount of objects is too much for a single kube-state-metrics deployment, at which point you have to deploy a kube-state-metrics instance per namespace and or collector anyways, which is already possible today. The latter also makes a lot of sense from a security/access control view, as kube-state-metrics then really only has access to what it must view, instead of what it could view.

dohnto:
@brancz thank you for quick response.

Ok I understand. It is probable, that we will end up running several kube-state-metrics instances (per namespace). But it would be awesome, if we could run kube-state-metrics per team (but one team owns few namespaces).

I appreciate that you implemented `--namespace` argument but it would be awesome, if this argument could be passed multiple times. Would you accept such extension? Or another way how to express that we want are interested not in single but multiple namespaces.

Or is it possible to achieve this using Roles, Rolebindings and ServiceAccounts? To limit the scope of what kube-state-metrics is able to see?

dohnto:
For a consistency, wouldn't it be better to add extra arguments 
```--namespaces=test,stage``` (note the plural, same as `--collectors`) and after deprecate the ```--namespace```?

Or can anyone help me with the interface the upstream prefers?

brancz:
for compatibility reasons, I'd prefer not to change the name of the flag, but extend the usage to accept lists as you suggested

andyxning:
@brancz @dohnto 
[...]

PR #394, opened 2018-03-10
Fix namespacing of HPA

dohnto:
Fixes #393

PR #395, opened 2018-03-10
[RFC] Allow to specify multiple namespace.

dohnto:
**I am no golang professional so I welcome any feedback or improvement tips. Also I am not very familiar with kubernetes golang library so maybe I am trying to solve it too naively. Please tell me.**

### What problem is it trying to solve
See #368 for rationale.

### What is this feature?
Allow to pass value of `--namespace` as a list to allow to monitor multiple namespaces. E. g.
```
$ ./kube-state-metrics # all namespaces
$ ./kube-state-metrics --namespace=kube-system # single namespace
$ ./kube-state-metrics --namespace=kube-system,default,ingress-nginx # multiple namespaces
```

### Is it backwards compatible?
Yes

### DIFF
--- a/collectors/node.go
+++ b/collectors/node.go
@@ -22,9 +22,7 @@ import (
 	"golang.org/x/net/context"
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/client-go/kubernetes"
-	"k8s.io/client-go/tools/cache"
 )
 
 var (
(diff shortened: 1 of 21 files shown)


main.go
```diff
@@ -136,6 +137,26 @@ func (c *collectorSet) Type() string {
...
+}
+
+func (n *namespaceList) Set(value string) error {
+	splittedNamespaces := strings.Split(value, ",")
+	*n = append(*n, splittedNamespaces...)
+	return nil
+}
+
+func (n *namespaceList) Type() string {
+	return "string"
```
brancz: This would be "namespaceList"

dohnto: Fixed in 239e9f2. I will squash after review.

collectors/node.go
```diff
@@ -125,21 +123,23 @@ func (l NodeLister) List() (v1.NodeList, error) {
...
 }
 
-func RegisterNodeCollector(registry prometheus.Registerer, kubeClient kubernetes.Interface, namespace string) {
+func RegisterNodeCollector(registry prometheus.Registerer, kubeClient kubernetes.Interface, namespaces []string) {
 	client := kubeClient.CoreV1().RESTClient()
 	glog.Infof("collect node with %s", client.APIVersion())
-	nlw := cache.NewListWatchFromClient(client, "nodes", metav1.NamespaceAll, fields.Everything())
-	ninf := cache.NewSharedInformer(nlw, &v1.Node{}, resyncPeriod)
+
+	ninfs := NewSharedInformerList(client, "nodes", []string{metav1.NamespaceAll}, &v1.Node{})
```
dohnto: I was not sure whether passing `[]string{metav1.NamespaceAll}` as a dummy namespace list is suitable here... shouln't there be some variable with such constant (it is used in several places - persistentvolumes). Also, should we iterate over the namespaces later?

Maybe matter of taste but I would still prefer to discuss it.

brancz: I think this way is perfectly fine, we were using the same constant before after all.

brancz:
Could you just rebase this PR? The conflict is due to your previous PR (https://github.com/kubernetes/kube-state-metrics/pull/394), so just make sure that's still ends up like that :slightly_smiling_face: .

dohnto:
I have rebased.

PR #396, opened 2018-03-12
collectors: Report pod's completion time

aramalipoor:
This metric alongside `pod_start_time` can help determine total Pod's lifetime. Useful for example to get average duration it takes Openshift's BuildConfigs to build images.

The `pod_completion_time` calculation logic is based [this piece of code](https://github.com/openshift/origin-web-console/blob/57086c97efd8e5983d250d3aef110f61e0a5eab0/app/scripts/filters/resources.js#L1102-L1116) in [origin-web-console](https://github.com/openshift/origin-web-console)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,6 @@
+
+* [FEATURE] Add `kube_pod_completion_time` metric.
+
 ## v1.2.0 / 2018-01-15
 
 After a testing period of 10 days, there were no additional bugs found or features introduced.
--- a/collectors/pod_test.go
+++ b/collectors/pod_test.go
@@ -66,6 +66,8 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_info gauge
 		# HELP kube_pod_start_time Start time in unix timestamp for a pod.
 		# TYPE kube_pod_start_time gauge
+		# HELP kube_pod_completion_time Completion time in unix timestamp for a pod.
+		# TYPE kube_pod_completion_time gauge
 		# HELP kube_pod_owner Information about the Pod's owner.
 		# TYPE kube_pod_owner gauge
 		# HELP kube_pod_status_phase The pods current phase.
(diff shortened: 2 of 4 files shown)


collectors/pod_test.go
```diff
@@ -429,6 +431,26 @@ func TestPodCollector(t *testing.T) {
...
+											Time: time.Unix(1501769018, 0),
+										},
+									},
+								},
+							},
+							v1.ContainerStatus{
+								State: v1.ContainerState{
+									Terminated: &v1.ContainerStateTerminated{
+										FinishedAt: metav1.Time{
+											Time: time.Unix(1501779018, 0),
```
brancz: Just to make sure the if condition works as intended, could you make this timestamp smaller than the first one (or add a third status with the middle one being that "latest")?

CHANGELOG.md
```diff
@@ -1,3 +1,6 @@
+
```
andyxning: @brancz 

This indicates to me that maybe we should add an `[Unreleased]` section to the section and make every PR that needs to be mentioned in the change log also update changelog.md. This has an advantage that this will definitely decrease your burden to write a complete release note once every release. I suggest we give it a try before we automate the changelog gen just like what Kubernetes main repo does.

andyxning: @aramalipoor 

Could you please add a new `[Unreleased]` section before this line just like what [keepchangelog](https://keepachangelog.com/en/1.0.0/#effort) suggests.

brancz: @andyxning happy to start this practice.

brancz: My bad for merging this already, do you just want to open a PR to add the title?

andyxning: I will update this in another PR.

aramalipoor: https://github.com/kubernetes/kube-state-metrics/pull/397

PR #403, opened 2018-03-21
collectors: Add NOP eventhandler to ensure we perform resyncs

### LINKED ISSUE #402: two kube-state-metrics significantly deviate in reports over time
Hi all,

I have very strong suspicion that kube-state-metrics never does resync and hence a long-running instance will deviate from the reality quite dramatically. At least this is what we see in production when we run two instances of kube-state-metrics watching same k8s cluster. They will end up showing different data quite soon.

I'll explain the problem using pods collector as an example. Please take the explanation with a graint of salt as I'm not big expert in kubernetes ecosystem.

So, pods collector uses [shared informer](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go#L200). The share informed gets created [here](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/collectors.go#L55).
```
func NewSharedInformerList(client rest.Interface, resource string, namespaces []string, objType runtime.Object) *SharedInformerList {
    sinfs := SharedInformerList{}
    for _, namespace := range namespaces {
        slw := cache.NewListWatchFromClient(client, resource, namespace, fields.Everything())
        sinfs = append(sinfs, cache.NewSharedInformer(slw, objType, resyncPeriod))
    }
    return &sinfs
}
```

One of the parameters which NewSharedInformer() receives is resyncPeriod (it's set to 5 min by default). As far as I understand the kubernetes model, resync period is here to limit potential inconsistency between kubernetes and a client (kube-state-metrics in our case). In other words, kube-state-metrics can deviate from the true state of affairs, but it's guaranteed to catch up within resync period. I'm no a big guru in kubernetes so my wording might not be precise, please excuse me in such case.

The piece of client-go which is responsible for syncing aclient with kube-apiservers is [Reflector](https://godoc.org/k8s.io/client-go/tools/cache#NewReflector). The condition which actually triggers a resync is [here](https://github.com/kubernetes/client-go/blob/master/tools/cache/reflector.go#L285). I'll quote it here:
```
            if r.ShouldResync == nil || r.ShouldResync() {
                glog.V(4).Infof("%s: forcing resync", r.name)
                if err := r.store.Resync(); err != nil {
                    resyncerrc <- err
                    return
                }
            }
```

So, whether reflector performs a resync depends on ShouldResync callback being set and the result of calling it. ShouldResync is populated through a sequence of passes which I don't want to mention here. The call of ShouldResync() also goes through the chain of calls and it ends up being in [shouldResync()](https://github.com/kubernetes/client-go/blob/master/tools/cache/shared_informer.go#L436) of sharedProcessor. I'll also quite the function here:

```
func (p *sharedProcessor) shouldResync() bool {
    p.listenersLock.Lock()
    defer p.listenersLock.Unlock()

    p.syncingListeners = []*processorListener{}

    resyncNeeded := false
    now := p.clock.Now()
    for _, listener := range p.listeners {
        // need to loop through all the listeners to see if they need to resync so we can prepare any
        // listeners that are going to be resyncing.
        if listener.shouldResync(now) {
            resyncNeeded = true
            p.syncingListeners = append(p.syncingListeners, listener)
            listener.determineNextResync(now)
        }
    }
    return resyncNeeded
}
```

So, the function roughly implements following logic: go through every available processorListener and ask whether it needs a resync or not. If any of them needs resyncing then return true. If no one needs or there are no listeners then return false.

The "there are no listeners" is key here. The only place which I found where listeners are populated is [addListener()](https://github.com/kubernetes/client-go/blob/5f85fe426e7aa3c1df401a7ae6c1ba837bd76be9/tools/cache/shared_informer.go#L384) function. This function is only called twice inside [AddEventHandlerWithResyncPeriod()](https://github.com/kubernetes/client-go/blob/5f85fe426e7aa3c1df401a7ae6c1ba837bd76be9/tools/cache/shared_informer.go#L326).

These findings lead us to a simple result: if there is no EventHandler than there will be no resyncs.

And this is indeed what happens in kube-state-metrics. It simply created NewSharedInformer and doesn't add any watchers because of it following the different logic of listing pods directly through a [inf.GetStore().List()](https://github.com/kubernetes/kube-state-metrics/blob/master/collectors/pod.go#L204). I'll quote things here once again:
```
    pinfs := NewSharedInformerList(client, "pods", namespaces, &v1.Pod{})
    podLister := PodLister(func() (pods []v1.Pod, err error) {
        for _, pinf := range *pinfs {
            for _, m := range pinf.GetStore().List() {
                pods = append(pods, *m.(*v1.Pod))
            }
        }
        return pods, nil
    })
```

As an attempt to fix the problem I applied this patch to kube-state-metrics:
```
diff --git a/collectors/collectors.go b/collectors/collectors.go
index 875b218..3a59e93 100644
--- a/collectors/collectors.go
+++ b/collectors/collectors.go
@@ -52,7 +52,9 @@ func NewSharedInformerList(client rest.Interface, resource string, namespaces []
        sinfs := SharedInformerList{}
        for _, namespace := range namespaces {
                slw := cache.NewListWatchFromClient(client, resource, namespace, fields.Everything())
-               sinfs = append(sinfs, cache.NewSharedInformer(slw, objType, resyncPeriod))
+               inf := cache.NewSharedInformer(slw, objType, resyncPeriod)
+               inf.AddEventHandler(cache.ResourceEventHandlerFuncs{AddFunc: func(obj interface{}) {}})
+               sinfs = append(sinfs, inf)
        }
        return &sinfs
 }
```

This was enough to start seeing periodical "forcing resync" in logs when running kube-state-metrics with -v=4.

```
# note that resyncPeriod is set to 5 sec here
$ ./kube-state-metrics --v=4 --kubeconfig=kubeconfig --logtostderr --host=127.0.0.1 --port=80 --telemetry-host=127.0.0.1 --telemetry-port=81 --collectors=pods
I0318 21:38:15.768770   16454 main.go:236] Using all namespace
I0318 21:38:15.773566   16454 main.go:274] Testing communication with server
I0318 21:38:15.814821   16454 main.go:279] Running with Kubernetes cluster version: v1.9. git version: v1.9.3. git tree state: clean. commit: d2835416544f298c919e2ead3be3d0864b52323b. platform: linux/amd64
I0318 21:38:15.814903   16454 main.go:281] Communication with server successful
I0318 21:38:15.815600   16454 pod.go:198] collect pod with v1
I0318 21:38:15.815626   16454 main.go:290] Starting kube-state-metrics self metrics server: 127.0.0.1:18081
I0318 21:38:15.815707   16454 main.go:360] Active collectors: pods
I0318 21:38:15.815714   16454 main.go:315] Starting metrics server: 127.0.0.1:18080
I0318 21:38:15.816033   16454 reflector.go:202] Starting reflector *v1.Pod (5s) from k8s.io/kube-state-metrics/collectors/collectors.go:65
I0318 21:38:15.816086   16454 reflector.go:240] Listing and watching *v1.Pod from k8s.io/kube-state-metrics/collectors/collectors.go:65
I0318 21:38:20.860003   16454 reflector.go:286] k8s.io/kube-state-metrics/collectors/collectors.go:65: forcing resync
I0318 21:38:25.860295   16454 reflector.go:286] k8s.io/kube-state-metrics/collectors/collectors.go:65: forcing resync
I0318 21:38:30.860556   16454 reflector.go:286] k8s.io/kube-state-metrics/collectors/collectors.go:65: forcing resync
I0318 21:38:35.860835   16454 reflector.go:286] k8s.io/kube-state-metrics/collectors/collectors.go:65: forcing resync

```
I have been running patched kube-state-metrics in production for a couple of days and didn't notice any deviation in reported data. So, it seems that the patch works. However, I would like somebody to double check my findings. I'm also not sure that the patch is the best possible solution.

Base on my observation of how kube-state-metrics can be wrong in production, I think this is an important bug to get the fix soon. There might be also logical problems in client-go. At least the need to create a handler for an informer to periodically resync seems questionable for me. But, I'm not a big expert so I'm not going to judge much here.

brancz:
Fixes #402 

@andyxning

### DIFF
--- a/collectors/collectors.go
+++ b/collectors/collectors.go
@@ -52,7 +52,14 @@ func NewSharedInformerList(client rest.Interface, resource string, namespaces []
 	sinfs := SharedInformerList{}
 	for _, namespace := range namespaces {
 		slw := cache.NewListWatchFromClient(client, resource, namespace, fields.Everything())
-		sinfs = append(sinfs, cache.NewSharedInformer(slw, objType, resyncPeriod))
+		inf := cache.NewSharedInformer(slw, objType, resyncPeriod)
+
+		// We are adding an empty EventHandler on purpose, as without
+		// EventHandlers a shared informer will never be resynced, therefore it
+		// is only a matter of time until it will serve stale metrics. (see
+		// https://github.com/kubernetes/kube-state-metrics/issues/402)
+		inf.AddEventHandler(cache.ResourceEventHandlerFuncs{AddFunc: func(obj interface{}) {}})
+		sinfs = append(sinfs, inf)
 	}
 	return &sinfs
 }


brancz:
Seems like we first have to verify again, per the discussion in #402

PR #406, opened 2018-03-26
fix cronjob parse

andyxning:
This should fix cronjob schedule parse for next schedule time calculation.

Sorry for changing this  logic over and over again. But, it should work now. 

The reason for the next schedule time ut test is not idempotent in all time zone is that:
* the time for the same unix timestamp is different in different time zone
* the cronjob is parsed in local time zone 

 For example, if the schedule cronjob config is `0 */6 * * *` and the last scheduled time zone is . In GMT+8, i.e., "Asia/Shanghai". 
* schedule time in local: "2018-03-11 12:34:56 +0800 CST"
* next schedule time in local: "2018-03-11 18:00:00 +0800 CST"
* next schedule time in utc: "1.5207624e+09"

Whereas in the time zone is UTC, i.e., travis CI
* schedule time in local: "2018-03-11 04:34:56 +0000 UTC"
* next schedule time in local: "2018-03-11 06:00:00 +0000 UTC"
* next schedule time in utc: "1.520748e+09"

As for now, [cronjob always use five elements tuple to represent the schedule cron config](https://kubernetes.io/docs/concepts/workloads/controllers/cron-jobs/#creating-a-cron-job), kube-state-metrics should be in consensus.

### DIFF
--- a/collectors/cronjob.go
+++ b/collectors/cronjob.go
@@ -143,7 +143,7 @@ func (cjc *cronJobCollector) Collect(ch chan<- prometheus.Metric) {
 }
 
 func getNextScheduledTime(schedule string, lastScheduleTime *metav1.Time, createdTime metav1.Time) (time.Time, error) {
-	sched, err := cron.Parse(schedule)
+	sched, err := cron.ParseStandard(schedule)
 	if err != nil {
 		return time.Time{}, fmt.Errorf("Failed to parse cron job schedule '%s': %s", schedule, err)
 	}
--- a/collectors/cronjob_test.go
+++ b/collectors/cronjob_test.go
@@ -17,6 +17,8 @@ limitations under the License.
 package collectors
 
 import (
+	"fmt"
+	"math"
 	"testing"
 	"time"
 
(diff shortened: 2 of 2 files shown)


collectors/cronjob.go
```diff
@@ -143,7 +143,9 @@ func (cjc *cronJobCollector) Collect(ch chan<- prometheus.Metric) {
 }
 
 func getNextScheduledTime(schedule string, lastScheduleTime *metav1.Time, createdTime metav1.Time) (time.Time, error) {
-	sched, err := cron.Parse(schedule)
+	sched, err := cron.ParseStandard(schedule)
+	fmt.Printf("-----%+v\n", schedule)
```
brancz: could you remove these debug outputs?

brancz:
looks like a build is still failing, and some debug messages are still in there

andyxning:
> looks like a build is still failing, and some debug messages are still in there

@brancz Yes. they are exist for debugging. Now i have test this in GMT+8 time zone and UTC. Could you please help to clone this branch and `make test-unit` in your time zone to test the ut change?

brancz:
all green on CI and locally

/approve
/lgtm

PR #411, opened 2018-04-03
update readme for cronjob collector

### LINKED ISSUE #409: Cronjob support, docs question
Is this section of the docs still correct?
>If users want to enable this feature when kubernetes version larger than 1.7, It must be configured, with the following parameter setting for apiserver. --runtime-config=batch/v2alpha1=true

It seems that given https://github.com/kubernetes/kube-state-metrics/commit/20c6b1f78a1174cbd95272bc72183d6f96c64db7

Enabling `batch/v2alpha1` in the api server is no longer needed.

r0fls:
fix #409

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,9 +92,10 @@ additional metrics!
 > **collectors**
 > * v2alpha1/cronjob
 >
-> If users want to enable this feature when kubernetes version larger than 1.7,
-> It must be configured, with the following parameter setting for apiserver.
-> `--runtime-config=batch/v2alpha1=true`
+> If users want to enable this feature in kubernetes version less than 1.8,
+> it must be configured with the following parameter setting for apiserver:
+> `--runtime-config=batch/v2alpha1=true`. For cluster versions >= 1.8, cronjobs
+> are enabled by default in batch/v2alpha1
 >
 > Any collectors and metrics based on alpha Kubernetes APIs are excluded from any stability guarantee,
 > which may be changed at any given release.
--- a/README.md
+++ b/README.md
@@ -87,15 +87,7 @@ additional metrics!
 >	* kube_node_status_capacity_nvidia_gpu_cards
 >	* kube_node_status_allocatable_nvidia_gpu_cards
 >
->	are in alpha stage and will be deprecated when the kubernetes gpu support is final in 1.9 version.
->
-> **collectors**
-> * v2alpha1/cronjob
->
-> If users want to enable this feature in kubernetes version less than 1.8,
-> it must be configured with the following parameter setting for apiserver:
-> `--runtime-config=batch/v2alpha1=true`. For cluster versions >= 1.8, cronjobs
-> are enabled by default in batch/v2alpha1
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -92,9 +92,10 @@ additional metrics!
 > **collectors**
 > * v2alpha1/cronjob
 >
-> If users want to enable this feature when kubernetes version larger than 1.7,
-> It must be configured, with the following parameter setting for apiserver.
-> `--runtime-config=batch/v2alpha1=true`
+> If users want to enable this feature in kubernetes version less than 1.8,
+> it must be configured with the following parameter setting for apiserver:
+> `--runtime-config=batch/v2alpha1=true`. For cluster versions >= 1.8, cronjobs
+> are enabled by default in batch/v2alpha1
```
brancz: I think we should just remove this statement all together, as kube-state-metrics only supports using the v1beta1 API.

r0fls: Sounds good, I left the following lines though: https://github.com/kubernetes/kube-state-metrics/pull/411/files#diff-04c6e90faac2675aa89e2176d2eec7d8R100

Since it seems they still apply to the GPU metrics.

andyxning: According to Kubernetes 1.10 release note:

> The alpha Accelerators feature gate is deprecated and will be removed in v1.11. Please use device plugins (https://github.com/kubernetes/features/issues/368) instead. They can be enabled using the DevicePlugins feature gate. (#57384, @mindprince)

gpu support for kube-state-metrics will exist at least until Kubernetes 1.10. It's ok to keep gpu related statement here. 

@r0fls BTW, could you please update the statement about the deprecation for gpu since it is not accurate for now:
`are in alpha stage and will be deprecated when the kubernetes gpu support is final in 1.9 version.` -> `are in alpha stage and will be deprecated when the kubernetes accelerator feature support is final in 1.11 version.`

andyxning: Others lgtm.

r0fls: ✅ done

andyxning: lgtm.

Waiting for a secondary review from @brancz

README.md
```diff
@@ -87,14 +87,7 @@ additional metrics!
...
 >
->	are in alpha stage and will be deprecated when the kubernetes gpu support is final in 1.9 version.
->
-> **collectors**
-> * v2alpha1/cronjob
->
-> If users want to enable this feature when kubernetes version larger than 1.7,
-> It must be configured, with the following parameter setting for apiserver.
-> `--runtime-config=batch/v2alpha1=true`
+>	are in alpha stage and will be deprecated when the kubernetes accelerator feature support is deprecated in version v1.11.
```
mindprince: > kubernetes accelerator feature support is deprecated in version v1.11.

Kubernetes accelerator support is already deprecated in 1.10. The PR to completely remove it is merged in master, so it will be completely removed in 1.11.

andyxning: @mindprince Thanks for the more accurate deprecation procudure about Kubernetes accelerator. Will fire another PR to update this info.

andyxning: @mindprince FRI #417

PR #417, opened 2018-04-07
fix kubernetes accelerator deprecation procedure

andyxning:
Fix https://github.com/kubernetes/kube-state-metrics/pull/411#discussion_r179186482
>>kubernetes accelerator feature support is deprecated in version v1.11.

> Kubernetes accelerator support is already deprecated in 1.10. The PR to completely remove it is merged in master, so it will be completely removed in 1.11.

/cc @mindprince @brancz


README.md
```diff
@@ -87,7 +87,7 @@ additional metrics!
 >	* kube_node_status_capacity_nvidia_gpu_cards
 >	* kube_node_status_allocatable_nvidia_gpu_cards
 >
->	are in alpha stage and will be deprecated when the kubernetes accelerator feature support is deprecated in version v1.11.
+>	are in alpha stage and will be deprecated when the kubernetes accelerator feature support is completely deprecated and removed in version v1.11(Kubernetes accelerator support is already deprecated in v1.10).
```
mindprince: I would say:

```
are deprecated and will be completely removed when the Kubernetes accelerator feature support is removed in version v1.11. (Kubernetes accelerator support is already deprecated in v1.10).
```

andyxning: @brancz @mindprince As for the deprecation statement about gpu metrics, i suggest kube-state-metrics announce it like this:
```
Kubernetes accelerator support is already deprecated in Kubernetes v1.10 and will be completely removed in version v1.11. As Kubernetes has a six month(i.e., two minor release) support window, kube-state-metrics will keep these gpu metrics until Kubernetes v1.13 is released. 
```

mindprince:
That policy doesn’t apply to alpha APIs. Kubernetes accelerator support was
alpha.

andyxning:
Thanks for pointing this out. However, in case users are still using Kubernetes 1.8/1.9/1.10, kube-state-metrics should support these gpu related metrics until kube-state-metrics do not support Kubernetes 1.8/1.9/1.10 anymore. This is only required for kube-state-metrics.

mindprince:
Okay, I just wanted to point out that you are not obligated to support it
because of some Kubernetes support policies but feel free to continue doing
so. Also, when you update your `vendor/` directories to Kubernetes 1.11,
your code will not work.

andyxning:
> Also, when you update your `vendor/` directories to Kubernetes 1.11,
your code will not work.

@mindprince Thanks for pointing this. 
I have also thought about this. If we decide to support for more two minor versions until v1.13, we should definitely add some guards if possible. Actually i am also fine to directly deprecated gpu related metrics from Kubernetes accelerator support once Kubernetes v1.11 is released. Just from a user perspective. Want to hear what @brancz think about this. :)

And after Kubernetes v1.11 is released, we need to add the official gpu related metrics based on device plugin.

README.md
brancz: not sure about that, I think we should just remove them as well, and people who really need them in that way can use the older version of kube-state-metrics

andyxning: Ok. Then we should state it as [what @mindprince said above](https://github.com/kubernetes/kube-state-metrics/pull/417#discussion_r179923665)

andyxning:
Never mind. It is totally user depends. If users complain about it and suggest kube-state-metrics to continue support gpu related metrics after Kubernetes v1.11, we can update the statement and do whatever we can continue to support it.

thomasjungblut:
> If users complain about it

Here I am :) we just upgraded to k8s 1.9 and used the device plugin architecture instead of the accelerator feature and all of our gpu resource charts broke. Any idea how to get that back?

brancz:
You are in luck, the next release (that we will start the process for very soon) will have support for generic pod requests/limits as well as node capacity/allocatable.

Issue #290, opened 2017-10-27
Add more resource instance to e2e test 


andyxning:
Currently, we only check for ClusterRoleBinding, ClusterRole, ServiceAccount, Deployment, RoleBinding, Role and Service.

We should add more resources when access kube-state-metrics `metrics` endpoint. This will make the e2e test more reliable and more help for us to discover problems.

andyxning:
@dohnto Glad to hear it. I would be very happy if you can spent some time on completing this with the requirements from the issue description. Once it is ready, i will help to review. :)

andyxning:
Since an enhanced e2e tests has been merged in #418 , we can currently close this.

PR #418, opened 2018-04-07
Extend e2e tests

dohnto:
Addressing #290. I tried to made commits readable as possible, can squash if prefered.

* I refactored e2e testsuite allowing user to overwrite whether he wants to install minikube & kubectl and also choose minikube's driver (locally when testing I want to use virtualbox and I don't want the script to download minikube and kubectl over and over to my system path).

* I added `cronjob daemonset hpa job limitrange persistentvolumeclaim replicationcontroller resourcequota statefulset` resources which are created

* I added a simple check that expected metrics are exposed

### DIFF
--- a/tests/README.md
+++ b/tests/README.md
@@ -1,0 +1,19 @@
+# End2end testsuite
+
+This folder contains simple e2e tests.
+When launched it spins up a kubernetes cluster using minikube, creates several kubernetes resources and launches a kube-state-metrics deployment.
+Then, it downloads kube-state-metrics' metrics and examines validity using `promtool` tool.
+
+The testsuite is run automatically using Travis.
+
+## Running locally
+
+In case you need to run e2e test manually on your local machine, you can configure the `e2e.sh` script by few environment variables.
+
+```bash
+export E2E_SETUP_MINIKUBE=        # set to empty string if you have already your own minikube binary, prevents from downloading one
+export E2E_SETUP_KUBECTL=         # set to empty string if you have already your own kubectl binary, prevents from downloading one
+export MINIKUBE_DRIVER=virtualbox # choose minikube's driver of your choice
+export SUDO=                      # if you don't need sudo, you can redefine the SUDO variable from default `sudo`
+./tests/e2e.sh
+```
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -20,17 +20,48 @@ set -o pipefail
 KUBERNETES_VERSION=v1.8.0
 KUBE_STATE_METRICS_LOG_DIR=./log
 KUBE_STATE_METRICS_IMAGE_NAME='quay.io/coreos/kube-state-metrics'
-KUBE_STATE_METRICS_IMAGE_NAME_PATTERN='quay.io\/coreos\/kube-state-metrics'
 PROMETHEUS_VERSION=2.0.0
+E2E_SETUP_MINIKUBE=${E2E_SETUP_MINIKUBE-yes}
+E2E_SETUP_KUBECTL=${E2E_SETUP_KUBECTL-yes}
+E2E_SETUP_PROMTOOL=${E2E_SETUP_PROMTOOL-yes}
+MINIKUBE_DRIVER=${MINIKUBE_DRIVER:-none}
+SUDO=${SUDO-sudo}
 
 mkdir -p $KUBE_STATE_METRICS_LOG_DIR
 
-# setup a Kubernetes cluster
-curl -sLo minikube https://storage.googleapis.com/minikube/releases/v0.25.2/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+function finish() {
+    echo "calling cleaup function"
+    # kill kubectl proxy in background
+    kill %1 || true
+    kubectl delete -f kubernetes/ || true
+    kubectl delete -f tests/manifests/ || true
+}
+
+function setup_minikube() {
+    curl -sLo minikube https://storage.googleapis.com/minikube/releases/v0.25.2/minikube-linux-amd64 \
+        && chmod +x minikube \
+        && $SUDO mv minikube /usr/local/bin/
+}
+
+function setup_kubectl() {
+    curl -sLo kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl \
+        && chmod +x kubectl \
+        && $SUDO mv kubectl /usr/local/bin/
+}
+
+function setup_promtool() {
+    wget -q -O /tmp/prometheus.tar.gz https://github.com/prometheus/prometheus/releases/download/v$PROMETHEUS_VERSION/prometheus-$PROMETHEUS_VERSION.linux-amd64.tar.gz
+    tar zxfv /tmp/prometheus.tar.gz -C /tmp/ prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool
+    $SUDO mv /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool /usr/local/bin/
+    rmdir /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64
+    rm /tmp/prometheus.tar.gz
+}
+
+[ -n "$E2E_SETUP_MINIKUBE" ] && setup_minikube
 
 minikube version
 
-curl -sLo kubectl https://storage.googleapis.com/kubernetes-release/release/$(curl -s https://storage.googleapis.com/kubernetes-release/release/stable.txt)/bin/linux/amd64/kubectl && chmod +x kubectl && sudo mv kubectl /usr/local/bin/
+[ -n "$E2E_SETUP_KUBECTL" ] && setup_kubectl
 
 export MINIKUBE_WANTUPDATENOTIFICATION=false
 export MINIKUBE_WANTREPORTERRORPROMPT=false
@@ -40,7 +71,7 @@ mkdir $HOME/.kube || true
 touch $HOME/.kube/config
 
 export KUBECONFIG=$HOME/.kube/config
-sudo minikube start --vm-driver=none --kubernetes-version=$KUBERNETES_VERSION --logtostderr
+$SUDO minikube start --vm-driver=$MINIKUBE_DRIVER --kubernetes-version=$KUBERNETES_VERSION --logtostderr
 
 minikube update-context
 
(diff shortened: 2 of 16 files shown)


dohnto:
Currently, the e2e is failing with promtool complaining that
> kube_job_status_start_time counter metrics should have "_total" suffix

brancz:
@dohnto in order to get rid of that, change the type to a gauge, instead of a counter.

andyxning:
>  in order to get rid of that, change the type to a gauge, instead of a counter.

Agreed. `kube_job_status_start_time` represents the start time which should be a gauge.

andyxning:
@dohnto Once this PR is ready to review, please do not hesitate to ping me and @brancz .

BTW, good job. 👍

dohnto:
Ok @andyxning @brancz , this PR is ready (there might be more PR's afterwards), you can review by commits if neccessary.

I am aware, that changelog notes there are missing, I will add it after the review.

andyxning:
Will review this later today.

tests/e2e.sh
```diff
@@ -127,9 +166,14 @@ echo "access kube-state-metrics metrics endpoint"
...
-wget -q -O /tmp/prometheus.tar.gz https://github.com/prometheus/prometheus/releases/download/v$PROMETHEUS_VERSION/prometheus-$PROMETHEUS_VERSION.linux-amd64.tar.gz
-tar zxfv /tmp/prometheus.tar.gz -C /tmp
-cat $KUBE_STATE_METRICS_LOG_DIR/metrics | /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool check metrics
+[ -n "$E2E_SETUP_PROMTOOL" ] && setup_promtool
+cat $KUBE_STATE_METRICS_LOG_DIR/metrics | promtool check metrics
+
+collectors="configmap cronjob daemonset deployment endpoint hpa job limitrange namespace node persistentvolumeclaim pod replicaset replicationcontroller resourcequota secret service statefulset"
+for collector in $collectors; do
+    echo "checking that kube_${collector}* metrics exists"
+    grep "^kube_${collector}_" $KUBE_STATE_METRICS_LOG_DIR/metrics
```
brancz: I feel that if “tests” get any more complicated than this, then we should start implementing them in Go.

brancz:
Generally looks good

/lgtm

tests/README.md
```diff
@@ -0,0 +1,19 @@
+# End2end testsuite
+
+This folder contains simple e2e tests.
+When launched it spins up a kubernetes cluster using minikube, creates several kubernetes resources and launches a kube-state-metrics deployment.
+Then, it downloads its metrics and examines validity using `promtool` tool.
```
andyxning: `s/its/kube-state-mtrics`

tests/e2e.sh
```diff
@@ -20,17 +20,49 @@ set -o pipefail
...
+function finish() {
+    echo "calling cleaup function"
+    # kill kubectl proxy in background
+    kill %1 || true
+    kubectl delete -f kubernetes/ || true
+    kubectl delete -f tests/manifests/ || true
+}
+
+function setup_minikube() {
+    # setup a Kubernetes cluster
```
andyxning: drop this comment as it is just downloading the minikube binary. :)

tests/e2e.sh
```diff
@@ -127,9 +166,14 @@ echo "access kube-state-metrics metrics endpoint"
 curl -s "http://localhost:8001/api/v1/proxy/namespaces/kube-system/services/kube-state-metrics:8080/metrics" >$KUBE_STATE_METRICS_LOG_DIR/metrics
 
 echo "check metrics format with promtool"
-wget -q -O /tmp/prometheus.tar.gz https://github.com/prometheus/prometheus/releases/download/v$PROMETHEUS_VERSION/prometheus-$PROMETHEUS_VERSION.linux-amd64.tar.gz
-tar zxfv /tmp/prometheus.tar.gz -C /tmp
-cat $KUBE_STATE_METRICS_LOG_DIR/metrics | /tmp/prometheus-$PROMETHEUS_VERSION.linux-amd64/promtool check metrics
+[ -n "$E2E_SETUP_PROMTOOL" ] && setup_promtool
+cat $KUBE_STATE_METRICS_LOG_DIR/metrics | promtool check metrics
+
+collectors="configmap cronjob daemonset deployment endpoint hpa job limitrange namespace node persistentvolumeclaim pod replicaset replicationcontroller resourcequota secret service statefulset"
```
andyxning: Maybe we should generate the collectors list from collectors package by leveraging the go source file names. This help us from avoids updating e2e test when new resource is added or deleted.

tests/e2e.sh
```diff
@@ -70,16 +102,21 @@ set -e
...
 kubectl version
 
+# ensure that we build docker image in minikube
+[ "$MINIKUBE_DRIVER" != "none" ] && eval $(minikube docker-env)
+
 # query kube-state-metrics image tag
 make container
 docker images -a
-ksm_image_tag=`docker images -a|grep 'quay.io/coreos/kube-state-metrics'|awk '{print $2}'|sort -u`
+ksm_image_tag=`docker images -a|grep 'quay.io/coreos/kube-state-metrics'|awk '{print $2}'|sort -u|head -n1`
```
andyxning: Why should we `head -n1`?

tests/e2e.sh
```diff
@@ -70,16 +102,21 @@ set -e
 
 kubectl version
 
+# ensure that we build docker image in minikube
```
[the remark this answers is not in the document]
dohnto: Not sure whether I understand. In travis we don't have to switch docker to minikube's docker since it is shared anyway (it uses localkube) but in other setups (virtualbox for in my case, I need to build the docker image in minikube's context in order to run the ksm deployment.

Just to note,  you cannot use the command `minikube docker-env` command when running with `--vm-driver=none` as described [here](https://github.com/kubernetes/minikube/issues/2443).

As I see this, the solution I picked is the only one working :)

tests/e2e.sh
dohnto: In my environment (I have build multiple kms with different versions) I receive:

```bash
ksm_image_tag=`docker images -a|grep 'quay.io/coreos/kube-state-metrics'|awk '{print $2}'|sort -u`
echo $ksm_image_tag
v1.3.0
v1.2.0
```

By selecting only the first one, we choose only the image which was build with latest version.

tests/e2e.sh
dohnto: Addressed in 6a44a4c.

tests/README.md
dohnto: Addressed in 6a44a4c.

tests/e2e.sh
dohnto: Ok what about
```
$ find collectors/ -name "*.go" -not -name "*_test.go" -not -name "collectors.go" | xargs -n1 basename | awk -F. '{print $1}'
```

tests/e2e.sh
andyxning: @dohnto There is no guarantee that the first one version is what we want. This has no problem for travis e2e environment as it is empty. This should only exist under local environment. A simple method is try to `docker rmi` all kube-state-metrics before running e2e test.

tests/e2e.sh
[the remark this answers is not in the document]
andyxning: I see. Thanks for the explanation. This is required when 
* no local docker daemon is already exists
* using minikube with none-none vm-drivers

Am i right?

tests/e2e.sh
dohnto: You are right, there is no guarantee, it is a best guess.

I think that the best option is parsing the `make container` or other way of getting the $(TAG) variable from `Makefile`.

andyxning: Seems also a reasonable way.

tests/e2e.sh
andyxning: seems good to me.

tests/e2e.sh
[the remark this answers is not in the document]
dohnto: I am note sure whether I understand. There are 2 scenarios: a) run with --vm-driver=none and b) run with --vm-driver **!=none**.

a) In Travis, you run `minikube start --vm-driver=none` and It will setup the minikube locally without any virtualization. The kubernetes (minikube) and the system share the docker daemon. If you create a kubernetes Deployment with some docker image, it is tested, that the image is already on machine (unless `:latest` tag or `imagePullPolicy: Always` is used, but in this scenario it is not). Everything is easy and transparent.

b)
I locally don't want to run minikube on my host system (it is not recommended), so I run `minikube start --vm-driver=virtualbox` and it runs the kubernetes inside a virtualbox VM. But I run the kubernetes Deployment with the same image as before. Kubelet will not find the docker image in local docker storage (because it is a brand new virtualmachine), so docker/kubelet will try to pull it from docker registry. That is why I put the line with `minikube docker-env`)

```
$ minikube docker-env
export DOCKER_TLS_VERIFY="1"
export DOCKER_HOST="tcp://192.168.99.100:2376"
export DOCKER_CERT_PATH="/home/tom/.minikube/certs"
export DOCKER_API_VERSION="1.23"
# Run this command to configure your shell:
# eval $(minikube docker-env)
```

If you run this, the docker client will use a remote docker server - the one running inside the VM where kubernetes is running. Therefore I run the command`eval` **before** building the image, I want the image to be present in docker storage **inside** the VM. After that the kubelet will use the docker image I just builded.

So we use the `minikube docker-env` only when `--vm-driver` is not `none`. So, you don't even have to have a docker daemon running, docker client binary is enough.

I hope I made it (my motivation & solution) more clear, if not, please ask again :)

tests/e2e.sh
dohnto: But maybe I can drop this change and fix it in a separate PR.

dohnto: Dropped in 84f8344.

tests/e2e.sh
dohnto: Addessed in 0e48a35.

tests/e2e.sh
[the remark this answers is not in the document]
andyxning: Excellent explanation.

dohnto:
I would like to squash after review but before merge if possible. It seems to many commits for a non-functional piece :)

andyxning:
@dohnto You can squash it now.

dohnto:
Done.

andyxning:
@dohnto Any reason to not squash it into one commit?

dohnto:
Squashed to a single commit.

dohnto:
Ping @andyxning @brancz

andyxning:
lgtm to me.
/cc @brancz

PR #419, opened 2018-04-08
use alpine as base image

andyxning:
This PR will move the base image for kube-state-metrics from `scratch` to `alpine`. Alpine has many useful binaries, such as `sh`. This will be helpful when we need to `exec` into an running kube-state-metrics container and run some debug command.

Last but not the least is that alpine image is small enough thus the kube-state-metircs image will not increase much.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM scratch
+FROM alpine:3.7
 
 COPY kube-state-metrics /
 VOLUME /tmp


brancz:
If any I would prefer busybox over alpine.

andyxning:
@brancz As for UNIX utilies, busybox is good enough and it is more smaller than alpine. But according to readme about alpine: 
> The image is only 5 MB in size and has access to a package repository that is much more complete than other BusyBox based images. 

Seems that alpine is more user friendly when it comes to package management. Maybe we will not  even use this feature, it should be fine to make it available in case we need it. But the precondition is that no much overhead will be added. Does it has some reasons that alpine is not the preferred option. Alpine is only larger than busybox with 3MB.

brancz:
Fine by me as well.

/lgtm
/approve

PR #420, opened 2018-04-08
fix pod unknown stats

### LINKED ISSUE #410: Not recognizing pods in an Unknown state
Hi,

We had two pods in our cluster go into an unknown state however kube-state-metrics isn't recognizing this.  See below:

Output from curl and kubectl:

kube_pod_status_phase{namespace="kube-system",phase="Unknown",pod="filebeat-ht37s"} 0
kube_pod_status_phase{namespace="kube-system",phase="Unknown",pod="filebeat-j1ppj"} 0

kubectl get pods -n kube-system | grep Unknown
filebeat-ht37s                              1/1       Unknown   3          103d
filebeat-j1ppj                              1/1       Unknown   5          103d

Kube-state-metrics version: 1.2
Docker version: 1.12
K8s version: 1.5.4
OS flavor: Ubuntu 16.04 LTS

andyxning:
Fix #410 

This PR will adds another pod unknown situation where `kubectl` uses to mark a pod in Unknown state.


andyxning:
/cc @brancz 

IMHO, it should be fine to call `addGauge` for `descPodStatusPhase` with label `Unknown` because if one of the criteria is met or both of the two criteria are met, the value for `descPodStatusPhase` with `Unknown` label will be always be `1` to indicating that pod is in `Unknown` state.

brancz:
could you add a test for this?

andyxning:
@brancz Done. PTAL.

PR #421, opened 2018-04-09
add go version check

### LINKED ISSUE #416: kms 1.3.0 fails to parse certificate due to Go version
Similar to https://github.com/prometheus/prometheus/issues/3983

For certificates with non-fqdn values in dnsName fields, applications built with certain Go versions (1.9.<something> -> 1.10) will fail to communicate with the apiserver.

```
F0406 16:58:08.867531       1 main.go:245] Failed to create client: ERROR communicating with apiserver: Get https://10.82.16.1:443/version: tls: failed to parse certificate from server: x509: cannot parse dnsName "K8s foo bar"
```

Since "K8s foo bar" is not an fqdn, tls can't parse it won't allow to establish a connection.

This issue was fixed/mitigated in Go 1.10.1
https://groups.google.com/forum/m/#!msg/golang-announce/IkPkOF8JqLs/TFBbWHJYAwAJ

andyxning:
Fix #416 

This PR will ad a go version check for `make build` to help us mitigate the problem encountered in https://github.com/kubernetes/kube-state-metrics/issues/416

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: required
 language: go
 
 go:
-- "1.10"
+- "1.10.1"
 
 install:
 - mkdir -p $HOME/gopath/src/k8s.io
--- a/Makefile
+++ b/Makefile
@@ -10,6 +10,7 @@ Commit = $(shell git rev-parse --short HEAD)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 PKG=k8s.io/kube-state-metrics
 
+
 IMAGE = $(REGISTRY)/kube-state-metrics
 MULTI_ARCH_IMG = $(IMAGE)-$(ARCH)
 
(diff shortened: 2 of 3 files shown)


andyxning:
/cc @brancz 

Ping @dannyk81

brancz:
I think this might be the wrong place to do something like this. The "fix" we should be doing to counter this is having a more unified build pipeline instead of coreos/redhat/google employees building and pushing images. At that point we just need to make sure to use the latest golang version and cut a patch release when it was built with an outdated one. Additionally we may always get into a situation where a new golang version is cut after a kube-state-metrics release. I feel these scripts create a higher burden than gain. We should do a patch release though to build with the new golang version.

brancz:
I disagree with the second point. The golang runtime is a dependency of ours if there is a bug in the dependency that's fixed we need to upgrade and do another release.

andyxning:
The root requirement is that i think we need to make the go version requirement for compiling kube-state-metrics well known not only to maintainers/developers but also end users. Document it somewhere is another option but it relay on that the statement is saw by users which is hard to promise.

brancz:
How about using docker to build the binaries in the first place in our makefile. Then we have a canonical source of the runtime automated into the process.

andyxning:
Nice suggestion. That requires a docker to be available locally. This should not be a problem as the wide spread of docker. :)

This can help avoid the hacked scripts.

Will file another PR to do this.

brancz:
Given that this project is pretty much useless without a container runtime and we're already using docker in our makefile it's fair to do so :slightly_smiling_face: .

PR #422, opened 2018-04-11
use golang image to compile ksm

andyxning:
Fix #421

This PR will add support for using golang container to compile kube-state-metrics.

PR #423, opened 2018-04-11
Add note about kube-state-metrics exposing raw data

### LINKED ISSUE #414: kube-state-metrics keeps failed pods around for kube_pod_* metrics
Some `kube_pod_*` metrics exist for Failed pods:
```
kube_pod_info
kube_pod_owner
kube_pod_labels
kube_pod_created
kube_pod_start_time
```

`kube_status_phase` shows that kube-state-metrics is populating this group of metrics with information about the failed (non running) pods:
screenshot from 2018-04-04 13-35-16

`kubectl` is indeed confirming there's only one pod running (and so are the `kube_pod_container_*` metrics).

I would expect the metrics above to only show `Pending|Running` pods, maybe `Unknown` but I'm not sure when the last value is set and seems like it could be a transient state more than anything else.

brancz:
This closes #414

cc @andyxning

### DIFF
--- a/README.md
+++ b/README.md
@@ -8,6 +8,15 @@ the Metrics section below.) It is not focused on the health of the individual
 Kubernetes components, but rather on the health of the various objects inside,
 such as deployments, nodes and pods.
 
+That kube-state-metrics is about generating metrics from Kubernetes API
+objects without modification. This ensures, that features provided by
+kube-state-metrics have the same grade of stability as the Kubernetes API
+objects themselves. In turn this means, that kube-state-metrics in certain
+situation may not show the exact same values as kubectl, as kubectl applies
+certain heuristics to display comprehensible messages. kube-state-metrics
+exposes raw data unmodified from the Kubernetes API, this way users has all the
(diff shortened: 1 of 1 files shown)


andyxning:
The reason for ci error is the same as https://github.com/kubernetes/kube-state-metrics/pull/422 which is confused. We have change nothing about this.

andyxning:
Seems a new minikube version has been released. This may be related to the ci error.

README.md
```diff
@@ -8,6 +8,15 @@ the Metrics section below.) It is not focused on the health of the individual
 Kubernetes components, but rather on the health of the various objects inside,
 such as deployments, nodes and pods.
 
+That kube-state-metrics is about generating metrics of off Kubernetes API
```
andyxning: `s/That kube-state-metrics/Kube-state-metrics`

`s/of off/of`

andyxning:
Small nits. Others lgtm.

README.md
andyxning: @brancz Comments need to be addressed. :)

PR #424, opened 2018-04-11
pin minikube v0.25.2

andyxning:
Minikube has released v0.26.0 and it seems that e2e test for kube-state-metrics is not compatible with the latest minikube release. Let's first pin minikube version to v0.25.2 and file an issue in minikube to track this.


andyxning:
/cc @brancz 

Once this PR merged, we need to rebase #423 and #422 .

PR #426, opened 2018-04-11
force bootstrapper to localkube for minikube

andyxning:
Force minikube use localkube as the bootstrapper instead of kubeadm. For more info: https://github.com/kubernetes/minikube/issues/2704

### DIFF
--- a/scripts/e2e.sh
+++ b/scripts/e2e.sh
@@ -26,7 +26,7 @@ PROMETHEUS_VERSION=2.0.0
 mkdir -p $KUBE_STATE_METRICS_LOG_DIR
 
 # setup a Kubernetes cluster
-curl -sLo minikube https://storage.googleapis.com/minikube/releases/v0.25.2/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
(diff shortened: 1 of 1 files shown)


andyxning:
@brancz We can wait the minikube maintainers to give more suggestion on this before merging this. If they chose a patch release to mitigate this problem, we should do nothing. :)

scripts/e2e.sh
```diff
@@ -26,7 +26,7 @@ PROMETHEUS_VERSION=2.0.0
 mkdir -p $KUBE_STATE_METRICS_LOG_DIR
 
 # setup a Kubernetes cluster
-curl -sLo minikube https://storage.googleapis.com/minikube/releases/v0.25.2/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
+curl -sLo minikube https://storage.googleapis.com/minikube/releases/latest/minikube-linux-amd64 && chmod +x minikube && sudo mv minikube /usr/local/bin/
```
brancz: I'd prefer to still explicitly bump minikube, there have been numerous occasions where the latest releases were broken to a point where it was just not usable at all.

andyxning: Agreed for now since kube-state-metrics do not need any new feature that latest minikube supports. Let's first pin at `v0.25.2` and seem whether Travis CI ubuntu image will support systemd alongside.

PR #427, opened 2018-04-11
Add total resources scraped metric

ayushpateria:
This adds `ksm_resources_scraped_total` which counts total resources scraped since the start of the metrics server.

### DIFF
--- a/collectors/collectors.go
+++ b/collectors/collectors.go
@@ -37,6 +37,13 @@ var (
 		[]string{"resource"},
 	)
 
+	ObjectsListedTotalMetric = prometheus.NewCounter(
+		prometheus.CounterOpts{
+			Name: "ksm_kubernetes_objects_listed_total",
+			Help: "Total Kubernetes objects listed since ksm start",
+		},
+	)
+
 	ResourcesPerScrapeMetric = prometheus.NewSummaryVec(
 		prometheus.SummaryOpts{
 			Name: "ksm_resources_per_scrape",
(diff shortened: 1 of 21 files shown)


collectors/collectors.go
```diff
@@ -37,6 +37,13 @@ var (
 		[]string{"resource"},
 	)
 
+	ResourcesScrapedTotalMetric = prometheus.NewCounter(
```
andyxning: `ResourcesScrapedTotalMetric` should be registered in `main.go` just like what we do for `ScrapeErrorTotalMetric`. 

FYI: https://github.com/kubernetes/kube-state-metrics/blob/master/main.go#L248-L253

andyxning:
Since we already have a scraped number per resource, Prometheus aggregator function should help to summarize the total number. Why should we still need another metrics to track total metrics? Or in the other words, what is the directly demand that kube-state-metrics can not satisfy for now?

collectors/collectors.go
ayushpateria: Yes, it's here - https://github.com/kubernetes/kube-state-metrics/pull/427/files#diff-7ddfb3e035b42cd70649cc33393fe32cR250

andyxning:
> Since we already have a scraped number per resource, Prometheus aggregator function should help to summarize the total number. Why should we still need another metrics to track total metrics? Or in the other words, what is the directly demand that kube-state-metrics can not satisfy for now?

Ping @brancz @ayushpateria

brancz:
I think it's slightly different, as the counter would allow doing a rate function, which the aggregate of the gauges would not allow that.

collectors/collectors.go
```diff
@@ -37,6 +37,13 @@ var (
...
+		prometheus.CounterOpts{
+			Name: "ksm_resources_scraped_total",
```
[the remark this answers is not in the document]
ayushpateria: Updated!

collectors/collectors.go
```diff
@@ -37,6 +37,13 @@ var (
 		[]string{"resource"},
 	)
 
+	ObjectsListedTotalMetric = prometheus.NewCounter(
```
brancz: As there are only a small amount of collectors it would make sense to add a label here to see which collector/resource type a specific counter comes from.

PR #430, opened 2018-04-13
Expose Pod Annotations as Prometheus labels

elcuyi:


elcuyi:
I appreciate your concern re labels explosion....
The problem is not so much with the `kubernetes.io/created-by` etc. labels as I'm not really interested in those. There are other pod annotations aside from the `kubernetes.io` ones that people might want to use in their prometheus queries.
If you're doing the same thing for pod labels `kube_pod_lables` why not for pod annotations?
I'm not saying you should merge this as is but if you can think of a better way to expose only those annotations one might be interested in then I'd be very interested in hearing it.

andyxning:
@elcuyi  The difference between labels and annotations are: labels are simple key value pairs, while annotations maybe structured data as for values. I am not opposed about adding annotations and actually we have do this for namespace resource. What i am thinking is that whether we need to add a command line knob to enable annotations collecting.

/cc @brancz

elcuyi:
@andyxning Agree with you on not wanting to have the JSON annotations such as `openshift.io/deployment-config` in prometheus. I just realized my pods didn't have them, but that's because they weren't created by openshift itself.
It would still be nice to be able to selectively extract pod annotations maybe based on settings, could just use some guidance on how to make which annotations to extract configurable.

andyxning:
> It would still be nice to be able to selectively extract pod annotations maybe based on settings, could just use some guidance on how to make which annotations to extract configurable.

Yes. But that will make kube-state-metrics complicated to be used or deployed. Will think about this for some time.

brancz:
I also feel that this is probably going to add excessive amounts of metrics. I would suggest to have a flag that is called "--extended-metrics", that enables high cardinality metrics like this.

andyxning:
@brancz @elcuyi Is anyone willing to work on this.

andyxning:
@brancz @elcuyi Is anyone willing to work on this.

brancz:
I'm not

andyxning:
@elcuyi How about you? :)

dohnto:
Maybe I can take a look on this extension? :)

brancz:
@dohnto yeah I'm thinking something like `--disable-metric` which can have multiple metrics separated by `,` and can be repeated. With that we can filter arbitrary metrics. That in addition to the ability to pick collectors should be sufficient I believe.

andyxning:
@dohnto Do you have time to help kube-state-metrics to improve this?

andyxning:
BTW, i suggest we make the annotation metrics by default disabled until end users explicitly collecting them by using `--allow-metric`.

dohnto:
Ok just to assure, is this really we are going to implement or this should be a user's responsibility to drop the metrics using `metric_relabel_config` in prometheus?

brancz:
@dohnto I recommend to use metric relabeling today, until we figure out how to do this on kube-state-metrics side as this will likely take some time.

andyxning:
@dohnto Are you still preparing to help us to refactor this PR according to the comments. :)

dohnto:
I am sorry but I am not going to make it for an another month. If still not implemented, I will continue after my return from my vacation.

andyxning:
@dohnto Never mind. There is no deadline for this feature. :) Feel free to work on it after your vacation. BTW, thanks for your contribution!

brancz:
Yes no worries at all, have a great vacation!

dohnto:
So meanwhile you have implemented #488 which seems close to my designed solution, but not quite. Can anyone explain to me, how to proceed now?

I like the idea of black/white listing. I don't like that you cannot do both (black and white) at the same time and we currently don't have mechanism how to add a new metric which will be disabled by default.

I expected that we will add `kube_pod_annotations` metric, which will be disabled by default and you would have to explicitly enable it using whitelist. But, this would, in current setup, make impossible usege of blacklisting.

andyxning:
> I like the idea of black/white listing. I don't like that you cannot do both (black and white) at the same time and we currently don't have mechanism how to add a new metric which will be disabled by default.

The functionality of current blacklist and whitelist is somewhat limited. But it can really resolve some problems and make end uers life more comfortable. Making blacklist and whitelist more flexible is preferred, imo. I.e., giving end users more flexible.

> I expected that we will add kube_pod_annotations metric, which will be disabled by default and you would have to explicitly enable it using whitelist. But, this would, in current setup, make impossible usege of blacklisting.

I am fine to do this.

PR #432, opened 2018-04-15
Clarify apiserver override example.

jmcarp:
* Drop `--in-cluster` flag, which doesn't exist as of latest release
* Pass `--apiserver` flag to match instructions


andyxning:
@jmcarp Thanks for help updating this! 
/lgtm
/approve

PR #433, opened 2018-04-16
Adds time_to_schedule and time_to_start metrics

mwhittington21:
This PR adds two metrics with the aim of tracking the beginning of the pod lifecycle better. They are named `kube_pod_time_to_schedule` and `kube_pod_time_to_start`.

I tried to get the results I wanted by just adding/subtracting various existing metrics in Prometheus but there's a bit of conditional logic required to make the most sense out of the created vs started timestamps for a pod object.

### DIFF
--- a/Documentation/pod-metrics.md
+++ b/Documentation/pod-metrics.md
@@ -26,4 +26,5 @@
 | kube_pod_container_resource_limits_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_created | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
 | kube_pod_spec_volumes_persistentvolumeclaims_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; |
-| kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
+| kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; |
+| kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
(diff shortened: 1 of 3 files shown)


mwhittington21:
I've now sorted my org membership for the CNCF so that it's reflected in my Linux Foundation account. Didn't realise I hadn't done that yet when I submitted the PR.

andyxning:
@mwhittington21 You need to first sign the CLA.

andyxning:
@mwhittington21 IIUC, you mean that there are some questions or inconvenience when query pod start or schedule time. Can you explain what exactly it is? 

BTW, generally i am opposed to do any aggregation that can be done by Promethues.

mwhittington21:
Sorry, what I meant was I am part of an org that has signed the CLA. I've now got that org listed in my profile at identity.linuxfoundation.org so it should pick up the fact that I've signed the CLA via my org now (I hope).

---

Yeah I'll explain the inconvenience in detail.

The timestamp for a pod moving into the "Scheduled" status is not tracked anywhere in kube-state-metrics. I cannot find a good way to report on a "time to schedule" metric for pods without access to this. The way I'm thinking would be to use some combination of timestamp() and min() by (pod) to calculate the metric, but nothing is really representative of what I'm after.

The time to start for a Pod can be calculated by taking the difference of the creation timestamp and the start timestamp, so this one can be done in prometheus via `clamp_min(kube_pod_start_time - kube_pod_created, 0)`. Doing this inside prometheus rather than kube-state-metrics removes the ability to get a little bit of extra ordering of events by comparing the start/creation timestamps vs the current timestamp but this precision loss isn't much.

---

If you want I can rework this PR to just add a `kube_pod_scheduled_time` metric that reports the timestamp that the pod was scheduled and then calculate the rest inside prometheus. That first use case is the only one that I can't find a way to do in prometheus alone.

brancz:
I would personally prefer the approach where we just expose the `kube_pod_scheduled_time` metric, as all of kube-state-metrics today simply mirrors the Kubernetes API objects and I'd prefer to keep it that way.

Beyond that, I wonder what exactly you want to do with this metric, it seems to me this is mostly about the scheduler, and the scheduler already has metrics for various latencies in regards to the scheduling algorithm and operations. Binding latency, algorithm latency, e2e scheduling latency are all already existing scheduler metrics coming directly from the scheduler with out the additional indirection of kube-state-metrics.

dadux:
@brancz 

The `kube_pod_scheduled_time` is particularly interesting for us for our SLAs. 
We run multi-tenanted k8s clusters, for different internal customers. One of our SLA would be for instance "we guarantee all your pods to be scheduled in less than 5mins".
A pod could be stuck in "Pending" for a while, say if the cluster is at capacity and scaling up.

Having that metric on a per pod basis - which the scheduler doesn't provide - would allow us to do some aggregation per customer/namespace etc... and alert if we go above a certain level.

brancz:
I get the SLA, but problem is that this metric won't help you track that SLA over time, as pods disappear and with them the scheduling information. Only the scheduler keeps the time-series alive over time.

mwhittington21:
We can track it over time by using range selectors in prometheus (e.g. metric[24h]) and then aggregating over that, thus capturing all pods over that time period.

collectors/pod.go
```diff
@@ -73,6 +73,12 @@ var (
 		[]string{"namespace", "pod"}, nil,
 	)
 
+	descPodScheduled = prometheus.NewDesc(
+		"kube_pod_scheduled",
```
brancz: As this comes from information under the status subresource convention is to name this `kube_pod_status_...`, I'd prefer `kube_pod_status_scheduled_time`.

mwhittington21:
Updated with the new metric name

brancz:
It seems something went wrong while rebasing.

mwhittington21:
Rebase fixed, thanks for pointing that out.

Documentation/pod-metrics.md
```diff
@@ -26,4 +26,4 @@
 | kube_pod_container_resource_limits_nvidia_gpu_devices | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; |
 | kube_pod_created | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
 | kube_pod_spec_volumes_persistentvolumeclaims_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; |
-| kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; |
+| kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
```
lksv: There is probably missing ENTER and one `|` to be displayed correctly

mwhittington21: That's what I get for using soft wrap in my text editor...

PR #436, opened 2018-04-17
Add hpa conditions

discordianfish:
Fixed after discussion in Slack, thanks to @DirectXMan12 
This now uses a explicitly versioned client as recommended anyways.

---
Not working yet, need to understand the versioning and would welcome somebody pointing me in the right direction to solve this in a backward compatible way. The HPA conditions are part of the HPA status since v2beta1. When I update the import though, I get this error:

```
E0417 12:51:00.051078   17716 reflector.go:386] k8s.io/kube-state-metrics/collectors/collectors.go:62: expected type *v2beta1.HorizontalPodAutoscaler, but watch event object had type *v1.HorizontalPodAutoscaler
```

### DIFF
--- a/collectors/hpa.go
+++ b/collectors/hpa.go
@@ -21,7 +21,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
-	autoscaling "k8s.io/api/autoscaling/v1"
+	autoscaling "k8s.io/api/autoscaling/v2beta1"
 	"k8s.io/client-go/kubernetes"
 )
 
@@ -60,6 +60,11 @@ var (
 		descHorizontalPodAutoscalerLabelsHelp,
 		descHorizontalPodAutoscalerLabelsDefaultLabels, nil,
 	)
+	descHorizontalPodAutoscalerCondition = prometheus.NewDesc(
+		"kube_hpa_status_condition",
+		"The condition of this autoscaler.",
+		[]string{"namespace", "hpa", "condition", "status"}, nil,
+	)
 )
 
 type HPALister func() (autoscaling.HorizontalPodAutoscalerList, error)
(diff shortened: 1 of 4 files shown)


collectors/node.go
```diff
@@ -266,28 +266,3 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n v1.Node) {
 	addResource(descNodeStatusAllocatableMemory, n.Status.Allocatable, v1.ResourceMemory)
 	addResource(descNodeStatusAllocatablePods, n.Status.Allocatable, v1.ResourcePods)
 }
-
-// addConditionMetrics generates one metric for each possible node condition
-// status. For this function to work properly, the last label in the metric
-// description must be the condition.
-func addConditionMetrics(ch chan<- prometheus.Metric, desc *prometheus.Desc, cs v1.ConditionStatus, lv ...string) {
```
brancz: did you mean to put this into `collectors.go`?

discordianfish: Ah moved it into it's own file and forget to add it. Fixed! Or would you rather have it in collectors.go?

brancz: Either works for me - no strong opinion.

dohnto: Moving to a separate file will probably break e2e tests due to changes I have made recently. But it should be easy to fix that.

discordianfish: I've just moved it to collectors.go

collectors/hpa.go
```diff
@@ -21,7 +21,7 @@ import (
 
 	"github.com/golang/glog"
 	"github.com/prometheus/client_golang/prometheus"
-	autoscaling "k8s.io/api/autoscaling/v1"
+	autoscaling "k8s.io/api/autoscaling/v2beta1"
```
andyxning: @discordianfish Why do we change api version for autoscaling from `v1` to `v2beta1`?

andyxning: Is `v2beta1` the newest api version for autoscaling?

discordianfish: The Conditions field wasn't available in v1 and according to @DirectXMan12 and @brancz's it's okay to update it since it's been supported for a while (iirc 1.6?)

andyxning: So `v2beta1` is the newest api version for autoscaling?

discordianfish: I think so, yes: https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#api-object

PR #438, opened 2018-04-18
Move gatherAndCompare and related code to separate package and expose it

cofyc:
hi, 

`gatherAndCompare` function is really useful to test `prometheus.Collector`.  I have used it [in kubelet collectors](https://github.com/kubernetes/kubernetes/pull/59170/files#diff-18db6236f1e4cf727350b7f4dafa9e50R17). Now I am working on exposing metrics in local volume provisioner: https://github.com/kubernetes-incubator/external-storage/pull/721. I suggest to move `gatherAndCompare` to separate package and expose it. Then I can vendor it instead of copying.
What do you think?

cc @brancz @andyxning


brancz:
Happy to promote sharing :slightly_smiling_face: lgtm on green

PR #441, opened 2018-04-24
Fix hpa condition labels

discordianfish:
The labels were switched accidentially. This fixes it.


brancz:
oops ... lgtm :+1: good catch!

PR #442, opened 2018-04-26
add pod resource 

ZhiqinYang:
sometimes we only need to pod resource,  but the container resources for limit cannot get when we don't add limit in yaml.

### DIFF
--- a/collectors/pod.go
+++ b/collectors/pod.go
@@ -188,10 +188,17 @@ var (
 	)
 )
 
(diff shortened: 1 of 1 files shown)


ZhiqinYang:
@brancz  
when no set limits I can't calculate the resource of pod in PromQL, so I add this!

brancz:
In that case you want to join information onto pod metrics to get the node it runs on and that way you can use the node metrics.

PR #451, opened 2018-05-03
add generic pod resource request and limit

### LINKED ISSUE #408: pod collector should surface all request and limit resources
The pod collector exposes a named metric for each standard resource.

The pod collector should support the ability to expose metrics for all request and limits on a pod/container including extended resources.  This is important for handling resources that are introduced via device plugins.

@brancz - this was a topic that came up in resource mgmt f2f discussion.  we want to engage with sig-instrumentation to find the right approach.

andyxning:
This PR
* Adds `disable-pod-none-generic-resource-metrics` command line argument to kube-state-metrics to disable old not labeled resource metrics.
* Adds generic new `kube_pod_container_resource_requests ` and `kube_pod_container_resource_limits `.

Fixes #408

### DIFF
--- a/README.md
+++ b/README.md
@@ -33,6 +33,7 @@ the raw metrics.
   - [Resource group version compatibility](#resource-group-version-compatibility)
   - [Container Image](#container-image)
 - [Metrics Documentation](#metrics-documentation)
+- [Metrics Deprecation](#metrics-deprecation)
 - [Kube-state-metrics self metrics](#kube-state-metrics-self-metrics)
 - [Resource recommendation](#resource-recommendation)
 - [kube-state-metrics vs. Heapster](#kube-state-metrics-vs-heapster)
@@ -102,6 +103,18 @@ additional metrics!
 > which may be changed at any given release.
 
 See the [`Documentation`](Documentation) directory for documentation of the exposed metrics.
+
+### Metrics Deprecation
+* **The following non-generic resource metrics for pods are marked deprecated. They will be removed in kube-state-metrics v2.0.0.**
+`kube_pod_container_resource_requests` and `kube_pod_container_resource_limits` are the replacements with `resource` labels
+representing the resource name and `unit` labels representing the resource unit.
+  * kube_pod_container_resource_requests_cpu_cores
+  * kube_pod_container_resource_limits_cpu_cores
+  * kube_pod_container_resource_requests_memory_bytes
+  * kube_pod_container_resource_limits_memory_bytes
+  * kube_pod_container_resource_requests_nvidia_gpu_devices
+  * kube_pod_container_resource_limits_nvidia_gpu_devices
+  
 
 ### Kube-state-metrics self metrics
 kube-state-metrics exposes its own metrics under `--telemetry-host` and `--telemetry-port` (default 81).
--- a/pkg/collectors/collectors.go
+++ b/pkg/collectors/collectors.go
@@ -19,12 +19,16 @@ package collectors
 import (
 	"time"
 
+	"regexp"
+
 	"github.com/prometheus/client_golang/prometheus"
 	"k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/fields"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/kube-state-metrics/pkg/options"
 )
 
 var (
@@ -45,7 +49,31 @@ var (
 		},
 		[]string{"resource"},
 	)
+
+	invalidLabelCharRE = regexp.MustCompile(`[^a-zA-Z0-9_]`)
 )
+
+var AvailableCollectors = map[string]func(registry prometheus.Registerer, kubeClient kubernetes.Interface, namespaces []string, opts *options.Options){
+	"cronjobs":                 RegisterCronJobCollector,
+	"daemonsets":               RegisterDaemonSetCollector,
+	"deployments":              RegisterDeploymentCollector,
+	"jobs":                     RegisterJobCollector,
+	"limitranges":              RegisterLimitRangeCollector,
+	"nodes":                    RegisterNodeCollector,
+	"pods":                     RegisterPodCollector,
+	"replicasets":              RegisterReplicaSetCollector,
+	"replicationcontrollers":   RegisterReplicationControllerCollector,
+	"resourcequotas":           RegisterResourceQuotaCollector,
+	"services":                 RegisterServiceCollector,
+	"statefulsets":             RegisterStatefulSetCollector,
+	"persistentvolumes":        RegisterPersistentVolumeCollector,
+	"persistentvolumeclaims":   RegisterPersistentVolumeClaimCollector,
+	"namespaces":               RegisterNamespaceCollector,
+	"horizontalpodautoscalers": RegisterHorizontalPodAutoScalerCollector,
+	"endpoints":                RegisterEndpointCollector,
+	"secrets":                  RegisterSecretCollector,
+	"configmaps":               RegisterConfigMapCollector,
+}
 
 type SharedInformerList []cache.SharedInformer
 
@@ -87,4 +115,32 @@ func addConditionMetrics(ch chan<- prometheus.Metric, desc *prometheus.Desc, cs 
 		desc, prometheus.GaugeValue, boolFloat64(cs == v1.ConditionUnknown),
 		append(lv, "unknown")...,
 	)
+}
+
+func kubeLabelsToPrometheusLabels(labels map[string]string) ([]string, []string) {
+	labelKeys := make([]string, len(labels))
+	labelValues := make([]string, len(labels))
+	i := 0
+	for k, v := range labels {
+		labelKeys[i] = "label_" + sanitizeLabelName(k)
+		labelValues[i] = v
+		i++
+	}
+	return labelKeys, labelValues
+}
+
+func kubeAnnotationsToPrometheusAnnotations(annotations map[string]string) ([]string, []string) {
+	annotationKeys := make([]string, len(annotations))
+	annotationValues := make([]string, len(annotations))
+	i := 0
+	for k, v := range annotations {
+		annotationKeys[i] = "annotation_" + sanitizeLabelName(k)
+		annotationValues[i] = v
+		i++
+	}
+	return annotationKeys, annotationValues
+}
+
+func sanitizeLabelName(s string) string {
+	return invalidLabelCharRE.ReplaceAllString(s, "_")
 }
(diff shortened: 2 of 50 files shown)


README.md
```diff
@@ -103,6 +104,18 @@ additional metrics!
 
 See the [`Documentation`](Documentation) directory for documentation of the exposed metrics.
 
+### Metrics Deprecation
+* **Pod none generic resource metrics below are marked as deprecated and will be finally removed in kube-state-metrics v1.5.0.**
```
brancz: I don't think this is something we can do. We should definitely mark them as deprecated and create a list of deprecated metrics, but we'll need to cut a 2.0 to drop metrics I think.

pkg/collectors/pod.go
```diff
@@ -256,6 +268,8 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerResourceLimitsNvidiaGPUDevices
```
brancz: the descriptions should be conditional as well

pkg/collectors/pod.go
```diff
@@ -145,6 +144,18 @@ var (
...
+	descPodContainerResourceRequests = prometheus.NewDesc(
+		"kube_pod_container_resource_requests",
+		"The number of requested request resource by a container.",
+		[]string{"namespace", "pod", "container", "node", "resource", "unit"}, nil,
+	)
+
+	descPodContainerResourceLimits = prometheus.NewDesc(
+		"kube_pod_container_resource_limits",
+		"The number of requested limit resource by a container.",
+		[]string{"namespace", "pod", "container", "node", "resource", "unit"}, nil,
```
brancz: I'm not sure the `unit` label will be confusing for use with device plugins as we can't tell the unit. @derekwaynecarr what's your thought on that?

andyxning: @brancz That's right. I have also think about this when prepare this PR but having not found any better names/values.

README.md
andyxning: > but we'll need to cut a 2.0 to drop metrics I think.

What is the requirement for us to prepare the 2.0 release? IMHO, kube-state-metrics has no  significant none compatible changes for now.

pkg/collectors/pod.go
andyxning: Will do.

andyxning: I have tested this. It will not produce any old metrics even we add them to the channel. But i think it will be more clear and safe to make them conditional.

pkg/collectors/pod.go
andyxning: /c @derekwaynecarr @brancz

README.md
brancz: > kube-state-metrics has no significant none compatible changes for now.

Exactly which is why if we do introduce breaking changes, we need to cut a new major version, which I would like to avoid at this point, and keep collecting deprecated metrics for some time until we're comfortable and then cut a new major release dropping all deprecated metrics. For anyone who doesn't want deprecated metrics we can already have a flag to disable them.

pkg/collectors/pod.go
brancz: I talked to @derekwaynecarr and devices have no unit and are always integers, so I think you can decide on either. I can see the unit label being valuable to see that something is indeed a device plugin, but also cpu, memory and storage seems pretty standard so it may not be all that necessary. Up to you either works for me.

andyxning: I suggest we add a `unit` label to help end users learn the unit of the value. I.e., it should be clear for users to learn that memory related limit and request are in `byte` instead of `GB` or `KB`. :)

README.md
andyxning: Whether dropping deprecated metrics should be considered as a none compatible breaking change is fine to me as long as we have state the deprecation clearly and leave a long enough time to let end users learn and prepare for that. So stating that deprecated metrics will be dropped in v2.0.0 release is also fine. Will do.

andyxning:
Comments addressed. @brancz @derekwaynecarr PTAL.

pkg/collectors/collectors.go
```diff
@@ -45,8 +49,32 @@ var (
...
+	"nodes":                    RegisterNodeCollector,
+	"pods":                     RegisterPodCollector,
+	"replicasets":              RegisterReplicaSetCollector,
+	"replicationcontrollers":   RegisterReplicationControllerCollector,
+	"resourcequotas":           RegisterResourceQuotaCollector,
+	"services":                 RegisterServiceCollector,
+	"statefulsets":             RegisterStatefulSetCollector,
+	"persistentvolumes":        RegisterPersistentVolumeCollector,
+	"persistentvolumeclaims":   RegisterPersistentVolumeClaimCollector,
+	"Namespaces":               RegisterNamespaceCollector,
```
brancz: this should be lower case

README.md
```diff
@@ -103,6 +104,18 @@ additional metrics!
 
 See the [`Documentation`](Documentation) directory for documentation of the exposed metrics.
 
+### Metrics Deprecation
+* **Pod none generic resource metrics below are marked as deprecated and will be finally removed in kube-state-metrics v2.0.0.**
```
derekwaynecarr: nit: this reads awkwardly.

"The following non-generic resource metrics for pods are marked deprecated.  They will be removed in kube-state-metrics v2.0.0"

pkg/collectors/pod.go
derekwaynecarr: for reference:
https://github.com/kubernetes/kubernetes/blob/8f20a815e5ef27d4d247e42bff90e42706397ca2/pkg/apis/core/helper/helpers.go#L263

validates that any extended resource is integral.

derekwaynecarr: ephemeral storage is bytes as well.

pkg/collectors/pod.go
```diff
@@ -429,35 +418,84 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
...
-			addGauge(descPodContainerResourceLimitsMemoryBytes, float64(mem.Value()),
-				c.Name, nodeName)
+		for resourceName, val := range req {
+			switch resourceName {
+			case v1.ResourceCPU:
+				addGauge(descPodContainerResourceRequests, float64(val.MilliValue())/1000,
+					c.Name, nodeName, sanitizeLabelName(string(resourceName)), "core")
+			case v1.ResourceStorage:
+				fallthrough
+			case v1.ResourceEphemeralStorage:
```
derekwaynecarr: this won't label `hugepages-*` properly as bytes.  any resource name that starts with `hugepages-` is measured in bytes.

pkg/options/options.go
```diff
@@ -24,26 +24,27 @@ import (
...
+	Kubeconfig                           string
+	Help                                 bool
+	Port                                 int
+	Host                                 string
+	TelemetryPort                        int
+	TelemetryHost                        string
+	Collectors                           CollectorSet
+	Namespaces                           NamespaceList
+	Version                              bool
+	DisablePodNoneGenericResourceMetrics bool
```
derekwaynecarr: Should this be `DisablePodNonGenericResourceMetrics`?

nit: it may be easier to think about this in the affirmative, i.e. "EnablePodSpecificResourceMetrics"

pkg/options/options.go
```diff
@@ -66,13 +67,14 @@ func (o *options) AddFlags() {
 	o.flags.Var(&o.Collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &DefaultCollectors))
 	o.flags.Var(&o.Namespaces, "namespace", fmt.Sprintf("Comma-separated list of namespaces to be enabled. Defaults to %q", &DefaultNamespaces))
 	o.flags.BoolVarP(&o.Version, "version", "", false, "kube-state-metrics build version information")
+	o.flags.BoolVarP(&o.DisablePodNoneGenericResourceMetrics, "disable-pod-none-generic-resource-metrics", "", false, "Disable pod none generic resource request and limit metrics")
```
derekwaynecarr: same comment here: disable-pod-non-generic-resource-metrics

README.md
andyxning: Thanks for pointing this out. English writing skill should be enhanced. :-)

pkg/collectors/pod.go
andyxning: @derekwaynecarr 
> ephemeral storage is bytes as well.

This has been the truth as when the resource name is `ResourceEphemeralStorage` and `ResourceStorage`, it will `fallthrough` to `ResourceMemory`.

> this won't label hugepages-* properly as bytes. any resource name that starts with hugepages- is measured in bytes.

I have actually not aware about this. Will do.

pkg/options/options.go
andyxning: > Should this be DisablePodNonGenericResourceMetrics?

Yes. typos. Will update it.

> it may be easier to think about this in the affirmative, i.e. "EnablePodSpecificResourceMetrics"

Either works for me. Will listen for other reviewers comments about this. :)

pkg/collectors/pod.go
andyxning: > cpu is millicores

The unit for cpu limit and resource has been converted from millicores to cores by devided by 1000. So maybe cpu is cores. WDYT?

@derekwaynecarr

pkg/collectors/pod.go
```diff
@@ -429,35 +418,84 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
...
+				addGauge(descPodContainerResourceLimits, float64(val.MilliValue())/1000,
+					c.Name, nodeName, sanitizeLabelName(string(resourceName)), "core")
+			case v1.ResourceStorage:
+				fallthrough
+			case v1.ResourceEphemeralStorage:
+				fallthrough
+			case v1.ResourceMemory:
+				addGauge(descPodContainerResourceLimits, float64(val.Value()),
+					c.Name, nodeName, sanitizeLabelName(string(resourceName)), "byte")
+			case v1.ResourceNvidiaGPU:
```
vishh: This resource is now deprecated.

andyxning: Yes. We have not update the client-go and we have already make an agreement on deprecating this once Kubernetes 1.11 is released and the client-go dependency is updated.

derekwaynecarr:
Can we make a similar change for node as well?

andyxning:
> Can we make a similar change for node as well?

Yeah. I will make another PR to do this.

pkg/collectors/pod.go
andyxning: > all extended resources (device plugin or otherwise in a qualified namespace) are integral

IIUC, you mean that we should
* check whether a resource name is an extended resource
* unit for extended resource should be integer as it is must be an integer

andyxning: kube-state-metrics should not need to check whether the value is integer as this has been done in api-server.

@derekwaynecarr

andyxning:
@derekwaynecarr comments addressed. PTAL.

/cc @brancz @vishh

PR #466, opened 2018-05-28
add generic node resource capacity and allocatable

andyxning:
* Adds `disable-node-none-generic-resource-metrics` command line argument to kube-state-metrics to disable old not labeled resource metrics.
* Adds generic new `kube_node_status_capacity` and `kube_node_status_allocatable`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,7 +114,17 @@ representing the resource name and `unit` labels representing the resource unit.
   * kube_pod_container_resource_limits_memory_bytes
   * kube_pod_container_resource_requests_nvidia_gpu_devices
   * kube_pod_container_resource_limits_nvidia_gpu_devices
-  
+* **The following non-generic resource metrics for nodes are marked deprecated. They will be removed in kube-state-metrics v2.0.0.**
+`kube_node_status_capacity` and `kube_node_status_allocatable` are the replacements with `resource` labels
+representing the resource name and `unit` labels representing the resource unit.
+  * kube_node_status_capacity_pods
+  * kube_node_status_capacity_cpu_cores
+  * kube_node_status_capacity_nvidia_gpu_cards
+  * kube_node_status_capacity_memory_bytes
+  * kube_node_status_allocatable_pods
+  * kube_node_status_allocatable_cpu_cores
+  * kube_node_status_allocatable_nvidia_gpu_cards
+  * kube_node_status_allocatable_memory_bytes
 
 ### Kube-state-metrics self metrics
 kube-state-metrics exposes its own metrics under `--telemetry-host` and `--telemetry-port` (default 81).
--- a/pkg/collectors/node.go
+++ b/pkg/collectors/node.go
@@ -23,7 +23,9 @@ import (
 	"k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/client-go/kubernetes"
+	"k8s.io/kube-state-metrics/pkg/constant"
 	"k8s.io/kube-state-metrics/pkg/options"
+	"k8s.io/kubernetes/pkg/apis/core/v1/helper"
 )
 
 var (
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -114,7 +114,17 @@ representing the resource name and `unit` labels representing the resource unit.
   * kube_pod_container_resource_limits_memory_bytes
   * kube_pod_container_resource_requests_nvidia_gpu_devices
   * kube_pod_container_resource_limits_nvidia_gpu_devices
-  
+* **The following non-generic resource metrics for nodes are marked deprecated. They will be removed in kube-state-metrics v2.0.0.**
```
brancz: I think it would be good if we document with which flags one can disable all deprecated metrics today, to already have the chance to adapt everything to kube-state-metrics v2.0 state.

pkg/collectors/node.go
```diff
@@ -252,19 +277,74 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n v1.Node) {
...
+			addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitByte))
+		case v1.ResourceNvidiaGPU:
+			addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitInteger))
+		case v1.ResourcePods:
+			addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitInteger))
+		default:
+			if helper.IsHugePageResourceName(resourceName) {
+				addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitByte))
+			}
+			if helper.IsExtendedResourceName(resourceName) {
```
brancz: I'm not sure I know enough to tell whether these two cases under the `default` block are enough. Is it possible that resources don't fit either of these helper conditions?

@derekwaynecarr can you verify that each of these cases are handled appropriately?

andyxning: There should be no false positive. There are at most some node level resources will not be collected.

README.md
andyxning: Yep. Considering more and more metrics maybe deprecated, maybe we should make a decision about whether we need to add:
* separate knob
* global knob
to control the whether to scrape the old deprecated metrics.

The first option will give end users more fine control on what deprecated metrics to disable and what deprecated metrics should still be collected.
The second option will make the deprecation of all old metrics more easily. However, the end user control requirement above maybe not satisfied.

brancz: Let's keep separate flags for each until someone complains, and generally as long as we document all of them it's fine in my opinion. It easily escalates when there is more than one way to do things.

pkg/collectors/node.go
andyxning: /cc @derekwaynecarr PTAL.

andyxning: /cc @derekwaynecarr

pkg/collectors/node.go
```diff
@@ -252,19 +277,74 @@ func (nc *nodeCollector) collectNode(ch chan<- prometheus.Metric, n v1.Node) {
...
+		switch resourceName {
+		case v1.ResourceCPU:
+			addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitCore))
+		case v1.ResourceStorage:
+			fallthrough
+		case v1.ResourceEphemeralStorage:
+			fallthrough
+		case v1.ResourceMemory:
+			addGauge(descNodeStatusCapacity, float64(val.MilliValue())/1000, sanitizeLabelName(string(resourceName)), string(constant.UnitByte))
+		case v1.ResourceNvidiaGPU:
```
derekwaynecarr: is there a tracking issue to remove this resource in a follow-on?

pkg/collectors/node.go
derekwaynecarr: in kube 1.11, there is a new type of resource that advertises attachable volume limits.

see: https://github.com/kubernetes/kubernetes/pull/64154

making this discoverable is important to monitoring for exhaustion.

it's alpha so we can do it in a follow-on.

pkg/collectors/node.go
andyxning: There is no tracking issue. But we have make a agreement that once Kubernetes 1.11 is released it will be removed in #417

PR #471, opened 2018-05-31
Dockerfile: Remove `VOLUME` directive.

brancz:
The `VOLUME` directive causes Docker to provision host storage to cover
the implicitly requested volume mount. For security reasons this may be
disallowed by Kubernetes clusters, causing kube-state-metrics containers
to fail to be created.

Besides that, this directive does not do anything and to my knowledge
never did.

@andyxning 

cc @ironcladlou

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,7 +1,6 @@
 FROM alpine:3.7
 
 COPY kube-state-metrics /
-VOLUME /tmp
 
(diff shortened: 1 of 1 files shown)


andyxning:
> Besides that, this directive does not do anything and to my knowledge
never did.

This also seems right to me.

/lgtm

PR #476, opened 2018-06-08
avoid dirty hack to prometheus client

andyxning:
This PR avoids dirty code change hack to prometheus client.

### DIFF
--- a/pkg/collectors/pod_test.go
+++ b/pkg/collectors/pod_test.go
@@ -442,6 +442,10 @@ func TestPodCollector(t *testing.T) {
 						PodIP:  "2.3.4.5",
 						ContainerStatuses: []v1.ContainerStatus{
 							v1.ContainerStatus{
+								Name:        "container2_1",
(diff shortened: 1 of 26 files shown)


pkg/collectors/pod_test.go
```diff
@@ -157,93 +157,93 @@ func TestPodCollector(t *testing.T) {
 			pods: []v1.Pod{
 				{
 					ObjectMeta: metav1.ObjectMeta{
-						Name:      "pod1",
-						Namespace: "ns1",
+						Name:      "pod3",
```
brancz: The fact that we need to do this to make the tests pass sounds like there is some global/cross-test state that should not be. Can you spot where this is coming from? I would rather get rid of that global state and not do all of these changes (if that's the case).

andyxning: Actually, we need only three changes to one of the pod unit test cases.

andyxning: /cc @brancz

PR #482, opened 2018-06-24
adjust client-go User-Agent

andyxning:
This RP will customize User-Agent client-go uses. This will make the User-Agent more useful and be in more Kubernetes's format:
`kube-state-metrics/v1.3.0 (linux/amd64) kube-state-metrics/64fdc24`

Fixes #

### DIFF
--- a/main.go
+++ b/main.go
@@ -116,6 +116,7 @@ func createKubeClient(apiserver string, kubeconfig string) (clientset.Interface,
 		return nil, err
 	}
 
+	config.UserAgent = version.GetVersion().String()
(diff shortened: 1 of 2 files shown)


main.go
```diff
@@ -116,6 +116,7 @@ func createKubeClient(apiserver string, kubeconfig string) (clientset.Interface,
 		return nil, err
 	}
 
+	config.UserAgent = fmt.Sprintf("%s", version.GetVersion())
```
brancz: `fmt.Sprintf` is not necessary here, `version.GetVersion()` already returns a string

andyxning: Good catch. Will do.

PR #488, opened 2018-07-06
Allow white- and blacklisting metrics to be exposed

brancz:

In large Kubernetes clusters single metrics can have a significant effect on the total number of metrics exposed by a single scrape. This PR allows white or blacklisting metrics to be exposed.

@andyxning

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,7 @@
 
 * [CHANGE] `kube_job_status_start_time` and `kube_job_status_completion_time` metric types changed from counter to gauge.
 * [CHANGE] `job` label to `job_name` as this collides with the Prometheus `job` label.
+* [FEATURE] Allow white- and black-listing metrics to be exposed.
 
 ## v1.3.1 / 2018-04-12
 
--- a/main.go
+++ b/main.go
@@ -31,10 +31,11 @@ import (
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	clientset "k8s.io/client-go/kubernetes"
+	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	"k8s.io/client-go/tools/clientcmd"
 
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
 	kcollectors "k8s.io/kube-state-metrics/pkg/collectors"
+	"k8s.io/kube-state-metrics/pkg/metrics"
 	"k8s.io/kube-state-metrics/pkg/options"
 	"k8s.io/kube-state-metrics/pkg/version"
 )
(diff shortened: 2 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,7 @@
 ## Unreleased
 
 * [CHANGE] `kube_job_status_start_time` and `kube_job_status_completion_time` metric types changed from counter to gauge.
+* [FEATURE] Allow white- and blacklisting metrics to be exposed.
```
andyxning: `s/white- and blacklisting/white- and black-listing`

main.go
```diff
@@ -216,3 +227,46 @@ func registerCollectors(registry prometheus.Registerer, kubeClient clientset.Int
...
+func (f GathererFunc) Gather() ([]*dto.MetricFamily, error) {
+	return f()
+}
+
+func filteredGatherer(r prometheus.Gatherer, whitelist options.MetricSet, blacklist options.MetricSet) prometheus.Gatherer {
+	whitelistEnabled := !whitelist.IsEmpty()
+	blacklistEnabled := !blacklist.IsEmpty()
+	filterEnabled := whitelistEnabled || blacklistEnabled
+
+	return GathererFunc(func() ([]*dto.MetricFamily, error) {
```
andyxning: I suggest we make this function a package-level function thus we can add some unit tests for it. We need to determine the relationship between whitelist and blacklist:
* Does we permit users specify both whitelist and blacklist
* Does kube-state-metrics should only prefer whitelist and ignore blacklist even blacklist is not empty. 

@brancz 

Currently, the implementation is kube-state-metrics prefer whitelist metrics even blacklist is not empty which i think is not the right logic. :)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+func (f GathererFunc) Gather() ([]*dto.MetricFamily, error) {
+	return f()
+}
+
+func FilteredGatherer(r prometheus.Gatherer, whitelist options.MetricSet, blacklist options.MetricSet) prometheus.Gatherer {
+	whitelistEnabled := !whitelist.IsEmpty()
+	blacklistEnabled := !blacklist.IsEmpty()
+	filterEnabled := whitelistEnabled || blacklistEnabled
+
+	return GathererFunc(func() ([]*dto.MetricFamily, error) {
```
andyxning: I suggest we split the `GathererFunc` into a inter-package level function, just like `FilteredGatherer` And add some ut for this.

andyxning: @brancz Once you have time, could you please update this PR? :)

pkg/metrics/metrics.go
```diff
@@ -0,0 +1,67 @@
...
+				name := *metricFamily.Name
+				if whitelistEnabled {
+					_, onWhitelist := whitelist[name]
+					if onWhitelist {
+						newMetricFamilies = append(newMetricFamilies, metricFamily)
+					}
+					continue
+				}
+
+				_, onBlacklist := blacklist[name]
```
dohnto: Why do you check whether `whitelistEnabled` is true, but you **don't** check `blacklistEnabled` here.

brancz: We have to opt for one strategy here. Both the whitelist and blacklist are explicit, so which one is the right one, the one where we explicitly said we definitely want this, or the one where we said we definitely don't want this. As whitelists are usually pretty hand picked I figured that would be the stronger of the two.

pkg/metrics/metrics.go
brancz: Can you explain what you meant me to do with `GathererFunc`? It's just so a function satisfies the `prometheus.Gatherer` interface.

pkg/metrics/metrics.go
brancz: I'm going to make sure to document he precedence behavior.

brancz: I changed blacklist and whitelist to be mutually exclusive. It really doesn't make any sense to have both here.

brancz:
This is ready for a re-review @andyxning @dohnto.

pkg/options/options.go
```diff
@@ -69,6 +73,8 @@ func (o *Options) AddFlags() {
 	o.flags.StringVar(&o.TelemetryHost, "telemetry-host", "0.0.0.0", `Host to expose kube-state-metrics self metrics on.`)
 	o.flags.Var(&o.Collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &DefaultCollectors))
 	o.flags.Var(&o.Namespaces, "namespace", fmt.Sprintf("Comma-separated list of namespaces to be enabled. Defaults to %q", &DefaultNamespaces))
+	o.flags.Var(&o.MetricWhitelist, "metric-whitelist", "Comma-separated list of metrics to be exposed. The whitelist and blacklist are mutually exclusive.")
```
andyxning: We should state this clearly. !

PR #506, opened 2018-08-11
Added init containers pod metrics.

### LINKED ISSUE #502: kube state metrics doesn't report init container pod metrics

/kind feature

**What happened**:

**Environment**:
- All Kubernetes version

dohnto:

Fixes #502

### DIFF
--- a/pkg/collectors/pod.go
+++ b/pkg/collectors/pod.go
@@ -379,8 +379,9 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 	}
 
 	var lastFinishTime float64
+	allContainerStatuses := append(p.Status.ContainerStatuses, p.Status.InitContainerStatuses...)
 
-	for _, cs := range p.Status.ContainerStatuses {
+	for _, cs := range allContainerStatuses {
 		addGauge(descPodContainerInfo, 1,
 			cs.Name, cs.Image, cs.ImageID, cs.ContainerID,
 		)
--- a/pkg/collectors/pod_test.go
+++ b/pkg/collectors/pod_test.go
@@ -140,6 +140,14 @@ func TestPodCollector(t *testing.T) {
 								ContainerID: "docker://ef789",
 							},
 						},
+						InitContainerStatuses: []v1.ContainerStatus{
+							v1.ContainerStatus{
+								Name:        "container4",
+								Image:       "k8s.gcr.io/hyperkube4",
+								ImageID:     "docker://sha256:ddd",
+								ContainerID: "docker://gh012",
+							},
+						},
 					},
 				},
 			},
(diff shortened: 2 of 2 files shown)


zouyee:
Why do we need  the init container metric that is a temporary object compare with container？

thedebugger:
@zouyee the current metric is inadequate to reason about pod phase. In our cluster, pods were in waiting phase when init containers were failing. But I do agree it seems excessive to send init metrics everytime because they are short-lived. Let me know if you have suggestions to optimize it.

I'm thinking we should creating the following metrics for init containers --

* kube_pod_init_container_info
* kube_pod_container_status_running
* kube_pod_init_container_status_waiting
* kube_pod_init_container_status_waiting_reason
* kube_pod_init_container_status_restarts_total
* kube_pod_init_container_status_terminated
* kube_pod_init_container_status_terminated_reason

Ready state is not useful for init container[2]. In principle, we should also add resource request and limit, but they can be added later. Let me know what you all think. Thanks!

References
1. https://kubernetes.io/docs/concepts/workloads/pods/init-containers/#detailed-behavior
2. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.11/#pod-v1-core

andyxning:
> But I do agree it seems excessive to send init metrics everytime because they are short-lived.

Having not recognized that init containers may be short-lived and it may even be done in a scrape interval. Thus no init container metrics may be emitted.

andyxning:
> the current metric is inadequate to reason about pod phase. In our cluster, pods were in waiting phase when init containers were failing.

Maybe in your case, we need to adjust pod phase metrics when init containers are failing. This may not cover all possible reasons that result in a pod waiting phase. But, imo, it is hard to enumerate all possible situations that can result in a pod waiting phase.

thedebugger:
> Having not recognized that init containers may be short-lived and it may even be done in a scrape interval. Thus no init container metrics may be emitted.

Hmm, we may not able to be able to catpure init container state transitions, but don't InitContainerStatus of a pod provide us the information to emit the metrics we need?

 >Maybe in your case, we need to adjust pod phase metrics when init containers are failing. This may not cover all possible reasons that result in a pod waiting phase. But, imo, it is hard to enumerate all possible situations that can result in a pod waiting phase.

Do you have specifics in mind that we can do to adjust pod phase metrics when init containers are failing? I agree it is hard. That is why, imo, it might be better to provide the same information as (or stay close to) kubernetes API. That would be error prone and path of "least surprise".

andyxning:
> Hmm, we may not able to be able to catpure init container state transitions, but don't InitContainerStatus of a pod provide us the information to emit the metrics we need?

That's a good suggestion. We need to make sure that we can get the status of init containers from pod status even init containers have been stopped. Any clue on this? @thedebugger

thedebugger:
@andyxning [InitContainerStatuses](https://github.com/kubernetes/kube-state-metrics/blob/96c8688fc87c6abfabf6e89dc87ea6ca41b5b968/vendor/k8s.io/api/core/v1/types.go#L3035) is an attribute in pod status that should tell us the final state of all init containers if that is what you are asking. That is what @dohnto is using in this PR

andyxning:
Feel free to ping me once the PR is ready to review. :) @dohnto @thedebugger

thedebugger:
@andyxning https://github.com/kubernetes/kube-state-metrics/pull/541

dohnto:
Closing in favour of #541

PR #509, opened 2018-08-13
Update release documentation

loburm:
Adds information related to staging container registry that should be used for release candidate images.


brancz:
We don’t usually change the image in the deployment manifest when we do release candidates.

andyxning:
Then we should document the latest release candidates to let end users know and give it a try.

brancz:
Yes we can do that.

PR #516, opened 2018-08-17
Update CONTRIBUTING.md

wangxy518:
the url has been changed.


zouyee:
`http://git.k8s.io/` is k8s community domain which redirect to github.

andyxning:
@wangxy518 Just like what @zouyee said, it is right for now to use `git.k8s.io` and it will redirect to `https://github.com/kubernetes/`.

PR #520, opened 2018-08-20
add kube_replicatset_owner metric

### LINKED ISSUE #510: There should be a way to get the Deployment of a ReplicaSet
/kind feature

I was looking for a way to connect Pods to their Deployment. I can use the `kube_pod_owner` metric to connect Pods to their owning ReplicaSet. But, I do not see a way to go from a ReplicaSet to its Deployment.

I'm not an expert in their Prometheus or Kubernetes, but I think the best way to do this would be to add a `kube_replicaset_owner` metric similar to `kube_pod_owner`, because that would be consistent with the existing design.

**Environment**:
- Kubernetes version (use `kubectl version`): client v1.9.7, server v1.9.6-gke.2
- Kube-state-metrics image version: quay.io/coreos/kube-state-metrics:v1.3.0

andyxning:
This PR adds `kube_replicaset_owner` metric.
Fixes #510


pkg/collectors/replicaset.go
```diff
@@ -107,19 +115,20 @@ type replicasetCollector struct {
...
 	ch <- descReplicaSetStatusObservedGeneration
 	ch <- descReplicaSetSpecReplicas
 	ch <- descReplicaSetMetadataGeneration
+	ch <- descReplicaSetOwner
 }
 
 // Collect implements the prometheus.Collector interface.
-func (dc *replicasetCollector) Collect(ch chan<- prometheus.Metric) {
-	rss, err := dc.store.List()
+func (rs *replicasetCollector) Collect(ch chan<- prometheus.Metric) {
```
mxinden: Small nit, looking at the other controllers this should be called `rc` or `rsc`, right?

andyxning: Good catch. `rsc` should be fine.

andyxning: Done. PTAL. @mxinden

brancz:
This lgtm, but putting a hold on this until we lift the feature freeze.

/hold
/lgtm

Issue #498, opened 2018-07-23
Performance Obtimization Proposal


mxinden:
/kind improvement

I have been brainstorming with @brancz on how we can improve kube-state-metrics' performance  in terms of response time and memory usage.

I am welcoming any feedback on this [Optimization Proposal](https://docs.google.com/document/d/1Z1q84680vG56SKDi1AMu0x3lFe4bi40DcpMCQDylyMg/edit?usp=sharing).

Related issues: https://github.com/kubernetes/kube-state-metrics/issues/493 https://github.com/kubernetes/kube-state-metrics/issues/257

andyxning:
Great!.
We should work with developers focus on client-go to work this out. I am a little busy recently. But i am willing give some help or ideas if possible.

I am willing to seem this happen in v1.5.0. :)

andyxning:
> by removing the informer cache, which we are currently building metrics from, basically with a cache of our own

IIUC, informer cache is inside client-go. @brancz

brancz:
correct, but the reflector is also in client-go and is independently usable, it just won't write into the informer cache, but directly into the metrics cache

andyxning:
> but the reflector is also in client-go and is independently usable, it just won't write into the informer cache, but directly into the metrics cache

This indicates that we need to change client-go or add metrics cache plugin to replace the informer cache, right?

brancz:
> This indicates that we need to change client-go or add metrics cache plugin to replace the informer cache, right?

No other program will ever need what we intend to build, and client-go's reflector is already pluggable :slightly_smiling_face: , so I don't think there is anything to do in client-go for this concrete proposal. I think there are valid things that we can look at separately to improve client-go but they don't directly influence the proposal.

> Is there somewhere to think about adding some kind of horizontal scaling?

I think this is a valid point, although orthogonal to the problems the proposal addresses. I would like to investigate whether we would fulfill the Kubernetes scalability requirements simply with functional sharding, basically by having one kube-state-metrics with all cluster-wide collectors enabled and one kube-state-metrics per namespace/tenant (maybe multiple namespaces). If that doesn't suffice we could look into (consistent) hashing of object UUIDs to distribute them among kube-state-metrics instances. Any further optimizations would require changes to the Kubernetes API to support sharding list/watch somehow.

I would do improvements roughly in the order I laid out, starting with this proposal.

directionless:
[...]

PR #534, opened 2018-09-03
 pkg/collectors: Introduce MetricsStore updated by Reflectors

mxinden:

As discussed in #498 kube-state-metrics faces the following two performance
issues when deployed in bigger Kubernetes environments:

1. Long response time on metrics scrape
2. High memory usage

This patch introduces a custom client go store, MetricsStore. It is updated by a
corresponding reflector. Instead of generating the Prometheus custom metrics on
demand, metrics are generated on new Kubernetes object arrival.

This pull request represents an early *work-in-progress* state to enable early
feedback. It targets the *performance-optimizations* branch, instead of *master*.

Known ToDos:

- Re-enable multi namespace support
- Re-enable white/black listing
- Adjust remaining `collectors/*_test.go`

**Premature benchmarking results**:

1000 configmaps, 1000 pods, 100 scrapes:

```shell
$ git checkout master
$ go test -bench .                    
goos: linux
goarch: amd64
pkg: k8s.io/kube-state-metrics
BenchmarkKubeStateMetrics-4            1        10267655552 ns/op
PASS
ok      k8s.io/kube-state-metrics       10.282s

```

[master_cpu.pdf](https://github.com/kubernetes/kube-state-metrics/files/2345809/master_cpu.pdf)
[master_mem.pdf](https://github.com/kubernetes/kube-state-metrics/files/2345810/master_mem.pdf)

```shell
$ git checkout metricsstore
$ go test -bench .
goos: linux
goarch: amd64
pkg: k8s.io/kube-state-metrics
BenchmarkKubeStateMetrics-4            1        1876562818 ns/op
--- BENCH: BenchmarkKubeStateMetrics-4
        main_test.go:39: starting kube-state-metrics benchmark with fixtureMultiplier 1000 and requestCount 100
PASS
ok      k8s.io/kube-state-metrics       1.892s
```

[metricsstore_cpu.pdf](https://github.com/kubernetes/kube-state-metrics/files/2345825/metricsstore_cpu.pdf)
[metricsstore_mem.pdf](https://github.com/kubernetes/kube-state-metrics/files/2345826/metricsstore_mem.pdf)

Relates to https://github.com/kubernetes/kube-state-metrics/issues/498

//CC @metalmatze @brancz @squat @s-urbaniak


pkg/collectors/collectors.go
```diff
@@ -50,28 +55,116 @@ var (
...
+	return collectors, nil
+}
+
+type Collector interface {
+	createListWatch(clientset.Interface) cache.ListWatch
+	generateMetrics(obj interface{}) []*metrics.Metric
+	expectedType() interface{}
+	setStore(*metricsstore.MetricsStore)
+	Collect() []*metrics.Metric
+}
```
s-urbaniak: nit: this public interface with both public and private methods looks iffy. Do you really need it that way?

pkg/collectors/collectors.go
```diff
@@ -50,28 +55,116 @@ var (
...
+
+// TODO: rename to collectorConstructor ?
+type newCollector func(*options.Options) Collector
```
s-urbaniak: question: why do you need to abstract this?

pkg/collectors/collectors.go
```diff
@@ -50,28 +55,116 @@ var (
...
+	opts *options.Options,
+) collectorFactory {
+
+	return collectorFactory{
+		kubeClient,
+		opts,
+	}
 }
 
+// TODO: Is the error return necessary?
```
s-urbaniak: the body doesn't return any error, hence no, I guess ;-)

pkg/collectors/collectors.go
```diff
@@ -90,19 +183,25 @@ func boolFloat64(b bool) float64 {
...
-	ch <- prometheus.MustNewConstMetric(
-		desc, prometheus.GaugeValue, boolFloat64(cs == v1.ConditionUnknown),
-		append(lv, "unknown")...,
-	)
+func addConditionMetrics(desc *metricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
+	ms := []*metrics.Metric{}
+	m, err := metrics.NewMetric(desc.Name, desc.LabelKeys, append(lv, "true"), boolFloat64(cs == v1.ConditionTrue))
+	if err != nil {
+		panic(err)
+	}
```
s-urbaniak: nit, maybe wrap this in a `mustNewMetric`

pkg/metrics/metrics.go
```diff
@@ -17,12 +17,62 @@ limitations under the License.
...
+
+	m := ""
+
+	m = m + name
+
+	m = m + labelsToString(labelKeys, labelValues)
+
+	m = m + fmt.Sprintf(" %v", value)
+
+	m = m + "\n"
```
s-urbaniak: nit: just create one `m := name + labelsToString(...) + fmt.Sprintf(" %g", value)` line.
note the `%g` suggestion which is the default for `%v` in conjunction with floats and seems to be more explicit.

brancz:
cc @andyxning @zouyee

main_test.go
```diff
@@ -17,6 +17,8 @@ limitations under the License.
 package main
 
 import (
+	// "fmt"
+	// "io/ioutil"
```
metalmatze: You should be able to remove these.

main.go
```diff
@@ -217,26 +219,17 @@ func metricsServer(registry prometheus.Gatherer, host string, port int) {
...
-			f(registry, informerFactories, opts)
-			activeCollectors = append(activeCollectors, c)
+type metricHandler struct {
+	c []kcollectors.Collector
+}
+
+func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	for _, c := range m.c {
+		for _, m := range c.Collect() {
+			_, err := fmt.Fprint(w, *m)
```
brancz: I'm certain that we can still get significant speed up if we group all metrics from one object in a buffer and then just call `io.CopyBuffer` (or just `copy`) with a large enough buffer then we also don't waste any allocations here which look like they should be taking a lot of time themselves here.

pkg/metrics/metrics.go
```diff
@@ -17,12 +17,62 @@ limitations under the License.
...
 	dto "github.com/prometheus/client_model/go"
 
 	"k8s.io/kube-state-metrics/pkg/options"
 )
 
+// Metric represents a single line entry in the /metrics export format
+type Metric string
+
+// NewMetric returns a new Metric
+func NewMetric(name string, labelKeys []string, labelValues []string, value float64) (*Metric, error) {
```
brancz: I'm not sure how performance critical this end is as it just fills the cache, but the size of the buffer to write these bytes into should be calculatable upfront and then all bytes just directly written into it. But I think on this one we should first figure out whether this is actually a problem. To do that we should add some metrics on the metrics generation.

pkg/metrics_store/metrics_store.go
```diff
@@ -0,0 +1,110 @@
...
+// Implementing k8s.io/kubernetes/client-go/tools/cache.Store interface
+
+func (s *MetricsStore) Add(obj interface{}) error {
+	o, err := meta.Accessor(obj)
+	if err != nil {
+		return err
+	}
+
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
```
brancz: We should take some lock profiles of this on a large cluster, this looks like there is going to be some contention.

andyxning:
@mxinden Thanks for the good job!. Could you please also add a performance enhance arch graph to demonstrate your thoughts? Actually, i am a little bit confused about the scheme.

mxinden:
As discussed in the sig-instrumentation meeting today, a docker image of the current state of this PR can be found here: quay.io/mxinden/kube-state-metrics:v1.4.0-perf.1

I will push new images while working further on this, incrementing the semver pre-release version.

mxinden:
> Could you please also add a performance enhance arch graph to demonstrate your thoughts?

@andyxning do mean something like a [sequence-diagram](https://en.wikipedia.org/wiki/Sequence_diagram)? More than happy to prepare something like that. Good idea.

pkg/collectors/collectors.go
```diff
@@ -50,28 +55,116 @@ var (
...
+		kubeClient,
+		opts,
+	}
 }
 
+// TODO: Is the error return necessary?
+func (f *collectorFactory) create(constructor newCollector) (Collector, error) {
+	collector := constructor(f.opts)
+
+	// TODO: We probably need one list watch per namespace?
```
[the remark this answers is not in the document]
zouyee: using graceful turn off?

pkg/collectors/collectors.go
```diff
@@ -50,28 +55,116 @@ var (
...
+	store := metricsstore.NewMetricsStore(collector.generateMetrics)
+	collector.setStore(store)
+
+	lw := collector.createListWatch(f.kubeClient)
+
+	expectedType := collector.expectedType()
+	reflector := cache.NewReflector(&lw, &expectedType, store, 0)
+
+	// TODO: Just a background context?
+	go reflector.Run(context.Background().Done())
```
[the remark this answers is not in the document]
zouyee: using graceful turn off?

mxinden:
@andyxning here is a high level sequence diagram. It omits many details, e.g. cardinality introduced through other Kubernetes objects than *pods*, *reflectors* per *namespace*, ... . Let me know what you think.

```
       +---------------+         +-----------+               +---------------+         +-------------------+
       | pod_reflector |         | pod_store |               | pod_collector |         | metrics_endpoint  |
       +---------------+         +-----------+               +---------------+         +-------------------+
-------------\ |                       |                             |                           |
| new pod p1 |-|                       |                             |                           |
|------------| |                       |                             |                           |
               |                       |                             |                           |
               | Add(p1)               |                             |                           |
               |---------------------->|                             |                           |
               |                       |                             |                           |
               |                       | generateMetrics(p1)         |                           |
               |                       |---------------------------->|                           |
               |                       |                             |                           |
               |                       |                   string(m) |                           |
               |                       |<----------------------------|                           |
               |---------------------\ |                             |                           |
               || store[p1.Name] = m |-|                             |                           |
               ||--------------------| |                             |                           |
               |                       |                             |                           |
               |                   nil |                             |                           |
               |<----------------------|                             |                           |
               |                       |                             |                           | ---------------\
               |                       |                             |                           |-| GET /metrics |
               |                       |                             |                           | |--------------|
               |                       |                             |                           |
               |                       |                             |                 Collect() |
               |                       |                             |<--------------------------|
               |                       |                             |                           |
               |                       |               ListMetrics() |                           |
               |                       |<----------------------------|                           |
               |                       |                             |                           |
               |                       | []string{metrics}           |                           |
               |                       |---------------------------->|                           |
               |                       |                             |                           |
               |                       |                             | concat(metrics)           |
               |                       |                             |-------------------------->|
               |                       |                             |                           |
```


<details>
 <summary>Code to reproduce diagram</summary>

Build via [text-diagram](http://weidagang.github.io/text-diagram/)

```
object pod_reflector pod_store pod_collector metrics_endpoint

note left of pod_reflector: new pod p1
pod_reflector -> pod_store: Add(p1)
pod_store -> pod_collector: generateMetrics(p1)
pod_collector -> pod_store: string(metrics)
pod_store -> pod_reflector: nil

note right of metrics_endpoint: GET /metrics
metrics_endpoint -> pod_collector: Collect()
pod_collector -> pod_store: ListMetrics()
pod_store -> pod_collector: []string{metrics}
pod_collector -> metrics_endpoint: concat(metrics)
```

</details>

brancz:
@mxinden can you add that to the docs under a new folder called “design” (ideally with the command to reproduce in a markdown comment :) )

mxinden:
@brancz good idea. I added `Documentation/design/metrics-store-performance-optimization.md` with both the diagram as well as the initial proposal from Google docs.

---

In regards to the CLA issue posted above by k8s-ci-robot: `mxinden/metricsstore` is rebased onto current `kubernetes/master`. This PR targets `kubernetes/performance-optimization`. As the latter is behind master, this PR includes commits which already landed on `kubernetes/master`, but which I am apparently not allowed to propose.

Options moving forward:
- fast-forward `kubernetes/performance-optimization`
- make PR target `kubernetes/master` instead

I am fine with either one.

brancz:
I fast forwarded `kubernetes/performance-optimization`.

/check-cla

ehashman:
Per the request in https://github.com/kubernetes/kube-state-metrics/pull/534#issuecomment-419251576 I've tested this PR in production against a ~200 node cluster with ~2000 running pods. I will follow up with more detailed stats once I let this run for a little longer, but preliminary results after 24 hours of uptime show about a **90% reduction in average CPU usage** (0.60 cores -> 0.06 cores) and a **70% reduction in average RAM usage** (1.3Gi -> 385Mi) from the 1.4.0 release I'm running simultaneously.

I noticed that responses from KSM are no longer gzipped. This has resulted in a **much larger response size (~3MB -> ~55MB)**. I use a readinessProbe every 10s against KSM and found as a result that 1.4.0 sends about 330kBps of traffic on average, but 1.4.0-perf.6 increases to 5.5MBps.

mxinden:
Thanks a lot @ehashman for running this in a bigger environment. Those numbers look great.

> responses from KSM are no longer gzipped

Right, I will look into that. Thanks for the hint.

I have pushed a bugfix since, here are the updated images:
- quay.io/mxinden/kube-state-metrics:v1.4.0-perf.8
- minden/kube-state-metrics:v1.4.0-perf.8

In case you have a chance, would you mind trying those out as well?

---

Our scalability team gave me access to a big cluster (~250 nodes) with some artificial workload on it. Next to the above mentioned improvements by @ehashman scrape request duration also improved and is a lot more stable now (top=v1.4.0; bottom=v1.4.0-perf.8):

scrape_duration

mxinden:
With the last commit I have added **gzip** support via http accept headers. I have used both the Golang `compress/gzip` and `https://github.com/NYTimes/gziphandler` (as suggested [here](https://github.com/prometheus/client_golang/issues/366)) packages with their standard configuration. The following graphs show CPU usage, memory usage and response time of the following kube-state-metrics versions:

- `v1.4.0`
- `v1.4.0-perf.8`
- `v1.4.0-nytgzip.1`
- `v1.4.0-gzip.3`

kube-state-metrics_cpu
kube-state-metrics_memory
kube-state-metrics_response-time


Overall in this environment `v1.4.0-perf.8` wins in CPU, memory and response time.

What are peoples thoughts on this? Happy for more input in this regard. You can try out the above mentioned tags here:

- quay.io/coreos/kube-state-metrics:v1.4.0
    minden/kube-state-metrics:v1.4.0

- quay.io/mxinden/kube-state-metrics:v1.4.0-gzip.3
    minden/kube-state-metrics:v1.4.0-gzip.3

- quay.io/mxinden/kube-state-metrics:v1.4.0-nytgzip.1
    minden/kube-state-metrics:v1.4.0-nytgzip.1

- quay.io/mxinden/kube-state-metrics:v1.4.0-perf.8
    minden/kube-state-metrics:v1.4.0-perf.8

mrsiano:
> here

Max, as you mentioned perf wins, there is any good reason to zip the results set if it's not too big?, may be it worth to have another cycle for huge results set ?

mxinden:
@mrsiano are you saying we should only activate the gzip option when we expect response sizes to be large?

mxinden:
Gziped content seems to be summable. In the long run we could also look into gziping the metrics of a Kubernetes object on creation / deletion / update time and saving them compressed in the cache. Once a scrape request comes in one should be able to concatenate the compressed metrics as one response.

That would save a lot of resources in environments where number of scrapes is bigger than number of object creations / deletions  / updates.

mrsiano:
we had another benchmark for large scale pods (10K) and seems like the gzip wins.
also there is much better memory usage for gzip.
the CPU avg is very close gzip has 2.27% vs perf8 2.39%

perf8 scrape response time
image

gzip scrape response time
image


@mxinden

brancz:
@mxinden it looks like something went wrong with rebasing.

brancz:
My suggestion let's work towards merging this into master and then keep iterating. The improvements we already have now are already pretty vast, so let's get those out and start doing pre-releases to get more widespread feedback.

mxinden:
> @mxinden it looks like something went wrong with rebasing.

@brancz this is related to `kubernetes:performance-optimizations` branch being behind master (See earlier [discussion](https://github.com/kubernetes/kube-state-metrics/pull/534#issuecomment-420738447) above). I will change the target of this pull request to `master`.

---

Moving on with pre-releases sounds good to me. Last thing that needs to be discussed is the so far missing exposition of `HELP` and `TYPE` headings. As of today the end-to-end tests run promtool on the `/metrics` output, which complains about the above. My suggestion is to ignore these failures for now via a simple regex. I already have a way in mind to bring them back without too much performance loss. As they are not necessary for Prometheus core functionality this is not a blocker.

mxinden:
From my side this PR is ready to merge into master. Are there any further comments blocking this pull request? If not I will squash the commits and ask one of the maintainers @brancz @andyxning to merge.

brancz:
Looks good from my side. Go ahead and squash.

mxinden:
@brancz squashed.

PR #535, opened 2018-09-04
Add kube_pod_container_status_last_terminated_reason

### LINKED ISSUE #344: Duration of kube_pod_container_status_terminated_reason metrics
I understand that this metrics is keeping track of pods which were terminated and the reason for it, but the metrics store data of how many days? Like the data will be of last two days or last 12 hours?

jutley:
This PR introduces a new metric: `kube_pod_container_status_last_terminated_reason`. 

Currently, we have the `kube_pod_container_status_terminated_reason`, but this always returns to 0 once a container starts back up. This means that we will only have a couple data points, if any at all, around the reason for a container termination. As a result, we cannot alert when a container crashes for a specific reason (we'd like to alert based on OOMs). 

This is brought up in this issue: https://github.com/kubernetes/kube-state-metrics/issues/344

Fixes #344

### DIFF
--- a/pkg/collectors/pod.go
+++ b/pkg/collectors/pod.go
@@ -135,6 +135,13 @@ var (
 		append(descPodLabelsDefaultLabels, "container", "reason"),
 		nil,
 	)
+	descPodContainerStatusLastTerminatedReason = prometheus.NewDesc(
+		"kube_pod_container_status_last_terminated_reason",
+		"Describes the last reason the container was in terminated state.",
+		append(descPodLabelsDefaultLabels, "container", "reason"),
+		nil,
+	)
+
 	descPodContainerStatusReady = prometheus.NewDesc(
 		"kube_pod_container_status_ready",
 		"Describes whether the containers readiness check succeeded.",
@@ -251,6 +258,7 @@ func (pc *podCollector) Describe(ch chan<- *prometheus.Desc) {
 	ch <- descPodContainerStatusRunning
 	ch <- descPodContainerStatusTerminated
 	ch <- descPodContainerStatusTerminatedReason
+	ch <- descPodContainerStatusLastTerminatedReason
 	ch <- descPodContainerStatusReady
 	ch <- descPodContainerStatusRestarts
 	ch <- descPodSpecVolumesPersistentVolumeClaimsInfo
@@ -380,6 +388,13 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 		return cs.State.Terminated.Reason == reason
 	}
 
+	lastTerminationReason := func(cs v1.ContainerStatus, reason string) bool {
+		if cs.LastTerminationState.Terminated == nil {
+			return false
+		}
+		return cs.LastTerminationState.Terminated.Reason == reason
+	}
+
 	var lastFinishTime float64
 
 	for _, cs := range p.Status.ContainerStatuses {
@@ -394,6 +409,9 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 		addGauge(descPodContainerStatusTerminated, boolFloat64(cs.State.Terminated != nil), cs.Name)
 		for _, reason := range containerTerminatedReasons {
 			addGauge(descPodContainerStatusTerminatedReason, boolFloat64(terminationReason(cs, reason)), cs.Name, reason)
+		}
+		for _, reason := range containerTerminatedReasons {
+			addGauge(descPodContainerStatusLastTerminatedReason, boolFloat64(lastTerminationReason(cs, reason)), cs.Name, reason)
 		}
 		addGauge(descPodContainerStatusReady, boolFloat64(cs.Ready), cs.Name)
 		addCounter(descPodContainerStatusRestarts, float64(cs.RestartCount), cs.Name)
--- a/pkg/collectors/pod_test.go
+++ b/pkg/collectors/pod_test.go
@@ -61,6 +61,8 @@ func TestPodCollector(t *testing.T) {
 		# TYPE kube_pod_container_status_terminated gauge
 		# HELP kube_pod_container_status_terminated_reason Describes the reason the container is currently in terminated state.
 		# TYPE kube_pod_container_status_terminated_reason gauge
+		# HELP kube_pod_container_status_last_terminated_reason Describes the last reason the container was in terminated state.
+		# TYPE kube_pod_container_status_last_terminated_reason gauge
 		# HELP kube_pod_container_status_waiting Describes whether the container is currently in waiting state.
 		# TYPE kube_pod_container_status_waiting gauge
 		# HELP kube_pod_container_status_waiting_reason Describes the reason the container is currently in waiting state.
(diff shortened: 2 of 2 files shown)


mxinden:
Thanks for following up with this @jutley! :+1:

olvesh:
I can see that #535 was merged, but there has been no release since august. Would it be ok to cut a release of kube-state-metrics soon?

brancz:
we are soon going to start cutting pre releases of the new release

andredantasrocha:
Hi @brancz any updates on the new release? This new metric will be very handy to us!

Thanks,
Andre

mxinden:
https://github.com/kubernetes/kube-state-metrics/pull/577 should be the
last thing needed, before we can cut the first alpha release.

aiman-alsari:
I've used the alpha version of 1.5.0 to try this out. Whilst it is useful for alerting on a single occurrence of a particular reason e.g OomKill, it makes it very difficult to do anything more complex like "only alert if there have been 2 OomKills in the space of an hour". Sometimes a single kill is a valid scenario, so isn't worth alerting on. To be able to do this cleanly, there needs to be a restart counter per reason rather than a simple flag as is now.

It is possible to do more complex queries by using a multiplication of the difference in restart count * the last reason flag, but it makes the PromQL a lot trickier in general.

Just my 2c

jutley:
@aiman-alsari I agree that having counts for each error type could be useful. Unfortunately, the Kubernetes API does not provide that type of information. All it offers is 1) current state, 2) last state, and 3) error count. Since this project creates Prometheus metrics that reflect data from the Kubernetes API, I don't think this suggestion fits within the project scope. 

If you'd like, you can still file an issue. I don't think posting on this PR will get much more attention since it has already been merged and is nearly released. 

As an alternative, you can create an alert that looks like:
```
kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 2
AND
kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1
```
This will fire if there have been at least 2 restarts in the last hour, and the last_terminated_reason is OOMKilled. It's not perfect, but will probably get you what you want in most cases. If there are other regular causes for the container to terminate, you can change the latter expression to:
```
min_over_time(kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}[1h]) == 1
```
This will cause the alert to only fire if the last terminated reason for the last hour is OOMKilled. It will take a minimum of an hour to fire, so in most cases I would probably prefer the first example.

gjtempleton:
@jutley Am I right in thinking that to get the points to match up neatly between the `kube_pod_container_status_restarts_total` and `kube_pod_container_status_last_terminated_reason` metric paths so that your first example would work the reason tag will need to be stripped out, e.g. with the `without` clause?

EvilCreamsicle:
I'm able to get almost what I need with this metric, but does anyone know if there is a metric that displays the actual container exit code?
I'm trying to filter on a specific segfault error, so I need to know if an error was an exit code 139.

brancz:
@EvilCreamsicle it is unlikely that kube-state-metrics will expose that fine granular detail, as it would cause too high churn of time-series data.

onrylmz:
> @aiman-alsari I agree that having counts for each error type could be useful. Unfortunately, the Kubernetes API does not provide that type of information. All it offers is 1) current state, 2) last state, and 3) error count. Since this project creates Prometheus metrics that reflect data from the Kubernetes API, I don't think this suggestion fits within the project scope.
> 
> If you'd like, you can still file an issue. I don't think posting on this PR will get much more attention since it has already been merged and is nearly released.
> 
> As an alternative, you can create an alert that looks like:
> 
> ```
> kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 2
> AND
> kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1
> ```
> 
> This will fire if there have been at least 2 restarts in the last hour, and the last_terminated_reason is OOMKilled. It's not perfect, but will probably get you what you want in most cases. If there are other regular causes for the container to terminate, you can change the latter expression to:
> 
> ```
> min_over_time(kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}[1h]) == 1
> ```
> 
> This will cause the alert to only fire if the last terminated reason for the last hour is OOMKilled. It will take a minimum of an hour to fire, so in most cases I would probably prefer the first example.

Only one remark, 

- `kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 2` 

- `kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1` 

return different results in terms of the labels. So, prometheus cannot match the two vectors provided from those expressions. I solved the issue with _ignoring_ keyword. 

`kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 1 AND ignoring(reason) kube_pod_container_status_last_terminated_reason{reason='ContainerCannotRun'} > 0`

Here is the reference link --> https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches

ZubAnt:
> > @aiman-alsari I agree that having counts for each error type could be useful. Unfortunately, the Kubernetes API does not provide that type of information. All it offers is 1) current state, 2) last state, and 3) error count. Since this project creates Prometheus metrics that reflect data from the Kubernetes API, I don't think this suggestion fits within the project scope.
> > If you'd like, you can still file an issue. I don't think posting on this PR will get much more attention since it has already been merged and is nearly released.
> > As an alternative, you can create an alert that looks like:
> > ```
> > kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 2
> > AND
> > kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1
> > ```
> > 
> > 
> > This will fire if there have been at least 2 restarts in the last hour, and the last_terminated_reason is OOMKilled. It's not perfect, but will probably get you what you want in most cases. If there are other regular causes for the container to terminate, you can change the latter expression to:
> > ```
> > min_over_time(kube_pod_container_status_last_terminated_reason{reason="OOMKilled"}[1h]) == 1
> > ```
> > 
> > 
> > This will cause the alert to only fire if the last terminated reason for the last hour is OOMKilled. It will take a minimum of an hour to fire, so in most cases I would probably prefer the first example.
> 
> Only one remark,
> 
> * `kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 2`
> * `kube_pod_container_status_last_terminated_reason{reason="OOMKilled"} == 1`
> 
> return different results in terms of the labels. So, prometheus cannot match the two vectors provided from those expressions. I solved the issue with _ignoring_ keyword.
> 
> `kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 1 AND ignoring(reason) kube_pod_container_status_last_terminated_reason{reason='ContainerCannotRun'} > 0`
> 
> Here is the reference link --> https://prometheus.io/docs/prometheus/latest/querying/operators/#one-to-one-vector-matches

I think, it is better
`(kube_pod_container_status_restarts_total - kube_pod_container_status_restarts_total offset 1h >= 1) * ignoring(reason) kube_pod_container_status_last_terminated_reason{reason='ContainerCannotRun'} > 0. `

Experimentally found that when using `ignoring` with operator `and` the second vector ignores the filter `reason` in right vector. The result is greater than 0, while it is not true

leoskyrocker:
@ZubAnt The second vector is not ignoring the `reason` filter in my case.

PR #538, opened 2018-09-13
Change error message

MIBc:
The test creates two resources(pod, configmap), not only pod added.


main_test.go
```diff
@@ -36,7 +36,7 @@ func BenchmarkKubeStateMetrics(t *testing.B) {
 	kubeClient := fake.NewSimpleClientset()
 
 	if err := injectFixtures(kubeClient, 1000); err != nil {
-		t.Errorf("error injecting pod add: %v", err)
+		t.Errorf("error injecting resources add: %v", err)
```
mxinden: Maybe even remove the `add`?

MIBc: OK

PR #541, opened 2018-09-18
Add init container metrics

### LINKED ISSUE #502: kube state metrics doesn't report init container pod metrics

thedebugger:

Fixes #502

### DIFF
--- a/pkg/collectors/pod.go
+++ b/pkg/collectors/pod.go
@@ -202,6 +202,48 @@ var (
 		append(descPodLabelsDefaultLabels, "volume", "persistentvolumeclaim"),
 		nil,
 	)
+	descPodInitContainerInfo = prometheus.NewDesc(
+		"kube_pod_init_container_info",
+		"Information about an init container in a pod.",
+		append(descPodLabelsDefaultLabels, "container", "image", "image_id", "container_id"),
+		nil,
+	)
+	descPodInitContainerStatusWaiting = prometheus.NewDesc(
+		"kube_pod_init_container_status_waiting",
+		"Describes whether the init container is currently in waiting state.",
+		append(descPodLabelsDefaultLabels, "container"),
+		nil,
+	)
+	descPodInitContainerStatusWaitingReason = prometheus.NewDesc(
+		"kube_pod_init_container_status_waiting_reason",
+		"Describes the reason the init container is currently in waiting state.",
+		append(descPodLabelsDefaultLabels, "container", "reason"),
+		nil,
+	)
+	descPodInitContainerStatusRunning = prometheus.NewDesc(
+		"kube_pod_init_container_status_running",
+		"Describes whether the init container is currently in running state.",
+		append(descPodLabelsDefaultLabels, "container"),
+		nil,
+	)
+	descPodInitContainerStatusTerminated = prometheus.NewDesc(
+		"kube_pod_init_container_status_terminated",
+		"Describes whether the init container is currently in terminated state.",
+		append(descPodLabelsDefaultLabels, "container"),
+		nil,
+	)
+	descPodInitContainerStatusTerminatedReason = prometheus.NewDesc(
+		"kube_pod_init_container_status_terminated_reason",
+		"Describes the reason the init container is currently in terminated state.",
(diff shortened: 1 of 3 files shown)


pkg/collectors/pod_test.go
```diff
@@ -98,311 +114,381 @@ func TestPodCollector(t *testing.T) {
 		# HELP kube_pod_spec_volumes_persistentvolumeclaims_readonly Describes whether a persistentvolumeclaim is mounted read only.
 		# TYPE kube_pod_spec_volumes_persistentvolumeclaims_readonly gauge
 	`
-	cases := []struct {
+
+	type cs_to_prom func(cs v1.ContainerStatus, ns string, pod string) string
+
+	pod_to_prom := func(p v1.Pod, fn_cs cs_to_prom, fn_ics cs_to_prom) string {
```
thedebugger: on the hindsight, refactoring this code in these methods may have made it more complex. But it makes changing things around a breeze and adding more pods/containers.

pkg/collectors/pod_test.go
```diff
@@ -98,311 +114,381 @@ func TestPodCollector(t *testing.T) {
...
+		if p.Status.InitContainerStatuses != nil {
+			for _, ics := range p.Status.InitContainerStatuses {
+				sb.WriteString(fn_ics(ics, p.Namespace, p.Name))
+				sb.WriteString("\n")
+			}
+		}
+		return sb.String()
+	}
+
+	mk_cs := func(name string, id int, ready bool, rc int32) v1.ContainerStatus {
```
thedebugger: i might need to change the name of these methods. let me know

thedebugger:
i'm going to test this in our cluster, but wanted to get it out to get feedback.

pkg/collectors/pod.go
```diff
@@ -382,21 +431,30 @@ func (pc *podCollector) collectPod(ch chan<- prometheus.Metric, p v1.Pod) {
 
 	var lastFinishTime float64
 
-	for _, cs := range p.Status.ContainerStatuses {
-		addGauge(descPodContainerInfo, 1,
+	collectContainerStatusMetric := func(cs v1.ContainerStatus, info *prometheus.Desc, waiting *prometheus.Desc, _waitingReason *prometheus.Desc, running *prometheus.Desc, terminated *prometheus.Desc, terminatedReason *prometheus.Desc, restarts *prometheus.Desc) {
```
mxinden: How about instead of passing in each metric description as arguments to the function, just passing in a single boolean e.g. `isInitContainer`. The function then decides which metric descriptions to use based on that. Should make this less verbose.

thedebugger: I agree it would reduce the params but not sure if that make it less verbose. 
``` 
if init {
desc1 = init_metrics_desc1
...
}
else {
desc1 = metrics_desc1
...
}
addGauge(metrics_desc1, 1)
```
If i can make the change if you strongly feel about it

mxinden: What do others think?

brancz: how about a third way? we pass a map instead of each single description, that way we have a single parameter and we can initialize the map just like the descriptions

thedebugger: Sounds good. I made the change.

thedebugger:
sorry, as I was rebasing it, the rebase changes got split across my commits. So I decided to squash them together for better readability. I'm going to test these changes in our cluster, but the PR is ready for final review. Let me know if you folks have more feedbacks.

brancz:
Sorry lost track of this.

/reopen

brancz:
Sorry for the delay. This needs a rebase.

tariq1890:
@thedebugger would you be able to update this PR?

thedebugger:
@brancz @tariq1890 hey, i can try doing it but it is going to be a lot of work since the code has changed a lot. Could we have a smaller turn around time so that we can break this cycle? Let me know.

brancz:
Usually turn around is very fast for this project, not sure why we dropped the ball here. Sorry!

PR #549, opened 2018-09-28
Add changelog note for breaking change in #371

### LINKED ISSUE #543: KSM is unable to authenticate to cluster in 1.3.x+ releases

/kind bug

**What happened**:

When I run kube-state-metrics on any release 1.3.0+, it launches, but when it attempts to scrape any resources, I get the error:

```
User "system:anonymous" cannot list <resource>s at the cluster scope: No policy matched.
```

However, all the various clusterbindingroles are set up correctly for the cluster. I am not sure why KSM is unable to authenticate.

The correct token is definitely loaded inside the KSM Pod and I've successfully used it to authenticate as the ServiceAccount against the API server with a curl command (it matches "system:serviceaccount:kube-system:kube-state-metrics" in this case). For some reason, it seems that KSM is ignoring the token?

KSM successfully authenticates with the API server and scrapes the cluster. This does work correctly in the 1.2.0 release (but I believe that certificate verification might not be working properly in that release, as I didn't have to load in our custom CA certs to get KSM 1.2.0 to work correctly).

I'm not really sure how to minimally reproduce this; it affects all our clusters (prod + QA) but there's nothing particularly special about them. We have RBAC enabled.

**Environment**:
- Kubernetes version (use `kubectl version`): 1.8.7
- Kube-state-metrics image version: 1.3.0, 1.3.1, 1.4.0

ehashman:

Adds a note about breaking changes to the CHANGELOG for the 1.2.0 -> 1.3.0 upgrade.

The changelog does not currently warn about the deprecation of the `--in-cluster` flag, meaning that users on 1.2.0 releases and earlier who use `--apiserver` with the default `--in-cluster=true` flag have authentication break on upgrade.

Fixes #543.

**Other notes:**

The behaviour from 1.2.0 can be emulated on upgrade by removing the `--apiserver` flag and setting the environment variable `KUBERNETES_SERVICE_HOST` to the API server value.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -29,6 +29,10 @@ After a testing period of 12 days, there were no additional bugs found or featur
 
 ## v1.3.0-rc.0 / 2018-03-23
 
+* [CHANGE] Removed `--in-cluster` flag in [#371](https://github.com/kubernetes/kube-state-metrics/pull/371).
+  Users can no longer specify `--apiserver` with `--in-cluster=true`. To
+  emulate this behaviour in future releases, set the `KUBERNETES_SERVICE_HOST`
+  environment variable to the value of the `--apiserver` argument.
 * [FEATURE] Allow to specify multiple namespace.
 * [FEATURE] Add `kube_pod_completion_time`, `kube_pod_spec_volumes_persistentvolumeclaims_info`, and `kube_pod_spec_volumes_persistentvolumeclaims_readonly` metrics to the Pod collector.
 * [FEATURE] Add `kube_node_spec_taint` metric.


mxinden:
Would you mind mentioning the fix via `KUBERNETES_SERVICE_HOST` in the changelog as well? Other than that this looks good to me, thanks for the follow up!

andyxning:
Please also add a notation about the deprecation PR #371

ehashman:
All feedback should be addressed.

andyxning:
@ehashman You should add a link to the `#371` PR to make it a link instead of just a statement.

ehashman:
I figured it would auto-link but apparently not...

andyxning:
@ehashman Thanks for taking care of this!

PR #553, opened 2018-10-10
Fix link href

colmose:
One of the hrefs in your README is broken currently. 

This updates with the correct link reference


colmose:
I signed it

colmose:
Thanks, @brancz!

If I had known the journey involved in fixing one href, I would have thought twice. 😂

PR #563, opened 2018-10-17
[WIP] pkg/options: Add flag to enable gzip encoding for responses

mxinden:

Dependent on the environment users might or might not want to encode the
responses of kube-state-metrics via gzip independent of the
`Accept-Header` that the client (Prometheus) sends.

This patch adds a corresponding command line flag.

As I am still evaluating whether to enable or disable gzip by default, this is a *work-in-progress* pull request. Thoughts are welcome.


brancz:
While I understand arguments for either, I'd prefer to stick with what we do now, which is respect the accept header. Optimizations should be left up to the user.

beorn7:
Only that you cannot configure that on the Prometheus side… I wouldn't surprise if Prometheus server folks say “this should be left to the monitored target to decide”. :o)

brancz:
I don’t really feel strongly either way, if we change from what we do currently it should at least show up in the changelog.

beorn7:
See https://github.com/kubernetes/kube-state-metrics/issues/498#issuecomment-431882415 for performance comparison with and without gzip encoding.

mxinden:
With previous benchmarking and @beorn7 [extensive tests](https://github.com/kubernetes/kube-state-metrics/issues/498#issuecomment-431882415) I think disabling gzip by default is the way to go. With the current patch, gzip would only be enabled when adding `--enable-gzip-encoding`.

@brancz @andyxning Let me know what you think.

brancz:
Let's do it, but also let's also make sure to call it out explicitly in the release notes.

CHANGELOG.md
```diff
@@ -1,5 +1,7 @@
 ## Unreleased
 
+- [CHANGE] Disable gzip compression of kube-state-metrics responses by default. Can be re-enabled via `--enable-gzip-encoding`. See #563 for more details.
```
mxinden: > let's also make sure to call it out explicitly in the release notes.

@brancz added note to CHANGELOG.md to make sure we don't forget to mention it.

PR #566, opened 2018-10-22
Pre-allocate the metrics slice in `GetAll`

beorn7:

This method is the heaviest allocator. Pre-allocation will not help a
lot, but I guess every little helps... (Plus, it's really easy.)

It's related to #498 . See analysis there.

@mxinden


beorn7:
Hmm, right, I thought I signed the CLA, but that was for Google, not LF… will look into it.

beorn7:
OK, CLA worked.

The downside of this, which I humbly mentioned in the commit description, is actually none. Without this change, _more_ allocations happen under the lock.

PR #567, opened 2018-10-24
pkg/metrics: Tune NewMetrics function

mxinden:

This patch tunes the `NewMetrics` function based on optimizations discussed in https://github.com/prometheus/common/pull/148 and https://github.com/kubernetes/kube-state-metrics/issues/498#issuecomment-431882415 (//CC @beorn7).

- Use strings.Builder minimizing memory allocations.

- Use strconv to render float64 in combination with a []byte buffer pool
instead of fmt.Sprintf reducing memory allocations.

- Only sort labels of metrics in test framework, not by default (e.g. in
production).

Benchmark patch:

```bash
go get golang.org/x/tools/cmd/benchcmp

git checkout perf-exp~1
go test -v -bench=NewMetric  -run=^$  -benchmem ./pkg/metrics/... > old.txt

git checkout perf-exp
go test -v -bench=NewMetric  -run=^$  -benchmem ./pkg/metrics/... > new.txt

benchcmp old.txt new.txt

benchmark                           old ns/op     new ns/op     delta
BenchmarkNewMetric/value-1-4        2652          612           -76.92%
BenchmarkNewMetric/value-35.7-4     2706          755           -72.10%

benchmark                           old allocs     new allocs     delta
BenchmarkNewMetric/value-1-4        32             6              -81.25%
BenchmarkNewMetric/value-35.7-4     32             6              -81.25%

benchmark                           old bytes     new bytes     delta
BenchmarkNewMetric/value-1-4        1616          528           -67.33%
BenchmarkNewMetric/value-35.7-4     1616          528           -67.33%
```

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -18,14 +18,28 @@ package metrics
 
 import (
 	"errors"
-	"fmt"
-	"sort"
+	"math"
+	"strconv"
 	"strings"
+	"sync"
 
 	"github.com/prometheus/client_golang/prometheus"
 	dto "github.com/prometheus/client_model/go"
 
 	"k8s.io/kube-state-metrics/pkg/options"
+)
+
+const (
+	initialNumBufSize = 24
+)
+
+var (
+	numBufPool = sync.Pool{
+		New: func() interface{} {
+			b := make([]byte, 0, initialNumBufSize)
+			return &b
+		},
+	}
 )
 
 // Metric represents a single line entry in the /metrics export format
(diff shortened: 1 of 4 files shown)


mxinden:
Applied all comments @beorn7, thanks for the review.

Also looked into converting `WriteString` of length two with two `WriteByte` as mentioned [here](https://github.com/golang/go/issues/26264), but improvement seems not worth it.

@beorn7 @brancz any further comments?

pkg/metrics/metrics.go
```diff
@@ -37,48 +51,75 @@ func NewMetric(name string, labelKeys []string, labelValues []string, value floa
...
+	labelsToString(&m, labelKeys, labelValues)
 
-	m = m + labelsToString(labelKeys, labelValues)
+	m.WriteByte(' ')
 
-	m = m + fmt.Sprintf(" %v", value)
+	writeFloat(&m, value)
 
-	m = m + "\n"
+	m.WriteString("\n")
```
beorn7: DIdn't catch this in my first round of review: This could be written as
```Go
m.WriteByte('\n')
```

mxinden: Right, sorry about that. Follow up: https://github.com/kubernetes/kube-state-metrics/pull/572

PR #572, opened 2018-10-29
pkg/metrics: Write newline as byte with strings.Builder

mxinden:

Write newline with `WriteByte` for higher performance.

Origins from https://github.com/kubernetes/kube-state-metrics/pull/567#discussion_r228943917

//CC @brancz @beorn7


brancz:
lgtm :+1:

mxinden:
Thanks for the super quick review @brancz!

PR #575, opened 2018-11-01
Making kube-state-metrics a library

LiliC:
`kube-state-metrics` has many useful parts which could be used as library, this is the first step towards that.

For now only the functions are exposed and some helper functions moved to a different place. In the long term I would prefer to end with the following structure:
```
pkg
├── collector // functions needed to create a collector, this would be used in the builder pkg and externally 
├── internal
│   └── builder // content of the current upstream resource collectors which would consume the collector pkg
├── metrics // what is currently there but also move the utils/helper functions from collectors. Also the metricHandler interface as that would be useful to expose and reuse elsewhere.

```
Any thoughts on the above are welcome, I can also open an issue to discuss it?

### DIFF
--- a/pkg/collectors/utils.go
+++ b/pkg/collectors/utils.go
@@ -58,7 +58,7 @@ func boolFloat64(b bool) float64 {
 // addConditionMetrics generates one metric for each possible node condition
 // status. For this function to work properly, the last label in the metric
 // description must be the condition.
-func addConditionMetrics(desc *metricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
+func addConditionMetrics(desc *MetricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
 	ms := []*metrics.Metric{}
 	m, err := metrics.NewMetric(desc.Name, desc.LabelKeys, append(lv, "true"), boolFloat64(cs == v1.ConditionTrue))
 	if err != nil {
--- a/pkg/collectors/utils.go
+++ b/pkg/collectors/utils.go
@@ -58,7 +58,7 @@ func boolFloat64(b bool) float64 {
 // addConditionMetrics generates one metric for each possible node condition
 // status. For this function to work properly, the last label in the metric
 // description must be the condition.
-func addConditionMetrics(desc *MetricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
+func addConditionMetrics(desc *metrics.MetricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
 	ms := []*metrics.Metric{}
 	m, err := metrics.NewMetric(desc.Name, desc.LabelKeys, append(lv, "true"), boolFloat64(cs == v1.ConditionTrue))
 	if err != nil {
--- a/pkg/collectors/utils.go
+++ b/pkg/collectors/utils.go
@@ -1,0 +1,108 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collectors
+
+import (
+	"regexp"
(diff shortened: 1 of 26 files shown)


LiliC:
cc @mxinden

pkg/collectors/utils.go
```diff
@@ -0,0 +1,108 @@
...
+	if b {
+		return 1
+	}
+	return 0
+}
+
+// addConditionMetrics generates one metric for each possible node condition
+// status. For this function to work properly, the last label in the metric
+// description must be the condition.
+func addConditionMetrics(desc *metrics.MetricFamilyDef, cs v1.ConditionStatus, lv ...string) []*metrics.Metric {
```
brancz: I'm expecting that people are going to want to use these as they are rather common cases, but let's start with a minimal API rather than adding helpers we're unsure about.

brancz:
I also like the suggestion of making the actual kube-state-metrics part an internal package to ensure users don't ever import it. Naming we can discuss once we re-organize :slightly_smiling_face: .

mxinden:
To test out the API of the library I wrote this little test in `tests/as_library/as_library_test.go`:

```golang

package aslibrary

import (
	"context"
	"strings"
	"testing"
	"time"

	"k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/runtime"
	"k8s.io/apimachinery/pkg/watch"
	clientset "k8s.io/client-go/kubernetes"
	"k8s.io/client-go/kubernetes/fake"
	"k8s.io/client-go/tools/cache"
	"k8s.io/kube-state-metrics/pkg/collectors"
	"k8s.io/kube-state-metrics/pkg/metrics"
	metricsstore "k8s.io/kube-state-metrics/pkg/metrics_store"
)

func TestAsLibrary(t *testing.T) {
	kubeClient := fake.NewSimpleClientset()

	service := v1.Service{
			Name:            "my-service",
			ResourceVersion: "123456",
		},
	}

	_, err := kubeClient.CoreV1().Services(metav1.NamespaceDefault).Create(&service)
	if err != nil {
		t.Fatal(err)
	}

	c := serviceCollector(kubeClient)

	// Wait for informers to sync
	time.Sleep(time.Second)

	m := c.Collect()

	if len(m) != 1 {
		t.Fatalf("expected one metric to be returned but got %v", len(m))
	}

	if !strings.Contains(string(*m[0]), service.ObjectMeta.Name) {
		t.Fatal("expected string to contain service name")
	}
}

func serviceCollector(kubeClient clientset.Interface) *collectors.Collector {
	store := metricsstore.NewMetricsStore(generateServiceMetrics)

	lw := cache.ListWatch{
			return kubeClient.CoreV1().Services(metav1.NamespaceDefault).List(opts)
		},
			return kubeClient.CoreV1().Services(metav1.NamespaceDefault).Watch(opts)
		},
	}

	r := cache.NewReflector(&lw, &v1.Service{}, store, 0)

	go r.Run(context.TODO().Done())

	return collectors.NewCollector(store)
}

func generateServiceMetrics(obj interface{}) []*metrics.Metric {
	sPointer := obj.(*v1.Service)
	s := *sPointer

	m, err := metrics.NewMetric("test_metric", []string{"name"}, []string{s.Name}, 1)
	if err != nil {
		panic(err)
	}

	return []*metrics.Metric{m}
}

```

This still seems like a lot of code a user would need to write. I will follow up with some ideas how we can improve this. Some of your proposed future restructures definitely sound good.

Please keep in mind, that https://github.com/kubernetes/kube-state-metrics/pull/577 will introduce a couple of breaking changes that would also affect the library API. In addition as mentioned above I would like to improve the usability of the libraries API. Thereby the library would currently have no stability guarantees. Maybe we can include something like my test above in this patch to make sure I am not breaking too much with https://github.com/kubernetes/kube-state-metrics/pull/577. I am also happy to follow up with a Golang [example](https://blog.golang.org/examples).

What are your thoughts @LiliC?

LiliC:
> This still seems like a lot of code a user would need to write. I will follow up with some ideas how we can improve this. Some of your proposed future restructurings definitely sound good.

Yes agreed. Also, there is serving metrics, I initially exposed the `metricHandler` as well, and the `ServeHTTP`, but the problem was cycling dependencies, as for me it made more sense to have that in the `metrics` pkg. Which is why I suggested the above `internal` pkg that would just consume these pkgs.

> Thereby the library would currently have no stability guarantees.

That is fine and makes perfect sense.

LiliC:
Added the test you suggested, thanks! Not sure what you had in mind about the example, as this already serves an example. But maybe some docs when the API is more stable?

@mxinden @brancz Please have another look, thanks!

brancz:
> But maybe some docs when the API is more stable?

I think that's a good idea once #577 lands.

Thanks for the contributions!

/lgtm

Issue #557, opened 2018-10-12
Next release date?


alapidas:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:

Wondering when the next release will be, looking to pick up https://github.com/kubernetes/kube-state-metrics/commit/d171c3e440ab4a492f7fc966fdbf98cb26417783

**What you expected to happen**:

**How to reproduce it (as minimally and precisely as possible)**:

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`):
- Kube-state-metrics image version

mxinden:
Most of the performance refactoring is done (https://github.com/kubernetes/kube-state-metrics/pull/534), but I need to re-enable the white-/blacklisting feature and figure out whether to go with gzip by default or not (https://github.com/kubernetes/kube-state-metrics/pull/563). Anyone else have any blockers for an alpha release?

brancz:
Once we're at feature parity again, let's move forward with the pre-release.

fedevegili:
Any update here? Related merges seems merged

mxinden:
@fedevegili we just merged the last optimization. I suggest merging everything into master next and then we can cut the first alpha.

PR #577, opened 2018-11-01
pkg/collectors: Renable white-/blacklisting and HELP text

mxinden:

- Reenable feature to filter the exposition of metric families by their
name. This is now done at startup time, thereby not slowing down the
critical path.

- Expose metrics grouped by their metric family and prefix each metric
family with their HELP text. In the future this can easily be extended
to also expose the TYPE text.

(- This also paved the path for sorting the exposed metrics in a
zero-cost manner if this is wanted (//CC @beorn7).)

Relates to https://github.com/kubernetes/kube-state-metrics/issues/557


pkg/collectors/builder.go
```diff
@@ -20,16 +20,16 @@ package collectors
 import (
 	"strings"
 
-	apps "k8s.io/api/apps/v1beta1"
-	autoscaling "k8s.io/api/autoscaling/v2beta1"
-	batchv1 "k8s.io/api/batch/v1"
-	batchv1beta1 "k8s.io/api/batch/v1beta1"
-	extensions "k8s.io/api/extensions/v1beta1"
+	// 	apps "k8s.io/api/apps/v1beta1"
```
mxinden: I have done the refactor only for service.go for now. Thereby one can ignore this change.

mxinden:
This pull request is still in a work-in-progress state. So far the refactoring was only applied to the `service.go` collector and the overall patch is missing unit- and benchmarking-tests.

I would like to welcome feedback on the overall architecture proposed here.

mxinden:
Also //CC @LiliC as this relates to https://github.com/kubernetes/kube-state-metrics/pull/575.

brancz:
As this is WIP and still needs a substantial amount of work, could we move forward with #575, and rebase these changes on top of that?

brancz:
I'm not sure I'm happy with the amount of concerns in this PR. Could we split this into multiple?

The PRs that I'm seeing in this (each individually a big improvements, but unless separated hard to review):

* reintroducing whitelist/blacklist
* reintroducing help text (and type?)
* changing the interface to simply consume `io.Writer` (I do like the change)

While I'm very much in favor of just accepting `io.Writer` as the input parameter, do you think this is something that we could postpone to after the v1.5.0 release? If not we will need to put this through another run of benchmarking as the changes proposed are rather invasive.

beorn7:
> - This also paved the path for sorting the exposed metrics in a zero-cost manner if this is wanted

Yes, I would love that.

Also :+1: for HELP and TYPE.

mxinden:
> I'm not sure I'm happy with the amount of concerns in this PR. Could we split this into multiple?

Good point @brancz. I have split up the work into the three suggested parts.

> While I'm very much in favor of just accepting io.Writer as the input parameter, do you think this is something that we could postpone to after the v1.5.0 release? If not we will need to put this through another run of benchmarking as the changes proposed are rather invasive.

Performance wise this patch includes multiple changes in the hot path that need benchmarking:

- Re-enabling white- and blacklisting forces us to have a different marshal logic. 
- Reintroducing HELP text forces us to save slices of slices of metrics and to loop multiple times over the same slice.
- Passing a io.Writer instead of returning strings might influence memory allocation in a unforeseeable way.

Thereby I would suggest including them before the first release candidate. I don't think we need to go through the same thorough testing as @beorn7 and I did before as the general architecture (caching, ...) stays the same. What do you think @brancz?

brancz:
That’s fair, let’s do them and do another round of benchmarking. While we are at it, what do you thing of splitting the micro benchmarks into metric rendering into buffer, rendering the metrics request and one that does the whole thing (this one is essentially what we have today)? Plus it would be a nice detail to have the benchmarks show the byte throughout in addition to the Op/s.

mxinden:
> splitting the micro benchmarks

@brancz what are you referring to with "micro benchmarks"? As of today we only have a single benchmark test.

brancz:
That’s what I meant, I should have put the micro in quotes :)

mxinden:
@brancz @all any comments on the structural changes in `pkg/collectors/builder.go` and `pkg/collectors/service.go`? Otherwise I would follow up to restructure the other collectors as well.

brancz:
Generally I understand the scheme and I think it makes sense, but could we think a bit about how we could reduce the amount of boilerplate that brings with it? I imagine the pod collector being very hard to read once converted. It seems there is a common theme for: value metrics, label metrics, condition metrics and boolean metrics. If we can somehow abstract those further so they're essentially one liners, I think I'd be happy with the approach. I haven't looked too much into that, but it feels like that should be possible.

main_test.go
```diff
@@ -34,13 +34,13 @@ import (
 )
 
 func BenchmarkKubeStateMetrics(t *testing.B) {
```
LiliC: Nit: I would rename this to be `(b *testing.B)` as it's a benchmark and makes things more clear onwards.

main_test.go
```diff
@@ -56,32 +56,87 @@ func BenchmarkKubeStateMetrics(t *testing.B) {
 	builder.WithKubeClient(kubeClient)
 	builder.WithNamespaces(options.DefaultNamespaces)
 
-	collectors := builder.Build()
+	// This test is not suitable to be compared in terms of time, as it includes
```
LiliC: Maybe you can just pause the timer while you wait for the cache to be filled, by using `b.StopTimer()` right before `time.Sleep` is called and after that call `b.StartTimer`.

main_test.go
```diff
@@ -56,32 +56,87 @@ func BenchmarkKubeStateMetrics(t *testing.B) {
...
 	req := httptest.NewRequest("GET", "http://localhost:8080/metrics", nil)
 
-	// Wait for informers to sync
+	t.Run("MakeRequests", func(b *testing.B) {
+		for i := 0; i < requestCount; i++ {
+			w := httptest.NewRecorder()
+			handler.ServeHTTP(w, req)
+
+			resp := w.Result()
+			if resp.StatusCode != 200 {
```
LiliC: In general, you are just checking that the status is OK, but not that the data that is being served is present. I guess what I am suggesting is that maybe you should check that the metrics you are benchmarking are actually there.

brancz: Agreed. Maybe without checking the "right" content, we can just use the `BenchmarkTest.SetBytes` functionality of the benchmark to set the amount of bytes processed, if things are completely broken then the bytes amount should be off as well, further validation would be done by e2e tests.

main_test.go
brancz: I do agree with @LiliC, there are two benchmarks that we want to measure here:

1) filling the cache
2) rendering the output from the cache

It could be interesting to have an additional one that does do both (like we have now) just for overall performance, but in order to see gain of optimizations for each of the two scenarios above it would be good to have the detailed ones as well. They're distinct problems, that need distinct optimizations, so they should have distinct benchmarks.

mxinden: Not sure I understand. Filling cache and rendering into http response writer are split up into two subtests, thereby their results are reported as two distinct tests. Would you prefer to complete separate once @brancz?

main_test.go
mxinden: I am a bit undecided here. On the one hand I would like  to separate concerns, thereby not do correctness testing in a benchmark test, on the other hand I see your point and would also like to test correct responses, especially under load.

In regards to using  `BenchmarkTest.SetBytes`: Yes, that would be an option but requires a user to monitor this metric, whereas we could also entirely fail if the result is wrong.

I will look into stopping the timer and then properly checking the result.

mxinden: We could compare the output to a [golden record](https://www.youtube.com/watch?v=yszygk1cpEc) but that would need to be touched every time we change a metric.

Thinking about this some more I would prefer to keep it as it is with the 200 check, as there is the correctness check further below (`TestFullScrapeCycle`). What do you think @LiliC @brancz?

main_test.go
brancz: Ah that makes perfect sense. If that automatically resets the timers and bytes processed then this lgtm

main_test.go
brancz: Yes the actual correctness should not be tested in the benchmark I agree, real e2e tests are best for that, our unit tests are already very close to the golden record type testing, and we have e2e tests that check that every collector does expose metrics.

I still think setting bytes is a useful thing, and possibly could show users early that things are completely wrong, but not necessarily a complete check. Nevertheless showing how much Mb/s we can do at exposition is interesting to know.

LiliC: > I am a bit undecided here. On the one hand I would like to separate concerns, thereby not do correctness testing in a benchmark test, on the other hand I see your point and would also like to test correct responses, especially under load.

Yes this was my thought process behind my initial comment suggestion, was not sure either.

> Thinking about this some more I would prefer to keep it as it is with the 200 check, as there is the correctness check further below (TestFullScrapeCycle).

Agreed, it makes perfect sense. 👍

mxinden:
This pull request is now targeting `kubernetes:performance-optimization-new`. Any further suggestions? Otherwise I will follow up with the rest in future pull requests.

mxinden:
@brancz tests are green. Any further comments?

main.go
```diff
@@ -40,6 +40,7 @@ import (
 	kcollectors "k8s.io/kube-state-metrics/pkg/collectors"
 	"k8s.io/kube-state-metrics/pkg/options"
 	"k8s.io/kube-state-metrics/pkg/version"
+	"k8s.io/kube-state-metrics/pkg/whiteblacklist"
```
brancz: Can we already call this allow/deny list as planned for 2.0 and just keep the flags called as they are called today?

mxinden: @brancz I would prefer to do this in one go to avoid confusion. Is that also ok?

brancz: also fine by me

Issue #585, opened 2018-11-15
Collisions between metrics of same name but different namespace


gregory-lyons:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
As of #534, metric stores use a map keyed by object name:
https://github.com/mxinden/kube-state-metrics/blob/234d788a0d5cfac2b5bf33153afd3b944c39b220/pkg/metrics_store/metrics_store.go#L41

This a poor key choice given that Kubernetes allows you to have objects with the [same name across different namespaces](https://kubernetes.io/docs/concepts/overview/working-with-objects/namespaces/#when-to-use-multiple-namespaces). For any objects with duplicate names, kube-state-metrics now reports metrics for only one of them at a time.

We observed this problem in our cluster, as many of our object names are duplicated across namespaces. We have seen that kube-state-metrics switches which object it is reporting, possibly due to watches expiring/refreshing at staggered times, but only ever one object at once.

**What you expected to happen**:
kube-state-metrics should display metrics for all objects (some combination or hash of name+namespace might be a better key).

**How to reproduce it (as minimally and precisely as possible)**:
Create two objects with same name in different namespaces, and observe that kube-state-metrics only reports one.
```
glyons@mbp-005245 ~ $ kubectl create ns ns1
namespace "ns1" created
glyons@mbp-005245 ~ $ kubectl create ns ns2
namespace "ns2" created
glyons@mbp-005245 ~ $ kubectl create secret generic -n ns1 my-secret
secret "my-secret" created
glyons@mbp-005245 ~ $ kubectl create secret generic -n ns2 my-secret
secret "my-secret" created
glyons@mbp-005245 ~ $ $GOPATH/bin/kube-state-metrics --port=8080 --telemetry-port=8081 --host=localhost --kubeconfig=/Users/glyons/.kube/config &
[1] 70921
glyons@mbp-005245 ~ $ curl -s localhost:8080/metrics | grep my-secret
kube_secret_info{namespace="ns1",secret="my-secret"} 1
kube_secret_type{namespace="ns1",secret="my-secret",type="Opaque"} 1
kube_secret_created{namespace="ns1",secret="my-secret"} 1.542316991e+09
kube_secret_labels{namespace="ns1",secret="my-secret"} 1
kube_secret_metadata_resource_version{namespace="ns1",resource_version="4437",secret="my-secret"} 1
[...]

PR #586, opened 2018-11-16
pkg/metrics_store: Cache objects by uid not name

mxinden:

Previously the metrics store would cache metrics of Kubernetes objects
by object name. A name is not unique as two namespaces can each have an
object with the same name.

Instead cache objects by their uid which is unique across the cluster.

Thanks @gregory-lyons for the catch.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/585

This probably relates to the missing metrics referenced in https://github.com/kubernetes/kube-state-metrics/issues/498#issuecomment-431931146.

### DIFF
--- a/pkg/metrics_store/metrics_store.go
+++ b/pkg/metrics_store/metrics_store.go
@@ -6,6 +6,7 @@ import (
 	"k8s.io/kube-state-metrics/pkg/metrics"
 
 	"k8s.io/apimachinery/pkg/api/meta"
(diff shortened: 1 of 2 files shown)


brancz:
@gouthamve @tomwilkie I remember we were talking about you running master right now. If you did end up running master, it's definitely broken before this PR, just letting you know so you can update :slightly_smiling_face: .

gregory-lyons:
@mxinden @brancz thanks for the quick fix!

PR #587, opened 2018-11-22
main_test.go: Extend pod fixture and fix bench set bytes

mxinden:

Extending pod fixture for better benchmarking.


brancz:
neat :+1: 

/lgtm

PR #589, opened 2018-11-22
pkg/collectors: Refactor Cron Job and Config Map collector

mxinden:

This patch refactors the cron job and config map collector to the format introduced in https://github.com/kubernetes/kube-state-metrics/pull/577.


brancz:
Can you tell why the test failed? The output doesn't seem obvious.

mxinden:
This is fixed with https://github.com/kubernetes/kube-state-metrics/pull/588. I will force push to retrigger the tests.

mxinden:
@brancz would you mind taking another look at this one?

PR #590, opened 2018-11-22
pkg/collectors: Refactor deployment and daemonset

mxinden:

This patch refactors the daemon set and deployment collector to the format introduced in #577.


main_test.go
```diff
@@ -144,7 +144,29 @@ func TestFullScrapeCycle(t *testing.T) {
 
 	body, _ := ioutil.ReadAll(resp.Body)
 
-	expected := `# HELP kube_pod_info Information about pod.
+	expected := `# HELP kube_daemonset_created Unix creation timestamp
```
brancz: I'm wondering whether this actually gives us any additional safety or just maintenance burden.

mxinden: Same here. This is not a lot of work to do for now. It helps me catch some bugs during refactoring. Once this is done, I would suggest ignoring the `# HELP` lines of most collectors here. What do you think?

brancz: Sounds good. Ignore my comment on the other PR then :)

PR #591, opened 2018-11-23
pkg/collectors: Refactor endpoint and hpa collector

mxinden:

This patch refactors the endpoint and horizontal pod autoscaler collector to the format introduced in #577.


main_test.go
```diff
@@ -144,7 +144,20 @@ func TestFullScrapeCycle(t *testing.T) {
 
 	body, _ := ioutil.ReadAll(resp.Body)
 
-	expected := `# HELP kube_pod_info Information about pod.
+	// TODO: Move this to a golden record file?
```
brancz: as I mentioned in #590 I don't think the maintenance burden of this is actually worth it, I'd vote for removing this check entirely, as the unit and e2e tests already validate the same, without the additional maintenance required

brancz:
this needs a rebase I think

brancz:
this needs a rebase

mxinden:
@brancz rebased. Would you mind taking another look?

mxinden:
@brancz Resolved final merge commit. Would you mind giving this another :+1: ?

PR #593, opened 2018-11-26
pkg/collectors: Refactor namespace and node collector

mxinden:

This patch refactors the namespace and node collector to the format introduced in #577.

### DIFF
--- a/main.go
+++ b/main.go
@@ -74,8 +74,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	// TODO: Probably not necessary to pass all of opts into builder, right?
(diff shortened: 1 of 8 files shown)


main.go
```diff
@@ -74,8 +74,7 @@ func main() {
 		os.Exit(0)
 	}
 
-	// TODO: Probably not necessary to pass all of opts into builder, right?
-	collectorBuilder := kcollectors.NewBuilder(context.TODO(), opts)
+	collectorBuilder := kcollectors.NewBuilder(context.TODO())
```
brancz: this should probably be a cancellable background context that is cancelled on shutdown, I'd say we take care of that in a follow up, but can you open an issue to track this?

PR #596, opened 2018-11-27
pkg/collectors: Refactor replication controller and resource quota collector

mxinden:

This patch refactors the replication controller and resource quota collector to the format introduced in #577.


brancz:
this needs a rebase

PR #597, opened 2018-11-28
pkg/collectors: Refactor secret and stateful set collector

mxinden:

This patch refactors the secret and stateful set collector to the format introduced in #577.


mxinden:
This is the last one :tada:

brancz:
woooohooooo :tada: 

/lgtm 

Very very nice work!

PR #598, opened 2018-11-28
*: Reintroduce TYPE text

mxinden:

Prometheus exposes the [type](https://prometheus.io/docs/concepts/metric_types/) of a metric family as a header line. This patch reintroduces this feature to the performance optimized version.

This patch is based on https://github.com/kubernetes/kube-state-metrics/pull/591. Once that is merged, this can follow along.

### DIFF
--- a/pkg/metrics/metrics.go
+++ b/pkg/metrics/metrics.go
@@ -41,6 +41,7 @@ var (
 type FamilyGenerator struct {
 	Name         string
 	Help         string
+	Type         MetricType
 	GenerateFunc func(obj interface{}) Family
 }
 
@@ -56,6 +57,16 @@ func (f Family) String() string {
 
 	return b.String()
 }
+
+// MetricType represents the type of a metric e.g. a counter. See
+// https://prometheus.io/docs/concepts/metric_types/.
+type MetricType string
+
+// MetricTypeGauge defines a Prometheus gauge.
+var MetricTypeGauge MetricType = "gauge"
+
+// MetricTypeCounter defines a Prometheus counter.
+var MetricTypeCounter MetricType = "counter"
 
 // Metric represents a single time series.
 type Metric struct {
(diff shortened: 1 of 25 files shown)


brancz:
Can you rebase this as #591 is merged?

tests/e2e.sh
```diff
@@ -166,14 +166,7 @@ curl -s "http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state
...
 [ -n "$E2E_SETUP_PROMTOOL" ] && setup_promtool
-# kube-state-metrics does not expose the HELP lines for each metric family on
-# its /metrics endpoint. In order to still check for other issues raised by
-# promtool, filter out the HELP text warnings.
-# `grep -v` fails when it returns an empty string. This is inverted at the
-# beginning via `!`.
-set +o pipefail
-! cat $KUBE_STATE_METRICS_LOG_DIR/metrics | promtool check metrics 2>&1 | grep -v "no help text"
-set -o pipefail
+cat $KUBE_STATE_METRICS_LOG_DIR/metrics | promtool check metrics
```
mxinden: Reverted the filtering, now that we should be fully Prometheus format compliant again.

pkg/metrics/metrics.go
```diff
@@ -57,6 +58,16 @@ func (f Family) String() string {
 	return b.String()
 }
 
+// MetricType represents the type of a metric e.g. a counter. See
+// https://prometheus.io/docs/concepts/metric_types/.
+type MetricType string
+
+// MetricTypeGauge defines a Prometheus gauge.
+var MetricTypeGauge MetricType = "gauge"
```
mxinden: If we don't use a type alias, we could use constants instead, but I am not sure whether that has a real performance impact.

brancz:
Am I missing something? This PR looks good to me, but still has `[WIP]`?

mxinden:
> Am I missing something? This PR looks good to me, but still has [WIP]?

Right, sorry about that. I marked it as such while it was still based on #591.

brancz:
Ok, was just checking, perfect! :+1: 

This whole effort has been a long time coming, congrats and nice work!

/lgtm

PR #608, opened 2018-11-30
tests: Add stability notice to lib pkg

mxinden:

This patch adds a stability notice to the `tests/lib` package making sure no one depends on kube-state-metrics as a library and expects a stable interface.

While we are already in the package, it also adds a license header to the `lib_test.go` file.

For now this pull request targets the `release-1.5` branch. This can later on be merged back into `master`.


brancz:
Thanks!

/lgtm

PR #613, opened 2018-12-08
Introduce sharding (and experimental auto-sharding)

brancz:

This PR introduces sharding capabilities for kube-state-metrics, as well as the experimental feature to have kube-state-metrics automatically detect sharding, by discovering its nominal position within a StatefulSet and using the number of replicas in the StatefulSet as the total number of shards. I added a further description of the mechanism to the documentation.

These features will allow kube-state-metrics to be easily auto-scaled with the horizontal-pod-autoscaler by using its latency of `/metrics` requests to be lower than a certain threshold, otherwise scaling it up, as the latency is directly related to the amount of objects a shard needs to handle and the more replicas in a setup, the lower the number of objects, therefore a lower latency, as each shard handles 1/nth of objects in the cluster.

kube-state-metrics is well optimized, but in order to scale it infinitely and on commodity hardware it is necessary to shard the work it needs to do.

Note, I do not intend to ship these features in the already started release series of 1.5.x of kube-state-metrics, but rather release following it.

@andyxning @zouyee @mxinden @s-urbaniak @squat @metalmatze


README.md
```diff
@@ -132,6 +139,23 @@ These numbers are based on [scalability tests](https://github.com/kubernetes/kub
 
 Note that if CPU limits are set too low, kube-state-metrics' internal queues will not be able to be worked off quickly enough, resulting in increased memory consumption as the queue length grows. If you experience problems resulting from high memory allocation, try increasing the CPU limits.
 
+#### Horizontal scaling (sharding)
+
+In order to horizontally scale kube-state-metrics some automated sharding capabilities have been implemented. It is configured with the following flags:
+
+* `--shard`
```
mxinden: Maybe worth mentioning whether these are zero indexed?

README.md
```diff
@@ -132,6 +139,23 @@ These numbers are based on [scalability tests](https://github.com/kubernetes/kub
...
+
+Sharding should be used carefully, and additional monitoring should be set up in order to ensure that sharding is setup as expected (eg. instances for each shard out of the total shards are configured).
+
+##### Automated sharding
+
+There is also an experimental feature, that allows kube-state-metrics to auto discover its nominal position if it is deployed in a StatefulSet, in order to automatically configure sharding. This is an experimental feature and may be broken or removed without notice.
+
+To enable automated sharding kube-state-metrics must be run by a `StatefulSet` and the pod names and namespace must be handed to the kube-state-metrics process via the `--pod` and `--pod-namespace` flags.
```
mxinden: Instead of kube-state-metrics reaching out to the API-Server to get its index in the total amount of shards, why not derive it from the pod name retrieved via an environment variable?

pkg/sharding/listwatch.go
```diff
@@ -0,0 +1,88 @@
+package sharding
+
+import (
+	"crypto/md5"
```
mxinden: Why use a `crypto` hash function in this context? Wouldn't one of https://golang.org/pkg/hash/ be faster?

pkg/sharding/listwatch.go
```diff
@@ -0,0 +1,88 @@
...
+
+type shardedListWatch struct {
+	sharding *sharding
+	lw       cache.ListerWatcher
+}
+
+func NewShardedListWatch(shard, totalShards uint64, lw cache.ListerWatcher) cache.ListerWatcher {
+	// This is an "optimization" as this configuration means no sharding is to
+	// be performed.
+	if shard == 0 && totalShards == 1 {
```
mxinden: I guess if sharding is not set in `main.go`, these two would default to zero. Don't we want to just do `return lw` in that case? Hope I am not missing something.

pkg/sharding/listwatch.go
```diff
@@ -0,0 +1,88 @@
...
+type sharding struct {
+	shard, totalShards uint64
+}
+
+func (s *sharding) keep(o metav1.Object) bool {
+	return sum64(md5.Sum([]byte(o.GetUID())))%s.totalShards == s.shard
+}
+
+// sum64 sums the md5 hash to an uint64.
+func sum64(hash [md5.Size]byte) uint64 {
```
mxinden: If we use the golang/hash package, we could use the `Sum64` interface function instead, right?

```golang
type Hash64 interface {
        Hash
        Sum64() uint64
}
```

compare-sharding.sh
```diff
@@ -0,0 +1,9 @@
+#!/usr/bin/bash
```
mxinden: I would prefer this script not to be on the root of the repo. How about the `tests` directory instead?

README.md
brancz: kube-state-metrics needs to know the total number of replicas, that’s what it requests the statefulset for. Detecting it’s nominal within it is done as you said.

mxinden: Right. Sorry about that. And this total # can not be templated from the statefulset status as an environment variable?

brancz: The downward API only works for data within the pod object, not higher level objects.

main.go
```diff
@@ -272,3 +287,37 @@ func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+func detectSharding(kubeClient clientset.Interface, podName, namespaceName string) (shard, totalShards uint64, err error) {
+	glog.Infof("Auto detecting shard.")
+	p, err := kubeClient.CoreV1().Pods(namespaceName).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return 0, 0, err
+	}
+
+	owners := p.GetOwnerReferences()
+	for _, o := range owners {
+		if o.APIVersion == "apps/v1" && o.Kind == "StatefulSet" {
```
zouyee: wether we need to detect pod actual Controller?

s-urbaniak: stylistic nit to reduce the amount of indentation:
```
if o.APIVersion != "apps/v1" || o.Kind != "StatefulSet" {
  continue
}
// unindented logic
```

main.go
```diff
@@ -272,3 +287,37 @@ func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		closer.Close()
 	}
 }
+
+func detectSharding(kubeClient clientset.Interface, podName, namespaceName string) (shard, totalShards uint64, err error) {
+	glog.Infof("Auto detecting shard.")
+	p, err := kubeClient.CoreV1().Pods(namespaceName).Get(podName, metav1.GetOptions{})
+	if err != nil {
+		return 0, 0, err
```
s-urbaniak: nit: let's add some context to indicate where we are, i.e. `fmt.Errorf("error detecting shard: %v", err)`

main.go
```diff
@@ -272,3 +287,37 @@ func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	}
+
+	owners := p.GetOwnerReferences()
+	for _, o := range owners {
+		if o.APIVersion == "apps/v1" && o.Kind == "StatefulSet" {
+			ss, err := kubeClient.AppsV1().StatefulSets(namespaceName).Get(o.Name, metav1.GetOptions{})
+			if err != nil {
+				return 0, 0, err
```
s-urbaniak: same here, i.e.`fmt.Errorf("error detecting shard statefulset: %v", err)`

main.go
```diff
@@ -272,3 +287,37 @@ func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+			}
+
+			nominalString := strings.TrimLeft(ss.Name+"-", p.Name)
+			nominal, err := strconv.Atoi(nominalString)
+			if err != nil {
+				return 0, 0, err
```
s-urbaniak: same here, i.e. `fmt.Errorf("error detecting shard index: %v", err)`

pkg/sharding/listwatch.go
s-urbaniak: Choosing the right hash function, woah ;-) I am literally not an expert, but I think the focus is here on good key distribution and speed. When we talk about "speed", the amount of data is not super critical here I guess as there are not that many objects to be calculated per second.

While looking at https://github.com/golang/groupcache/blob/c65c006176ff7ff98bb916961c7abbc6b0afc0aa/consistenthash/consistenthash.go#L42, crc32 might be fine, but https://medium.com/@dgryski/consistent-hashing-algorithmic-tradeoffs-ef6b8e2fcae8 points out some pros/cons.

pkg/sharding/listwatch.go
brancz: shard defaults to 0 and total shards defaults to 1

pkg/sharding/listwatch.go
brancz: I chose this exact hashing mechanism as it's been well proven in the Prometheus `hashmod` system.

main.go
brancz: Not strictly necessary I think, it's a heuristic anyways, as theoretically there could be any number of statefulsets in the ownerreferences, and we just pick the first occurrence. I'll add it anyways though.

brancz:
I addressed the comments, but this also needs a reconcile loop, in cases where the statefulset is resized. I'd say we address that in a follow up as it requires a larger refactoring of the main.go in order to allow reloads of internals (including proper handling and passing of context).

mxinden:
Sounds good @brancz.

kubernetes/autosharding/kube-state-metrics-statefulset.yaml
```diff
@@ -0,0 +1,44 @@
...
+  replicas: 1
+  template:
+    metadata:
+      labels:
+        k8s-app: kube-state-metrics
+    spec:
+      serviceAccountName: kube-state-metrics
+      containers:
+      - name: kube-state-metrics
+        image: quay.io/coreos/kube-state-metrics:v1.4.0
```
tariq1890:
```suggestion
        image: quay.io/coreos/kube-state-metrics:v1.5.0
```

tariq1890: updated my review comment.

main.go
```diff
@@ -28,14 +28,20 @@ import (
 	"os"
 	"strconv"
 	"strings"
+	"sync"
 
+	"github.com/golang/glog"
```
tariq1890: We should be using `klog`.

main.go
```diff
@@ -243,11 +253,105 @@ func serveMetrics(collectors []*coll.Collector, host string, port int, enableGZI
...
+	defer m.mtx.Unlock()
+
+	if m.cancel != nil {
+		m.cancel()
+	}
+	if totalShards == 1 {
+		glog.Infof("sharding disabled")
+	}
+	if totalShards > 1 {
+		glog.Infof("configuring sharding of this instance to be shard %d out of %d", shard, totalShards)
```
tariq1890:
```suggestion
		klog.Infof("configuring sharding of this instance to be shard %d out of %d", shard, totalShards)
```

main.go
```diff
@@ -243,11 +253,105 @@ func serveMetrics(collectors []*coll.Collector, host string, port int, enableGZI
...
+				return
+			}
+
+			if old.ResourceVersion == cur.ResourceVersion {
+				return
+			}
+
+			shard, totalShards, err := shardingSettingsFromStatefulSet(ss, m.opts.Pod)
+			if err != nil {
+				glog.Errorf("detect sharding settings from StatefulSet: %v", err)
```
tariq1890:
```suggestion
				klog.Errorf("detect sharding settings from StatefulSet: %v", err)
```

main.go
```diff
@@ -243,11 +253,105 @@ func serveMetrics(collectors []*coll.Collector, host string, port int, enableGZI
...
+func (m *metricHandler) Run(ctx context.Context) error {
+	autoSharding := len(m.opts.Pod) > 0 && len(m.opts.Namespace) > 0
+
+	if !autoSharding {
+		m.configureSharding(ctx, m.opts.Shard, m.opts.TotalShards)
+		<-ctx.Done()
+		return ctx.Err()
+	}
+
+	glog.Infof("Auto detecting sharding settings.")
```
tariq1890:
```suggestion
	klog.Infof("Auto detecting sharding settings.")
```

main.go
```diff
@@ -243,11 +253,105 @@ func serveMetrics(collectors []*coll.Collector, host string, port int, enableGZI
...
+	i.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(o interface{}) {
+			ss := o.(*appsv1.StatefulSet)
+			if ss.Name != statefulSetName {
+				return
+			}
+
+			shard, totalShards, err := shardingSettingsFromStatefulSet(ss, m.opts.Pod)
+			if err != nil {
+				glog.Errorf("detect sharding settings from StatefulSet: %v", err)
```
tariq1890:
```suggestion
				klog.Errorf("detect sharding settings from StatefulSet: %v", err)
```

main.go
```diff
@@ -276,3 +380,42 @@ func (m *metricHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+
+		ss, err := kubeClient.AppsV1().StatefulSets(namespaceName).Get(o.Name, metav1.GetOptions{})
+		if err != nil {
+			return nil, errors.Wrapf(err, "retrieve shard's StatefulSet: %s/%s", namespaceName, o.Name)
+		}
+
+		return ss, nil
+	}
+
+	return nil, fmt.Errorf("no suitable statefulset found for auto detecting sharding for Pod %s/%s", namespaceName, podName)
```
tariq1890:
```suggestion
	return nil, errors.Errorf("no suitable statefulset found for auto detecting sharding for Pod %s/%s", namespaceName, podName)
```

pkg/options/options.go
```diff
@@ -82,6 +86,13 @@ func (o *Options) AddFlags() {
 	o.flags.StringVar(&o.TelemetryHost, "telemetry-host", "0.0.0.0", `Host to expose kube-state-metrics self metrics on.`)
 	o.flags.Var(&o.Collectors, "collectors", fmt.Sprintf("Comma-separated list of collectors to be enabled. Defaults to %q", &DefaultCollectors))
 	o.flags.Var(&o.Namespaces, "namespace", fmt.Sprintf("Comma-separated list of namespaces to be enabled. Defaults to %q", &DefaultNamespaces))
+	o.flags.Uint64Var(&o.Shard, "shard", uint64(0), "The instances shard nominal (zero indexed) within the total number of shards. (default 0)")
+	o.flags.Uint64Var(&o.TotalShards, "total-shards", uint64(1), "The total number of shards. Sharding is disabled when total shards is set to 1.")
+
+	autoshardingNotice := "When set, it is expected that --pod and --pod-namespace are both set. This is used for auto-detecting sharding. If set this has preference over statically configured sharding. This is exprimental, it may be removed without notice."
```
tariq1890:
```suggestion
	autoshardingNotice := "When set, it is expected that --pod and --pod-namespace are both set. This is used for auto-detecting sharding. If set, this has preference over statically configured sharding. This is experimental, it may be removed without notice."
```

brancz:
I have yet to address all comments here, but autosharding including re-sharding seems to work now! :tada: I have yet to also add e2e tests for this. But it's progressing.

brancz:
I finally figured out why the benchmark seemed so much worse with these patches even though there is no change in the http serving parts: the go compiler can do unfair optimizations with the code on master. Once we change that, the result becomes what I would have expected. I opened #784 to address this.

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,244 @@
...
+			if shardingUnchanged {
+				return
+			}
+
+			m.ConfigureSharding(ctx, shard, totalShards)
+		},
+	})
+	go i.Run(ctx.Done())
+	if !cache.WaitForCacheSync(ctx.Done(), i.HasSynced) {
+		return fmt.Errorf("waiting for informer cache to sync failed")
```
tariq1890:
```suggestion
		return errors.New("waiting for informer cache to sync failed")
```

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,244 @@
...
+	}
+
+	i := cache.NewSharedIndexInformer(
+		cache.NewFilteredListWatchFromClient(m.kubeClient.AppsV1().RESTClient(), "statefulsets", m.opts.Namespace, labelSelectorOptions),
+		&appsv1.StatefulSet{}, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc},
+	)
+	i.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(o interface{}) {
+			ss := o.(*appsv1.StatefulSet)
+			if ss.Name != statefulSetName {
```
tariq1890: Let's address the code duplication in AddFunc and UpdateFunc.

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,244 @@
...
+	nominal, err := strconv.Atoi(nominalString)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to detect shard index for Pod %s of StatefulSet %s, parsed %v", podName, statefulSetName, nominalString)
+	}
+
+	return int32(nominal), nil
+}
+
+func detectStatefulSet(kubeClient kubernetes.Interface, podName, namespaceName string) (*appsv1.StatefulSet, error) {
+	p, err := kubeClient.CoreV1().Pods(namespaceName).Get(podName, metav1.GetOptions{})
```
tariq1890: Is it possible to use a field selector within `metav1.GetOptions`so that pods could be pre-filtered by client-go?

pkg/metricshandler/metrics_handler.go
brancz: I don't think this is worth it, and the "skip" logic is slightly different for add and update, as in update we also get to compare to the previous object. I think this duplication is ok in its current extend.

pkg/metricshandler/metrics_handler.go
brancz: we already select exact pods, and not list

brancz:
That minikube failure does not look good. I think we'll be seeing that on all builds.

brancz:
Yep, we're seeing it everywhere. This is a master build with the same failure: https://travis-ci.org/kubernetes/kube-state-metrics/jobs/562200973

pkg/options/options.go
```diff
@@ -37,6 +37,10 @@ type Options struct {
 	TelemetryHost                        string
 	Collectors                           CollectorSet
 	Namespaces                           NamespaceList
+	Shard                                int32
+	TotalShards                          int
+	Pod                                  string
+	Namespace                            string
```
tariq1890: Let's rename this to `PodNamespace`. Since we already have `Namespaces`, `PodNamespace` would be a bit more clearer to the uninformed reader.

docs/cli-arguments.md
```diff
@@ -42,12 +42,16 @@ Usage of ./kube-state-metrics:
       --metric-blacklist string                     Comma-separated list of metrics not to be enabled. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.
       --metric-whitelist string                     Comma-separated list of metrics to be exposed. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.
       --namespace string                            Comma-separated list of namespaces to be enabled. Defaults to ""
+      --pod string                                  Name of the pod. When set, it is expected that --pod and --pod-namespace are both set. This is used for auto-detecting sharding. If set, this has preference over statically configured sharding. This is experimental, it may be removed without notice.
```
tariq1890: Name of the pod isn't too clear. We need to mention that this is the pod that corresponds to the ksm binary itself.

Thoughts on this - "Name of the pod that contains the kube-state-metrics container"?

docs/cli-arguments.md
```diff
@@ -42,12 +42,16 @@ Usage of ./kube-state-metrics:
...
+      --pod-namespace string                        Name of the namespace. When set, it is expected that --pod and --pod-namespace are both set. This is used for auto-detecting sharding. If set, this has preference over statically configured sharding. This is experimental, it may be removed without notice.
```
tariq1890: Isn't it safe to assume the default pod and pod namespaces names to be - `kube-state-metrics` and `kube-system` respectively? 

If so, we can remove the requirement for both pod name and pod namespace to be set.

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	if !autoSharding {
+		m.ConfigureSharding(ctx, m.opts.Shard, m.opts.TotalShards)
+		<-ctx.Done()
+		return ctx.Err()
+	}
+
+	klog.Infof("Auto detecting sharding settings.")
+	ss, err := detectStatefulSet(m.kubeClient, m.opts.Pod, m.opts.Namespace)
+	if err != nil {
+		return errors.Wrap(err, "detect StatefulSet")
```
tariq1890: I see several instances of these error message styles. Can that be fixed everywhere?
```suggestion
		return errors.Wrap(err, "error when attempting to detect StatefulSet")
```

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	// In case we gzipped the response, we have to close the writer.
+	if closer, ok := writer.(io.Closer); ok {
+		closer.Close()
+	}
+}
+
+func shardingSettingsFromStatefulSet(ss *appsv1.StatefulSet, podName string) (nominal int32, totalReplicas int, err error) {
+	nominal, err = detectNominalFromPod(ss.Name, podName)
+	if err != nil {
+		return 0, 0, errors.Wrap(err, "detecting Pod nominal")
```
tariq1890: Do we need to wrap the error here?

tests/compare-sharding.sh
```diff
@@ -0,0 +1,24 @@
...
+kubectl -n kube-system port-forward pod/kube-state-metrics-1 8084:8080 &
+kubectl -n kube-system port-forward pod/kube-state-metrics-2 8086:8080 &
+
+sleep 3
+
+RESULT_UNSHARDED="$(mktemp)"
+RESULT_SHARDED_UNSORTED="$(mktemp)"
+RESULT_SHARDED="$(mktemp)"
+
+curl localhost:8080/metrics | grep -v "^#" | sort | tee "${RESULT_UNSHARDED}"
```
tariq1890: can we have a ports array and loop through that?

tests/compare-sharding.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/bash
+
+trap 'kill 0' SIGTERM
+
+kubectl -n kube-system port-forward deployment/kube-state-metrics 8080:8080 &
```
tariq1890: Let's move `kube-system` and `kube-state-metrics` to a variable. This way, we are able to correlate the `pod_name` and `pod_namespace` parameters specified in the README with those mentioned in this test.

tests/compare-sharding.sh
```diff
@@ -0,0 +1,24 @@
...
+RESULT_UNSHARDED="$(mktemp)"
+RESULT_SHARDED_UNSORTED="$(mktemp)"
+RESULT_SHARDED="$(mktemp)"
+
+curl localhost:8080/metrics | grep -v "^#" | sort | tee "${RESULT_UNSHARDED}"
+curl localhost:8082/metrics | grep -v "^#" | tee "${RESULT_SHARDED_UNSORTED}"
+curl localhost:8084/metrics | grep -v "^#" | tee -a "${RESULT_SHARDED_UNSORTED}"
+curl localhost:8086/metrics | grep -v "^#" | tee -a "${RESULT_SHARDED_UNSORTED}"
+
+sort "$RESULT_SHARDED_UNSORTED" | tee "${RESULT_SHARDED}"
```
tariq1890:
```suggestion
sort "${RESULT_SHARDED_UNSORTED}" | tee "${RESULT_SHARDED}"
```

tests/compare-sharding.sh
```diff
@@ -0,0 +1,24 @@
...
+
+diff <(echo "$RESULT_UNSHARDED") <(echo "$RESULT_SHARDED")
```
tariq1890:
```suggestion
diff <(echo "${RESULT_UNSHARDED}") <(echo "${RESULT_SHARDED}")
```

tariq1890:
@brancz Can you rebase?

@LiliC Requesting your eyes on this as well.

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	m.storeBuilder.WithContext(ctx)
+	m.stores = m.storeBuilder.Build()
+	m.curShard = shard
+	m.curTotalShards = totalShards
+}
+
+func (m *MetricsHandler) Run(ctx context.Context) error {
+	autoSharding := len(m.opts.Pod) > 0 && len(m.opts.Namespace) > 0
+
+	klog.Infof("autosharding enabled: %t; pod=%v; pod_namespace=%v", autoSharding, m.opts.Pod, m.opts.Namespace)
```
LiliC: Hmm not sure I like that this log is always logging. Also is it true that autosharding is enabled, how do we know that? From the options warning, it says you need to enable it? Can you elaborate, thanks! 

Nits: Also can we do the following:
- set this log level to debug
- default values for those fields are empty strings, meaning this msg will not make much sense.

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	if !autoSharding {
```
LiliC: Hmm should this not be before the log?

pkg/options/options.go
```diff
@@ -84,6 +88,13 @@ func (o *Options) AddFlags() {
...
 	o.flags.Var(&o.MetricWhitelist, "metric-whitelist", "Comma-separated list of metrics to be exposed. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.")
 	o.flags.Var(&o.MetricBlacklist, "metric-blacklist", "Comma-separated list of metrics not to be enabled. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.")
+	o.flags.Int32Var(&o.Shard, "shard", int32(0), "The instances shard nominal (zero indexed) within the total number of shards. (default 0)")
+	o.flags.IntVar(&o.TotalShards, "total-shards", 1, "The total number of shards. Sharding is disabled when total shards is set to 1.")
+
+	autoshardingNotice := "When set, it is expected that --pod and --pod-namespace are both set. This is used for auto-detecting sharding. If set, this has preference over statically configured sharding. This is experimental, it may be removed without notice."
+
+	o.flags.StringVar(&o.Pod, "pod", "", "Name of the pod. "+autoshardingNotice)
+	o.flags.StringVar(&o.Namespace, "pod-namespace", "", "Name of the namespace. "+autoshardingNotice)
```
LiliC: Will this maybe `""` default to `default` namespace, because k8s? 🤔 :)

brancz: It doesn’t, that’s kubectl Logic doing that. Here we’re referencing a concrete object so this isn’t correct.

docs/cli-arguments.md
brancz: We can't assume that I think and I'd prefer to have to explicitly set this instead of a default. I feel a default would be more error prone. I prefer to handle this via documented manifests.

pkg/metricshandler/metrics_handler.go
brancz: The way it is is best practice. We're essentially building a stack trace.

tests/compare-sharding.sh
brancz: I don't think we can, as they are slightly different each time. Also I like the explicitness, and it's just a script for manual verification.

tests/compare-sharding.sh
brancz: hmm .. how strongly do you feel about this, I quite like the explicitness, and feel introducing variables would make this harder to understand

kubernetes/autosharding/kube-state-metrics-cluster-role.yaml
```diff
@@ -0,0 +1,60 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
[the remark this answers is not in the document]
brancz: The autosharding manifests use a statefulset instead of deployment to allow the identification of a nominal within a statefulset (as in the "self replica" out of the "total replicas"). As  well as the permissions for the pods to discover their owning statefulset.

pkg/metricshandler/metrics_handler.go
brancz: changed the log line to be less confusing

pkg/metricshandler/metrics_handler.go
brancz: It's only logged once at startup and is very valuable information, I'd prefer to keep it at info level. Otherwise I split the log line into two (one enabled, one disabled) to make it less confusing.

kubernetes/autosharding/kube-state-metrics-cluster-role-binding.yaml
```diff
@@ -0,0 +1,13 @@
+apiVersion: rbac.authorization.k8s.io/v1
+# kubernetes versions before 1.8.0 should use rbac.authorization.k8s.io/v1beta1
```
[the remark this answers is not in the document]
brancz: removed :)

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,243 @@
...
+	cancel func()
+
+	// mtx protects stores, curShard, and curTotalShards
+	mtx            *sync.RWMutex
+	stores         []*metricsstore.MetricsStore
+	curShard       int32
+	curTotalShards int
+}
+
+func New(opts *options.Options, kubeClient kubernetes.Interface, storeBuilder *store.Builder, enableGZIPEncoding bool) *MetricsHandler {
```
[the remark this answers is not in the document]
brancz: Done.

kubernetes/autosharding/kube-state-metrics-cluster-role.yaml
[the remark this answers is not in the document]
brancz: Let me know if it still needs to be clarified elsewhere. I added a comment to the readme referencing the manifests.

docs/cli-arguments.md
brancz: Yes I think that's a good idea. I also adapted the comment to additionally say that people are most likely going to need to do this via the downward api.

kubernetes/autosharding/kube-state-metrics-cluster-role.yaml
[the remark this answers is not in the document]
brancz: Discussed out of band: we separated the "regular" deployment option and autosharding into separate directories.

brancz:
I believe I've fixed all comments and or replied to them and CI is green :green_apple: . Please take another look. :slightly_smiling_face:

pkg/metricshandler/metrics_handler.go
```diff
@@ -0,0 +1,257 @@
...
+	}
+
+	return nominal, totalReplicas, nil
+}
+
+func detectNominalFromPod(statefulSetName, podName string) (int32, error) {
+	nominalString := strings.TrimPrefix(podName, statefulSetName+"-")
+	nominal, err := strconv.Atoi(nominalString)
+	if err != nil {
+		return 0, errors.Wrapf(err, "failed to detect shard index for Pod %s of StatefulSet %s, parsed %v", podName, statefulSetName, nominalString)
```
tariq1890: Let's use the string directive here.
```suggestion
		return 0, errors.Wrapf(err, "failed to detect shard index for Pod %s of StatefulSet %s, parsed %s", podName, statefulSetName, nominalString)
```

tests/compare-sharding.sh
tariq1890: I can see that. Works for me.

pkg/metricshandler/metrics_handler.go
tariq1890: Sure.

tariq1890:
@brancz I think we should treat the moving of k8s manifests to a different folder as a breaking change. I would prefer if we kept the current directory setup as-is and created a separate folder for the autosharding scenarios.

I know that there are quite a few blogs and books that rely on the assumption that k8s manifests are located in the `kubernetes` folder.

tests/compare-sharding.sh
tariq1890: I can agree with that.

brancz:
I think you're right. I reverted the regular deployment manifests to stay in the same place. PTAL :slightly_smiling_face:

kubernetes/autosharding/kube-state-metrics-statefulset.yaml
```diff
@@ -0,0 +1,47 @@
...
+          valueFrom:
+            fieldRef:
+              apiVersion: v1
+              fieldPath: metadata.namespace
+        ports:
+        - name: http-metrics
+          containerPort: 8080
+        - name: telemetry
+          containerPort: 8081
+        readinessProbe:
```
tariq1890: Can we add `livenessProbe` too?

kubernetes/autosharding/kube-state-metrics-statefulset.yaml
```diff
@@ -0,0 +1,53 @@
...
+          containerPort: 8081
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        readinessProbe:
+          httpGet:
+            path: /healthz
```
tariq1890: the readiness should hit the base correct?
```suggestion
            path: /
```

tariq1890:
/lgtm

Thanks @brancz! Great to have this merged :).

PR #616, opened 2018-12-11
Add labels in Exposed Metrics documentation

MIBc:
Documentation lacks some labels in node and pod metrics .


andyxning:
@MIBc Thanks for updating this. @mxinden Maybe we need to come up with a way to extract those labels automatically. That can help us to decrease the maintanence burden. 

Also, i think we can merge this since it is only related with docs update.

PR #617, opened 2018-12-11
*: Cut v1.5.0-beta.0

mxinden:

Cut v1.5.0-beta.0 of kube-state-metrics. I am not entirely sure what the time between alpha and beta was in the past but given that there are no bug reports so far I would suggest moving on to a beta release.

What are your thoughts?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+## v1.5.0-beta.0 / 2018-12-11
+
+After a testing period of 11 days, there were no additional bugs found or features introduced.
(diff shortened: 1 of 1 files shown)


brancz:
Sounds good!

/lgtm

brancz:
The image has been published to `quay.io/coreos/kube-state-metrics:v1.5.0-beta.0`.

mxinden:
@brancz wow, thanks for the quick release!

andyxning:
Great!

PR #618, opened 2018-12-13
Clear TODO comment in builder.go

MIBc:
Clear comment --- "TODO : What if not ok?". Because the collector's name has been checked in CollectorSet.Set function. It will exit if collector' name is not in DefaultCollectors. 
As a result, all collectors in Build.enabledCollectors will find its constructor function. 
So I think it doesn't exist "not ok" situation.

### DIFF
--- a/pkg/collectors/builder.go
+++ b/pkg/collectors/builder.go
@@ -107,7 +107,6 @@ func (b *Builder) Build() []*Collector {
 			activeCollectorNames = append(activeCollectorNames, c)
 			collectors = append(collectors, collector)
 		}
-		// TODO: What if not ok?
 	}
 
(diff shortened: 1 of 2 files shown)


pkg/collectors/builder.go
```diff
@@ -107,7 +107,6 @@ func (b *Builder) Build() []*Collector {
 			activeCollectorNames = append(activeCollectorNames, c)
 			collectors = append(collectors, collector)
 		}
-		// TODO: What if not ok?
```
mxinden: Good catch!

MIBc:
I will squash it .

MIBc:
@mxinden  I have squashed commit.

mxinden:
Cool, thanks @MIBc. @brancz @andyxning would one of you mind merging?

PR #633, opened 2019-01-16
Deduplicate metric name

mxinden:

Instead of defining the metric name both in the metric family generator
as well as the metric itself, with this patch the metric name is
injected into the metric family by the metric family generator.

### DIFF
--- a/internal/collector/configmap.go
+++ b/internal/collector/configmap.go
@@ -37,12 +37,11 @@ var (
 			Help: "Information about configmap.",
 			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) metric.Family {
 				return metric.Family{
-					&metric.Metric{
-						Name:        "kube_configmap_info",
+					Metrics: []*metric.Metric{{
 						LabelKeys:   []string{},
 						LabelValues: []string{},
 						Value:       1,
-					},
+					}},
 				}
 			}),
 		},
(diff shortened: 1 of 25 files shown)


mxinden:
> Out of curiosity have you found any mistakes while doing this due to the existing duplication? I'd also be interested in diffing this against v1.5.0.

@brancz I have not noticed any inconsistencies, but I have also not diffed it against a kube-state-metrics 1.5.

Collector tests stayed the same, hence I don't expect anything to not be aligned.

brancz:
What are your thoughts on maybe adding some micro-benchmark output as a CI job? It's hard to tell wether changes like this might have a large or minor impact on metrics generation/scraping (in this case I'd guess it's more about generation, but still just a guess until measured :slightly_smiling_face: ). Generally happy to move forward with this though, as it majorly improves the codebase readability and probability for mistakes.

/lgtm

mxinden:
> What are your thoughts on maybe adding some micro-benchmark output as a CI job? It's hard to tell wether changes like this might have a large or minor impact on metrics generation/scraping (in this case I'd guess it's more about generation, but still just a guess until measured slightly_smiling_face ). Generally happy to move forward with this though, as it majorly improves the codebase readability and probability for mistakes.

@brancz what do you think of https://github.com/kubernetes/kube-state-metrics/pull/643 ?

PR #635, opened 2019-01-17
clarify the import location of Store interface in comments

tariq1890:
Minor correction in comments to indicate that we are importing from client-go and not upstream k8s :)


brancz:
Thanks

/lgtm

PR #638, opened 2019-01-22
Add replicaset label metrics

kminehart:
**What this PR does / why we need it**: This PR adds the `kube_replicaset_labels` metric. In some edge case scenarios this can be useful.

In our environment, we heavily use labels for tenancy. Developers want to be able to see how often they "deploy" an update; this can normally be done using `kube_deployment_created`, unless the deployment is updated. When the deployment is updated, a new replicaset is created, so we use `kube_replicaset_created`, but we can not determine what team owns the replicaset without being able to access its labels

I'm having minikube issues so I'm unable to run the e2e tests locally :sweat_smile:


kminehart:
I'm unsure why the travis build failed, it does not seem relevant to the changes I've made.

/assign @mxinden

brancz:
> I'm having minikube issues so I'm unable to run the e2e tests locally :sweat_smile: 

Please let us know how we can improve docs to make this easier for the next contributor :)

/lgtm

PR #642, opened 2019-01-27
remove dependency on kubernetes/kubernetes

tariq1890:
It is considered best practice to not rely on kubernetes/kubernetes as a dependency as it's not really meant for external consumption.

Moreover, the methods and variables imported in this project are simple logic which can just reside in the project itself IMO.

This PR addresses that


internal/collector/pod.go
```diff
@@ -228,8 +226,8 @@ var (
 					{phase == v1.PodFailed, string(v1.PodFailed)},
 					// This logic is directly copied from: https://github.com/kubernetes/kubernetes/blob/d39bfa0d138368bbe72b0eaf434501dcb4ec9908/pkg/printers/internalversion/printers.go#L597-L601
 					// For more info, please go to: https://github.com/kubernetes/kube-state-metrics/issues/410
-					{phase == v1.PodRunning && !(p.DeletionTimestamp != nil && p.Status.Reason == node.NodeUnreachablePodReason), string(v1.PodRunning)},
-					{phase == v1.PodUnknown || (p.DeletionTimestamp != nil && p.Status.Reason == node.NodeUnreachablePodReason), string(v1.PodUnknown)},
+					{phase == v1.PodRunning && !(p.DeletionTimestamp != nil && p.Status.Reason == "NodeLost"), string(v1.PodRunning)},
```
brancz: let's move `"NodeLost"` into a constant

tariq1890: Sure

internal/collector/utils.go
```diff
@@ -84,3 +88,32 @@ func kubeAnnotationsToPrometheusAnnotations(annotations map[string]string) ([]st
 func sanitizeLabelName(s string) string {
```
zouyee: It maybe better to add related unit test.

tariq1890: @zouyee @brancz ?

brancz: I think @zouyee meant to transfer the unit tests relevant from what is currently the dependency as well. I agree that would be good.

tariq1890: I am okay with doing that. But that would be potentially adding a lot of redundant code since the test coverage is already high with the current test suite. I can add extra units fill the existing gaps. If the redundant tests are okay, then I can proceed with porting over the test cases from k/k to here. Thoughts?

brancz: I prefer the redundant tests.

tariq1890: Sounds good! Thanks for letting me know  :)

tariq1890: I will update by PR once I get the time today.

tariq1890: Done!

tariq1890:
I have ported over the tests from k/k to here.

PR #644, opened 2019-01-28
pkg/metric/generator: Refactor GenerateFunc call and return pointer

mxinden:

**pkg/metric/generator: Introduce Generate method calling GenerateFunc**

Instead of making an external component responsible of adding the family name to
each family, enforce this in the generator via the `Generate` function.

**pkg/metric/generator: Make GenerateFunc return pointer to Family**

Instead of returning a Family right away, make GenerateFunc return a pointer to
a Family, matching the convention across the k8s project and removing the need
to do so later in the code path.

### DIFF
--- a/internal/collector/configmap.go
+++ b/internal/collector/configmap.go
@@ -35,8 +35,8 @@ var (
 			Name: "kube_configmap_info",
 			Type: metric.MetricTypeGauge,
 			Help: "Information about configmap.",
-			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) metric.Family {
-				return metric.Family{
+			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) *metric.Family {
+				return &metric.Family{
 					Metrics: []*metric.Metric{{
 						LabelKeys:   []string{},
 						LabelValues: []string{},
@@ -49,7 +49,7 @@ var (
 			Name: "kube_configmap_created",
 			Type: metric.MetricTypeGauge,
 			Help: "Unix creation timestamp",
-			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) metric.Family {
+			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) *metric.Family {
 				ms := []*metric.Metric{}
 
 				if !c.CreationTimestamp.IsZero() {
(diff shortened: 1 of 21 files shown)


mxinden:
@LiliC @wanghaoran1988 what are your thoughts on this?

Updating your downstreams shouldn't be that bad:

```bash
find *.go -type f -exec sed -i -e 's/metric\.Family)/*metric.Family)/g' {} \;
find *.go -type f -exec sed -i -e 's/metric\.Family /*metric.Family /g' {} \;
find *.go -type f -exec sed -i -e 's/metric\.Family{/\&metric.Family{/g' {} \;
```

wanghaoran1988:
LGTM

internal/collector/configmap.go
```diff
@@ -35,8 +35,8 @@ var (
 			Name: "kube_configmap_info",
 			Type: metric.MetricTypeGauge,
 			Help: "Information about configmap.",
-			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) metric.Family {
-				return metric.Family{
+			GenerateFunc: wrapConfigMapFunc(func(c *v1.ConfigMap) *metric.Family {
```
LiliC: Think we said at some point we also want to rename this `GenerateFunc` -> `Generator` or something along those lines?

brancz: I think I'd generally be in favor of this, I think the common thing to do in go is have the `Genreator` be an interface and then have `GenerateFunc` be a typed function that implements the `Generator` interface.

mxinden: @LiliC yes, we talked about renaming `GenerateFunc` to `Generate` so that a `FamilyGenerator` `Generate`s families.

Instead I chose to introduce the `Generate` method on the `FamilyGenerator` (see comments):

```golang
// Generate calls the FamilyGenerator.GenerateFunc and gives the family its
// name. The reasoning behind injecting the name at such a late point in time is
// deduplication in the code, preventing typos made by developers as
// well as saving memory.
func (g *FamilyGenerator) Generate(obj interface{}) *Family {
	family := g.GenerateFunc(obj)
	family.Name = g.Name
	return family
}
```

By doing so we have the advantage of:

> Instead of making an external component responsible of adding the family name to
each family, enforce this in the generator via the Generate function.

Not sure we need the `Generator` interface abstraction at this point. If so, we could still do so at caller side, enforcing the signature of the above cited `Generate` function.

@LiliC @brancz what do you think?

mxinden:
@brancz would you mind giving this another look?

PR #645, opened 2019-01-28
Enable use as kustomize base layer

mgoodness:

- Primarily adds a `kustomization.yaml` file to enable use as a [kustomize](https://github.com/kubernetes-sigs/kustomize) base layer.
- Also adds resources from the `apps` group to the Role and ClusterRole.
- Updates the Deployment to `apps/v1`

### DIFF
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -1,5 +1,5 @@
-apiVersion: apps/v1beta2
-# Kubernetes versions after 1.9.0 should use apps/v1
+apiVersion: apps/v1
+# Kubernetes version 1.8.x should use apps/v1beta2
 # Kubernetes versions before 1.8.0 should use apps/v1beta1 or extensions/v1beta1
 kind: Deployment
 metadata:
--- a/kubernetes/kustomization.yaml
+++ b/kubernetes/kustomization.yaml
@@ -1,12 +1,6 @@
 commonLabels:
   app.kubernetes.io/name: kube-state-metrics
 
-imageTags:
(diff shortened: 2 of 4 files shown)


kubernetes/kube-state-metrics-deployment.yaml
```diff
@@ -1,5 +1,5 @@
-apiVersion: apps/v1beta2
-# Kubernetes versions after 1.9.0 should use apps/v1
+apiVersion: apps/v1
+# Kubernetes version 1.8.x should use apps/v1beta2
```
tariq1890:
```suggestion
# Kubernetes versions after 1.8.x should use apps/v1beta2
```

mgoodness: > Kubernetes versions after 1.8.x should use apps/v1beta2

Doesn't that imply that versions 1.8 and higher should use `apps/v1beta2`, when it's really only the 1.8 series? Versions 1.9+ should use (the default) `apps/v1`.

kubernetes/kustomization.yaml
```diff
@@ -0,0 +1,19 @@
+commonLabels:
+  app.kubernetes.io/name: kube-state-metrics
+
+imageTags:
+- name: quay.io/coreos/kube-state-metrics
+  newTag: v1.5.0
```
brancz: Do these actually need to be specified here? I'd prefer if we don't have more places we need to update these at.

mgoodness: They're not necessary; I'll delete 'em.

bgrant0607:
cc @monopole

brancz:
Thanks for adding this!

/lgtm

PR #646, opened 2019-01-29
Get GCP email account from gcloud's 'value' format

eyalev:
https://cloud.google.com/sdk/gcloud/reference/topic/formats


brancz:
Thanks!

PR #650, opened 2019-01-31
Add collector for csr

### LINKED ISSUE #649: Add collector for CSR

/kind feature
As a SRE, I'd like monitor the csr status, like if it's pending for a long time, and not approved, there will be problem, eg. the node create csr for it's client/server certs, pending will cause the node outage.

**What happened**:

**Environment**:

wanghaoran1988:

Fixes #649


internal/collector/csr.go
```diff
@@ -0,0 +1,147 @@
...
+		},
+		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
+			return kubeClient.CertificatesV1beta1().CertificateSigningRequests().Watch(opts)
+		},
+	}
+}
+
+// addCSRConditionMetrics generates one metric for each possible csr condition status
+func addCSRConditionMetrics(cs certv1beta1.CertificateSigningRequestStatus) []*metric.Metric {
+	approved := false
```
mxinden: A CSR can either be approved or denied, right? If so, I would encode this here by only having either a `approved` or `denied` boolean, and derive the opposite via the `!` operator.

internal/collector/csr.go
```diff
@@ -0,0 +1,147 @@
...
+	denied := false
+	if len(cs.Conditions) > 0 {
+		for _, s := range cs.Conditions {
```
mxinden: Just to make sure, if a CSR has multiple conditions, do all of the apply or only the most recent one?

wanghaoran1988: If it has multiple condition, then do all the apply, it can has "approve" and "denied" condition at the same time now.

internal/collector/csr.go
brancz: yes, this should be treated as an enum metric

internal/collector/csr.go
```diff
@@ -0,0 +1,151 @@
...
+				approved = true
+			}
+			if s.Type == certv1beta1.CertificateDenied {
+				denied = true
+			}
+		}
+	}
+	//"approved" if a certificate request has the
+	// "Approved" condition and no "Denied" conditions; false otherwise.
+	approved = approved && !denied
```
brancz: interpretation of data should be done at query time, not in kube-state-metrics, we should just expose each state with a 0 or 1 depending on whether it is in that state

wanghaoran1988: A little confused, what do you mean by "query time" ? do you mean I can just remove this line?  when the csr  have "approved" and "denied" condition at the same time,  should I expose the metrics both 1 like this:
kube_csr_condition{csr="certificate-test",csr_condition="denied"} 1
kube_csr_condition{csr="certificate-test",csr_condition="approved"} 1

mxinden: @brancz if we don't interpret the data, there is the possibility for both approved and denied being true.

This upstream section seems to be the source of truth: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/certificate_controller_utils.go#L23

internal/collector/csr.go
```diff
@@ -0,0 +1,151 @@
...
+			return kubeClient.CertificatesV1beta1().CertificateSigningRequests().Watch(opts)
+		},
+	}
+}
+
+// addCSRConditionMetrics generates one metric for each possible csr condition status
+func addCSRConditionMetrics(cs certv1beta1.CertificateSigningRequestStatus) []*metric.Metric {
+	approved := false
+	denied := false
+	if len(cs.Conditions) > 0 {
```
mxinden: No need for this, if it is 0, the `for` loop will loop 0 times.

internal/collector/csr.go
```diff
@@ -0,0 +1,151 @@
...
+		for _, s := range cs.Conditions {
```
mxinden: How about using [`GetCertApprovalCondition`](https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/certificates/certificate_controller_utils.go#L28) here instead?

internal/collector/csr.go
mxinden: :+1:

internal/collector/csr.go
brancz: Correct, but here we want to expose the raw data. Essentially the `approved && !denied` is something that is up to a user to do in PromQL.

wanghaoran1988: This make sense, and it's how I did originally

internal/collector/csr.go
wanghaoran1988: good catch

internal/collector/csr.go
wanghaoran1988: I'd like a duplication here, or it will bring go dependency here, WDYT?

mxinden:
In regards to the _outdated_ discussion here: https://github.com/kubernetes/kube-state-metrics/pull/650#discussion_r253916475

If we want to expose the raw data, how about also exposing the count of denied and approved conditions instead of just `true` or `false`?

@brancz @wanghaoran1988 what do you think?

brancz:
Just `true` and `false`. Count is then something a user can do in PromQL :slightly_smiling_face: .

internal/collector/csr.go
```diff
@@ -0,0 +1,146 @@
...
+		}
+		if s.Type == certv1beta1.CertificateDenied {
+			denied = true
+		}
+	}
+
+	return []*metric.Metric{
+		&metric.Metric{
+			LabelValues: []string{"issued"},
+			Value:       boolFloat64(len(cs.Certificate) > 0),
```
brancz: this is a state assessed based on data, not raw data, this should be done in PromQL

internal/collector/csr.go
```diff
@@ -0,0 +1,146 @@
...
+			LabelKeys:   []string{"csr_condition"},
+		},
+		&metric.Metric{
+			LabelValues: []string{"pending"},
+			Value:       boolFloat64(!(approved || denied)),
```
brancz: Same as abbove

wanghaoran1988:
Yes, we can do the "count" in PromQL

mxinden:
I am not quite sure we are on the same page @brancz.

For the following CertificateSigningRequestStatus

``` golang
CertificateSigningRequestStatus {
    []Condition {
        { approved }, { approved }, { denied }, { denied }, { denied },
    }
}
```

I am suggesting to expose the number of approvals and denies like:

```
kube_csr_condition{ condition="approved" } 2
kube_csr_condition{ conditiion="denied" } 3
```

Instead of what we currently do:

```
kube_csr_condition{ condition="approved" } 1
kube_csr_condition{ conditiion="denied" } 1
```

What do you think?

wanghaoran1988:
> { approved }, { approved }, { denied }, { denied }, { denied },

@mxinden This will not happen, we can only have one for each like:  { approved }, { denied }

internal/collector/csr.go
wanghaoran1988: @brancz So how about expose a metrics kube_csr_cert_length ?

wanghaoran1988:
@mxinden Did some test with running "kubectl certificate approve/deny" many time against on one csr, finally, it only have one "approve" and one "denied" condition.

internal/collector/csr.go
brancz: that works :+1:

Documentation/README.md
```diff
@@ -62,7 +62,7 @@ Per group of metrics there is one file for each metrics. See each file for speci
 * [Secret Metrics](secret-metrics.md)
 * [ConfigMap Metrics](configmap-metrics.md)
 * [Ingress Metrics](ingress-metrics.md)
-
+* [CSR Metrics](csr-metrics.md)
```
tariq1890:
```suggestion
* [CertificateSigningRequest Metrics](csr-metrics.md)
```
Suggesting this, as the convention seems to be the resource type name rather than resource abbreviation

internal/collector/csr.go
```diff
@@ -0,0 +1,152 @@
+/*
```
tariq1890: This should be renamed to `certificatesigningrequest.go` IMO

internal/collector/csr.go
```diff
@@ -0,0 +1,152 @@
...
+	certv1beta1 "k8s.io/api/certificates/v1beta1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descCSRLabelsName          = "kube_csr_labels"
```
tariq1890:
```suggestion
	descCSRLabelsName          = "kube_certificatesigningrequest_labels"
```

Documentation/csr-metrics.md
```diff
@@ -0,0 +1,8 @@
+# CSR Metrics
+
+| Metric name| Metric type | Labels/tags | Status |
+| ---------- | ----------- | ----------- | ----------- |
+| kube_csr_created| Gauge | `csr`=&lt;csr-name&gt;| STABLE |
```
tariq1890:
```suggestion
| kube_certificatesigningrequest_created| Gauge | `csr`=&lt;csr-name&gt;| STABLE |
```

I understand that this may be concerning given the length, but I see instances like `kube_persistentvolumeclaim_resource_requests_storage_bytes`

internal/collector/builder.go
```diff
@@ -134,6 +135,7 @@ var availableCollectors = map[string]func(f *Builder) *coll.Collector{
 	"secrets":                  func(b *Builder) *coll.Collector { return b.buildSecretCollector() },
 	"services":                 func(b *Builder) *coll.Collector { return b.buildServiceCollector() },
 	"statefulsets":             func(b *Builder) *coll.Collector { return b.buildStatefulSetCollector() },
+	"csr":                      func(b *Builder) *coll.Collector { return b.buildCsrCollector() },
```
tariq1890:
```suggestion
	"certificatesigningrequests":                      func(b *Builder) *coll.Collector { return b.buildCsrCollector() },
```

tariq1890:
@mxinden @brancz I've added my non-maintainer review comments (apologize if it's inappropriate) for ensuring nomenclature consistency.

tariq1890:
@wanghaoran1988 make sure you run either `go mod tidy` or `go mod vendor`. That  would fix one of the travic CI fails :).

wanghaoran1988:
@tariq1890 Thanks, I am trying to fix the failure, the command you told me is really handy. I don't have any prefer of the metrics name, if @mxinden agree, I can change name  to "certificatesigningrequests".

internal/collector/csr.go
```diff
@@ -0,0 +1,152 @@
...
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
+			Name: "kube_csr_condition",
+			Type: metric.MetricTypeCounter,
```
tariq1890:
```suggestion
			Type: metric.MetricTypeGauge,
```

@wanghaoran1988 This is the cause of the e2e fail. Hopefully your travis fails should pass after that.

brancz:
I agree we should make the csr -> certificatesigningrequest change

tariq1890:
@wanghaoran1988 Looks like the PR needs to be rebased.

wanghaoran1988:
Sorry, I just back from paternity leave, will have an update this week.

wanghaoran1988:
@mxinden @tariq1890 Could you take a look ? I rebased and renamed.

docs/certificatessigningrequest-metrics.md
```diff
@@ -0,0 +1,8 @@
+# CSR Metrics
```
tariq1890:
```suggestion
# CertificateSigningRequest Metrics
```

docs/certificatessigningrequest-metrics.md
```diff
@@ -0,0 +1,8 @@
...
+
+| Metric name| Metric type | Labels/tags | Status |
+| ---------- | ----------- | ----------- | ----------- |
+| kube_certificatesigningrequest_created| Gauge | `csr`=&lt;csr-name&gt;| STABLE |
```
tariq1890: s/csr/certificatesigningrequest

internal/collector/certificatesigningrequest.go
```diff
@@ -0,0 +1,152 @@
...
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+)
+
+var (
+	descCSRLabelsName          = "kube_certificatesigningrequest_labels"
+	descCSRLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
+	descCSRLabelsDefaultLabels = []string{"csr"}
```
tariq1890:
```suggestion
	descCSRLabelsDefaultLabels = []string{"certificatesigningrequest"}
```

wanghaoran1988:
comments addressed, @tariq1890 PTAL

docs/certificatessigningrequest-metrics.md
tariq1890: @wanghaoran1988 Let's change this as well.

internal/collector/builder.go
```diff
@@ -112,27 +113,28 @@ func (b *Builder) Build() []*coll.Collector {
...
+	"persistentvolumes":          func(b *Builder) *coll.Collector { return b.buildPersistentVolumeCollector() },
+	"poddisruptionbudgets":       func(b *Builder) *coll.Collector { return b.buildPodDisruptionBudgetCollector() },
+	"pods":                       func(b *Builder) *coll.Collector { return b.buildPodCollector() },
+	"replicasets":                func(b *Builder) *coll.Collector { return b.buildReplicaSetCollector() },
+	"replicationcontrollers":     func(b *Builder) *coll.Collector { return b.buildReplicationControllerCollector() },
+	"resourcequotas":             func(b *Builder) *coll.Collector { return b.buildResourceQuotaCollector() },
+	"secrets":                    func(b *Builder) *coll.Collector { return b.buildSecretCollector() },
+	"services":                   func(b *Builder) *coll.Collector { return b.buildServiceCollector() },
+	"statefulsets":               func(b *Builder) *coll.Collector { return b.buildStatefulSetCollector() },
+	"certificatesigningrequests": func(b *Builder) *coll.Collector { return b.buildCsrCollector() },
```
tariq1890: This list actually follows alphabetical order. Let's continue that :)

tariq1890: Sorry to have missed this earlier.

internal/collector/certificatesigningrequest_test.go
```diff
@@ -0,0 +1,220 @@
...
+				kube_certificatesigningrequest_labels{certificatesigningrequest="certificate-test",label_cert="test"} 1
+				kube_certificatesigningrequest_cert_length{certificatesigningrequest="certificate-test"} 0
+`,
+			MetricNames: []string{"kube_certificatesigningrequest_created", "kube_certificatesigningrequest_condition", "kube_certificatesigningrequest_labels", "kube_certificatesigningrequest_cert_length"},
+		},
+	}
+	for i, c := range cases {
+		c.Func = metric.ComposeMetricGenFuncs(csrMetricFamilies)
+		if err := c.run(); err != nil {
+			t.Errorf("unexpected collecting result in %vth run:\n%s", i, err)
```
tariq1890:
```suggestion
			t.Errorf("unexpected error when collecting result in %vth run:\n%s", i, err)
```

tariq1890:
@wanghaoran1988 Thanks for your patience! This should be my last round of review comments. 

LGTM pending those comments and @brancz 's final word of approval.

wanghaoran1988:
@tariq1890 comments addressed

docs/certificatessigningrequest-metrics.md
```diff
@@ -0,0 +1,8 @@
+# CertificatesSigningRequest Metrics
```
tariq1890:
```suggestion
# CertificateSigningRequest Metrics
```

tariq1890: there is an extra `s`.

pkg/options/collector.go
```diff
@@ -26,26 +26,27 @@ var (
...
-		"pods":                     struct{}{},
-		"replicasets":              struct{}{},
-		"replicationcontrollers":   struct{}{},
-		"resourcequotas":           struct{}{},
-		"secrets":                  struct{}{},
-		"services":                 struct{}{},
-		"statefulsets":             struct{}{},
+		"configmaps":                 struct{}{},
+		"cronjobs":                   struct{}{},
+		"certificatesigningrequests": struct{}{},
```
tariq1890: can this go to the top?

docs/certificatessigningrequest-metrics.md
wanghaoran1988: done

pkg/options/collector.go
wanghaoran1988: done

tariq1890:
/assign @mxinden for final approval. 

Possible to fit this into release 1.6 once merged?

brancz:
lgtm but leaving final decision up to @mxinden whether to include in 1.6 or not

mxinden:
Let's include this in kube state metrics 1.6.

/lgtm

/approve

tariq1890:
Yes! Thanks once again @wanghaoran1988 :)

PR #651, opened 2019-01-31
[WIP] Native pod termination counter - proposed implementation

wozniakjan:
I would like to continue discussion from https://github.com/kubernetes/kube-state-metrics/issues/344. While using `LastTerminationState` allows to catch more OOMKills, it still misses those where during the scrape period container dies more than once with different termination state. I am not entirely sure this implementation would follow conventions set by other metrics but it does allow never to miss an OOMKill because it can increment own counter on each observed pod change.

It is split into two commits
- first one allowing unit tests to take more than one object
- implementing the observed termination counter

Looking forward to any comments and hoping we can come up with good solution together :)

Fixes #

### DIFF
--- a/internal/collector/pod.go
+++ b/internal/collector/pod.go
@@ -18,6 +18,7 @@ package collector
 
 import (
 	"strconv"
+	"sync"
 
 	"k8s.io/kube-state-metrics/pkg/constant"
 	"k8s.io/kube-state-metrics/pkg/metric"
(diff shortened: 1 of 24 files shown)


main_test.go
```diff
@@ -209,6 +209,16 @@ kube_pod_container_status_terminated_reason{namespace="default",pod="pod0",conta
 kube_pod_container_status_terminated_reason{namespace="default",pod="pod0",container="container3",reason="Completed"} 0
 kube_pod_container_status_terminated_reason{namespace="default",pod="pod0",container="container3",reason="Error"} 0
 kube_pod_container_status_terminated_reason{namespace="default",pod="pod0",container="container3",reason="ContainerCannotRun"} 0
+# HELP kube_pod_container_status_terminated_reasons_total Counter with observed container restarts and reason the container was terminated.
```
zouyee: This `kube_pod_container_status_terminated_reasons_total ` can be obtained by  `kube_pod_container_status_terminated_reason` calculations. Do we really need it? @brancz

main_test.go
```diff
@@ -209,6 +209,16 @@ kube_pod_container_status_terminated_reason{namespace="default",pod="pod0",conta
...
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container2",reason="Error"} 0
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container2",reason="OOMKilled"} 0
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container3",reason="Completed"} 0
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container3",reason="ContainerCannotRun"} 0
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container3",reason="Error"} 0
+kube_pod_container_status_terminated_reasons_total{namespace="default",pod="pod0",container="container3",reason="OOMKilled"} 0
 # HELP kube_pod_container_status_last_terminated_reason Describes the last reason the container was in terminated state.
 # TYPE kube_pod_container_status_last_terminated_reason gauge
 kube_pod_container_status_last_terminated_reason{namespace="default",pod="pod0",container="container2",reason="OOMKilled"} 1
```
zouyee: Once add new object metrics, we also need to complete the [doc](https://github.com/kubernetes/kube-state-metrics/tree/master/Documentation) in the corresponding places. Thanks for your work!

internal/collector/pod.go
```diff
@@ -32,10 +33,35 @@ import (
...
+	}
+}
+
 var (
-	descPodLabelsDefaultLabels = []string{"namespace", "pod"}
-	containerWaitingReasons    = []string{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff"}
-	containerTerminatedReasons = []string{"OOMKilled", "Completed", "Error", "ContainerCannotRun"}
+	descPodLabelsDefaultLabels    = []string{"namespace", "pod"}
+	containerWaitingReasons       = []string{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff"}
+	containerTerminatedReasons    = []string{"OOMKilled", "Completed", "Error", "ContainerCannotRun"}
```
zouyee: These terminated reasons which is only for runtime type [`docker`](https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/dockershim/docker_container.go#L352), Do we need to mark it?

main_test.go
wozniakjan: thanks for the reminder, I will wait for the discussion about this proposal to continue for a little bit and then based on the output, either add docs if the PR provides additional value or close if it isn't worth having.

wozniakjan: thanks for the reminder, added to the docs :)

mxinden:
From an architectural point of view this patch makes kube-state-metrics a stateful application. So far the outputs of a freshly started kube-state-metrics would equal the outputs of a long running kube-state-metrics. With this patch this is not ensured anymore.

Is this inconsistency worth introducing for the given benefit?

wozniakjan:
Ack, thanks all for your time providing feedback, closing with no merge and will put this elsewhere

PR #661, opened 2019-02-07
cleanup api groups and add ingresses to the cluster role RBAC

tariq1890:
Using appsv1 for replicasets and daemonset as it's more appropriate. Adding also adding ingresses to the extensions RBAC

### DIFF
--- a/internal/collector/daemonset.go
+++ b/internal/collector/daemonset.go
@@ -19,7 +19,7 @@ package collector
 import (
 	"k8s.io/kube-state-metrics/pkg/metric"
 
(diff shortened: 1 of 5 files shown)


brancz:
This removes compatibility with 1.8. Is there any benefit to this? Otherwise I think I’d prefer to keep compatibility.

tariq1890:
Oh yes. I forgot about `1.8`. I can close this and add a PR for just adding the `ingresses` to RBAC then

brancz:
Sounds good to me!

PR #662, opened 2019-02-07
adding the CI check for go modules consistency

tariq1890:
Updated Travis CI to check for go mod consistency.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -26,6 +26,8 @@ jobs:
   include:
     # Format
     - script: make format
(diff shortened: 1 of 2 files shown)


brancz:
Let’s also add a “go mod verify”

Otherwise lgtm

tariq1890:
@brancz added!

brancz:
Good catch @mxinden. Then we should be good to go.

tariq1890:
Thanks for pointing that out @mxinden !

brancz:
Awesome! Thank you!

/lgtm
/approve

PR #671, opened 2019-02-13
use cancelable context instead of context.TODO()

### LINKED ISSUE #600: main.go: Pass cancellable context to collector builder

/kind feature

**What happened**:

In `main.go` we pass a `context.TODO()` to the collectors builder.

https://github.com/kubernetes/kube-state-metrics/blob/af34d12ac49e6e7431e44efcfc9b41fddce95453/main.go#L77

Instead we should pass a cancellable context that is cancelled on shutdown.

Originated here https://github.com/kubernetes/kube-state-metrics/pull/593#discussion_r236237555.

tariq1890:
Fixes #600

### DIFF
--- a/main.go
+++ b/main.go
@@ -60,6 +60,9 @@ func main() {
 	opts := options.NewOptions()
 	opts.AddFlags()
 
+	ctx, cancel := context.WithCancel(context.Background())
+	defer cancel()
+
 	err := opts.Parse()
 	if err != nil {
 		glog.Fatalf("Error: %s", err)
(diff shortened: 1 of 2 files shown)


brancz:
The final state of this that I imagined is that we use create a closable `net.Listener` to serve the various http servers, and create a [`run.Group`](https://github.com/oklog/run) for each and then have a [graceful shutdown handler](https://medium.com/@kpbird/golang-gracefully-stop-application-23c2390bb212) to actually gracefully shutdown the servers. This could then be used to also hot re-start servers when we want to live-reconfigure kube-state-metrics as we eventually want to with the features of #613.

As of what you implemented here though this already looks good to me and the rest can be a follow up. Thanks!

/lgtm

PR #674, opened 2019-02-16
add persistent volume capacity metric

### LINKED ISSUE #555: Add metric for persistent volume size
Feature

 /kind feature

**What happened**:

It would be helpful to see the size of persistent volumes.

**Environment**:

r0fls:
**What this PR does / why we need it**: Add `kube_persistentvolume_capacity` to track PV capacity.

Fixes #555

### DIFF
--- a/internal/collector/persistentvolume.go
+++ b/internal/collector/persistentvolume.go
@@ -109,7 +109,7 @@ var (
 			}),
 		},
 		{
-			Name: "kube_persistentvolume_capacity",
+			Name: "kube_persistentvolume_capacity_bytes",
 			Type: metric.MetricTypeGauge,
 			Help: "Persistentvolume capacity in bytes.",
 			GenerateFunc: wrapPersistentVolumeFunc(func(p *v1.PersistentVolume) *metric.Family {
--- a/internal/collector/persistentvolume.go
+++ b/internal/collector/persistentvolume.go
@@ -108,6 +108,21 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_persistentvolume_capacity",
+			Type: metric.MetricTypeGauge,
+			Help: "persistentvolume capacity in Gigabytes.",
+			GenerateFunc: wrapPersistentVolumeFunc(func(p *v1.PersistentVolume) *metric.Family {
+				storage := p.Spec.Capacity[v1.ResourceStorage]
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							Value: float64(storage.Size()),
+						},
+					},
+				}
+			}),
+		},
 	}
 )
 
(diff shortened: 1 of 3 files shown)


internal/collector/persistentvolume.go
```diff
@@ -108,6 +108,21 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_persistentvolume_capacity",
+			Type: metric.MetricTypeGauge,
+			Help: "persistentvolume capacity in Gigabytes.",
```
brancz: prometheus metrics best practice is to use base units, so this should be bytes (we're not loosing precision because we're using float64)

internal/collector/persistentvolume.go
```diff
@@ -108,6 +108,21 @@ var (
...
+			GenerateFunc: wrapPersistentVolumeFunc(func(p *v1.PersistentVolume) *metric.Family {
+				storage := p.Spec.Capacity[v1.ResourceStorage]
```
brancz: could there ever be an additoinal capacity? if so we should make the `v1.ResourceStorage` a label instead, similar to what we do with resource_requests on pod/containers

internal/collector/persistentvolume.go
mxinden:
```suggestion
			Help: "Persistentvolume capacity in gigabytes.",
```

Plus the suggestion by @brancz.

internal/collector/persistentvolume.go
r0fls: I do not believe there can ever be an additional capacity, because `p.Spec.Capacity` is a `v1.ResourceList`, which in turn is a `map[ResourceName]resource.Quantity`. So, there will only be one `resource.Quantity` for each `ResourceName` in the map (in this case the resource name is `v1.ResourceStorage`.

The name `v1.ResourceList` is a little misleading in that regard, but since it's a map to `resource.Quantity` (which is never a list) I think we're good.

r0fls:
Looks like the build probably needs restarting:

>No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.

internal/collector/persistentvolume.go
```diff
@@ -108,6 +108,21 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_persistentvolume_capacity",
```
brancz: if resource storage is the only capacity possible then this should be `kube_persistentvolume_capacity_bytes`

brancz:
Sorry about that, looks like a recent merge created a conflict. Could you rebase this? Thanks!

r0fls:
@brancz no problem, rebased :heavy_check_mark:

r0fls:
By the way it was just that the Documentation directory was renamed docs

Issue #569, opened 2018-10-24
kube-state-metrics breaking release aka 2.0


brancz:
We have accumulated a number of deprecated metrics and odd behaviors that I believe may justify a 2.0 release. I'd like to take this issue to discuss whether people think this is a good idea and collect what we would potentially like to break should we do a breaking release.

Off the top of my head breaking changes I would like to do:

* [x] remove deprecated metrics https://github.com/kubernetes/kube-state-metrics/issues/975
* [x] rename black-/whitelist to allow/deny-list https://github.com/kubernetes/kube-state-metrics/issues/975
* [x] use same ports in all cases (currently the flag defaults to 80/81, but the dockerfile specifies 8080 and 8081) https://github.com/kubernetes/kube-state-metrics/issues/976
* [x] rename hpa metrics to use full `horizontalpodautoscaler` nomenclature, to match the rest of the exposed metrics https://github.com/kubernetes/kube-state-metrics/issues/977
* [x] rename `--namespace` flag to `--namespaces` https://github.com/kubernetes/kube-state-metrics/issues/978
* [x] remove non-identifying labels from pod metrics (eg. [node labels](https://github.com/kubernetes/kube-state-metrics/blob/aed94850b839d111d15308c97e1db7ada8017c6f/internal/collector/pod.go#L511))
* [x] ~consider renaming kube-state-metrics to kubernetes-exporter~ we decided against this one
* [x] `kube_secret_metadata_resource_version`, `kube_configmap_metadata_resource_version` and `kube_ingress_metadata_resource_version` expose the resource version as a string in its set of labels. This value can change often and would therefore create huge cardinality. This should be a number or not existing at all. https://github.com/kubernetes/kube-state-metrics/pull/997
* [x] rename use of "collector" in user facing things (flags, help text, etc.) to "resources", as the collector type pattern does not exist anymore. https://github.com/kubernetes/kube-state-metrics/issues/980
* [x] rename storage class labels reclaimPolicy to reclaim_policy and volumeBindingMode to volume_binding_mode. https://github.com/kubernetes/kube-state-metrics/pull/1107
* [x] require `kube_*_labels` metrics to have explicit lists of labels to expose passed in via a flag to prevent unnecessary high cardinality https://github.com/kubernetes/kube-state-metrics/issues/1047
* [x] reworking resource metrics to try to adhere as best as possible to Prometheus "sums of series should make sense" rule. https://github.com/kubernetes/kube-state-metrics/pull/1168

[...]

PR #681, opened 2019-02-25
Export Job's owner

zuzzas:

This PR adds a Job's object owner metric similar to the Pod's owner. It is quite helpful when trying to correlate Jobs and CronJobs.

### DIFF
--- a/internal/collector/job.go
+++ b/internal/collector/job.go
@@ -271,13 +271,7 @@ var (
 				ms := []*metric.Metric{}
 
 				owners := j.GetOwnerReferences()
-				if len(owners) == 0 {
-					ms = append(ms, &metric.Metric{
-						LabelKeys:   labelKeys,
-						LabelValues: []string{"<none>", "<none>", "<none>"},
-						Value:       1,
-					})
-				} else {
+				if len(owners) > 0 {
 					for _, owner := range owners {
 						if owner.Controller != nil {
 							ms = append(ms, &metric.Metric{
--- a/internal/collector/job.go
+++ b/internal/collector/job.go
@@ -17,6 +17,8 @@ limitations under the License.
 package collector
 
 import (
+	"strconv"
+
 	"k8s.io/kube-state-metrics/pkg/metric"
 
 	v1batch "k8s.io/api/batch/v1"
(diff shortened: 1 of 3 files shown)


internal/collector/job.go
```diff
@@ -255,6 +257,44 @@ var (
...
+			Help: "Information about the Job's owner.",
+			GenerateFunc: wrapJobFunc(func(j *v1batch.Job) *metric.Family {
+				labelKeys := []string{"owner_kind", "owner_name", "owner_is_controller"}
+				ms := []*metric.Metric{}
+
+				owners := j.GetOwnerReferences()
+				if len(owners) == 0 {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   labelKeys,
+						LabelValues: []string{"<none>", "<none>", "<none>"},
```
brancz: no metric should be appended if none exists, in prometheus that's queryable through the `absent` function

zuzzas: Aye, I've had the same concern, but I've taken the Pod collector as a reference. There is no such machinery there.
https://github.com/kubernetes/kube-state-metrics/blob/master/internal/collector/pod.go#L121

zuzzas: Might be good to leave at it is to preserve uniformity in how we present metrics.

internal/collector/job_test.go
```diff
@@ -42,11 +42,15 @@ var (
 )
 
 func TestJobCollector(t *testing.T) {
+	var test = true
```
mxinden: Tiny thing: I would prefer:

```golang
var trueValue = true
```

What do you think?

internal/collector/job.go
mxinden: Good point. @brancz as changing this in `pod.go` would be a breaking change, should we do this with the v2.0 release?

internal/collector/job_test.go
zuzzas: Taken straight from the `pod_test.go`, but I'll change it.

internal/collector/job.go
brancz: Let's do it right in this case, and add it to the 2.0 release tracking issue: #569.

andyxning: @zuzzas IIUC, you need to delete the logic for appending `<none>` value metrics when the owners are empty as what @brancz proposed in comment https://github.com/kubernetes/kube-state-metrics/pull/681/files#r259773399

zuzzas: That I understood, @andyxning. My question was about the case in Pod collector, which works differently. Right now my issues are resolved any I'm adding a new feature.

zuzzas: I've done that in a separate commit: https://github.com/kubernetes/kube-state-metrics/pull/681/commits/894b9af069cb1dddd6a93d3bc6bd3622ffdbbe21

brancz:
Thanks!

/lgtm

xiaoxi1989:
why not add owner info to pod metric? 
like:
`kube_pod_status_phase{endpoint="https-main",instance="10.128.0.119:8443",job="kube-state-metrics",namespace="grpc-test",phase="Pending",pod="cxxtest-b78d5b566-gkhds",service="kube-state-metrics", deploy="cxxtest"}`

brancz:
there can be multiple owners

PR #682, opened 2019-02-25
add collector for networkpolicies

tariq1890:


zouyee:
As with other PRs, what is your scenario with these metrics?  May need to give a rough description.

internal/collector/builder.go
```diff
@@ -124,9 +125,10 @@ var availableCollectors = map[string]func(f *Builder) *coll.Collector{
 	"jobs":                     func(b *Builder) *coll.Collector { return b.buildJobCollector() },
 	"limitranges":              func(b *Builder) *coll.Collector { return b.buildLimitRangeCollector() },
 	"namespaces":               func(b *Builder) *coll.Collector { return b.buildNamespaceCollector() },
+	"networkpolicies":          func(b *Builder) *coll.Collector { return b.buildNetworkPolicyCollector() },
 	"nodes":                    func(b *Builder) *coll.Collector { return b.buildNodeCollector() },
 	"persistentvolumeclaims":   func(b *Builder) *coll.Collector { return b.buildPersistentVolumeClaimCollector() },
-	"persistentvolumes":        func(b *Builder) *coll.Collector { return b.buildPersistentVolumeCollector() },
+	"persistexntvolumes":       func(b *Builder) *coll.Collector { return b.buildPersistentVolumeCollector() },
```
brancz: this seems like it was a mistake

tariq1890:
No specific use case I have in mind. I just thought that there would be a potential use for tracking `networkPolicy` between the Pod communications as kube-state-metrics is commonly used Pod metrics collection and analysis.

brancz:
I think I'm happy to add it, if we have a concrete use case.

PR #688, opened 2019-02-28
replace git describe with VERSION file for maintaining version tagging

tariq1890:
Fixes #684

PR #690, opened 2019-03-01
improve the release documentation

tariq1890:
Release docs need to be updated given the changes in #688


brancz:
Good catch, thanks for updating!

/lgtm

PR #693, opened 2019-03-02
fix all golint and gosimple issues in kube-state-metrics

tariq1890:
Fix all golint issues in kube-state-metrics.

This should bring the goreportcard to 100%

### DIFF
--- a/internal/collector/builder.go
+++ b/internal/collector/builder.go
@@ -64,10 +64,8 @@ func NewBuilder(
 
 // WithEnabledCollectors sets the enabledCollectors property of a Builder.
 func (b *Builder) WithEnabledCollectors(c []string) {
-	copy := []string{}
-	for _, s := range c {
-		copy = append(copy, s)
-	}
+	var copy []string
+	copy = append(copy, c...)
 
 	sort.Strings(copy)
 
--- a/internal/collector/configmap.go
+++ b/internal/collector/configmap.go
@@ -33,7 +33,7 @@ var (
 	configMapMetricFamilies = []metric.FamilyGenerator{
 		{
 			Name: "kube_configmap_info",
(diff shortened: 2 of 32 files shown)


internal/collector/configmap.go
```diff
@@ -47,7 +47,7 @@ var (
 		},
 		{
 			Name: "kube_configmap_created",
-			Type: metric.MetricTypeGauge,
+			Type: metric.Gauge,
```
tariq1890: This actually fixes a golint issue. I guess this is concise and self-explanatory which is typical of idiomatic Go. 

`metric.MetricTypeGauge` is long and redundant.

internal/collector/builder.go
```diff
@@ -64,10 +64,8 @@ func NewBuilder(
 
 // WithEnabledCollectors sets the enabledCollectors property of a Builder.
 func (b *Builder) WithEnabledCollectors(c []string) {
-	copy := []string{}
-	for _, s := range c {
-		copy = append(copy, s)
-	}
+	var copy []string
+	copy = append(copy, c...)
```
tariq1890: This fixes an issue reported by `gosimple`

andyxning:
@tariq1890 Thanks for taking care of this. Could you also please add golint or gosimple checks just like `make gofmtcheck`.

tariq1890:
@andyxning I did have a plan to introduce golangci-lint to the CI. It is a fast and efficient linter that can run multiple linters like `golint`, `gosimple` etc. I can do that in a follow-up PR.

If you think it's more appropriate to add golangci-lint to this PR, I can do that as well.

brancz:
Let's add the CI steps in a follow up. Thanks for doing this!

/lgtm

PR #701, opened 2019-03-18
use kubernetes 1.14 dependencies in kube-state-metrics

tariq1890:
This PR uses the latest k8s 1.14.0 deps.


brancz:
Looks like this didn't update quite everything we needed?

tariq1890:
Yes. Looks like it. The errors happen within the client-go vendor layer. Some irreconcilable conflict with transitive seems to be at play here. Will look into this once more when I get more time on my hands.

pkg/options/options.go
```diff
@@ -59,7 +61,9 @@ func NewOptions() *Options {
 func (o *Options) AddFlags() {
 	o.flags = pflag.NewFlagSet("", pflag.ExitOnError)
 	// add klog flags
-	o.flags.AddGoFlagSet(flag.CommandLine)
+	klogFlags := flag.NewFlagSet("klog", flag.ExitOnError)
```
tariq1890: This may need reviewing. After taking in the new k8s deps, this seemed to be the only way to get this code to work.

For some reason, the flagSet never included the `logtostderr` flag with the current code after the dependency update.

tariq1890:
This PR is ready for review (after a lot of trials. Looks like client-go using Godeps makes the module vendoring a bit tricky.). I know that the doc updates are still pending. Just wanted to get the thoughts of the maintainers on this.

/assign @brancz

tariq1890:
/hold

I am gonna update this PR to take in k8s `1.14.0` now that 1.14.0 is out.

brancz:
cc @mxinden do you think we should include this in the v1.6 release series?

mxinden:
Fine by me. @tariq1890 any changes in particular that are relevant for kube-state-metrics with the version bump?

brancz:
I'm happy with including this.

/hold
/lgtm

@tariq1890 feel free to remove the hold label to confirm.

tariq1890:
@brancz @mxinden  I will have to update the docs for this. Also, this PR includes a code change in a non-vendor file. I've added a comment in the GH review. Could you reply to that?

pkg/options/options.go
tariq1890: @mxinden @brancz  Please review this. I did a bit of code digging and this was how I got it to work. It throws a panic with the new depped libraries otherwise. 

If something about this does not seem okay, I'll try to find some time to take a deeper look. (I am in between switching employers, so I will take a while. Apologies for that)

brancz: This looks good to me. I understand this way it doesn't panic anymore, right?

tariq1890: Yes that is correct

PR #705, opened 2019-03-24
Fix metric labels for kube_job_failed

lachlancooper:
**What this PR does / why we need it**: Fixes an inconsistency with the way the `condition` label key is applied to the `kube_job_failed` metric.


mxinden:
Thanks for the catch!

/approve

PR #707, opened 2019-03-28
update the prometheus_client_golang dependency to the latest version: v0.9.2

tariq1890:
Using the latest version of client_golang from prometheus.


main.go
```diff
@@ -139,7 +139,7 @@ func main() {
 	collectorBuilder.WithKubeClient(kubeClient)
 
 	ksmMetricsRegistry := prometheus.NewRegistry()
-	ksmMetricsRegistry.Register(prometheus.NewProcessCollector(os.Getpid(), ""))
+	ksmMetricsRegistry.Register(prometheus.NewProcessCollector(prometheus.ProcessCollectorOpts{}))
```
tariq1890: As per the documentation, my understanding is that this diff is functionally equivalent.

andyxning: @tariq1890 Yes. It is the recommended usage according to the doc.

tariq1890:
Thanks @andyxning

PR #702, opened 2019-03-19
*: Cut v1.6.0-rc.0

mxinden:
**Which issue(s) this PR fixes**:
Relates to https://github.com/kubernetes/kube-state-metrics/issues/697

PR #709, opened 2019-04-03
internal/collector+pkg/metrics_store: Reduce allocation & improve hot path

mxinden:
This pull request includes two performance optimizations found while testing the kube-state-metrics *v1.6.0* release candidate. Given that v1.6.0-rc.0 doesn't seem to introduce any performance regressions, these patches don't need to be included in the release.

### 1. internal/collector: Preallocate metric slice length when known upfront

Profiling kube-state-metrics shows that a lot of CPU time is spent on
`runtime.growslice`. This is related to metric generation when
converting a Kuberntes object to a slice of metrics in
`metric.FamilyGenerator`s. Oftentimes the length of the metric slice is
known upfront.

When the needed size of the metrics slice is known upfront, allocate an
appropriately sized metric slice.

### 2. pkg/metrics_store: Cache byte slices instead of strings

When converting a string to a byte slice, in Golang would do a full copy
which introduces memory allocations. On an HTTP scrape request WriteAll
in the metrics store component is called, which writes all cached
metrics into the given io.Writer. io.Writer takes a byte slice, whereas
metrics are cached as strings. The connect the two, the metrics store
component converts the metric strings to byte slices. This results in
`runtime.stringtoslicebyte` [1] being a prominent CPU user.

Instead of caching metrics as strings, cache them as byte slices,
removing the additional translation from the hot-path.

#### Unit benchmark

The first patch gets us some reasonable memory and runtime optimizations. Running `internal.collector.BenchmarkPodCollector` with (*new*) and without (*old*) returns the following:

```
benchmark                   old ns/op     new ns/op     delta
BenchmarkPodCollector-4     25942         24475         -5.65%

benchmark                   old allocs     new allocs     delta
BenchmarkPodCollector-4     419            395            -5.73%

benchmark                   old bytes     new bytes     delta
BenchmarkPodCollector-4     20112         19544         -2.82%
```

#### End to end test

I have compared kube-state-metrics *v1.6.0-rc.0-0* (vanilla https://github.com/kubernetes/kube-state-metrics/pull/702),  *v1.6.0-rc.0-4* (vanilla + patch 1) and *v1.6.0-rc.0-5* (vanilla + patch 1 & 2) on a 9 node cluster with AWS m5.large compute instances. During the comparison a test script would create and delete pods ranging from 250 - 5000 pods total. In addition the cluster had 8000 secrets constantly. Kube-state-metrics instances are scraped every 2s.

##### Scrape duration

Both patches seem to be neutral in regards to the Prometheus scrape duration.

image

##### CPU usage

Patch 1 seems to have only a subtle impact on the CPU usage, patch 2 constantly performs better, given that we only need to convert the metric string to a byte slice once and not on every scrape.

image

[1] https://golang.org/src/runtime/string.go?s=4063:4115#L145

### DIFF
--- a/pkg/metrics_store/metrics_store.go
+++ b/pkg/metrics_store/metrics_store.go
@@ -24,10 +24,10 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 )
 
-// FamilyStringer represents a metric family that can be converted to its string
+// FamilyByteSlicer represents a metric family that can be converted to its string
 // representation.
-type FamilyStringer interface {
-	String() string
+type FamilyByteSlicer interface {
+	ByteSlice() []byte
 }
 
 // MetricsStore implements the k8s.io/client-go/tools/cache.Store
@@ -40,7 +40,7 @@ type MetricsStore struct {
 	// metric families, containing a slice of metrics. We need to keep metrics
 	// grouped by metric families in order to zip families with their help text in
 	// MetricsStore.WriteAll().
-	metrics map[types.UID][]string
+	metrics map[types.UID][][]byte
 	// headers contains the header (TYPE and HELP) of each metric family. It is
 	// later on zipped with with their corresponding metric families in
 	// MetricStore.WriteAll().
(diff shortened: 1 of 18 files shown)


pkg/metrics_store/metrics_store.go
```diff
@@ -40,23 +40,23 @@ type MetricsStore struct {
 	// metric families, containing a slice of metrics. We need to keep metrics
 	// grouped by metric families in order to zip families with their help text in
 	// MetricsStore.WriteAll().
-	metrics map[types.UID][]string
+	metrics map[types.UID][][]byte
```
mxinden: Ping @LiliC: These patches make the metrics store cache byte slices instead of strings for performance reasons. You might need to adjust your logic in the operator framework to work with these patches.

LiliC: @mxinden Thanks for the heads up!

brancz:
This needs a run of `go fmt`.

mxinden:
CI is fixed. Thanks for the pointer @brancz. 

Would you mind giving this another review?

PR #710, opened 2019-04-03
go.mod: Pin k8s.io/client-go to v11.0.0

mxinden:
The previous patch updating kube-state-metrics to Kubernetes v1.14 (https://github.com/kubernetes/kube-state-metrics/pull/701) resulted in k8s.io/client-go module set to `v2.0.0-alpha.0.0.20190313235726-6ee68ca5fd83+incompatible`. While `6ee68ca5fd83` is the right reference to client-go release (v11), `v2.0.0-alpha` is clearly not. The k8s.io/client-go code checked in into /vendor follows the commit sha, not the version.

This patch pins k8s.io/client-go to `v11.0.0` via `GO111MODULE=on go get k8s.io/client-go@v11.0`. As /vendor was already correctly updated with the previous patch, this patch does not include any changes to /vendor.

So far I have not been able to understand, why go modules would pin the version to an old tag, but vendor the new version. Maybe @tariq1890 do you have an idea?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -34,7 +34,7 @@ require (
 	gopkg.in/inf.v0 v0.9.0 // indirect
 	k8s.io/api v0.0.0-20190313235455-40a48860b5ab
 	k8s.io/apimachinery v0.0.0-20190313205120-d7deff9243b1
-	k8s.io/client-go v2.0.0-alpha.0.0.20190313235726-6ee68ca5fd83+incompatible
+	k8s.io/client-go v11.0.0+incompatible
 	k8s.io/klog v0.2.0
 	k8s.io/kube-openapi v0.0.0-20190320154901-5e45bb682580 // indirect
 	k8s.io/utils v0.0.0-20190308190857-21c4ce38f2a7 // indirect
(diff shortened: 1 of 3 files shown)


tariq1890:
@mxinden I am afraid I don't have an answer either. go modules seems to change that automatically. When I did work on the PR, I had actually typed out `v11.0.0` and running `go mod tidy` changed it to what we see right now.

It is seen in prometheus as well which pins `v10.0.0`. https://github.com/prometheus/prometheus/blob/f8932dde660161a9b390628eec617b41039c52c9/go.mod#L114

tariq1890:
Also, can we run Travis against this PR?

mxinden:
Given that `/vendor` ends up with the right code, I am also fine with leaving it as is. What do you all think?

tariq1890:
I vote for leaving the code as-is. Thoughts @brancz @andyxning ?

Also will let @simonpasquier chime in and provide his thoughts.

brancz:
I say stay as-is as well.

/lgtm

mxinden:
/hold

> I say stay as-is as well.

So you vote for not merging?! @brancz?

simonpasquier:
> So far I have not been able to understand, why go modules would pin the version to an old tag, but vendor the new version.

It could be because https://github.com/kubernetes/kube-state-metrics/commit/ca4b467ac15cecfadba24ef02bb92f5525ed56a2 predates the creation of the `v11.0.0` tag.

```
~/workspace/src/k8s.io/client-go $ git for-each-ref --sort=taggerdate --format '%(refname) %(taggerdate)' refs/tags/v11.0.0
refs/tags/v11.0.0 Wed Apr 3 10:54:30 2019 +0530
```

But it doesn't explain why the same issue occurs in the Prometheus `go.mod` as the `v8.0.0` tag is way older than the change to `go.mod`. It might also be an issue with the local cache on @tariq1890's machine. Anyhow it is hard to tell and not a big deal since the SHA1 commit matches.

mxinden:
Thanks @simonpasquier! Given that this is not an issue I am closing here.

PR #711, opened 2019-04-03
README.md: Update compatibility matrix for kube-state-metrics v1.6

mxinden:
xref: https://github.com/kubernetes/kube-state-metrics/pull/702#issuecomment-478964510

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,14 +53,14 @@ All additional compatibility is only best effort, or happens to still/already be
 #### Compatibility matrix
 At most 5 kube-state-metrics releases will be recorded below.
 
-| kube-state-metrics | client-go  | **Kubernetes 1.9** | **Kubernetes 1.10** | **Kubernetes 1.11** | **Kubernetes 1.12** |
-|--------------------|------------|--------------------|---------------------|---------------------|---------------------|
-| **v1.2.0**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.3.0**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.3.1**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.4.0**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.5.0**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **master**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
+| kube-state-metrics | client-go  | **Kubernetes 1.10** | **Kubernetes 1.11** | **Kubernetes 1.12** | **Kubernetes 1.13** | **Kubernetes 1.14** |
+|--------------------|------------|---------------------|---------------------|---------------------|---------------------|---------------------|
+| **v1.2.0**         |  v6.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.3.1**         |  v6.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.4.0**         |  v8.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -53,14 +53,15 @@ All additional compatibility is only best effort, or happens to still/already be
...
-| **v1.2.0**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.3.0**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.3.1**         |  v6.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.4.0**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **v1.5.0**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
-| **master**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
+| kube-state-metrics | client-go  | **Kubernetes 1.10** | **Kubernetes 1.11** | **Kubernetes 1.12** | **Kubernetes 1.14** |
+|--------------------|------------|---------------------|---------------------|---------------------|---------------------|
+| **v1.2.0**         |  v6.0.0    |         ✓           |         ✓           |         ✓           |         -           |
```
tariq1890: I believe we'd need to remove one entry from this table as the version support matrix records just 5 releases (as per L54).

We could do either of the two
i) Remove L58. `v1.2.0`
ii) Remove L59. `v1.3.0`

I lean towards removing L59 as `v1.3.1` supersedes `v1.3.0`. Thoughts @mxinden ?

tariq1890:
Thanks so much for looking into this @mxinden :) (I didn't have access to my laptop yesterday.)

README.md
andyxning: @mxinden @tariq1890 Remove v1.2.0 is the right way to do this.

andyxning: After next release, we can both remove v1.3.0 and v1.3.1 as they have no big differences.

brancz: I'd say let's remove v1.3.1, and maybe even change each of the lines to `v1.3.x` (as we should never bump the client-go version in a patch unless for good reason and then we should have an item in this matrix specific to that). Also am I the only one who feels it's odd that we don't have a Kubernetes 1.13 column?

tariq1890: Thanks for catching that @brancz. I think we should definitely add a Kubernetes 1.13 column.

mxinden:
I have updated the compatibility matrix again. I have added column for Kubernetes v1.13. In addition I removed the row for kube-state-metrics v1.3.0, given that it is superseded by v1.3.1. I don't think we should support v1.3.0 any longer given that there is v1.3.1 which, given that it is a patch release, does not introduce any breaking changes.

I feel this is inline with the comments by @brancz and @tariq1890. @andyxning let me know if you have objections.

 Can I get another review for this?

README.md
```diff
@@ -53,14 +53,14 @@ All additional compatibility is only best effort, or happens to still/already be
...
-| **master**         |  v8.0.0    |         ✓          |         ✓           |         ✓           |         ✓           |
+| kube-state-metrics | client-go  | **Kubernetes 1.10** | **Kubernetes 1.11** | **Kubernetes 1.12** | **Kubernetes 1.13** | **Kubernetes 1.14** |
+|--------------------|------------|---------------------|---------------------|---------------------|---------------------|---------------------|
+| **v1.2.0**         |  v6.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.3.1**         |  v6.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.4.0**         |  v8.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.5.0**         |  v8.0.0    |         ✓           |         ✓           |         ✓           |         -           |         -           |
+| **v1.6.0**         |  v11.0.0   |         ✓           |         ✓           |         ✓           |         -           |         -           |
```
brancz: there should be all checkmarks on this line

mxinden: Fixed.

andyxning:
lgtm

PR #721, opened 2019-04-14
use a more idiomatic way of handling error messages

tariq1890:
This PR does two things

i) use `pkg/errors` instead of `fmt` as it provides better methods and accessing methods from `errors` is more idiomatic than `fmt`.
ii) Fixes error message generations where the wrong formatting directives were used
iii) Also fixes some import ordering.


tariq1890:
This PR can be merged after the `release 1.6` cycle.

mxinden:
> This PR can be merged after the release 1.6 cycle.

I don't see how this would be blocked by the release. [_release-1.6_](https://github.com/kubernetes/kube-state-metrics/commits/release-1.6) branch was already cut.

brancz:
Agreed with @mxinden.

/hold cancel
/lgtm

PR #722, opened 2019-04-16
use xenial dist in travis ci

andyxning:
This PR updates travis ci distribution to 16.04.
Fixes #

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,5 +1,5 @@
 sudo: required
-
+dist: xenial
 language: go
 
 go:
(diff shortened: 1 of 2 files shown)


andyxning:
FYI: https://blog.travis-ci.com/2018-11-08-xenial-release

brancz:
neat :+1: 

/lgtm

as a side note we might want to start looking into kind, I've seen great speed-ups in other setups, but also if this works for us I don't see a huge reason to change right now

andyxning:
> as a side note we might want to start looking into kind, I've seen great speed-ups in other setups, but also if this works for us I don't see a huge reason to change right now

It seems that kind is not so general as minikube. Since minikube v1.0.0 has been released and it should be stable enough, pinning on minikube should be ok for now.

PR #737, opened 2019-05-05
Add additional ingress metrics

ebabani:
Report on ingress host, path, service backend and port.

**What this PR does / why we need it**: It adds an additional ingress metric, with information about the hostname, path and service backend. This allows additional use cases, such as alerting on 2 ingresses using the same host and path.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*: None


andyxning:
/lgtm
/approve

Thanks for this valuable patch! @ebabani

PR #742, opened 2019-05-07
collector: De-duplicate common code for building collectors

brancz:
This refactors some duplication that is unnecessary in the collector builder.

I extracted this as I continued the work on #613 and this is a refactoring that can be done individually to not bloat the PR even more.

@mxinden @tariq1890 @andyxning @zouyee


mxinden:
I _git bisect_ ed the memory allocation increase. This was introduced in https://github.com/kubernetes/kube-state-metrics/pull/740. Given that the _container termination reason_ has a great influence on the cardinality of the Pod metrics (`#pods * #containers * #termination-reasons`) I guess the 9 % memory allocation increase for the pod collector is reasonable.

Issue #746, opened 2019-05-09
New metric for deployment revision


bvoss:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind feature

**What happened**:
Currently the annotation 'deployment.kubernetes.io/revision' is not exposed as metric.

**What you expected to happen**:
Expose the revision number contained in the annotation 'deployment.kubernetes.io/revision' as metric 'kube_deployment_revision'

**Anything else we need to know?**:
Why would you need this?
It allows you to easy set up a dashboard to show the deployment changes / rollbacks in a cluster.

Issue #574, opened 2018-10-30
Feature Request: Include Deployment and Pod Annotations


bluechips23:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**: Deployment and Pod Annotations are not included today in kube_state_metrics

**What you expected to happen**: Include Deployment and Pod Annotations

**Anything else we need to know?**: 
Deployment and Pod annotations contain some additional metadata information which are important. These metadata could be included when generating alerts using Prometheus. For instance, if the annotation has contact information (email addresses, etc), then Prometheus can fetch those to send alerts. Labels have a 63 character limit and do not support characters such as "@", "," or ";", making them unsuitable to include any email addresses. Annotations seem like an ideal place to store these information.

Please consider to see if deployment and pod annotations can be included.

jacobstr:
These could be useful in conjunction with approaches such as this one: https://karlstoney.com/2018/07/07/managing-your-costs-on-kubernetes/

The data exposed by `kube-state-metrics` is used to generate cost estimates for pods. Pushing a subset of labels/annotations into the metric dimensions would allow cost breakdown on a logical application level.

brancz:
@jacobstr at this point I need to make sure you understand that Prometheus is imprecise by design. I cannot recommend you to do billing on the data, if you need any form of precision :slightly_smiling_face: .

Having annotations of pods as labels has been requested before and is in some form theoretically possible but must be disabled by default as it will cause a huge amount of data in most clusters for Prometheus, that I'm not comfortable enabling by default.

bluechips23:
@brancz Having it as an optional feature is fine with it being disabled by default. We just want to leverage the ability to capture some metadata in addition to the metrics we are collecting. Labels can only do so much, whereas annotations are versatile. For some teams having annotations with a side effect of additional data may not be worth it, but for our objective, it could be worth it. At the least we want to have an option to evaluate and understand the pros and cons.

jacobstr:
@brancz yes - understood re: accuracy. Our use case is largely for internal dashboards to provide some reasonable ballparks for internal development teams. Today, folks are fairly abstracted from the cost of `2500m cpu * 16 replicas`. It's not meant for billing, but to provide some visibility into costs. Apologies to the op - I don't want mean to hijack since his use case is quite different but it might make this story more compelling with some divergent examples benefiting from this functionality.

[...]

PR #747, opened 2019-05-09
Fixes New metric for deployment revision #746

bvoss:
Add metric for deployment revision based on the annotation 'deployment.kubernetes.io/revision'

### DIFF
--- a/docs/deployment-metrics.md
+++ b/docs/deployment-metrics.md
@@ -14,3 +14,4 @@
 | kube_deployment_metadata_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_labels | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_created | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
+| kube_deployment_revision | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
--- a/internal/collector/deployment.go
+++ b/internal/collector/deployment.go
@@ -34,6 +34,7 @@ var (
 	descDeploymentLabelsName          = "kube_deployment_labels"
 	descDeploymentLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descDeploymentLabelsDefaultLabels = []string{"namespace", "deployment"}
+	annotationDeploymentRevision      = "deployment.kubernetes.io/revision"
 
 	deploymentMetricFamilies = []metric.FamilyGenerator{
 		{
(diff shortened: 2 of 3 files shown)


internal/collector/deployment.go
```diff
@@ -227,6 +229,27 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_deployment_revision",
+			Type: metric.Gauge,
+			Help: "Sequence number representing a specific revision of the deployment controller represented by the annotaion 'deployment.kubernetes.io/revision'.",
```
tariq1890:
```suggestion
			Help: "Sequence number representing a specific revision of the deployment controller represented by the annotation 'deployment.kubernetes.io/revision'.",
```

internal/collector/deployment_test.go
```diff
@@ -65,6 +65,8 @@ func TestDeploymentCollector(t *testing.T) {
 		# TYPE kube_deployment_spec_strategy_rollingupdate_max_surge gauge
 		# HELP kube_deployment_labels Kubernetes labels converted to Prometheus labels.
 		# TYPE kube_deployment_labels gauge
+		# HELP kube_deployment_revision Sequence number representing a specific revision of the deployment controller represented by the annotaion 'deployment.kubernetes.io/revision'.
```
tariq1890:
```suggestion
		# HELP kube_deployment_revision Sequence number representing a specific revision of the deployment controller represented by the annotation 'deployment.kubernetes.io/revision'.
```

internal/collector/deployment.go
```diff
@@ -227,6 +229,27 @@ var (
...
+			GenerateFunc: wrapDeploymentFunc(func(d *v1.Deployment) *metric.Family {
+				if d.ObjectMeta.Annotations == nil {
+					return &metric.Family{}
+				}
+				revision, err := strconv.Atoi(d.ObjectMeta.Annotations["deployment.kubernetes.io/revision"])
```
tariq1890: I would move `"deployment.kubernetes.io/revision" ` to a constant variable.

tariq1890:
I am not sure how I feel about exposing metrics from annotations. Thoughts @brancz @mxinden ?

mxinden:
Instead of exposing specific annotations, I would advocate towards exposing all annotations like we do with Kubernetes labels.

This has been raised in https://github.com/kubernetes/kube-state-metrics/issues/574 before. The cardinality increase introduced through this metric could have a high impact on big clusters, thus I think making the metric optional and disabled by default is the way forward.

@tariq1890 @bvoss @brancz what do you think?

bvoss:
@mxinden @tariq1890 @brancz 
Actual we have the problem with a large cluster, so all annotations isn't an option for us.

I understood, that the annotation and its value is essential for kubectl rollout --revision and so more likely part of the API and stable.
Honestly I don't know why it's not a property like in [StatefulSetStatus](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.14/#statefulsetstatus-v1-apps).

So I thought it would be ok - in this special case - to expose this annotation value.

Nevertheless the general annotation approach is also fine for me, if it's possible to configure a white-list for annotation to be contained in the metric.

internal/collector/deployment.go
bvoss: Fixed typo and changed from Sequence number into Number, because the revision is not really a sequence.

internal/collector/deployment.go
bvoss: Fixed typo and changed from Sequence number into Number, because the revision is not really a sequence.

internal/collector/deployment_test.go
bvoss: Fixed typo and changed from Sequence number into Number, because the revision is not really a sequence.

docs/deployment-metrics.md
```diff
@@ -14,3 +14,4 @@
 | kube_deployment_metadata_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_labels | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_created | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
+| kube_deployment_revision | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
```
brancz: Do you have some docs that show that this annotation is indeed considered stable? I'm only asking because I wonder why this is not a real field.

tariq1890:
I think this can be closed since the kube annotations PR was merged.

tariq1890:
Thanks for the feature request and taking a stab at this @bvoss :).

brancz:
I think there was a distinction here. The value of the annotation was actually parsed and exposed as the metric's value.

PR #748, opened 2019-05-10
use distroless/static as base image to further reduce image size

tariq1890:
Motivation: Given the Alpine CVE https://www.alpinelinux.org/posts/Docker-image-vulnerability-CVE-2019-5021.html (touted as the security-focused distro ironically), I would like to go one step further in preventive action by using ~~`scratch`~~ `distroless/static` as the base image. This will reduce the attack vectors  and the overall image size.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,11 +1,15 @@
 FROM alpine:3.9
 
-COPY kube-state-metrics /
-
-ENTRYPOINT ["/kube-state-metrics", "--port=8080", "--telemetry-port=8081"]
-
 RUN adduser -D kube-state-metrics
 
+FROM gcr.io/distroless/static
+
+COPY kube-state-metrics /
+
+COPY --from=0 /etc/passwd /etc/passwd
+
 USER kube-state-metrics
+
+ENTRYPOINT ["/kube-state-metrics", "--port=8080", "--telemetry-port=8081"]
 
 EXPOSE 8080 8081


mxinden:
In regards to the CVE, given that the latest image is build after March 7th 2019, this should not apply to the latest build. Please correct me if I am wrong.

In regards to moving to _scratch_ as a base image, would it make sense to align this effort with the current [enhancement proposal](https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md) updating the base image of the Kubernetes core components? Maybe @tallclair has some advice for us.

tariq1890:
Yes you are right @mxinden. We are safe. This PR is just an extra cautionary measure.

The distroless image actually ended up being up bigger in size than with alpine as base.

brancz:
I agree with @mxinden, even if it does end up being larger, I think we should use the distroless base.

tariq1890:
Alrighty, it is a trivial change. I will update this PR then?

tariq1890:
I have made the changes.

Correction: `distroless/static` is actually smaller than `alpine`. So win-win :).

Dockerfile
```diff
@@ -1,11 +1,7 @@
-FROM alpine:3.9
+FROM gcr.io/distroless/static
 
 COPY kube-state-metrics /
 
 ENTRYPOINT ["/kube-state-metrics", "--port=8080", "--telemetry-port=8081"]
 
-RUN adduser -D kube-state-metrics
```
brancz: why remove this?

tariq1890: I had initially thought that it wasn't necessary to explicitly create a non-root user in distroless, but that does not seem to be the case.

After some research, I see that I would need to use multi-stage builds. For the sake of making a clean change, this means changing up the Makefiles also a bit and other scripts that depend on it. I'll need to change this PR up significantly.

tariq1890:
/hold

This PR will need more work. Thanks for all your inputs @brancz @mxinden :)

Dockerfile
reetasingh: what is the use case for multi-stage builds?

andyxning:
@brancz @tariq1890 @mxinden I think we need to be consensus with the Kubernetes community in choosing base image.

tariq1890:
@andyxning As per this [doc](https://github.com/kubernetes/enhancements/blob/master/keps/sig-release/20190316-rebase-images-to-distroless.md), `distroless/static` is k8s-endorsed.

Dockerfile
tariq1890: Great question! This is the rationale essentially. https://medium.com/@lizrice/non-privileged-containers-based-on-the-scratch-image-a80105d6d341

tariq1890:
/hold cancel

This PR is ready for review again. I am leaving the makefiles untouched since we are using it for mult-arch builds. Hopefully, these changes should be agreeable :).

brancz:

/lgtm

PR #756, opened 2019-05-15
use latest version of k8s.io/klog

tariq1890:
updating to `v0.3.0`


andyxning:
Thanks for taking care of this! @tariq1890 

/lgtm
/approve

andyxning:
@tariq1890 Seems that `vendor/modules.txt` is maintained manually not by the golang module system. Could we remove it?

tariq1890:
I am note sure about that. I believe it's auto-generated by go mod.

andyxning:
@tariq1890 Yes, you're right. `go mod vendor` will generate it.

PR #762, opened 2019-05-17
Add pod init container metrics

### LINKED ISSUE #502: kube state metrics doesn't report init container pod metrics

tariq1890:
This should finish the work started by #541. Fixes #502 

@thedebugger I've added you as co-author to this PR.


tariq1890:
@brancz PTAL

PR #769, opened 2019-05-28
Add note on costing to README

shalvah:
My company ran into surprisingly high costs after a month of using kube-state-metrics on GCP because of the large number of resources that triggered a lot of events. Adding a friendly warning for everyone out there.


brancz:
Thanks!

/lgtm

brancz:
As a side note, may I suggest using Prometheus instead of ingesting this data into a cloud provider? Prometheus has easily ingest several million metrics, at relatively low cost.

PR #773, opened 2019-05-29
feat: implement black/white- listing of metrics by regex

### LINKED ISSUE #759: Allow wildcards/grouping of metrics to whitelist/blacklist

/kind feature

**What happened**:
I cannot whitelist/blacklist a bulk of metrics that are being exposed.

I expected to use wildcards to allow blacklisting for example `*_labels*` metrics.

¯\_(ツ)_/¯

I'm running an EKS cluster and as all my infra is in AWS, I was hoping to be able to use CloudWatch to monitor the state of my cluster. 
I'm using the prometheus-to-cloudwatch (https://github.com/cloudposse/prometheus-to-cloudwatch) scraper which seems to be working ok but the problem is that I'm getting a huge amount of metrics (around 220k metrics) and many of them are completely useless.
I want to be able to disable all the metrics that are meaningless (in the sense of cluster's health, i.e. the metrics `kube_*_info` or `kube_*_labels`) and reduce the amount of money spent on Cloudwatch; as an idea, ~220K metrics per month costs ~800USD.

**Environment**:
```Client Version: version.Info{Major:"1", Minor:"14", GitVersion:"v1.14.0", GitCommit:"641856db18352033a0d96dbc99153fa3b27298e5", GitTreeState:"clean", BuildDate:"2019-03-26T00:04:52Z", GoVersion:"go1.12.1", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"12+", GitVersion:"v1.12.6-eks-d69f1b", GitCommit:"d69f1bf3669bf00b7f4a758e978e0e7a1e3a68f7", GitTreeState:"clean", BuildDate:"2019-02-28T20:26:10Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}
```

`1.6.0`

tariq1890:
Fixes #759 

I believe this could be somewhat of a rough cut and may need thorough vetting and review.


pkg/whiteblacklist/whiteblacklist.go
```diff
@@ -85,13 +86,20 @@ func (l *WhiteBlackList) Exclude(items []string) {
 
 // IsIncluded returns if the given item is included.
 func (l *WhiteBlackList) IsIncluded(item string) bool {
-	_, exists := l.list[item]
+	var matched bool
+	for whiteListPattern := range l.list {
+		var err error
+		matched, err = regexp.MatchString(whiteListPattern, item)
+		if matched && err == nil {
```
tariq1890: I don't feel too great about the `err` check here. An alternative here would be validating the regex pattern (whether it's parseable or not) beforehand. That way, we wouldn't need to account for `err` check and use something like `regexp.MustCompile`

brancz: yeah I think we should validate these at startup and fail immediately if not a valid regex

tariq1890: This has been done.

brancz: Doing this in two places seems error prone. Could we architect it in a way where we do the parsing just once and create the whitelist/blacklist filter at that point already?

tariq1890: I hear you. I did think of rearchitecting this. Might as well do it in this PR.

brancz:
Just for my future self looking at this PR next time. As is this PR compiles the regex multiple times, and I'd prefer if we only did that once at startup, and fail starting if unsuccessful, but pass the compiles regex through to the white/blacklisting mechanism.

tariq1890:
Yes. Makes total sense.

tariq1890:
Wow. That was unexpected! Not sure how the approved label got added.

tariq1890:
@brancz I have changed this PR up a bit to make sure we only parse the regex once. Given that wbList may be mutated based on these options - `opts.DisableNodeNonGenericResourceMetrics` and `opts.DisablePodNonGenericResourceMetrics`, I needed to have the list populated first and then parse the list items into compiled regexes.  This is why I had created `whiteBlackList.Parse()` which performs validation and regex compilation. Let me know if you have any concerns.

brancz:
That's exactly what I imagined. lgtm :+1: 

/lgtm

PR #777, opened 2019-06-03
Add Metrics About StorageClass

### LINKED ISSUE #775: storageclass metrics can be supported?
kube-state-metrics not supported cluster's resource: storageclass,will you support it later?

xieyanker:

Add Some Metrics About StorageClass

Fixes #775

### DIFF
--- a/internal/collector/storageclass.go
+++ b/internal/collector/storageclass.go
@@ -60,22 +60,6 @@ var (
 			}),
 		},
 		{
-			Name: "kube_storageclass_metadata_resource_version",
-			Type: metric.Gauge,
-			Help: "Resource version representing a specific version of storageclass.",
-			GenerateFunc: wrapStorageClassFunc(func(s *storagev1.StorageClass) *metric.Family {
-				return &metric.Family{
-					Metrics: []*metric.Metric{
-						{
-							LabelKeys:   []string{"resource_version"},
-							LabelValues: []string{s.ObjectMeta.ResourceVersion},
-							Value:       1,
-						},
-					},
-				}
-			}),
-		},
-		{
 			Name: descStorageClassLabelsName,
 			Type: metric.Gauge,
 			Help: descStorageClassLabelsHelp,
--- a/internal/collector/storageclass.go
+++ b/internal/collector/storageclass.go
@@ -1,0 +1,122 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package collector
+
+import (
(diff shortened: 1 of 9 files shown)


internal/collector/storageclass.go
```diff
@@ -0,0 +1,124 @@
+/*
+Copyright 2017 The Kubernetes Authors All rights reserved.
```
mxinden:
```suggestion
Copyright 2019 The Kubernetes Authors All rights reserved.
```

internal/collector/storageclass.go
```diff
@@ -0,0 +1,124 @@
...
+		{
+			Name: "kube_storageclass_metadata_resource_version",
+			Type: metric.Gauge,
+			Help: "Resource version representing a specific version of storageclass.",
+			GenerateFunc: wrapStorageClassFunc(func(s *storagev1.StorageClass) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{"resource_version"},
+							LabelValues: []string{string(s.ObjectMeta.ResourceVersion)},
```
mxinden: As far as I know `ResourceVersion` is increased for each change propagated from etcd. Thus this metric would create a new timeseries for each change of the object. How often do you expect this Kubernetes object to change?

tariq1890:
You need to add `storageclasses` to the list of allowed resources in the `kube-state-metrics-cluster-role.yaml` file.

xieyanker:
> You need to add `storageclasses` to the list of allowed resources in the `kube-state-metrics-cluster-role.yaml` file.

Thank you, got it

tariq1890:
Can you address @mxinden's comment and also squash the commit history once this is good to merge?

internal/collector/storageclass.go
xieyanker: The `ResourceVersion` can be changed only when object is updated, but we won't update it usually, so I think it will not change freequently.

internal/collector/storageclass.go
```diff
@@ -0,0 +1,122 @@
...
+		{
+			Name: "kube_storageclass_metadata_resource_version",
+			Type: metric.Gauge,
+			Help: "Resource version representing a specific version of storageclass.",
+			GenerateFunc: wrapStorageClassFunc(func(s *storagev1.StorageClass) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{"resource_version"},
+							LabelValues: []string{s.ObjectMeta.ResourceVersion},
```
brancz: I'm still a little sceptic about this. If at all, why wouldn't this be the metric's value?

xieyanker: I refer to other code which use `ResourceVersion` as metric, I found all of them use `1` as metric's value. 
for example: `kube_secret_metadata_resource_version`, `kube_configmap_metadata_resource_version` and `kube_ingress_metadata_resource_version`.

brancz: Wow, you're totally right the other metrics do that, but it's wrong, we should have never merged those. Thanks for noticing! We should add these to the breaking release issue.

brancz: I think we should skip this one or model it in a different way for now. Do you have a concrete alerting use case in mind?

xieyanker: I remove `resource_version` for the time being, and I think it's not necessary to alert about `storageclass`, I only want to monitor its info.

brancz:
Thanks!

/lgtm

PR #780, opened 2019-06-05
.travis.yaml: Add job names

LiliC:

Adding the names of the jobs makes it easier to spot which jobs failed.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -30,15 +30,28 @@ jobs:
   include:
     # Lint
     - script: make lint
+      name: Lint
+
     # Check that all the go modules manifests, vendors, and packages are in sync
     - script: make validate-modules
+      name: Validate vendor is in sync with go modules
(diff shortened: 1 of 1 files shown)


brancz:
for some reason this makes travis not run them in parallel, hence why they weren't named before

tariq1890:
Then it should be reported to the Travis peeps IMO. Having Job names is useful. Hopefully, we can revisit this later.

brancz:
We have to migrate away from travis unfortunately anyways as it stopped working with prow ...

brancz:
IIRC this was intended behavior

andyxning:
We can not see any travis ci output for this PR. Adding names to each stage is definitely more readable. reopen this to do a further diagnose.

/reopen

andyxning:
Seems the branch has been deleted. :(

PR #783, opened 2019-06-07
Run check in CI for generated docs

LiliC:

The generated docs were out of date, so I reran the `make generate`. Also added a check for the generated docs, which will be now run as part of doc travis job.


docs/cli-arguments.md
```diff
@@ -37,13 +37,15 @@ Usage of ./kube-state-metrics:
...
+      --log_file_max_size uint                      Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
       --logtostderr                                 log to standard error instead of files (default true)
       --metric-blacklist string                     Comma-separated list of metrics not to be enabled. The whitelist and blacklist are mutually exclusive.
       --metric-whitelist string                     Comma-separated list of metrics to be exposed. The whitelist and blacklist are mutually exclusive.
       --namespace string                            Comma-separated list of namespaces to be enabled. Defaults to ""
       --port int                                    Port to expose metrics on. (default 80)
       --skip_headers                                If true, avoid header prefixes in the log messages
-      --stderrthreshold severity                    logs at or above this threshold go to stderr
+      --skip_log_headers                            If true, avoid headers when openning log files
```
tariq1890:
```suggestion
      --skip_log_headers                            If true, avoid headers when opening log files.
```

docs/cli-arguments.md
```diff
@@ -37,13 +37,15 @@ Usage of ./kube-state-metrics:
...
+      --stderrthreshold severity                    logs at or above this threshold go to stderr (default 2)
```
tariq1890:
```suggestion
      --stderrthreshold severity                    logs at or above this threshold go to stderr (default 2).
```

docs/horizontalpodautoscaler-metrics.md
```diff
@@ -8,4 +8,4 @@
 | kube_hpa_status_current_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_status_desired_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_status_condition         | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
-| kube_hpa_labels                   | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
\ No newline at end of file
+| kube_hpa_labels                   | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
```
tariq1890: I am not able to determine what this change exactly is.

Makefile
```diff
@@ -122,7 +124,7 @@ generate: build-local embedmd
 	@$(GOPATH)/bin/embedmd -w `find . -path ./vendor -prune -o -name "*.md" -print`
 
 embedmd:
-	go get github.com/campoy/embedmd@v1.0.0
+	GO111MODULE=on go get github.com/campoy/embedmd@v1.0.0
```
tariq1890: I am not sure if this is desirable. This changes the `go mod` artifacts and will affect the `go mod verify` checks in the CI.

docs/cli-arguments.md
LiliC: Ah I see, it was using the release, instead of master, let me fix this!

Makefile
LiliC: Yep you are right. But otherwise, go get fails.

docs/cli-arguments.md
LiliC: @tariq1890  Actually! I take that back, it's correct, it comes from klog, it's a typo there. If you do `kube-state-metrics --help` you will see it there :) 

So if we want to use the `make generate` to update these docs, we need to keep it as is.

docs/horizontalpodautoscaler-metrics.md
LiliC: Think it's a space, but these docs are generated so it's correct. :)

docs/cli-arguments.md
LiliC: See https://github.com/kubernetes/klog/pull/62 :)

docs/horizontalpodautoscaler-metrics.md
tariq1890: Cool! Thanks

docs/cli-arguments.md
tariq1890: Thanks !

Makefile
tariq1890: I am not sure we want to add this `@go mod tidy` to the `doccheck` target. Can we try this with `GO111MODULE=off` instead? I don't see why this needs to be added to the `go mod` artifacts, assuming we are using `embdedmd` as a a CLI convenience in CI.

LiliC: Interestingly enough with "off" hit this problem locally: `cannot use path@version syntax in GOPATH mode`

https://github.com/golang/go/issues/29415

LiliC: But I agree that was a hack, doing go mod tidy, I am sure there is a better way! :)

Makefile
```diff
@@ -122,7 +124,7 @@ generate: build-local embedmd
 	@$(GOPATH)/bin/embedmd -w `find . -path ./vendor -prune -o -name "*.md" -print`
 
 embedmd:
-	go get github.com/campoy/embedmd@v1.0.0
+	GO111MODULE=off go get github.com/campoy/embedmd
```
tariq1890: Yup I was going to suggest this. This `go mod` limitation is very annoying. You can see what the prometheus team had to do in order to get around this issue:

https://github.com/prometheus/prometheus/blob/a5d39361ab0b53081fef5bdc6c722148716f175d/Makefile.common#L187-L201

tariq1890: ^ That's just an FYI btw. I am definitely okay with go getting the embedmd without the versionedPath. Hopefully it works :)

LiliC: That looks interesting 😄

I removed the version restrain, test pass now, tbh the embedmd project does not change much. But up to you?

Makefile
```diff
@@ -37,7 +37,9 @@ ifndef HAS_GOLANGCI
 endif
 	golangci-lint run
 
-doccheck:
+doccheck: generate
+	@echo "- Checking if the generated documentation is up to date..."
+	@git diff --exit-code --
 	@echo "- Checking if the documentation is up to date..."
```
tariq1890: Can we change this to something like `Checking if the documentation is in sync with the code"` ? Do you agree that this would perhaps add more clarity on what's going on ?

Makefile
tariq1890: Fine by me.

Makefile
LiliC: Makes sense, done ✔️

tariq1890: Whoops, I actually meant changing this `"- Checking if the documentation is up to date..." `to `"- Checking if the documentation is up to date..."`

The informational line that you had added in this PR was fine IMO.

LiliC: ha, fixed and unfixed. Should be good now.

PR #785, opened 2019-06-11
Makefile: Enable benchmarking against master

LiliC:
 Adding a benchmark against master helps us catch performance
    regressions on the PR that introduces them, instead of at release
    time, or only in a later PRs.


brancz:
This is great! Thanks!

/lgtm

PR #786, opened 2019-06-11
Remove collector package

LiliC:

Initially, this started because we noticed, while using kube-state-metrics as a library, that there is an unnecessary pointer to an interface. Wanted to remove that pointer, but noticed that the collector package does not need to exist, as it's just a wrapper around store and instead we can just do a call `.WriteAll(w)` on the `store` object directly.

The majority of this PR is renaming things from collector -> store.

**Note:** None of the user-facing code was changed, as that would be a breaking change. So logs, flags that mention collectors stay unchanged. IMHO user-facing "collectors" flag should be renamed to "resources", but like I said that would be a user breaking change.


internal/store/builder.go
```diff
@@ -0,0 +1,264 @@
+/*
```
brancz: Any way we can preserve the history of this one? it's really just a rename right?

LiliC: It's not just a rename, it also renames the individual build functions to Store suffix, e.g. `buildDaemonSetCollection` -> `buildDaemonSetStore `

guthub is bad at displaying the diff, but I split it multiple commits, so hopefully, it's easier. :)

brancz: Ah I see. In that case the diff is probably too large for git to acknowledge, that this might be a rename, so it would have had to happen in two commits. In that case I'm ok with it as is.

LiliC:
> Are any of the vendoring changes necessary? It seems like if those are just updated, then we should do that in a separate PR.

I was wondering this as well when I noticed the big diff of added files. I think I ran the `go mod vendor` in case any of the deps in the collector package were removed, but seems like that triggered this update. Will remove this locally and test.

LiliC:
@brancz Update: the changes in the vendor directly are the result of running `go mod vendor`, as there is a check for that in `validate-modules` target, hence the change to the vendor dir.

brancz:
Interesting. In that case this lgtm. Thanks!

/lgtm

LiliC:
Travis seems to be having some issues with starting the build. 🤷‍♀

PR #793, opened 2019-06-18
builder/main: allow collectors not enabled by default

milesbxf:

This fixes an issue where it was impossible to specify a collector that
was available but not selected by default.

Instead of checking whether chosen collectors are valid at flag parse
time, this moves the check into the builder, where we can reference it
against the availableStores in the builder. As a bonus, the error
message also prints out a list of available collectors:

```
kube-state-metrics --collectors non-existent-collector
I0618 15:23:34.517532   50719 main.go:88] Using collectors non-existent-collector
F0618 15:23:34.519132   50719 main.go:90] Error: collector non-existent-collector does not exist. Available collectors: persistentvolumeclaims,configmaps,limitranges,nodes,namespaces,persistentvolumes,pods,replicasets,services,cronjobs,deployments,ingresses,horizontalpodautoscalers,jobs,poddisruptionbudgets,secrets,certificatesigningrequests,daemonsets,endpoints,storageclasses,replicationcontrollers,resourcequotas,statefulsets
```

Fixes #

### DIFF
--- a/go.sum
+++ b/go.sum
@@ -74,9 +74,7 @@ github.com/stretchr/testify v1.3.0/go.mod h1:M5WIy9Dh21IEIfnGCwXGc5bZfKNJtfHm1UV
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2 h1:VklqNMn3ovrHsnt90PveolxSbWFaJdECFbxSq0Mqo2M=
 golang.org/x/crypto v0.0.0-20190308221718-c2843e01d9a2/go.mod h1:djNgcEr1/C05ACkg1iLfiJU5Ep61QUkGW8qpdssI0+w=
 golang.org/x/net v0.0.0-20180724234803-3673e40ba225/go.mod h1:mL1N/T3taQHkDXs73rZJwtUhF3w3ftmwwsq0BUmARs4=
-golang.org/x/net v0.0.0-20180906233101-161cd47e91fd h1:nTDtHvHSdCn1m6ITfMRqtOd/9+7a3s8RBNOZ3eYZzJA=
 golang.org/x/net v0.0.0-20180906233101-161cd47e91fd/go.mod h1:mL1N/T3taQHkDXs73rZJwtUhF3w3ftmwwsq0BUmARs4=
-golang.org/x/net v0.0.0-20181201002055-351d144fa1fc h1:a3CU5tJYVj92DY2LaA1kUkrsqD5/3mLDhx2NcNqyW+0=
 golang.org/x/net v0.0.0-20181201002055-351d144fa1fc/go.mod h1:mL1N/T3taQHkDXs73rZJwtUhF3w3ftmwwsq0BUmARs4=
 golang.org/x/net v0.0.0-20190311183353-d8887717615a h1:oWX7TPOiFAMXLq8o0ikBYfCJVlRHBcsciT5bXOrH628=
 golang.org/x/net v0.0.0-20190311183353-d8887717615a/go.mod h1:t9HGtf8HONx5eT2rtn7q6eTqICYqUVnKs3thJo3Qplg=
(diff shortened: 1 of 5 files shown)


internal/store/builder.go
```diff
@@ -138,6 +146,23 @@ var availableStores = map[string]func(f *Builder) *metricsstore.MetricsStore{
...
+		if s == name {
+			return true
+		}
+	}
+	return false
+}
+
+func availableCollectors() []string {
+	c := []string{}
+	for name, _ := range availableStores {
```
brancz: this is a map already, so we can just check on existance of the collector key in this map directly in `collectorExists`

milesbxf: Fixed

main.go
```diff
@@ -86,7 +86,9 @@ func main() {
 		storeBuilder.WithEnabledResources(options.DefaultCollectors.AsSlice())
```
brancz: even if it should be impossible, we should check the error here as well

main.go
```diff
@@ -83,10 +83,13 @@ func main() {
 
 	if len(opts.Collectors) == 0 {
 		klog.Info("Using default collectors")
-		storeBuilder.WithEnabledResources(options.DefaultCollectors.AsSlice())
+		err = storeBuilder.WithEnabledResources(options.DefaultCollectors.AsSlice())
 	} else {
 		klog.Infof("Using collectors %s", opts.Collectors.String())
-		storeBuilder.WithEnabledResources(opts.Collectors.AsSlice())
+		err = storeBuilder.WithEnabledResources(opts.Collectors.AsSlice())
```
tariq1890: Can you remove the code duplication in L86 and L89? We could assign a variable to hold either `options.DefaultCollectors` or `ops.Collectors` and then call `storeBuilder.WithEnabledResources` . once.

brancz:
Thanks for the fix!

/lgtm
/approve

main.go
```diff
@@ -83,10 +83,13 @@ func main() {
...
+	}
+	if err != nil {
+		klog.Fatalf("Error: %s", err)
```
tariq1890: Something more descriptive here? and you should be using %v instead of %s.

brancz:
/hold

@tariq1890 comment on the error message is valid

internal/store/builder.go
tariq1890: Are we still calling this new method `availableCollectors()`?

milesbxf:
@tariq1890 @brancz done - thanks for looking at it!

go.sum
```diff
@@ -10,13 +10,10 @@ github.com/dgrijalva/jwt-go v0.0.0-20160705203006-01aeca54ebda h1:NyywMz59neOoVR
 github.com/dgrijalva/jwt-go v0.0.0-20160705203006-01aeca54ebda/go.mod h1:E3ru+11k8xSBh+hMPgOLZmtrrCbhqsmaPHjLKYnJCaQ=
 github.com/evanphx/json-patch v4.1.0+incompatible h1:K1MDoo4AZ4wU0GIU/fPmtZg7VpzLjCxu+UwBD1FvwOc=
 github.com/evanphx/json-patch v4.1.0+incompatible/go.mod h1:50XU6AFN0ol/bzJsmQLiYLvXMP4fmwYFNcr97nuDLSk=
-github.com/fsnotify/fsnotify v1.4.7 h1:IXs+QLmnXW2CcXuY+8Mzv/fWEsPGWxqefPtCP5CnV9I=
```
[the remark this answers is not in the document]
brancz: agreed, let's handle separately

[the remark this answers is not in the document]
tariq1890: Whoops spoke too soon. The CI checks are working as expected :). Can you run `go mod tidy` and `go mod vendor` @milesbxf ?

PR #795, opened 2019-06-19
update client_golang to v1.0.0

tariq1890:
update `prometheus/client_golang` to v1.0.0.


go.mod
```diff
@@ -5,29 +5,23 @@ require (
 	github.com/Azure/go-autorest v11.1.0+incompatible // indirect
 	github.com/dgrijalva/jwt-go v0.0.0-20160705203006-01aeca54ebda // indirect
 	github.com/evanphx/json-patch v4.1.0+incompatible // indirect
-	github.com/gogo/protobuf v0.0.0-20170330071051-c0656edd0d9e // indirect
```
andyxning: Since this module is deleted from `go.mod`, why it still exists under `vendor` directory?

andyxning: I have tested this locally and confirmed that this change is by design.

PR #798, opened 2019-06-20
update kubernetes versions to point to 1.15

tariq1890:
This PR updates the vendors of kube-state-metrics to point to k8s-1.15 libraries.


tariq1890:
sig/auth?

README.md
```diff
@@ -55,14 +55,14 @@ All additional compatibility is only best effort, or happens to still/already be
 #### Compatibility matrix
 At most 5 kube-state-metrics releases will be recorded below.
```
LiliC: Nit: It says on line 56 that five will be displayed below, so maybe we can remove the k8s 1.10 column. IIRC that's what usually happened. Or change the number to 6. :) 

> At most 5 kube-state-metrics releases will be recorded below.

tariq1890: Thanks for reviewing! As per that statement, we should be keeping only 5 releases of kube-state-metrics and it doesn't really mention kubernetes versions.

However, you do bring up a valid point about the support version matrix. I don't think it's practical to have an ever-expanding list of kubernetes versions as columns. I would update the documentation to reflect the k8s version limit in a separate PR.

brancz: If we're updating already, why not just remove kubernetes 1.10 here?

LiliC: Ah sorry about that, you are right!

> I would update the documentation to reflect the k8s version limit in a separate PR.

sgtm 👍 thanks!

tariq1890: @brancz Sorry, should have been more clear with my last comment. I meant that I would remove the K8s 1.10 column and update the documentation in another PR as I feel it's not on this PR's scope.

If you and @LiliC feel it's better to remove it in this PR, I can do that.

LiliC: @tariq1890 in another PR makes more sense to me

tariq1890:
@brancz Requesting your review.

PR #801, opened 2019-06-21
enable useless code check and more strict comment check

### LINKED ISSUE #800: The linter does not really catch all of the issues in kube-state-metrics.
The Golangci linting tool that we use does not seem to perform a complete scan of the code base for reporting issues with the code quality.

See this [PR](https://github.com/kubernetes/kube-state-metrics/pull/799) for one such example.

andyxning:
By default, golangci-lint will not enable check for exported function comment check and some other issues. Enable this manually for kube-state-metrics.
Fixes #800

### DIFF
--- a/internal/store/certificatesigningrequest_test.go
+++ b/internal/store/certificatesigningrequest_test.go
@@ -27,18 +27,6 @@ import (
 )
 
 func TestCsrStore(t *testing.T) {
-	const metadata = `
-		# HELP kube_certificatesigningrequest_labels Kubernetes labels converted to Prometheus labels.
-		# TYPE kube_certificatesigningrequest_labels gauge
-		# HELP kube_certificatesigningrequest_created Unix creation timestamp
-		# TYPE kube_certificatesigningrequest_created gauge
-		# HELP kube_certificatesigningrequest_condition The number of each certificatesigningrequest condition
-		# TYPE kube_certificatesigningrequest_condition gauge
-		# HELP kube_certificatesigningrequest_cert_length Length of the issued cert
-		# TYPE kube_certificatesigningrequest_cert_length gauge
-		# HELP kube_certificatesigningrequest_annotations Kubernetes annotations converted to Prometheus labels.
-		# TYPE kube_certificatesigningrequest_annotations gauge
-	`
 	cases := []generateMetricsTestCase{
 		{
 			Obj: &certv1beta1.CertificateSigningRequest{
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -861,21 +861,56 @@ var (
 			}),
 		},
 		{
-			Name: "kube_pod_container_resource_requests_cpu_cores",
+			Name: "kube_pod_init_container_resource_requests",
 			Type: metric.Gauge,
-			Help: "The number of requested cpu cores by a container.",
+			Help: "The number of requested request resource by the init container.",
 			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
 				ms := []*metric.Metric{}
 
-				for _, c := range p.Spec.Containers {
+				for _, c := range p.Spec.InitContainers {
 					req := c.Resources.Requests
-					if cpu, ok := req[v1.ResourceCPU]; ok {
-						ms = append(ms, &metric.Metric{
-							LabelKeys:   []string{"container", "node"},
-							LabelValues: []string{c.Name, p.Spec.NodeName},
-							Value:       float64(cpu.MilliValue()) / 1000,
-						})
+
+					for resourceName, val := range req {
+						switch resourceName {
+						case v1.ResourceCPU:
+							ms = append(ms, &metric.Metric{
+								Value:       float64(val.MilliValue()) / 1000,
+								LabelValues: []string{c.Name, p.Spec.NodeName, sanitizeLabelName(string(resourceName)), string(constant.UnitCore)},
+							})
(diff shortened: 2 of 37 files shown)


pkg/metrics_store/metrics_store.go
```diff
@@ -14,6 +14,7 @@ See the License for the specific language governing permissions and
 limitations under the License.
 */
 
+// Package metricsstore provides a memory tunned store for Kubernetes objects.
```
brancz: tuned

andyxning: Good catch. Will done.

tariq1890:
Any updates on this?

tariq1890:
@andyxning Do you know what needs to be done for `go vet`. The linting job overlooks these two go vet issues
```
Line 71: error: call of append copies lock value: regexp.Regexp contains sync.Mutex (vet)
Line 106: error: range var r copies lock: regexp.Regexp contains sync.Mutex (vet)
```

internal/store/pod.go
```diff
@@ -860,6 +860,64 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_init_container_resource_requests",
```
brancz: how come this is part of this PR?

internal/store/certificatesigningrequest_test.go
```diff
@@ -27,18 +27,6 @@ import (
 )
 
 func TestCsrStore(t *testing.T) {
-	const metadata = `
```
brancz: Why are we removing these? We do want these in output that's checked no?

internal/store/pod.go
andyxning: Because we documented `kube_pod_init_container_resource_requests` in the doc but we forgot for collect it in pod.go.  I have also changed the way we collect tested metrics. `kube_pod_init_container_resource_requests` is just added in the `metadata` field of Prometheus metrics output instead really collect it and contains it in the ouput.

internal/store/certificatesigningrequest_test.go
andyxning: I am afraid we do not need `metadata` for now. I have not checked out the details but golangci-tools reports that `metadata` is not referenced anymore.

LiliC: This seems to be a leftover from a refactor that was done on the collectors package. https://github.com/kubernetes/kube-state-metrics/blob/release-1.4/pkg/collectors/daemonset_test.go#L120

internal/store/pod.go
tariq1890: @andyxning Thanks for catching this. Looks like a bad metric has been added: `"kube_init_pod_container_resource_requests_cpu_cores"`

Can you remove that completely? CPU core requests are deprecated for pod containers and should never be added to pod init containers since they are new.

tariq1890:
@andyxning Thanks for this PR. This helps immensely. Could we also rename the commits and PR titles to accurately reflect the fixes this PR does?

internal/store/certificatesigningrequest_test.go
brancz: But wouldn't that mean that instead of removing it we want to prepend it to all "want" cases in the tests?

LiliC: @brancz It wasn't just the input, but the entire way the testing happens that was changed. The util testing tool for collectors changed. Before the [GatherAndCompare](https://github.com/kubernetes/kube-state-metrics/blob/release-1.4/pkg/collectors/testutils/testutils.go#L34 ) function was used which is removed now. That used the prometheus registry helper functions for gathering and comparing the metrics. Instead, the [new testutils](https://github.com/kubernetes/kube-state-metrics/blob/f24d831a4743b80f294193943b3f91a8b8308286//internal/store/testutils.go#L39:42) helpers are used now.

But I agree the tests should be testing the presence of the metadata! I just don't think that's necessarily in the scope of this PR.

brancz: Yes, however, I'd prefer to use the `metadata` variable to resolve the linting issues, instead of removing it.

andyxning: Maybe we should first merge this and then another PR needs to adjust the testing utils to leverage the `metadata`.  1.7.0 should be cut after the `metadata` refactor is done.

brancz: I'm not suggesting the testutils to actually make use of it now. Just use the `metadata` variable like it is in other tests already. I'd prefer to keep this around and then fix things as needed when we actually use it in the testutils again.

brancz: Sorry I'm not sure I was clear enough in what I meant. I meant use `metadata` in the `expected` just so we don't loose it, that way when we adapt the testutils to actually test metadata again we will have less to get back and or see missing or wrong metadata tests immediately.

LiliC: @andyxning if you rebase your PR the above problem of removing metadata won't be needed anymore it was fixed in https://github.com/kubernetes/kube-state-metrics/pull/848

tariq1890:
@andyxning Would you be able to rebase this?

LiliC:
@tariq1890 what do we have left in this PR, I think we did a lot already in other PRs. If @andyxning doesn't have time, lets close this an open a separate PR to bring in the leftovers, lets wait until next week, wdyt?

tariq1890:
@LiliC That sounds fair to me.

Issue #802, opened 2019-06-25
Switch from deprecated workload APIs due to it's removal in v1.16


LiliC:
Currently, kube-state-metrics uses the deprecated `extensions/v1beta1` workload APIs, but in v1.16 onwards these API are removed, so we should switch to the newer versions instead. This effects [ingress](https://github.com/kubernetes/kube-state-metrics/blob/e61670627018010350a963dcb4e276ac10a5d57f/internal/store/ingress.go#L22) and [replicaset](https://github.com/kubernetes/kube-state-metrics/blob/e61670627018010350a963dcb4e276ac10a5d57f/internal/store/replicaset.go#L24).

PR that removes it -> https://github.com/kubernetes/kubernetes/pull/70672

LiliC:
Happy to make the needed changes. :)

brancz:
Go ahead :)

LiliC:
@brancz Probably https://github.com/kubernetes/kube-state-metrics/issues/796 release should happen before the above change?

brancz:
How long as the networking API for the ingress replacement existed? If at least since kube 1.11 then I'm ok with having it in kube-state-metrics 1.7.

qw1mb0:
This API is available only from 1.14: https://github.com/kubernetes/kubernetes/pull/74057

brancz:
Ok so the ingress one we'll need to keep until 1.18 then, but replicasets we can move now.

tariq1890:
@LiliC I see that `ingresses` weren't included in this PR https://github.com/kubernetes/kubernetes/pull/70672 and aren't slated to be removed. 

I guess it can be closed. Thanks for bringing it up.

LiliC:
@tariq1890 We can open a follow-up issue for switching from the deprecated `ingress` workloads API, so we remember to do that before the v1.18 release. If you think that will be of any help?

PR #805, opened 2019-06-26
*: Cleanup references to removed API

LiliC:

Just some cleanup that was leftover as part of https://github.com/kubernetes/kube-state-metrics/issues/802.

@tariq1890 you left the issue open, is there anything else that needs to be done as part of https://github.com/kubernetes/kube-state-metrics/issues/802 that you found.


tariq1890:
@LiliC The `ingresses` have yet to be moved. So I thought we'll wait until the next k8s release.

PR #806, opened 2019-06-26
use common method for generating prom labels from kube labels and ann…

tariq1890:
…otations

Fixes #

### DIFF
--- a/pkg/whiteblacklist/whiteblacklist.go
+++ b/pkg/whiteblacklist/whiteblacklist.go
@@ -26,7 +26,7 @@ import (
 // WhiteBlackList encapsulates the logic needed to filter based on a string.
 type WhiteBlackList struct {
 	list        map[string]struct{}
-	rList       []regexp.Regexp
+	rList       []*regexp.Regexp
 	isWhiteList bool
 }
 
@@ -62,13 +62,13 @@ func New(w, b map[string]struct{}) (*WhiteBlackList, error) {
 
 // Parse parses and compiles all of the regexes in the whiteBlackList.
 func (l *WhiteBlackList) Parse() error {
-	var regexes []regexp.Regexp
+	var regexes []*regexp.Regexp
 	for item := range l.list {
 		r, err := regexp.Compile(item)
 		if err != nil {
 			return err
 		}
-		regexes = append(regexes, *r)
+		regexes = append(regexes, r)
 	}
 	l.rList = regexes
 	return nil


internal/store/utils.go
```diff
@@ -58,30 +63,18 @@ func addConditionMetrics(cs v1.ConditionStatus) []*metric.Metric {
 	return ms
 }
 
-func kubeLabelsToPrometheusLabels(labels map[string]string) ([]string, []string) {
-	labelKeys := make([]string, len(labels))
-	labelValues := make([]string, len(labels))
+func kubeMetadataToPrometheusLabels(keyPrefix string, metadata map[string]string) ([]string, []string) {
+	labelKeys := make([]string, len(metadata))
```
andyxning: `s/labelKeys/metadataKeys`
`s/labelValues/metadataValues`

andyxning: `sanitizeLabelName` may also refactored to `sanitizeMetadataName` to keep in sync.

internal/store/certificatesigningrequest.go
```diff
@@ -38,7 +38,7 @@ var (
 			Type: metric.Gauge,
 			Help: descCSRLabelsHelp,
 			GenerateFunc: wrapCSRFunc(func(j *certv1beta1.CertificateSigningRequest) *metric.Family {
-				labelKeys, labelValues := kubeLabelsToPrometheusLabels(j.Labels)
+				labelKeys, labelValues := kubeMetadataToPrometheusLabels(labelPrefix, j.Labels)
```
brancz: Nit but I think i would prefer these to be explicit functions for labels/annotations.

LiliC: I think if we rename it to something generic like `kubeMetadataToPrometheusMapping` or `kubeMetadataToPrometheusValue` or something along those lines, and add a comment to the function saying it can return either label or annotation, I think this would make sense IMHO.

brancz: The prefix is used to prefix the label when it's converted to a prometheus label.

tariq1890: I can understand that. If that is preferred (thoughts @andyxning ?), I can just  change this PR to rename the method `kubeAnnotationsToPrometheusAnnotations` to `kubeAnnotationsToPrometheusLabels`.

internal/store/utils.go
tariq1890: Here it has to be labelKeys and labelVaules IMO as the method returns prometheus label k-v pairs.

internal/store/certificatesigningrequest.go
andyxning: > I can just change this PR to rename the method `kubeAnnotationsToPrometheusAnnotations` to `kubeAnnotationsToPrometheusLabels`.

Seems reasonable. I agree.

brancz: Yep, I prefer that as well.

brancz:
How come closing this? The underlying idea was good I think.

tariq1890:
@brancz Force-push fail lol

andyxning:
Open it in another PR once you have time. :)

brancz:
He already did and it was already merged :)

andyxning:
Something messy must have occurred to me. :) Just see it in the merged PR.

PR #822, opened 2019-07-12
Support for job_status_duration_time metrics

### LINKED ISSUE #821: Support for job_status_duration_time metrics
/kind feature

**What happened**:
I feel worthwhile to be able to get the  duration time of the job.
For example, by continuously monitoring the duration time of jobs generated from Cronjob, you can notice performance degradation and unintended long-term execution.

Ready for implementation :)
https://github.com/int-tt/kube-state-metrics/commit/a3a4a2468dd1d41384b7a9da39d4acce48fbf4f2
**Environment**:

int-tt:

Fixes #821

### DIFF
--- a/internal/store/job.go
+++ b/internal/store/job.go
@@ -263,6 +263,24 @@ var (
 			}),
 		},
 		{
+			Name: "kube_job_status_duration_time",
+			Type: metric.Gauge,
+			Help: "DurationTime represents time when the job was duration.",
+			GenerateFunc: wrapJobFunc(func(j *v1batch.Job) *metric.Family{
+				ms := []*metric.Metric{}
+				if j.Status.CompletionTime != nil && j.Status.StartTime != nil{
+					ms = append(ms, &metric.Metric{
+
+						Value: j.Status.CompletionTime.Sub(j.Status.StartTime.Time).Seconds(),
+					})
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
 			Name: "kube_job_owner",
 			Type: metric.Gauge,
 			Help: "Information about the Job's owner.",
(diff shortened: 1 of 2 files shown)


internal/store/job.go
```diff
@@ -262,6 +262,24 @@ var (
...
+		{
+			Name: "kube_job_status_duration_time",
+			Type: metric.Gauge,
+			Help: "DurationTime represents time when the job was duration.",
+			GenerateFunc: wrapJobFunc(func(j *v1batch.Job) *metric.Family{
+				ms := []*metric.Metric{}
+				if j.Status.CompletionTime != nil && j.Status.StartTime != nil{
+					ms = append(ms, &metric.Metric{
+
+						Value: j.Status.CompletionTime.Sub(j.Status.StartTime.Time).Seconds(),
```
brancz: we typically don't do any computation in kube-state-metrics, why is using the metrics for each of these values not sufficient?

int-tt: Although Duration can be obtained with the `kubectl get jobs` command, since it is internally calculated, this method has been followed.
```
NAME                                              COMPLETIONS   DURATION   AGE
example-cronjob-1560267000                       0/1           19d        19d
example-cronjob-1561735800                       1/1           15s        2d15h
example-cronjob-1561822200                       1/1           22m        39h
example-cronjob-1561908600                       1/1           17m        15h
```
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubectl/describe/versioned/describe.go#L2092-L2094

brancz: not saying that's a wrong calculation, rather saying we can already do that with

```
kube_job_status_completion_time - kube_job_status_start_time
```

internal/store/job.go
```diff
@@ -262,6 +262,24 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_job_status_duration_time",
+			Type: metric.Gauge,
+			Help: "DurationTime represents time when the job was duration.",
```
tariq1890: This description doesn't make sense. Can you change this?

LiliC:
@int-tt are you still willing to continue with this PR?

Issue #830, opened 2019-07-18
Large Number of VPA Objects Results in Segfault with verticalpodautoscalers Enabled


stewbernetes:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 

/kind bug

**What happened**:
I am attempting to take advantage of the metrics being emitted for VPA with the 1.7.0 release. 

When I enable all collectors plus the VPA collector as follows:
```      containers:
      - name: kube-state-metrics
        image: quay.io/coreos/kube-state-metrics:v1.7.0
        args:
          - '--collectors=certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,limitranges,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,verticalpodautoscalers'
          - '--v=5'
```

I experience the following crash:
```
I0718 14:46:48.473233       1 main.go:90] Using collectors certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,limitranges,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,verticalpodautoscalers
I0718 14:46:48.473295       1 main.go:99] Using all namespace
I0718 14:46:48.473302       1 main.go:140] metric white-blacklisting: blacklisting the following items:
W0718 14:46:48.473323       1 client_config.go:541] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0718 14:46:48.474207       1 main.go:185] Testing communication with server
I0718 14:46:48.487955       1 main.go:190] Running with Kubernetes cluster version: v1.11. git version: v1.11.7. git tree state: clean. commit: 65ecaf0671341311ce6aea0edab46ee69f65d59e. platform: linux/amd64
I0718 14:46:48.487972       1 main.go:192] Communication with server successful
I0718 14:46:48.488128       1 main.go:201] Starting kube-state-metrics self metrics server: 0.0.0.0:8081
[...]

PR #831, opened 2019-07-18
Fix crash due to missing ingress http block

ebabani:
It's acceptable although uncommon for an ingress to have a rule and a host,
but no HTTP block for that particular rule.

Fixes a crash when an ingress exists with rules but no `HTTP` block.

Fixes #

### DIFF
--- a/internal/store/ingress.go
+++ b/internal/store/ingress.go
@@ -104,12 +104,14 @@ var (
 			GenerateFunc: wrapIngressFunc(func(i *v1beta1.Ingress) *metric.Family {
 				ms := []*metric.Metric{}
 				for _, rule := range i.Spec.Rules {
-					for _, path := range rule.HTTP.Paths {
-						ms = append(ms, &metric.Metric{
-							LabelKeys:   []string{"host", "path", "service_name", "service_port"},
-							LabelValues: []string{rule.Host, path.Path, path.Backend.ServiceName, path.Backend.ServicePort.String()},
-							Value:       1,
-						})
+					if rule.HTTP != nil {
+						for _, path := range rule.HTTP.Paths {
+							ms = append(ms, &metric.Metric{
+								LabelKeys:   []string{"host", "path", "service_name", "service_port"},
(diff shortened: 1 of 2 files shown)


tariq1890:
Thanks for the fix @ebabani . @brancz This is definitely a bug fix. Do you think we should backport this to release 1.7 and release 1.6?

brancz:
I'm thinking 1.7, https://github.com/kubernetes/kube-state-metrics/issues/830 looks like a candidate for a v1.7.1 as well

tariq1890:
Fine with me. I'll retarget this PR.

tariq1890:
@ebabani Can you rebase this PR against the `release-1.7` branch?

tariq1890:
Putting the PR on /hold until it's rebased.

ebabani:
This does not need to be backported to 1.6.x, these metrics were first added in 1.7. I'll rebase the PR

tariq1890:
I still see non-release commits from the master branch in this PR. Can you fix that?

ebabani:
Very odd, but should be fixed now. Sorry about that @tariq1890

tariq1890:
I will wait on @brancz for final approval :).

internal/store/ingress.go
```diff
@@ -104,10 +104,18 @@ var (
...
+								LabelKeys:   []string{"host", "path", "service_name", "service_port"},
+								LabelValues: []string{rule.Host, path.Path, path.Backend.ServiceName, path.Backend.ServicePort.String()},
+								Value:       1,
+							})
+						}
+					} else {
 						ms = append(ms, &metric.Metric{
 							LabelKeys:   []string{"host", "path", "service_name", "service_port"},
-							LabelValues: []string{rule.Host, path.Path, path.Backend.ServiceName, path.Backend.ServicePort.String()},
+							LabelValues: []string{rule.Host, "", "", ""},
```
brancz: If there is no path, I don't think this metric should be exposed at all I think, so there shouldn't be the else path. Any reason we would need it? (remember, that labels with empty values are just thrown away by Prometheus)

ebabani: I'm good with not exposing this metric. Will make the change.

PR #833, opened 2019-07-18
*: Cut v1.7.1

tariq1890:
Includes the following two fixes:

#831 #832


tariq1890:
@loburm @piosz Requesting gcr images to be built and pushed for v1.7.1.

loburm:
I think we should somehow change release process. I can't push new gcr image, because there is no 1.7.1 tag. And I assume you don't want to create new tag, because we don't have yet the image?

tariq1890:
@LiliC Perhaps it's the latter.

@loburm We are definitely open to making changes.

PR #834, opened 2019-07-19
pods: Add pod_restart_policy metric

### LINKED ISSUE #263: Difference between a restart=Never and restart=Always is very important for pod states
Currently pod metrics don't distinguish well between restart Never and Always pods.  Since the two workloads are very, very different, it doesn't seem unreasonable to distinguish them.

`kube_pod_created` really lacks the equivalent in `kube_pod_terminated` - once a pod is terminated, the majority of its function on the cluster is complete.  None of the existing pod metrics represent that in any way.

Example queries that are difficult / impossible:

1. number of pods that would fall under the terminating quota at this time
2. sum of pod resource limits in the terminating quota
3. average runtime of pods per namespace
4. average runtime of pods by resource request

smarterclayton:
Report pod_restart_policy{..., type="Always|OnFailure|Never"} 1 for each pod, which allows an admin to know how many batch vs service workload pods are running on the cluster.

Fixes #263

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -26,6 +26,7 @@
 | kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
 | kube_pod_container_resource_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | DEPRECATED |
 | kube_pod_created | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
+| kube_pod_restart_policy | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `type`=&lt;Always|Never|OnFailure&gt; | STABLE |
 | kube_pod_init_container_info | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `image`=&lt;image-name&gt; <br> `image_id`=&lt;image-id&gt; <br> `container_id`=&lt;containerid&gt; | STABLE |
 | kube_pod_init_container_status_waiting | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_init_container_status_waiting_reason | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;ContainerCreating\|CrashLoopBackOff\|ErrImagePull\|ImagePullBackOff\|CreateContainerConfigError&gt; | STABLE |
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -197,6 +197,22 @@ var (
 			}),
 		},
 		{
+			Name: "kube_pod_restart_policy",
+			Type: metric.Gauge,
+			Help: "Describes the restart policy in use by this pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{"type"},
+							LabelValues: []string{string(p.Spec.RestartPolicy)},
+							Value:       float64(1),
+						},
+					},
+				}
+			}),
+		},
+		{
 			Name: "kube_pod_status_scheduled_time",
 			Type: metric.Gauge,
 			Help: "Unix timestamp when pod moved into scheduled status",
--- a/main_test.go
+++ b/main_test.go
@@ -162,6 +162,9 @@ kube_pod_labels{namespace="default",pod="pod0"} 1
 # HELP kube_pod_created Unix creation timestamp
 # TYPE kube_pod_created gauge
 kube_pod_created{namespace="default",pod="pod0"} 1.5e+09
+# HELP kube_pod_restart_policy Describes the restart policy in use by this pod.
+# TYPE kube_pod_restart_policy gauge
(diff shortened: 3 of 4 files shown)


internal/store/pod.go
```diff
@@ -196,6 +196,26 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_restart_policy_always",
+			Type: metric.Gauge,
+			Help: "Set to 1 if the pod has restartPolicy Always, 0 otherwise.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				value := float64(0)
```
tariq1890: Can we make this more readable ? I feel the suggestion below would be idiomatic and suggestive.
```suggestion
				var value float64
```

tariq1890:
@smarterclayton You need change the expectedLength to 39 over [here](https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/pod_test.go#L1399)

main_test.go
```diff
@@ -19,6 +19,7 @@ package main
 import (
 	"bytes"
 	"context"
+	"fmt"
```
tariq1890: Unintentional I suppose?

main_test.go
```diff
@@ -313,6 +317,7 @@ kube_pod_container_resource_limits_memory_bytes{namespace="default",pod="pod0",c
 	sort.Strings(gotFiltered)
 
 	if len(expectedSplit) != len(gotFiltered) {
+		fmt.Printf("%s\n", string(body))
```
tariq1890: Debug statement?

internal/store/pod.go
```diff
@@ -196,6 +196,26 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_restart_policy_always",
+			Type: metric.Gauge,
+			Help: "Set to 1 if the pod has restartPolicy Always, 0 otherwise.",
```
tariq1890:
```suggestion
			Help: "Set to 1 if the pod's restart policy is specified as Always, 0 otherwise.",
```

main_test.go
smarterclayton: The statement isn’t very useful as is - if you want to fix the problem you end up copying and pasting then viewing diff.  I’m not going to force it but most people adding something have to print the output anyway

tariq1890: I understand and that's certainly fair. We can definitely look into adding a helpful diff statement for it. Also, sounds like a potential beginner-friendly issue that we can raise :).

LiliC: Any reason why we can't incorporate this in the next line to look something like:
```go
t.Fatalf("expected different output length\nfull output:\n%s\n", string(body))
```

smarterclayton: One disadvantage is that it won't be flush, so you can't cut and paste it.  You have to copy, demargin it, then paste it.  The closest equivalent is the way we do the RBAC generation tests in k/k - the test verifies you match the expected output, but if you don't match, makes it easy to add the expected output.  In those tests we typically left align (again, for ease of consumption).

smarterclayton: I'll drop it for now.

LiliC: I fine with keeping it, thanks for the explanation!

LiliC:
As @tariq1890 mentioned, https://github.com/kubernetes/kube-state-metrics/blob/d9068c21f930e414064c0a282a1465d8104184fc/internal/store/pod_test.go#L1429 the 38 here needs to be changed to 39, the error seems to be outdated and should have been 38 there in the first place.

Do you mind making that 39 a variable called something like `expectedFamilies` to be used both in the error msg and the comparison, so we don't run into that problem again. Thanks!

smarterclayton:
Interesting, that doesn't fail locally for me.

smarterclayton:
Oh, that's a benchmark, not a regular test.

tariq1890:
We need to add the new metrics over here as well - https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

LiliC:
As a side note:
@tariq1890 good catch on the docs! This should have been detected by our docs tests, I will open an issue to explain why it wasn't.

tariq1890:
We need to add the new metric [here](https://github.com/kubernetes/kube-state-metrics/blob/01e5aae021d75260de149a5a1be574dc5b692f6c/internal/store/pod_test.go#L38) as well. Sorry :(, I forgot to mention that.

internal/store/pod.go
```diff
@@ -196,6 +196,26 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_restart_policy_always",
+			Type: metric.Gauge,
+			Help: "Set to 1 if the pod's restart policy is Always, 0 otherwise.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				var value float64
+				if p.Spec.RestartPolicy == v1.RestartPolicyAlways {
```
brancz: I'm ok with only having a metric on this condition for now, but in order to allow potential expansion in the future, I'd prefer if we renamed this metric to `kube_pod_restart_policy` and make `restart_policy="Always"` a label, then still do the same with the value as this already does.

tariq1890: Makes sense. So iiuc, is what you’re suggesting is the option 1 approach as laid out by @smarterclayton in #263.

smarterclayton: Yeah, that's why I went with this path - you sounded like you preferred option 2 :)

tariq1890: Sorry about the confusion @smarterclayton :(.

docs/pod-metrics.md
```diff
@@ -26,6 +26,7 @@
 | kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
 | kube_pod_container_resource_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | DEPRECATED |
 | kube_pod_created | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
+| kube_pod_restart_policy_always | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; |
```
tariq1890:
```suggestion
| kube_pod_restart_policy_always | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
```

internal/store/pod.go
brancz: Yeah sorry, that was a misunderstanding.

smarterclayton: Np, updated.

smarterclayton:
Updated, PTAL

PR #835, opened 2019-07-19
feat: add kube_pod_status_unschedulable metric

### LINKED ISSUE #829: Feature request: add pod unschedulable reason
/kind feature

When node run out of resources, I think we should support `Unschedulable` reason for Pod
```
status:
  conditions:
  - lastProbeTime: null
    lastTransitionTime: 2019-07-18T07:04:38Z
    message: '0/3 nodes are available: 3 Insufficient cpu.'
    reason: Unschedulable
    status: "False"
    type: PodScheduled
```

pigletfly:

add kube_pod_status_unscheduled

**Which issue(s) this PR fixes** :

Fixes #829

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -40,3 +40,4 @@
 | kube_pod_spec_volumes_persistentvolumeclaims_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
 | kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+| kube_pod_status_unschedulable | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -238,6 +238,31 @@ var (
 			}),
 		},
 		{
+			Name: "kube_pod_status_unschedulable",
+			Type: metric.Gauge,
+			Help: "Describes the unschedulable status for the pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				for _, c := range p.Status.Conditions {
+					switch c.Type {
+					case v1.PodScheduled:
+						if c.Status == v1.ConditionFalse {
+							ms = append(ms, &metric.Metric{
+								LabelKeys:   []string{},
+								LabelValues: []string{},
+								Value:       1,
+							})
+						}
+					}
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
 			Name: "kube_pod_status_phase",
 			Type: metric.Gauge,
 			Help: "The pods current phase.",
--- a/internal/store/pod_test.go
+++ b/internal/store/pod_test.go
@@ -1157,6 +1157,30 @@ kube_pod_container_status_last_terminated_reason{container="container7",namespac
 		{
 			Obj: &v1.Pod{
 				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pod2",
+					Namespace: "ns2",
+				},
+				Status: v1.PodStatus{
+					Conditions: []v1.PodCondition{
+						{
+							Type:    v1.PodScheduled,
+							Status:  v1.ConditionFalse,
+							Reason:  "Unschedulable",
+							Message: "0/3 nodes are available: 3 Insufficient cpu.",
+						},
+					},
+				},
+			},
+			Want: `
+				# HELP kube_pod_status_unschedulable Describes the unschedulable status for the pod.
+				# TYPE kube_pod_status_unschedulable gauge
+				kube_pod_status_unschedulable{namespace="ns2",pod="pod2"} 1
+			`,
+			MetricNames: []string{"kube_pod_status_unschedulable"},
+		},
+		{
+			Obj: &v1.Pod{
+				ObjectMeta: metav1.ObjectMeta{
 					Name:      "pod1",
 					Namespace: "ns1",
 				},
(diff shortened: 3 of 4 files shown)


docs/pod-metrics.md
```diff
@@ -40,3 +40,4 @@
 | kube_pod_spec_volumes_persistentvolumeclaims_info | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
 | kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+| kube_pod_status_unscheduled_reason | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;Unschedulable&gt; <br> `message`=&lt;0/3 nodes are available: 3 Insufficient cpu&gt; <br> | STABLE |
```
andyxning: Since the metrics is named by `kube_pod_status_unscheduled_reason`, the `reason` label is redundant. BTW, `message` label can be renamed to `reason` label.

internal/store/pod.go
```diff
@@ -221,6 +221,31 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_status_unscheduled_reason",
+			Type: metric.Gauge,
+			Help: "Describes the reason the pod is unschedulable.",
```
andyxning: `Describes the reason the pod is unschedulable.` -> `Describes the reason why the pod is unschedulable.`

docs/pod-metrics.md
andyxning: Seems this has nothing to do with `container`.

pigletfly: Good catch, in this case, I think the reason is specific, it's `Unschedulable`.So we can call this `kube_pod_status_unscheduled_message`? Or keep it with `kube_pod_status_unscheduled_reason`?

tariq1890:
@pigletfly You would need to implement this in a way similar to we have implemented the following metrics

i) `"kube_pod_container_status_waiting"`
ii) `"kube_pod_container_status_terminated"`

pigletfly:
> @pigletfly You would need to implement this in a way similar to we have implemented the following metrics
> 
> i) `"kube_pod_container_status_waiting"`
> ii) `"kube_pod_container_status_terminated"`

Actually, when a Pod is in unscheduled state, the container in the Pod hasn't been created yet.Should we implement two metrics:

```
kube_pod_status_unscheduled
kube_pod_status_unscheduled_reason
```

tariq1890:
Why not just have it in one single metric? The reason metric is only populated when the pod is set to unschedulable. If pod is not in the unschedulable state, the label values would be empty and then it would just be dropped by prometheus.

And by implementing the metric family similar to the container_status_waiting and container_status_terminated, I meant use a known subset of reasons (like we have those two metrics) to limit the cardinality.

docs/pod-metrics.md
tariq1890: We should not be storing the message in metrics . Just the reason should suffice.

pigletfly:
@tariq1890 @andyxning PTAL

internal/store/pod_test.go
```diff
@@ -1072,6 +1074,28 @@ kube_pod_container_status_last_terminated_reason{container="container7",namespac
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "pod2",
+					Namespace: "ns2",
+				},
+				Status: v1.PodStatus{
+					Conditions: []v1.PodCondition{
+						{
+							Type:    v1.PodScheduled,
+							Status:  v1.ConditionFalse,
+							Reason:  "Unschedulable",
```
tariq1890: Is this the only known value for `reason`?

pigletfly: yes, according to https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/core/types.go#L2213, `Unschedulable`  is the only known reason for this situation.

tariq1890:
Sorry for the delay on this PR. If there is only one known value. It doesn't make much sense to store that as a label or even have this metric- `"kube_pod_status_unscheduled_reason:`

We might just be better off having this metric `"kube_pod_status_unscheduled"` in that case.

pigletfly:
@tariq1890 OK,it sounds more reasonable.I will change the metric to `kube_pod_status_unscheduled`

tariq1890:
Hi @pigletfly : Based on this [comment](https://github.com/kubernetes/kube-state-metrics/pull/850/files#r310180381), it wouldn't actually make sense to have this metric `"kube_pod_status_unscheduled"`. I should have been more attentive of this and I apologize for this oversight :(.

internal/store/pod.go
```diff
@@ -237,6 +237,31 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_status_unscheduled",
```
brancz: this is a condition happening during scheduling right? if so wouldn't this be "unschedulable"?

pigletfly: yeah, this is a condition happening during scheduling. As described in https://github.com/kubernetes/kube-state-metrics/issues/829,the condition type is `PodScheduled`,but the status is `False`, because of insufficient resource or specific node selector or node affinity strategy.

brancz: Yes I'm just saying `unscheduled` has a different meaning than `unschedulable`. The first is something eviction like, the latter describes this situation, so we should change the metric name to `kube_pod_status_unschedulable`.

brancz: I do wonder if we need a separate metric at all though, as we already have a metric that reflects a pod being in pending phase.

pigletfly: `kube_pod_status_unschedulable` will be a better metric name,is there a way to describe the state that pod can't be scheduled because of lack of resources without adding this metric?If there is a way, the `kube_pod_status_unschedulable` shouldn't be added.

brancz: kube_pod_status_phase{phase=“Pending”} should capture the same. If this metric is the active one for 1min then the Pod can’t be scheduled.

pigletfly: if the image is big, the pulling state is also in `kube_pod_status_phase{phase=“Pending”}`, the `Pending` state includes both image pulling and unschedulable.My origin intention is to distinguish the two state.

brancz: I see. In that case this should be ok to add.

pigletfly:
changed the metric name to `kube_pod_status_unschedulable`. @brancz PTAL

tariq1890:
Thanks for your contribution @pigletfly :)

PR #841, opened 2019-07-21
Add metrics for tls host and secretname

ebabani:
**What this PR does / why we need it**: It adds metrics for the ingress tls host section, indicating which secret is being used for which host.

This is useful to alert in case multiple ingresses are defining different tls secrets for the same host, which could lead to undefined behaviour by some ingress controllers.

Fixes #

### DIFF
--- a/docs/ingress-metrics.md
+++ b/docs/ingress-metrics.md
@@ -8,3 +8,4 @@
 | kube_ingress_metadata_resource_version  | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `resource_version`=&lt;ingress-resource-version&gt; | STABLE |
 | kube_ingress_path | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `host`=&lt;ingress-host&gt; <br> `path`=&lt;ingress-path&gt; <br> `service_name`=&lt;service name for the path&gt; <br> `service_port`=&lt;service port for hte path&gt; | STABLE |
 | kube_ingress_annotations | Gauge | `annotation_INGRESS_ANNOTATION`=&lt;INGRESS_ANNOTATION&gt; <br> `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; | EXPERIMENTAL |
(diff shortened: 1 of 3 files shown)


docs/ingress-metrics.md
```diff
@@ -8,3 +8,4 @@
 | kube_ingress_metadata_resource_version  | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `resource_version`=&lt;ingress-resource-version&gt; | STABLE |
 | kube_ingress_path | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `host`=&lt;ingress-host&gt; <br> `path`=&lt;ingress-path&gt; <br> `service_name`=&lt;service name for the path&gt; <br> `service_port`=&lt;service port for hte path&gt; | STABLE |
 | kube_ingress_annotations | Gauge | `annotation_INGRESS_ANNOTATION`=&lt;INGRESS_ANNOTATION&gt; <br> `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; | EXPERIMENTAL |
+| kube_ingress_tls | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `tls_host`=&lt;tls hostname&gt; <br> `secret_name`=&lt;tls secret name&gt;| EXPERIMENTAL |
```
tariq1890: This doesn't have to be EXPERIMENTAL, if the host and secret names are stable fields on the IngressSpec.

ebabani: Thanks, will update.

internal/store/ingress.go
```diff
@@ -136,6 +136,26 @@ var (
...
+		{
+			Name: "kube_ingress_tls",
+			Type: metric.Gauge,
+			Help: "Ingress TLS host and secret information.",
+			GenerateFunc: wrapIngressFunc(func(i *v1beta1.Ingress) *metric.Family {
+				ms := []*metric.Metric{}
+				for _, tls := range i.Spec.TLS {
+					for _, host := range tls.Hosts {
+						ms = append(ms, &metric.Metric{
+							LabelKeys:   []string{"tls_host", "secret_name"},
```
brancz: this should be `secret` instead of `secret_name` according to the convention (this allows for easier joining of resources)

ebabani: It is done

PR #842, opened 2019-07-21
Deterministic label ordering

brancz:

Prometheus caches the byte representation of a time-series in a metrics endpoint, so non-deterministic behavior in this ordering causes lots of cache misses in Prometheus' scrape logic. Beyond that, testing these metrics is painful when it's not deterministic.

Discovered this while working on #613 .

@tariq1890 @LiliC @mxinden @andyxning

### DIFF
--- a/internal/store/utils_test.go
+++ b/internal/store/utils_test.go
@@ -167,6 +167,15 @@ func TestKubeLabelsToPrometheusLabels(t *testing.T) {
 			expectKeys:   []string{"label__app5"},
 			expectValues: []string{"starts_with_underscore"},
 		},
+		{
+			kubeLabels: map[string]string{
+				"an":    "",
+				"order": "",
+				"test":  "",
+			},
+			expectKeys:   []string{"label_an", "label_order", "label_test"},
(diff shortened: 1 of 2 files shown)


internal/store/utils_test.go
```diff
@@ -167,6 +167,15 @@ func TestKubeLabelsToPrometheusLabels(t *testing.T) {
 			expectKeys:   []string{"label__app5"},
 			expectValues: []string{"starts_with_underscore"},
 		},
+		{
+			kubeLabels: map[string]string{
+				"an":    "",
+				"order": "",
+				"test":  "",
+			},
+			expectKeys:   []string{"label_an", "label_order", "label_test"},
```
brancz: if run multiple times against the previous code, this will fail due to non-deterministic ordering in the above map at instantiation time

tariq1890:
/hold

Whoops can you squash the commits?

tariq1890:
Also, let's have a slightly more descriptive commit message.

tariq1890:
/hold cancel
Perfect! Thanks @brancz

PR #844, opened 2019-07-21
vendor: remove unnecessary dependencies from the project.

tariq1890:
I meant to PR this a long time ago. (Didn't realize I had this branch for so long)

Kube-state-metrics still uses glog transitively through `openshift/origin`. This PR corrects that and cleans the extraneous deps.


Makefile
```diff
@@ -29,7 +29,14 @@ validate-modules:
 	@git diff --exit-code -- go.sum go.mod vendor/
 
 licensecheck:
-	./tests/check_license.sh
+	@echo ">> checking license header"
```
tariq1890: Taken from [prometheus](https://github.com/prometheus/prometheus/blob/818373a5f8227cd80d173e58209688ecdac9e05e/Makefile.common#L125).

Makefile
```diff
@@ -29,7 +29,14 @@ validate-modules:
...
+	@licRes=$$(for file in $$(find . -type f -iname '*.go' ! -path './vendor/*') ; do \
+               awk 'NR<=5' $$file | grep -Eq "(Copyright|generated|GENERATED)" || echo $$file; \
```
tariq1890: `NR<=5` needed because we need to account for files with go build tags.

PR #847, opened 2019-07-24
Makefile: Use extended regex instead of basic regex to find metric names

### LINKED ISSUE #839: Improve doccheck make target
Currently, the dockcheck make target makes sure the documentation is in sync with the code, but it is not correct. It seems like we search through all the test files to get the metric names by matching the [headers](https://github.com/kubernetes/kube-state-metrics/blob/master/Makefile#L45).

The improvement of this check would be to source the metric names from the go files instead of the test files. This way we can make sure that the newly added metrics are well documented as well as unit tested.

We should search for all the metric names in the files of the `internal/store` directory, that do not end with `_test.go`.

spangenberg:
The + sign in the regex is not part of basic regular expressions.
Using extended regular expressions however enabled the use of it.

Fixes #839

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,7 +49,7 @@ doccheck: generate
 	@git diff --exit-code
 	@echo "- Checking if the documentation is in sync with the code..."
 	@grep -hoE '(kube_[^ |]+)' docs/* --exclude=README.md| sort -u > documented_metrics
-	@sed -n 's/.*# TYPE \(kube_[^ ]\+\).*/\1/p' internal/store/*_test.go | sort -u > tested_metrics
+	@sed -nE 's/.*# TYPE (kube_[^ ]+).*/\1/p' internal/store/*_test.go | sort -u > tested_metrics
 	@diff -u0 tested_metrics documented_metrics || (echo "ERROR: Metrics with - are present in tests but missing in documentation, metrics with + are documented but not tested."; exit 1)
 	@echo OK
 	@rm -f tested_metrics documented_metrics


LiliC:
My initial issue was not clear enough and I explained to @spangenberg the problem. We discussed this offline and what we could do is just find all the metric names in the .go files instead. So for example here it would be the string after `Name: "` https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/daemonset.go#L56

That way we don't relly on the metric names added to the test files. Does that sgty @tariq1890?

tariq1890:
Sure. I was thinking of even using go to perform this check. We could, perhaps use the go analyze library from the go stdlib to extract the metrics in our source code.

Does this sound feasible?

brancz:
The way we structure metrics now is should be perfectly possible to extract all metrics from all available resource types in go.

LiliC:
@tariq1890 @brancz happy to have that! But I think right now we can fix this and discuss that implementation further in another issue, unless you think it would be an easy implementation?

brancz:
What @LiliC said sounds good to me, but that’s not what this patch is right now. Please update :)

tariq1890:
Sure, I am okay with this PR. I am not sure what Frederic (not tagging him since he's on vacation) meant by Please update. What requires updating here? I believe the patch does what's specified in the PR title.

LiliC:
@tariq1890 
> Sure, I am okay with this PR. I am not sure what Frederic (not tagging him since he's on vacation) meant by Please update. What requires updating here? I believe the patch does what's specified in the PR title.

Yes, it does what it says in the title, but didn't solve the initial problem in the issue it wants to close, which was to get the metric names, with which to run the doc check against, from a different source. But that was sort of solved by https://github.com/kubernetes/kube-state-metrics/pull/848, as we do test with metadata now. But in the cases when a user doesn't add a unit test for that metric we will miss docs for it, which is why I would prefer the metric list name to come from the `internal/store/pod.go` etc.

tariq1890:
Ah, I got it. Thanks for the clarification @LiliC. I am okay with merging this PR if we remove the fixes directive in the OP's comment. @spangenberg Could you do that?

So IIUC, It's preferable to source the metrics names from the source go file (as opposed to the test go file), so that we can ensure they are unit-tested as well documented. If this is right, let's create an issue articulate what needs to be solved and track the effort :).

LiliC:
I took in @spangenberg suggestion about using extended regex in https://github.com/kubernetes/kube-state-metrics/pull/851 so we can close this one, thanks Daniel! :) (Happy to do add you as a co-commiter in that one, if you want to? )

spangenberg:
@LiliC sure, thanks!
I'll close this PR in favor of #851.

PR #848, opened 2019-07-26
Enable headers and metadata in tests

LiliC:
This PR enables headers in all unit tests. See https://github.com/kubernetes/kube-state-metrics/pull/801#discussion_r298561311 for background on this issue.

So this was not as easy to just enable as we first thought. As explained in https://github.com/kubernetes/kube-state-metrics/pull/801#discussion_r298561311 we are now no longer using the `.GatherAndCompare()` so this had to enabled differently.

I used `metric.ExtractMetricFamilyHeaders()` to generate the headers and later on whitelisted based on the `MetricNames input`, same way as we do currently with the metrics themselves.  

**Couple of notes:**
- We sort the metrics for easier comparison and with that, we also sort the headers whenever there is whitelisting present.
- In some places, we do not use the `MetricNames` to whitelist the metrics. I think we should adjust to using that everywhere. Can do that in a follow-up PR if we agree. Because of this we sometimes get not sorted headers.
- There are some minor fixes like naming.

With this, the doc check should start working again. :) 

cc @tariq1890 @brancz


internal/store/configmap_test.go
```diff
@@ -54,6 +41,12 @@ func TestConfigMapStore(t *testing.T) {
 				},
 			},
 			Want: `
+				# HELP kube_configmap_annotations Kubernetes annotations converted to Prometheus labels.
```
LiliC: These are the sortings I was referring to in the description. We first get all the `HELP` and then the `TYPE`, is this a problem for anyone? We still get the correct headers and correct metrics, it's just they are sorted.

internal/store/cronjob_test.go
```diff
@@ -229,12 +256,12 @@ func TestCronJobStore(t *testing.T) {
 ` +
 				fmt.Sprintf("kube_cronjob_next_schedule_time{cronjob=\"ActiveCronJob1NoLastScheduled\",namespace=\"ns1\"} %ve+09\n",
 					float64(ActiveCronJob1NoLastScheduledNextScheduleTime.Unix())/math.Pow10(9)),
-			// TODO: Do we need to specify metricnames?
```
LiliC: I removed this TODO as it didn't need to be there.

But the question is valid: should we be consistent, we either want to always enable whitelisting or not.

internal/store/storageclass_test.go
```diff
@@ -59,15 +50,17 @@ func TestStorageClassStore(t *testing.T) {
 		{
 			Obj: &storagev1.StorageClass{
 				ObjectMeta: metav1.ObjectMeta{
-					Name:              "test_kube_storageclass_created",
+					Name:              "test_kube_storageclass-created",
```
LiliC: This was causing problems with the new regex :D. Also it's now consistent with the rest of the naming (see metric test above).

internal/store/testutils.go
```diff
@@ -114,7 +121,7 @@ func filterMetrics(ms []string, names []string) []string {
 
 	regexps := []*regexp.Regexp{}
 	for _, n := range names {
-		regexps = append(regexps, regexp.MustCompile(fmt.Sprintf("^%v", n)))
+		regexps = append(regexps, regexp.MustCompile(fmt.Sprintf(".*%v.*$", n)))
```
LiliC: This is so we catch the header filters as well.

docs/pod-metrics.md
```diff
@@ -36,7 +36,6 @@
 | kube_pod_init_container_status_last_terminated_reason | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;OOMKilled\|Error\|Completed\|ContainerCannotRun\|DeadlineExceeded&gt; | STABLE |
 | kube_pod_init_container_status_ready | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_init_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
-| kube_pod_init_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
```
LiliC: We don't have this metric anymore, this was discovered while running doccheck test.

internal/store/configmap_test.go
brancz: Not a problem. Prometheus handles this "correctly".

brancz:
Really awesome work! Thank you so much for taking care of this!

/lgtm

PR #850, opened 2019-07-30
[Fea] Get failure information for failed pods.

alanbiquet:
This PR enables, for a pod in Failed phase, to extract information (reason and message) regarding the failure.

The rationale behind this PR is to be able to define and generate _detailed_ alerts using prometheus/alertmanager when pods fail, typically due to eviction.

*Note:*
After some investigation, it seems not possible to list all the possible values for PodStatus.reason and PodStatus.message (see [Kubernetes API](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#podstatus-v1-core)). 
Therefore, the information retrieval in pod.go file is kind of generic and will be able to collect any strings for failure reason and message.

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -279,6 +279,27 @@ var (
 			}),
 		},
 		{
+			Name: "kube_pod_failure_information",
+			Type: metric.Gauge,
+			Help: "The failed pods failure information (reason and message).",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				phase := p.Status.Phase
+				if phase == v1.PodFailed {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"reason", "message"},
+						LabelValues: []string{p.Status.Reason, p.Status.Message},
+						Value:       1,
+					})
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
 			Name: "kube_pod_status_ready",
 			Type: metric.Gauge,
 			Help: "Describes whether the pod is ready to serve requests.",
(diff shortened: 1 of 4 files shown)


alanbiquet:
Hello @tariq1890
Could you please have a look at this PR please?

internal/store/pod.go
```diff
@@ -278,6 +278,27 @@ var (
...
+			Type: metric.Gauge,
+			Help: "The failed pods failure information (reason and message).",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				phase := p.Status.Phase
+				if phase == v1.PodFailed {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"reason", "message"},
+						LabelValues: []string{p.Status.Reason, p.Status.Message},
```
tariq1890: We can't use message in the metrics and if we are to use `Reason`, we need to maintain a known list of Pod failure reasons to prevent potential high cardinality problems.

A similar approach is used in metrics like
i) kube_pod_container_status_terminated
ii) kube_pod_container_status_waiting

tariq1890:
Thanks for your PR @alanbiquet , I have provided my comment.

internal/store/pod.go
alanbiquet: Hi @tariq1890 
Thanks for your review.

Afaik after digging into the code of the Kubernetes project, I could not find an exhaustive list of reasons for failed pods. Also, could you please explain why message are not allowed into metrics? I'm just curious about this. 

Anyway, it was nice to work on this great project. I will close this PR once you replied because I think it will not be possible to merge it according to the previous points.

All the best,
Alan

tariq1890: The potential for high cardinality is higher when we record message strings. Messages are very much unbounded and it is not possible or practical to set a known list for that. Which Is why it makes sense to just have a known list of reasons rather than the message.

tariq1890: By unbounded, I mean that the possible permutations of different messages are very high and can thus lead to high cardinality.

tariq1890: While I don’t think we should close the PR right away, we could still having a metric like “kube_pod_status_failed”.

I am just waiting on the other maintainers to confirm this.

LiliC: @tariq1890 we have the `kube_pod_status_phase` with label `"phase"="Failed"`, just doesn't include the failure msg.

```
kube_pod_status_phase{namespace="default",pod="pod0",phase="Failed"} 0
```

tariq1890: Whoops. Forgot about that :). Thanks

PR #851, opened 2019-07-31
Makefile: Improve doccheck target

LiliC:

Source the metric names from go files instead of test go files, this way
we make sure that all metrics that exist are documented.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/839

cc @tariq1890 

Note: I am sure there is a better way of doing this, happy to accept improvements, but this worked for me :)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,6 @@
 kube-state-metrics
 documented_metrics
-tested_metrics
+code_metrics
 
 # Created by https://www.gitignore.io/api/go
 
--- a/Makefile
+++ b/Makefile
@@ -49,10 +49,10 @@ doccheck: generate
 	@git diff --exit-code
 	@echo "- Checking if the documentation is in sync with the code..."
 	@grep -hoE '(kube_[^ |]+)' docs/* --exclude=README.md| sort -u > documented_metrics
-	@sed -n 's/.*# TYPE \(kube_[^ ]\+\).*/\1/p' internal/store/*_test.go | sort -u > tested_metrics
-	@diff -u0 tested_metrics documented_metrics || (echo "ERROR: Metrics with - are present in tests but missing in documentation, metrics with + are documented but not tested."; exit 1)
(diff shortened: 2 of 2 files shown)


LiliC:
The test is failing as it caught the undocumented and untested `kube_init_pod_container_resource_requests_cpu_cores` metric.

LiliC:
@tariq1890 I noticed your comment https://github.com/kubernetes/kube-state-metrics/pull/801#discussion_r298700970 so I guess we should just remove that metric altogether?

tariq1890:
@LiliC The CI fail is a very good sign then!

tariq1890:
Please squash commit history as well!

LiliC:
My rebase went wrong :) But should be good now. I added Daniel as a co-author as he contributed part of the solution for this.

LiliC:
🎉 test passed https://travis-ci.org/kubernetes/kube-state-metrics/jobs/566124770 :)

LiliC:
@tariq1890 PTAL thanks! :)

tariq1890:
@LiliC You would need to update the `.gitignore`

LiliC:
@tariq1890 good catch! 👍

tariq1890:
@LiliC Could you help me understand how this addresses the scenario where the metrics aren't unit tested?

It looks like your change just sources the metrics from the non test go files.

Side note: Need to squash history again, lol.

LiliC:
> @LiliC Could you help me understand how this addresses the scenario where the metrics aren't unit tested?

Yeah good point, my logic was a bit off there, I didn't think of the scenario where the metrics are documented but not unit tested.

Was thinking of adding both tested_metrics and documented_metrics diff check here, but not sure it belongs in this test as it's clearly about docs. What are you thoughts?

tariq1890:
Sure. I am okay with scoping that out of this PR. Can you amend the first comment then in this thread?

tariq1890:
Thanks @LiliC and @spangenberg. Much appreciated :)

PR #852, opened 2019-07-31
internal/store/pod.go: Remove unnecessary metric

LiliC:
This was found in https://github.com/kubernetes/kube-state-metrics/pull/851 PR. As per the comment in https://github.com/kubernetes/kube-state-metrics/pull/801#discussion_r298700970  we do not need this metric at all:
> CPU core requests are deprecated for pod containers and should never be added to pod init containers since they are new. This is my bad.

cc @tariq1890

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -900,29 +900,6 @@ var (
 			}),
 		},
 		{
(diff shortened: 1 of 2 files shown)


tariq1890:
Yes! Thanks so much for looking into this :)

Issue #854, opened 2019-08-01
Filter annotation "annotation_kubectl_kubernetes_io_last_applied_configuration"


towolf:
Kube-state-metrics 1.7.1 adds new kube_*_annotations metrics.

I disabled most of them because that sounded really spammy. 

`--metric-blacklist="kube_[^i][a-z]+_annotations" `

But I wanted to keep one exception: `kube_ingress_annotations`. There is no other way to capture the ingess class, than via the annotations: https://kubernetes.github.io/ingress-nginx/user-guide/multiple-ingress/

We use `kubectl apply` in our CI/CD pipeline and now every `_annotations` metric contains the full object itself in the `annotation_kubectl_kubernetes_io_last_applied_configuration` label.

Now, we don't store sensitive stuff in the objects, but I think this specific annotation should be excluded. 

**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind bug

**What happened**:

see above 

...

towolf:
Also, annotations are an uncontrolled field, where anything can go, very few contraints are placed on them. Just exposing them across the board is a very bold move.

LiliC:
Yes agreed, I am seeing those as well, we should filter the `annotation_kubectl_kubernetes_io_last_applied_configuration` out in the all the annotations. Unless anyone finds those useful? cc @tariq1890 @brancz 

A full example of how the metric looks like with the `annotation_kubectl_kubernetes_io_last_applied_configuration` annotation included:

```
kube_deployment_annotations{namespace="default",deployment="hello-world",annotation_kubectl_kubernetes_io_last_applied_configuration="{\"apiVersion\":\"apps/v1\",\"kind\":\"Deployment\",\"metadata\":{\"annotations\":{},\"name\":\"hello-world\",\"namespace\":\"default\"},\"spec\":{\"replicas\":2,\"selector\":{\"matchLabels\":{\"run\":\"load-balancer-example\"}},\"template\":{\"metadata\":{\"labels\":{\"run\":\"load-balancer-example\"}},\"spec\":{\"containers\":[{\"image\":\"gcr.io/google-samples/node-hello:1.0\",\"name\":\"hello-world\",\"ports\":[{\"containerPort\":8080,\"protocol\":\"TCP\"}]}]}}}}\n",annotation_deployment_kubernetes_io_revision="1"} 1
```

towolf:
[...]

PR #857, opened 2019-08-03
filter out last applied config deploy annotations

tariq1890:
This is the near-term fix for the issue reported in #854.

### DIFF
--- a/internal/store/utils.go
+++ b/internal/store/utils.go
@@ -71,6 +71,10 @@ func kubeLabelsToPrometheusLabels(labels map[string]string) ([]string, []string)
 }
 
 func kubeAnnotationsToPrometheusLabels(annotations map[string]string) ([]string, []string) {
+	// Remove Last applied configuration annotation since it can potentially store sensitive data.
+	// Refer to https://github.com/kubernetes/kube-state-metrics/issues/854 for more info.
+	// TODO: We need to rethink how we expose annotations since it's pretty unbounded.
+	delete(annotations, v1.LastAppliedConfigAnnotation)
 	annotationKeys := make([]string, len(annotations))
 	annotationValues := make([]string, len(annotations))
 	i := 0
--- a/internal/store/deployment_test.go
+++ b/internal/store/deployment_test.go
@@ -20,7 +20,8 @@ import (
 	"testing"
 	"time"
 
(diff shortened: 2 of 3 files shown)


tariq1890:
cc @LiliC @brancz

tariq1890:
Thanks for the review @LiliC .

internal/store/utils.go
```diff
@@ -75,6 +75,10 @@ func kubeLabelsToPrometheusLabels(labels map[string]string) ([]string, []string)
 }
 
 func kubeAnnotationsToPrometheusLabels(annotations map[string]string) ([]string, []string) {
+	// Remove Last applied configuration annotation since it can potentially store sensitive data.
+	// Refer to https://github.com/kubernetes/kube-state-metrics/issues/854 for more info.
+	// TODO: We need to rethink how we expose annotations since it's pretty unbounded.
+	delete(annotations, "kubectl_kubernetes_io_last_applied_configuration")
```
LiliC: This should actually be `"kubectl.kubernetes.io/last-applied-configuration"` because we don't yet transform it to the Prometheus label format yet. We could also just take the constant `v1.LastAppliedConfigAnnotation` here to make it clear. https://github.com/kubernetes/api/blob/d0d4f3afa3ab/core/v1/annotation_key_constants.go#L71:2

tariq1890: Done :). Thanks.

LiliC:
I think we wanted this to go in the 1.7 branch? Should we do that first or cherry-pick it out?

brancz:
I changed the base branch. This needs a rebase now.

LiliC:
So we (@brancz and I) found a few more annotations that kubernetes applies in the metrics, so we decided to just revert the annotations feature altogether. We cannot know all of the kubernetes annotations and what might potently be in there so reverting the feature might be best right now. Will do a PR for the revert.

We should rethink the annotation metrics feature later on, will open an issue for this.

2rs2ts:
Might I suggest using a prefix you define for the annotations to add? Like `kubernetes.io/kube-state-metrics/<tag name>`?

I'd recommend drawing some inspiration from Datadog's autodiscovery annotations: https://docs.datadoghq.com/agent/autodiscovery/integrations/?tab=kubernetespodannotations#configuration

PR #858, opened 2019-08-04
Add liveness and readiness probe in deployment file

### LINKED ISSUE #810: Implement a readiness endpoint

>
> /kind bug

/kind feature

**What happened**:
Currently there is a healthz endpoint which implements somewhat liveness but is used as a readinessProbe in the kubernetes deployment files.
A possible option would be to use the healthz endpoint as a path for livenessProbe and have a readyz endpoint for readiness checks.
Probably the readiness probe should return 200 if there is a connection to the cluster possible and should fail otherwise (not sure if the app becomes unhealthy otherwise or would just fail)

MIBc:
This is a fix for the issue reported in #810

### DIFF
--- a/kubernetes/kube-state-metrics-deployment.yaml
+++ b/kubernetes/kube-state-metrics-deployment.yaml
@@ -24,9 +24,15 @@ spec:
           containerPort: 8080
         - name: telemetry
           containerPort: 8081
-        readinessProbe:
+        livenessProbe:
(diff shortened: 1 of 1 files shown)


brancz:
/lgtm

Thanks!

MIBc:
I have updated commit message to erase invalid-commit-message label. @brancz

PR #859, opened 2019-08-05
Revert "add kube_*_annotations metrics for all objects"

LiliC:
This reverts commit 5ae00bb93fb6c224324d02d8b57c1fb1147948c9.

As discussed in https://github.com/kubernetes/kube-state-metrics/pull/857#issuecomment-518170731 due to the need to filter out too many annotations created by kubernetes we want to revert this as it was marked as `EXPERIMENTAL` we can do that, and rethink the annotation metrics in a separate PR.

Closes https://github.com/kubernetes/kube-state-metrics/issues/854

cc @tariq1890 @brancz 

Note: we should forward this to master branch as well as open issue to rethink annotation metrics.


internal/store/namespace.go
```diff
@@ -32,6 +32,10 @@ var (
 	descNamespaceLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descNamespaceLabelsDefaultLabels = []string{"namespace"}
 
+	descNamespaceAnnotationsName          = "kube_namespace_annotations"
```
LiliC: Seems like we already had the `kube_namespace_annotations` metric before this addition, but this metric is marked as STABLE in our docs. Any suggestions for this? We can either remove it, but because it's marked as stable, I would say we try to filter out?

LiliC: We discussed this and we want to remove this.

LiliC:
Tested it locally, and this gets rid of all the annotation metrics.

brancz:
/lgtm

PR #866, opened 2019-08-12
*: Add list/watch success/error count total metrics

LiliC:

When kube-state-metrics does not have the correct roles to list or watch on
a resource, it just logs this error but not actually error out. This
is a problem as the pod never restarts and it is hard to catch any
problems as other resource metrics continue to be created correctly. We
only see this error in the noisy kube-state-metrics logs.

This registers two metrics `kube_state_metrics_watch_total` and
`kube_state_metrics_list_total`. With the following labels:
"result" label is the type of action count relates to, error or success.
"resource" label contains the resource <apiVersion.Kind>.

This way we can do a rate alert when kube-state-metrics error rate is
too high.

Example of the metrics:
```
kube_state_metrics_list_total{resource="*v1.Namespace",result="success"} 1
kube_state_metrics_list_total{resource="*v1.Node",result="error"} 52
kube_state_metrics_watch_total{resource="*v1beta1.Ingress",result="success"} 1
```

*Note:* We can also use this new metric in our e2e tests, but I would suggest we migrate our e2e tests from bash to `go`. As parsing the metric on only certain resources that have errors, finding the rate etc. is complex in bash (at least my regex skills are not that good), but we can easily do that in go by using the prometheus library. 
I would suggest to split this up in follow-up PRs, one PR to convert the e2e tests to go, and there we can add this test as well. I can put a TODO maybe in the e2e test to not forget to add this. WDYT?

cc @brancz @tariq1890

### DIFF
--- a/README.md
+++ b/README.md
@@ -108,6 +108,9 @@ See the [`docs`](docs) directory for more information on the exposed metrics.
 kube-state-metrics exposes its own general process metrics under `--telemetry-host` and `--telemetry-port` (default 81).
 
 kube-state-metrics also exposes list and watch success and error metrics. These can be used to calculate the error rate of list or watch resources.
+If you encounter those errors in the metrics, it is most likely a configuration or permission issue, and the next thing to investigate would be looking
+at the logs of kube-state-metrics.
+
 Example of the above mentioned metrics:
 ```
 kube_state_metrics_list_total{resource="*v1.Node",result="success"} 1
--- a/README.md
+++ b/README.md
@@ -91,10 +91,10 @@ additional metrics!
 > For now, the following metrics and collectors
 >
 > **metrics**
->	* kube_pod_container_resource_requests_nvidia_gpu_devices
->	* kube_pod_container_resource_limits_nvidia_gpu_devices
->	* kube_node_status_capacity_nvidia_gpu_cards
->	* kube_node_status_allocatable_nvidia_gpu_cards
+>	* `kube_pod_container_resource_requests_nvidia_gpu_devices`
+>	* `kube_pod_container_resource_limits_nvidia_gpu_devices`
+>	* `kube_node_status_capacity_nvidia_gpu_cards`
+>	* `kube_node_status_allocatable_nvidia_gpu_cards`
 >
 >	are removed in kube-state-metrics v1.4.0.
 >
@@ -104,7 +104,16 @@ additional metrics!
 See the [`docs`](docs) directory for more information on the exposed metrics.
 
 ### Kube-state-metrics self metrics
+
 kube-state-metrics exposes its own general process metrics under `--telemetry-host` and `--telemetry-port` (default 81).
+
+kube-state-metrics also exposes list and watch success and error metrics. These can be used to calculate the error rate of list or watch resources.
+Example of the above mentioned metrics:
+```
+kube_state_metrics_list_total{resource="*v1.Node",result="success"} 1
+kube_state_metrics_list_total{resource="*v1.Node",result="error"} 52
+kube_state_metrics_watch_total{resource="*v1beta1.Ingress",result="success"} 1
+```
 
 ### Resource recommendation
 
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -24,8 +24,6 @@ import (
 
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
-	"k8s.io/klog"
-
 	appsv1 "k8s.io/api/apps/v1"
 	autoscaling "k8s.io/api/autoscaling/v2beta1"
 	batchv1 "k8s.io/api/batch/v1"
@@ -39,11 +37,12 @@ import (
 	vpaclientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
+	"k8s.io/klog"
 
 	"k8s.io/kube-state-metrics/pkg/metric"
 	metricsstore "k8s.io/kube-state-metrics/pkg/metrics_store"
 	"k8s.io/kube-state-metrics/pkg/options"
-	"k8s.io/kube-state-metrics/pkg/util/watcher"
+	"k8s.io/kube-state-metrics/pkg/watch"
 )
 
 type whiteBlackLister interface {
@@ -60,22 +59,23 @@ type Builder struct {
 	ctx              context.Context
 	enabledResources []string
 	whiteBlackList   whiteBlackLister
-	metrics          *watcher.ListWatchMetrics
+	metrics          *watch.ListWatchMetrics
 }
 
 // NewBuilder returns a new builder.
-func NewBuilder(
-	ctx context.Context,
-	r *prometheus.Registry,
-) *Builder {
-	metrics := watcher.NewListWatchMetrics(r)
+func NewBuilder(ctx context.Context) *Builder {
 	return &Builder{
-		ctx:     ctx,
-		metrics: metrics,
+		ctx: ctx,
 	}
 }
 
+// WithMetrics sets the metrics property of a Builder.
+func (b *Builder) WithMetrics(r *prometheus.Registry) {
+	b.metrics = watch.NewListWatchMetrics(r)
+}
+
 // WithEnabledResources sets the enabledResources property of a Builder.
+
 func (b *Builder) WithEnabledResources(c []string) error {
 	for _, col := range c {
 		if !collectorExists(col) {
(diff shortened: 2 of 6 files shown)


README.md
```diff
@@ -104,8 +104,17 @@ additional metrics!
...
 kube-state-metrics exposes its own general process metrics under `--telemetry-host` and `--telemetry-port` (default 81).
 
+kube-state-metrics also exposes list and watch success and error metrics. These can be used to calculate the error rate of list or watch resources.
+Example of the above mentioned metrics:
+```
+kube_state_metrics_list_total{resource="*v1.Node",result="success"} 1
+kube_state_metrics_list_total{resource="*v1.Node",result="error"} 52
+kube_state_metrics_watch_total{resource="*v1beta1.Ingress",result="success"} 1
+```
+
```
brancz: We can probably mention directly that if this happens it's most likely a configuration/permission issue, and logs are the next thing to investigate.

main.go
```diff
@@ -80,8 +80,8 @@ func main() {
 		opts.Usage()
 		os.Exit(0)
 	}
-
-	storeBuilder := store.NewBuilder(ctx)
+	ksmMetricsRegistry := prometheus.NewRegistry()
+	storeBuilder := store.NewBuilder(ctx, ksmMetricsRegistry)
```
brancz: for the better or worse, this should continue to use the same builder pattern I think, at least for consistency if nothing else :)

internal/store/builder.go
```diff
@@ -56,14 +59,18 @@ type Builder struct {
...
 	whiteBlackList   whiteBlackLister
+	metrics          *watcher.ListWatchMetrics
 }
 
 // NewBuilder returns a new builder.
 func NewBuilder(
 	ctx context.Context,
+	r *prometheus.Registry,
 ) *Builder {
+	metrics := watcher.NewListWatchMetrics(r)
```
brancz: nit: this can just be moved to line 73 no?

main.go
LiliC: meaning, to change this to `WithRegistry(ksmMetricsRegistry)`?

tariq1890: @LiliC Is it possible to follow the same pattern as we do currently for this?

```
func (b *Builder) buildDaemonSetStore() *metricsstore.MetricsStore {
	return b.buildStore(daemonSetMetricFamilies, &appsv1.DaemonSet{}, createDaemonSetListWatch)
}

func (b *Builder) buildDeploymentStore() *metricsstore.MetricsStore {
	return b.buildStore(deploymentMetricFamilies, &appsv1.Deployment{}, createDeploymentListWatch)
}

func (b *Builder) buildEndpointsStore() *metricsstore.MetricsStore {
	return b.buildStore(endpointMetricFamilies, &v1.Endpoints{}, createEndpointsListWatch)
}
```
So that we pass the method that returns the new ListerWatcher impl as a function param instead of injecting it as a new struct field param. I prefer not changing the struct signature of `whiteBlackLister` for the sake of restricting its scope to only white & black listing.

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
...
+// NewInstrumentedListerWatcher returns a new InstrumentedListerWatcher.
+func NewInstrumentedListerWatcher(lw cache.ListerWatcher, metrics *ListWatchMetrics, resource string) cache.ListerWatcher {
+	return &InstrumentedListerWatcher{
+		lw:       lw,
+		metrics:  metrics,
+		resource: resource,
+	}
+}
+
+// List is a wraper func around the cache.ListerWatcher.List func. It also increases the counter whenever any
```
tariq1890:
```suggestion
// List is a wrapper func around the cache.ListerWatcher.List func. It also increases the counter whenever any
```

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
...
+	if err != nil {
+		i.metrics.ListTotal.WithLabelValues("error", i.resource).Inc()
+		return
+	}
+
+	i.metrics.ListTotal.WithLabelValues("success", i.resource).Inc()
+	return
+}
+
+// Watch is a wraper func around the cache.ListerWatcher.Watch func. It also increases the counter whenever any
```
tariq1890:
```suggestion
// Watch is a wrapper func around the cache.ListerWatcher.Watch func. It also increases the counter whenever any
```

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
...
+// errors occur when trying to watch a resource.
```
tariq1890: So this increases either one of the counters based on the outcome. We should be mentioning that rather.

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
...
+
+// NewListWatchMetrics takes in a prometheus registry and initializes
+// and registers the kube_state_metrics_list_total and kube_state_metrics_watch_total
+// metrics. It returns those registered metrics.
+func NewListWatchMetrics(r *prometheus.Registry) *ListWatchMetrics {
+	var m ListWatchMetrics
+	m.WatchTotal = prometheus.NewCounterVec(
+		prometheus.CounterOpts{
+			Name: "kube_state_metrics_watch_total",
+			Help: "Number of total resource watches in kube-state-metrics",
```
tariq1890: Would this be better?
```suggestion
			Help: "Total number of resource watches attempted in kube-state-metrics",
```

main.go
LiliC: @tariq1890 we are changing the `Builder` struct, not the `whiteBlackLister` struct, just the github diff displays it silly.

tariq1890:
@LiliC Is it possible to implement the `MetricsProvider` interface of `client-go` and register the metrics that way?

For reference: https://github.com/prometheus/prometheus/blob/master/discovery/kubernetes/client_metrics.go#L35

LiliC:
@tariq1890 I don't see the error/success per resource list/watch in there, am I missing something. Can you explain a bit more about what you were thinking?

tariq1890:
I think as long as the metric is a Counter, we should still able to implement the `MetricsProvider` interface. You can define the counter metric like how you already have with the success/error and resourceType labels added have our `MetricsProvider` type struct obj return these metrics when needed.

Does this approach seem feasible? Where I am coming from is that - I am trying to avoid as much wrapping and signature changes as we can, if we can help it.

main.go
tariq1890: Whoops, sorry about that.

brancz:
Those metrics neither have resource nor success/error labels. I think starting with our own here is reasonable.

tariq1890:
Yes, I agree with you there. What I meant was simply implementing the client Go metrics provider interface in kube state metrics. The implementing the method in our case can return the counter metric with the labels that we need.

pkg/util/watcher/watch.go
tariq1890: @LiliC Can we address this?

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
...
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package watcher
+
+import (
+	"github.com/prometheus/client_golang/prometheus"
+
```
tariq1890:
```suggestion
```

tariq1890: We should group third-party imports together IMO.

pkg/util/watcher/watch.go
```diff
@@ -0,0 +1,99 @@
+/*
```
tariq1890: I am not sure if `pkg/util` is the best place for this? This definitely doesn't feel generic enough to be `util`. Thoughts @LiliC @brancz ?

tariq1890:
Sorry, clicked the approve button by mistake.

tariq1890:
> We can also use this new metric in our e2e tests, but I would suggest we migrate our e2e tests from bash to go. As parsing the metric on only certain resources that have errors, finding the rate etc. is complex in bash (at least my regex skills are not that good), but we can easily do that in go by using the prometheus library.

+1 to migrating e2e tests to go.

pkg/util/watcher/watch.go
LiliC: Can you give a suggestion on the wording here, thanks.

pkg/util/watcher/watch.go
LiliC: hmm any suggestions where it should live?

brancz:
> Yes, I agree with you there. What I meant was simply implementing the client Go metrics provider interface in kube state metrics. The implementing method in our case can return the counter metric with the labels that we need.

We can only control static labels in our implementation, we cannot influence the way these metrics are being used internally. This means the very thing that this PR does, which is show the rate of errors compared to successes is fundamentally not possible with the MetricsProvider. I very much think we should improve the MetricsProvider interface upstream, but this can solve immediate outages and setup problems.

LiliC:
@tariq1890 @brancz PTAL, I think I addressed all the comments, let me know if I left anything out, thanks!

main_test.go
```diff
@@ -27,6 +27,7 @@ import (
 	"testing"
 	"time"
 
+	"github.com/prometheus/client_golang/prometheus"
```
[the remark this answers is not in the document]
LiliC: Somehow locally when I ran the linter it did not show me this error only the one in the `main.go`, strange. Anyways fixed. :)

brancz:
I think we’ve got consensus.

/lgtm

PR #869, opened 2019-08-20
lint: enforce goimports local prefixes

tariq1890:
Raising this PR because I was wondering why the `goimports` settings were never being enforced. Turns out, it was a YAML format fail in `.golangci.yml` *facepalm*. This PR corrects that.

### DIFF
--- a/internal/store/hpa_test.go
+++ b/internal/store/hpa_test.go
@@ -19,10 +19,10 @@ package store
 import (
 	"testing"
 
-	v12 "k8s.io/api/core/v1"
-
 	autoscaling "k8s.io/api/autoscaling/v2beta1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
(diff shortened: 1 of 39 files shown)


internal/store/hpa_test.go
```diff
@@ -19,10 +19,10 @@ package store
 import (
 	"testing"
 
-	v12 "k8s.io/api/core/v1"
-
 	autoscaling "k8s.io/api/autoscaling/v2beta1"
+	v12 "k8s.io/api/core/v1"
```
LiliC: v12? :D

internal/store/utils.go
```diff
@@ -31,7 +30,6 @@ import (
 )
 
 var (
-	resyncPeriod       = 5 * time.Minute
```
LiliC: We are not using this anymore, interesting.

internal/store/hpa_test.go
tariq1890: 0___0 . Thanks for catching that.

PR #877, opened 2019-08-30
store: add kube_node_role metric

abursavich:
This adds a `kube_node_role` metric for node roles based on `node-role.kubernetes.io/<ROLE>` node labels.

This is implemented like `kube_node_spec_taint` where each taint is a separate metric in the same family, as opposed to `kube_node_labels` where each label is mapped to the same metric and family.

I've tried to implement this with recording rules on top of `kube_node_labels`, but an empty `node-role.kubernetes.io/<ROLE>` label value is valid and, as far as i can tell, there's no way to differentiate between a missing label and an empty value in a query. Further, every possible role must be known in advance to query for the expected labels.

N/A

### DIFF
--- a/docs/node-metrics.md
+++ b/docs/node-metrics.md
@@ -4,6 +4,7 @@
 | ---------- | ----------- | ----------- | ----------- |
 | kube_node_info | Gauge | `node`=&lt;node-address&gt; <br> `kernel_version`=&lt;kernel-version&gt; <br> `os_image`=&lt;os-image-name&gt; <br> `container_runtime_version`=&lt;container-runtime-and-version-combination&gt; <br> `kubelet_version`=&lt;kubelet-version&gt; <br> `kubeproxy_version`=&lt;kubeproxy-version&gt; <br> `provider_id`=&lt;provider-id&gt; | STABLE |
 | kube_node_labels | Gauge | `node`=&lt;node-address&gt; <br> `label_NODE_LABEL`=&lt;NODE_LABEL&gt;  | STABLE |
+| kube_node_role | Gauge | `node`=&lt;node-address&gt; <br> `role`=&lt;NODE_ROLE&gt; | EXPERIMENTAL |
 | kube_node_spec_unschedulable | Gauge | `node`=&lt;node-address&gt;|
 | kube_node_spec_taint | Gauge | `node`=&lt;node-address&gt; <br> `key`=&lt;taint-key&gt; <br> `value=`&lt;taint-value&gt; <br> `effect=`&lt;taint-effect&gt; | STABLE |
 | kube_node_status_phase| Gauge | `node`=&lt;node-address&gt; <br> `phase`=&lt;Pending\|Running\|Terminated&gt; | DEPRECATED |
--- a/internal/store/node_test.go
+++ b/internal/store/node_test.go
@@ -68,7 +68,7 @@ func TestNodeStore(t *testing.T) {
 					Name:              "127.0.0.1",
 					CreationTimestamp: metav1.Time{Time: time.Unix(1500000000, 0)},
 					Labels: map[string]string{
-						"type": "master",
+						"node-role.kubernetes.io/master": "",
 					},
 				},
 				Spec: v1.NodeSpec{
@@ -105,6 +105,7 @@ func TestNodeStore(t *testing.T) {
 		# HELP kube_node_created Unix creation timestamp
 		# HELP kube_node_info Information about a cluster node.
 		# HELP kube_node_labels Kubernetes labels converted to Prometheus labels.
+		# HELP kube_node_role The role of a cluster node.
 		# HELP kube_node_spec_unschedulable Whether a node can schedule new pods.
 		# HELP kube_node_status_allocatable The allocatable for different resources of a node that are available for scheduling.
 		# HELP kube_node_status_allocatable_cpu_cores The CPU resources of a node that are available for scheduling.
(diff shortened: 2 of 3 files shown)


brancz:
Nominally this is not really aligned with how kube state metrics works, however I’m inclined to accept something like this as node labels are treated with stability guarantees in kubernetes (at least this label) and it is indeed difficult to do what you are trying to accomplish.

@LiliC @tariq1890

abursavich:
@brancz, I'm not sure this would be more palatable, but in the interest of exploration... another option would be to export a `kube_node_label` metric with stable label dimensions and higher series cardinality. This could be in addition to or in place of the `kube_node_labels` metric.

```
# BEFORE
kube_node_labels{label_foo="bar",label_qux="",node="127.0.0.1"} 1

# AFTER
kube_node_label{label="foo",value="bar",node="127.0.0.1"} 1
kube_node_label{label="qux",value="",node="127.0.0.1"} 1
```
I have actually wished it was that way for the purposes of other queries in the past, but not just for nodes... for deployments, pods, etc. 

As an additional benefit, you would get the real label names without special characters replaced.

tariq1890:
I am on the fence here on this. While I wish this was made more contractually bound via APIObject, I also don't want it to be a blocker. I'll leave it to @LiliC to provide her opinions on this.

LiliC:
> While I wish this was made more contractually bound via APIObject

I agree!

Asking the node team on this, to see if they have plans to change this. If it will stay this way I would say let's add it, unless there is any other place in k8s we think this metric might be better off?

> something like this as node labels are treated with stability guarantees in kubernetes

Dear @kubernetes/sig-node-api-reviews can you let us know if you have any plans to change the `node-role.kubernetes.io/<ROLE>` in the near or far future, thank you!

abursavich:
I would like to be very explicit that this is a specific case of two generic issues which are the result of mapping dynamic object labels to dynamic metric dimensions, instead of fixing the dimensions and allowing the series cardinality to be dynamic per object.

1. It is not possible to query for objects with labels that exist but have empty values, as prometheus treats these the same as metrics without the labels.
2. It is not possible to use prometheus' value matching and parsing functionality on object label keys since they are munged into prometheus label keys.

Reiterating my previous comment, this is an alternative solution to the underlying problems without tying kube-state-metrics to any particular label conventions:

```
# BEFORE
kube_node_labels{label_foo="bar",label_node_role_kubernetes_io_master="",node="127.0.0.1"} 1

# AFTER
kube_node_label{label="foo",value="bar",node="127.0.0.1"} 1
kube_node_label{label="node-role.kubernetes.io/master",value="",node="127.0.0.1"} 1
```

One downside would be that queries on multiple labels would become more complicated joins:
```
# BEFORE
kube_node_labels{label_foo="bar",label_baz="qux"}

# AFTER
kube_node_label{foo="bar"} and on(node) kube_foo_label{baz="qux"}
```

derekwaynecarr:
Related discussion about appropriate use of node role labels:
https://github.com/kubernetes/enhancements/pull/1144

derekwaynecarr:
The tldr is that the node role label is not going away, but its an implementation detail of the cluster topology and not something that kubernetes components should imply behavior around in favor of more intent based labels (exclude-disruption, etc.).  That said, the role of the node should be stable for the life of the node.

docs/node-metrics.md
```diff
@@ -4,6 +4,7 @@
 | ---------- | ----------- | ----------- | ----------- |
 | kube_node_info | Gauge | `node`=&lt;node-address&gt; <br> `kernel_version`=&lt;kernel-version&gt; <br> `os_image`=&lt;os-image-name&gt; <br> `container_runtime_version`=&lt;container-runtime-and-version-combination&gt; <br> `kubelet_version`=&lt;kubelet-version&gt; <br> `kubeproxy_version`=&lt;kubeproxy-version&gt; <br> `provider_id`=&lt;provider-id&gt; | STABLE |
 | kube_node_labels | Gauge | `node`=&lt;node-address&gt; <br> `label_NODE_LABEL`=&lt;NODE_LABEL&gt;  | STABLE |
+| kube_node_role | Gauge | `node`=&lt;node-address&gt; <br> `role`=&lt;NODE_ROLE&gt;  | STABLE |
```
LiliC:
```suggestion
| kube_node_role | Gauge | `node`=&lt;node-address&gt; <br> `role`=&lt;NODE_ROLE&gt;  | EXPERIMENTAL |
```

brancz:
Got it, thanks @derekwaynecarr. I think we're good with moving forward with this. The only thing I would like to do as this is a new metric and in general a new method for kube-state-metrics, mark this as experimental for now. Then this lgtm to merge. Can you mark this metric as experimental in the docs @abursavich ? :slightly_smiling_face:

internal/store/node_test.go
```diff
@@ -68,7 +68,7 @@ func TestNodeStore(t *testing.T) {
 					Name:              "127.0.0.1",
 					CreationTimestamp: metav1.Time{Time: time.Unix(1500000000, 0)},
 					Labels: map[string]string{
-						"type": "master",
+						"node-role.kubernetes.io/master": "true",
```
LiliC: This is never true, I think this is always empty.

docs/node-metrics.md
abursavich: Done.

internal/store/node_test.go
abursavich: I've seen in both ways (e.g. in OpenShift the convention is to use "true"), but empty is a better test :)

PR #881, opened 2019-09-10
compile kube-state-metrics with go1.13

tariq1890:


tariq1890:
/hold

Unhold when #882 is merged.

LiliC:
@tariq the other PR is merged, we need to rebase and retest now.

tariq1890:
@LiliC Looks like #882 needs to be backported to the `release-1.8` branch to work :(.

Given that #882 is a harmless change and necessary, I'd vote for the backport. That way, we can unblock this PR.

LiliC:
@tariq1890 what’s the reason it needs to be backported? Not objecting at all, just curious.

tariq1890:
So the benchmark tests are failing because of an unsuccessful branch switch. This is happening because the `go 1.13` runtime auto-adds the go version to the `go.mod` manifest file if not present. The only way to get around this inconvenience is to add the `go version` pin to the `release-1.8` branch which does not have it yet. This way `go 1.13` won't introduce the unexpected change and cause the branch checkout to fail while the benchmark tests are running.

LiliC:
@tariq1890  Makes sense, (what go is doing doesn't :D) but makes sense to backport. Go ahead!

tariq1890:
@LiliC Ready for review :)

PR #889, opened 2019-09-19
deployment status conditions

### LINKED ISSUE #886: deployment condition metrics

/kind feature

**What happened**:

A deployment rollout stalled, no metrics from Kube State Metrics made this condition clear.

KSM to have deployment status conditions available, similar to node status conditions.

```
kube_deployment_status_condition{deployment="example", namespace="default", condition="ReplicaFailure", status="true"} 1
```

Look at the metrics produced by KSM and see no deployment status conditions are created.

**Environment**:
- Kube-state-metrics image version: v1.6.0

terinjokes:

Deployments, like Nodes, have status conditions observing the
current state. While the state of Available and Progressing conditions
can likely be inferred by other metrics, the state of ReplicaFailure can
not be inferred.

This changelist adds a new metric `kube_deployment_status_condition`
that observes all the conditions on a deployment for each condition
status. This is analogous to the status conditions observed by nodes and
horizontal pod autoscalers, and allows kube-state-metrics to observe
status conditions added by third-parties.

As an example, for a deployment that has stalled, the following new
metrics observed would allow an operator to detect the condition:

    kube_deployment_status_condition{deployment="example", namespace="default", condition="ReplicaFailure", status="true"} 1
    kube_deployment_status_condition{deployment="example", namespace="default", condition="ReplicaFailure", status="false"} 0
    kube_deployment_status_condition{deployment="example", namespace="default", condition="ReplicaFailure", status="unknown"} 0

Fixes #886

### DIFF
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -123,6 +123,30 @@ var (
 			}),
 		},
 		{
+			Name: "kube_deployment_status_condition",
+			Type: metric.Gauge,
+			Help: "The current status conditions of a deployment.",
+			GenerateFunc: wrapDeploymentFunc(func(d *v1.Deployment) *metric.Family {
+				ms := make([]*metric.Metric, len(d.Status.Conditions)*len(conditionStatuses))
+
+				for i, c := range d.Status.Conditions {
+					conditionMetrics := addConditionMetrics(c.Status)
+
+					for j, m := range conditionMetrics {
+						metric := m
+
+						metric.LabelKeys = []string{"condition", "status"}
+						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						ms[i*len(conditionStatuses)+j] = metric
+					}
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
 			Name: "kube_deployment_spec_replicas",
 			Type: metric.Gauge,
 			Help: "Number of desired pods for a deployment.",
--- a/docs/deployment-metrics.md
+++ b/docs/deployment-metrics.md
@@ -7,6 +7,7 @@
 | kube_deployment_status_replicas_unavailable | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_status_replicas_updated | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_status_observed_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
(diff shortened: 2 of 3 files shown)


internal/store/deployment.go
```diff
@@ -122,6 +122,35 @@ var (
...
+		// Third party plugin may report customized condition for deployments,
+		// and Kubernetes may add new apps conditions in future.
+		{
+			Name: "kube_deployment_status_condition",
+			Type: metric.Gauge,
+			Help: "The condition of a deployment.",
+			GenerateFunc: wrapDeploymentFunc(func(d *v1.Deployment) *metric.Family {
+				ms := make([]*metric.Metric, len(d.Status.Conditions)*len(conditionStatuses))
+
+				// Collect node conditions and while default to false.
```
tariq1890: Hah, did I just spot a copypasta (Man, am I guilty of this!) here :) ? Let's have a more descriptive and accurate comment OR not have a comment at all. Your call.

(I also agree that the comment in `internal/store/node.go` needs to be better).

internal/store/deployment.go
```diff
@@ -122,6 +122,35 @@ var (
 				}
 			}),
 		},
+		// This all-in-one metric family contains all conditions for extensibility.
+		// Third party plugin may report customized condition for deployments,
+		// and Kubernetes may add new apps conditions in future.
+		{
+			Name: "kube_deployment_status_condition",
+			Type: metric.Gauge,
+			Help: "The condition of a deployment.",
```
tariq1890:
```suggestion
			Help: "Describes the current state of a deployment.",
```

The equivalent node metric's description isn't that great unfortunately :(. Something that needs to be fixed.

internal/store/deployment.go
```diff
@@ -122,6 +122,35 @@ var (
 				}
 			}),
 		},
+		// This all-in-one metric family contains all conditions for extensibility.
```
tariq1890: Do we need these comments?

internal/store/deployment.go
terinjokes: what, no, i wouldn't copy and paste and forget to update a comment! :sob: 

hpa has no comments, and I don't think this one is accurate anyways. I'll remove

internal/store/deployment.go
terinjokes: I attempted to come to a middle ground based on the other help texts for deployments.

internal/store/deployment.go
terinjokes: Removed

internal/store/deployment.go
```diff
@@ -122,6 +122,30 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_deployment_status_condition",
+			Type: metric.Gauge,
+			Help: "The current status conditions of a deployment.",
```
LiliC: Only nit I would say we remove the `.` at the end. :)

terinjokes: All the other help texts have periods.

PR #890, opened 2019-09-19
correct hpa condition status

terinjokes:
:rotating_light: This may be considered a breaking change. :rotating_light: 

The status and condition labels for horizontal pod autoscalers status
condition metrics are reversed. This CL corrects the order.

### DIFF
--- a/internal/store/hpa.go
+++ b/internal/store/hpa.go
@@ -133,7 +133,7 @@ var (
 					for j, m := range metrics {
 						metric := m
 						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append(metric.LabelValues, string(c.Type))
+						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
 						ms[i*len(conditionStatuses)+j] = metric
 					}
 				}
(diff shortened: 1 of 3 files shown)


LiliC:
Seems like our docs are outdated for this label, do you mind updating this file as well https://github.com/kubernetes/kube-state-metrics/blob/master/docs/horizontalpodautoscaler-metrics.md 

Thanks!

internal/store/hpa.go
```diff
@@ -132,7 +132,7 @@ var (
 
 					for j, m := range metrics {
 						metric := m
-						metric.LabelKeys = []string{"condition", "status"}
```
brancz: labels are intended to be lexicographic-ally sorted, let's instead change the order these are passed in

brancz:
Bigger question, would we want to do a patch release or wait for the next minor release?

LiliC:
@brancz depends when this was introduced, do we want to backport this bug fix to previous releases as well then? If not, I would do this in next minor release then.

brancz:
We've historically never backported fixes to anything but the latest minor release series. I'm also for not doing a patch release for this, as it's somewhat breaking and people are a little more careful with minor releases than patch releases.

terinjokes:
The original commit is linked to in my commit message (I'm on mobile, otherwise I would link here) too, but this was introduced in early 2018.

internal/store/hpa.go
terinjokes: Sorted.

terinjokes:
@LiliC I updated the row, but broke the table dimensions. I can repair it, if we don't mind all the additional whitespace.

tariq1890:
Thanks so much for this fix :)

@terinjokes The rendered output seems to be fine to me, so I think it's not required. Is that not the case for you?

terinjokes:
@tariq1890 It is, was just mentioning that the markdown wasn't in case it mattered.

tariq1890:
Not a hard requirement.

tariq1890:
I agree with @brancz on this. 
> in my opinion this is a bug fix, and does need a clear changelog statement

Let's have a more descriptive commit message :).

@brancz @LiliC Since v1.8.0 is still in its RC stages, we could incorporate this into the `release-1.8` branch.

terinjokes:
@taragu I'm able to expand the commit message. What do you have in mind?

tariq1890:
Can you amend the commit message? Something along the lines of mapping the label keys and values correctly in the hpa condition status metric.

tariq1890:
Thanks @terinjokes !

/lgtm

Issue #892, opened 2019-09-20
Clarify and document stability guarantees for new metrics 


lilic:
Currently, all new metrics get the STABLE status, we tend to only mark metrics are not stable when the resource is not GA, from what I can tell.

I would suggest to instead start with all new metrics in a not stable version, not sure experimental is the right word for that. This gives us some time to catch any problems and change any labels or remove metrics. After some period of time, we can graduate those metrics into a stable status.

This goes also in line with what the SIG instrumentation is doing with kubernetes metrics stability.

cc @tariq1890 @brancz thoughts?

brancz:
> This goes also in line with what the SIG instrumentation is doing with kubernetes metrics stability.

I think this is the key here. We should make sure that we align the messaging, wording, etc. with the metrics stability framework. There the wording is rather just "alpha" and "stable". There are no set rules for what goes in what category but we have a much better idea in kube-state-metrics as the values rely on APIs that already have maturity levels. I think roughly that means metrics in kube-state-metrics must be aligned with the maturity level of the API it relies on.

tariq1890:
This was what I had in mind as well. So I agree with @brancz

> I think roughly that means metrics in kube-state-metrics must be aligned with the maturity level of the API it relies on.

brancz:
Sorry just for clarification, what I meant is:

* Metrics that are based on non-GA API objects cannot be in stable state
* Metrics that _are_ based GA API object _could_ be introduced as non-stable state, because we have gotten naming/labeling wrong a number of times even if the API object was stable (thinking Pod and Node API for example)

So basically only if based on a GA API _and_ gone through some vetting, can a metric become stable.

tariq1890:
What would vetting mean in this context? Do you mean letting a metric bake over a couple of releases?

brancz:
Precisely.

tariq1890:
Sounds good to me :). Thanks for starting this thread @LiliC .

LiliC:
> Do you mean letting a metric bake over a couple of releases?

Yes, I think that might help us prevent some wrong metrics, like labels, or just any high cardinality we can't catch etc. And makes it easier for us to remove metrics.


> Let's decide on specific timelines then. For e.g:- How to long wait until a metric graduate to STABLE state over release cycles etc.

Hmm, my gut says 2-3 minor releases?  But open to suggestions.

tariq1890:
[...]

PR #893, opened 2019-09-20
store: collector for NetworkPolicy objects

### LINKED ISSUE #891: [feature-request] NetworkPolicy metrics

> /kind feature

**What happened**:

Kubernetes has [NetworkPolicy](https://kubernetes.io/docs/concepts/services-networking/network-policies/) objects to control how pods communicate with each other. We'd find it very useful to have kube-state-metrics on these.

We probably wouldn't want to try and export the whole spec (with ingress/egress rules, ip specs, pod selectors etc.), at least not initially - but it'd be useful to export the following information as a starting point:
* `kube_networkpolicy_info` (with metadata)
* `kube_networkpolicy_rules` (count of ingress/egress rules, switched by label)

**Environment**:

Unless there are any existing efforts, I'd be happy to take this on :)

milesbxf:
Exports information about NetworkPolicy objects

Fixes #891


milesbxf:
Network policy metrics don't seem to be exposed in the e2e test rig. Anything obvious I've missed? Do I need to enable the collector somewhere?

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+			Name: "kube_networkpolicy_spec_ingress_rules",
+			Type: metric.Gauge,
+			Help: "Number of ingress rules on the networkpolicy",
+			GenerateFunc: wrapNetworkPolicyFunc(func(n *networkingv1.NetworkPolicy) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{},
+							LabelValues: []string{},
+							Value:       float64(len(n.Spec.Ingress)),
```
brancz: typically we don't do any aggregations in kube-state-metrics, why not expose a metric per rule?

LiliC: I agree this can be done afterwards very easily.

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+			Name: "kube_networkpolicy_spec_egress_rules",
+			Type: metric.Gauge,
+			Help: "Number of egress rules on the networkpolicy",
+			GenerateFunc: wrapNetworkPolicyFunc(func(n *networkingv1.NetworkPolicy) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{},
+							LabelValues: []string{},
+							Value:       float64(len(n.Spec.Egress)),
```
LiliC: Same as above, we should adjust this.

internal/store/networkpolicy_test.go
```diff
@@ -0,0 +1,59 @@
+package store
```
LiliC: Missing copyright header.

internal/store/networkpolicy.go
brancz: How do you mean afterwards?

LiliC: Sorry, meant -> I agree with you @brancz as if needed the aggregation can be done later, and should not be done here, but we should expose raw metrics.

docs/README.md
```diff
@@ -69,6 +69,7 @@ Per group of metrics there is one file for each metrics. See each file for speci
 - [Ingress Metrics](ingress-metrics.md)
 - [CertificateSigningRequest Metrics](certificatessigningrequest-metrics.md)
 - [VerticalPodAutoscaler Metrics](verticalpodautoscaler-metrics.md)
+- [NetworkPolicy Metrics](networkpolicy-metrics.md)
```
tariq1890: Something we haven't been enforcing well enough - lexicographical order. Let's place this above Node metrics.

docs/networkpolicy-metrics.md
```diff
@@ -0,0 +1,9 @@
+# Network Policy Metrics
+
+
+| Metric name                           | Metric type | Labels/tags                                 | Status       |
+| ------------------------------------- | ----------- | ------------------------------------------- | ------------ |
+| kube_networkpolicy_created            | Gauge       | `networkpolicy`=&lt;networkpolicy name&gt;  | EXPERIMENTAL |
```
tariq1890: These should be `STABLE` imo. They are all well-established fields in NetworkPolicy.

tariq1890:
@milesbxf Can you add `networkpolicies` to the list of default collectors [here](https://github.com/kubernetes/kube-state-metrics/blob/c9d5deb4fa2305a4bf6216f8a6165f71e59869a3/pkg/options/collector.go#L28)? That should fix the e2e failures.

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+)
+
+var (
+	descNetworkPolicyLabelsDefaultLabels = []string{"networkpolicy"}
+
+	networkpolicyMetricFamilies = []metric.FamilyGenerator{
+		{
+			Name: "kube_networkpolicy_created",
+			Type: metric.Gauge,
+			Help: "Unix creation timestamp",
```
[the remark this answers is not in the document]
tariq1890: No strong opinions from here. None of the other metrics follow this convention @LiliC .

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/kube-state-metrics/pkg/metric"
+)
+
+var (
+	descNetworkPolicyLabelsDefaultLabels = []string{"networkpolicy"}
```
tariq1890: NetworkPolicy is a namespaced kubernetes resource.
```suggestion
	descNetworkPolicyLabelsDefaultLabels = []string{"namespace", "networkpolicy"}
```

internal/store/networkpolicy_test.go
tariq1890: Good catch @LiliC! I am surprised we don't have a Copyright header checker added to our linting workflows.

internal/store/networkpolicy_test.go
```diff
@@ -0,0 +1,59 @@
...
+				"kube_networkpolicy_labels",
+				"kube_networkpolicy_spec_egress_rules",
+				"kube_networkpolicy_spec_ingress_rules",
+			},
+		},
+	}
+	for i, c := range cases {
+		c.Func = metric.ComposeMetricGenFuncs(networkpolicyMetricFamilies)
+		if err := c.run(); err != nil {
+			t.Errorf("unexpected collecting result in %vth run:\n%s", i, err)
```
tariq1890:
```suggestion
			t.Errorf("unexpected collecting result in %dth run:\n%v", i, err)
```

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+
+func wrapNetworkPolicyFunc(f func(*networkingv1.NetworkPolicy) *metric.Family) func(interface{}) *metric.Family {
+	return func(obj interface{}) *metric.Family {
+		networkPolicy := obj.(*networkingv1.NetworkPolicy)
+
+		metricFamily := f(networkPolicy)
+
+		for _, m := range metricFamily.Metrics {
+			m.LabelKeys = append(descNetworkPolicyLabelsDefaultLabels, m.LabelKeys...)
+			m.LabelValues = append([]string{networkPolicy.Name}, m.LabelValues...)
```
tariq1890:
```suggestion
			m.LabelValues = append([]string{networkPolicy.Namespace, networkPolicy.Name}, m.LabelValues...)
```

tariq1890:
Thanks very much for this PR @milesbxf :)

internal/store/networkpolicy_test.go
```diff
@@ -0,0 +1,59 @@
+package store
+
+import (
+	"testing"
+
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/kube-state-metrics/pkg/metric"
```
tariq1890: This should have been caught by our linters. First-party import are to be separated from third-party ones.
```suggestion

	"k8s.io/kube-state-metrics/pkg/metric"
```

internal/store/networkpolicy.go
```diff
@@ -0,0 +1,125 @@
...
+package store
+
+import (
+	networkingv1 "k8s.io/api/networking/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+	"k8s.io/kube-state-metrics/pkg/metric"
```
tariq1890: Wonder why the linters don't catch this. 🤔 
```suggestion

	"k8s.io/kube-state-metrics/pkg/metric"
```

docs/README.md
milesbxf: Would you like me to just sort the whole list alphabetically?

docs/networkpolicy-metrics.md
milesbxf: 👍 I wasn't sure what the convention was

internal/store/networkpolicy.go
milesbxf: I considered a raw metric per rule, but I was struggling to find a way to translate the `NetworkPolicy` rule spec into a sensible metric. For example, given this policy (from the docs at https://kubernetes.io/docs/concepts/services-networking/network-policies/): 
```
ingress:
  - from:
    - ipBlock:
        cidr: 172.17.0.0/16
        except:
        - 172.17.1.0/24
    - namespaceSelector:
        matchLabels:
          project: myproject
    - podSelector:
        matchLabels:
          role: frontend
    ports:
    - protocol: TCP
      port: 6379
  egress:
  - to:
    - ipBlock:
        cidr: 10.0.0.0/24
    ports:
    - protocol: TCP
      port: 5978
```

How would we sensibly represent this? The `from`/`to` blocks can have any number of sub-rules, with further nesting. I'm also not sure what the usefulness of this would be - it feels a bit like reimplementing the Kube API in metrics. Admittedly it'd be cool to be able to do things like count the number of times a particular label combination appears in NetworkPolicy rules, but I can't think how this could work 🤔 

Are there other resources we currently instrument in kube-state-metrics with complex nested structures like this?

docs/networkpolicy-metrics.md
LiliC: @tariq1890 See my open issue :D https://github.com/kubernetes/kube-state-metrics/issues/892

@milesbxf sorry about this, we are still discussing, this we need to also document it better! :)

internal/store/networkpolicy.go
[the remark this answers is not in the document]
LiliC: Yeah more like a nit for me :)

docs/README.md
tariq1890: That would be great actually @milesbxf :)

docs/networkpolicy-metrics.md
tariq1890: @LiliC Thanks for pointing that out. I think we can agree that these docs are fine as-is. So I'm resolving.

internal/store/networkpolicy.go
milesbxf: @tariq1890 @brancz @LiliC any thoughts on this? ☝️

brancz: I don't think there is actually any precedence with this complexity. I think for this case I'd be ok with exposing counts, and come back to it should there be additional use cases. Can we make sure to mark those metrics as experimental?

tariq1890:
@milesbxf Just checking in on this PR. Would you be able to rebase? I think we can just expose the counts here as mentioned by @brancz [here](https://github.com/kubernetes/kube-state-metrics/pull/893#discussion_r329600395)

milesbxf:
> @milesbxf Just checking in on this PR. Would you be able to rebase? I think we can just expose the counts here as mentioned by @brancz [here](https://github.com/kubernetes/kube-state-metrics/pull/893#discussion_r329600395)

I've rebased 👍 I noticed we moved to libsonnet for generating example YAMLs, so I've updated that too. I'll work through some of the other comments, and let you know once I think it's ready to finally review.

It looks like #914 only updated one of the generated examples and not
the libsonnet config itself, so I've added `admissionregistration.k8s.io`
and regenerated - do you think it's worth adding a CI check to catch this by making sure that `make examples` has been correctly run?

tariq1890:
@milesbxf You might need to rebase again because of this PR #927 and add the tool dependencies in the new way. Sorry for the trouble :(

milesbxf:
> @milesbxf You might need to rebase again because of this PR #927 and add the tool dependencies in the new way. Sorry for the trouble :(

@tariq1890 done ✅

tariq1890:
you would need to add the `gojsontoyaml` dep to `tools.go` so that it is registered in  the `go.mod` manifest as well.

milesbxf:
> you would need to add the `gojsontoyaml` dep to `tools.go` so that it is registered in the `go.mod` manifest as well.

Ah of course. Done as well 👍

milesbxf:
I've also run `go mod vendor` which should fix the failing CI check 🙈

tariq1890:
@milesbxf I pulled your branch and I see that you haven't checked in the changes of `go mod vendor` completely.

There are some files in the `vendor` dir that would be deleted and changes in the `modules.txt`. Please rerun `go mod vendor` and make sure the resultant changes are git added.

milesbxf:
> @milesbxf I pulled your branch and I see that you haven't checked in the changes of `go mod vendor` completely.
> 
> There are some files in the `vendor` dir that would be deleted and changes in the `modules.txt`. Please rerun `go mod vendor` and make sure the resultant changes are git added.

Turns out I was running it with Go 1.12. I've rerun with 1.13 and pushed, let's hope this works 🙏

tariq1890:
Awesome! Last request - let’s squash the commits :)

tariq1890:
Oh no! It looks like merge conflicts need to be resolved :’(.

We will need to rebase once again.

milesbxf:
@tariq1890 only just saw this - thanks so much for seeing this through and getting it merged! ❤️

tariq1890:
You're most welcome! Thank you for your hard work with this :)

Issue #875, opened 2019-08-30
Rewrite the end-to-end tests to go


lilic:
Currently, we have all our end-to-end tests in bash, this makes it painstaking to add new tests, with go we can easily use the client_golang library to do checks.

We can use the [prometheus-operator test framework](https://github.com/coreos/prometheus-operator/tree/master/test/framework) as an example to create the framework.

Below are some test cases I came up with, these are the already existing ones we have plus additional ones, should be good for the first iteration:

1. Wait until all pods are running
2. Loop and wait until kube-state-metrics is running by accessing the healthz endpoint
3. Access kube-state-metrics metrics endpoint, so 8080 to get the metrics data
4. Make sure metrics is not 0
5. Lint the metrics using prometheus client_golang library
6. Check that all the collector metrics are available, we can take the constants from options pkg `DefaultCollectors`  
7. Deleting resources causes metric to be updated
8. Make sure kube-state-metrics own metrics and healtz are present, the go metrics etc.
9. Check the metrics error rate is not high. The `kube_state_metrics_list_total ` and `kube_state_metrics_watch_total` 

cc @tariq1890 As discussed sharing some ideas I wrote down a while ago.

brancz:
> We can use the prometheus-operator test framework as an example to create the framework.

I personally think what the prometheus operator does would be a bit of an overkill here. We just need to launch the kube-state-metrics pod and interact with it via a proxy, that won't need a framework like the prometheus operator does. We should just do it from scratch here I think.

Otherwise very happy to finally get rid of the bash tests!

YaoZengzeng:
@LiliC @brancz 

I'd like to help to fix this issue if needed :)

LiliC:
@YaoZengzeng Yes, this still needs addressing, I know @tariq1890 wanted to work on it, but didn't have the time, so maybe you can sync with him? :) Thanks!!

tariq1890:
I haven't started this yet, but always happy to have contributors who are onboard to help :).

LiliC:
@YaoZengzeng Let me know if something is not clear in the issue, feel free to reach out on slack as well. Also feel free to open an early WIP PR. Thanks again!

YaoZengzeng:
@LiliC I will, thanks! :)

YaoZengzeng:
After my research and combined with what @brancz said, I think we can still use the script to install minikube (as well as `kubectl proxy`) and launch kube-state-metrics. 

[...]

PR #896, opened 2019-09-21
framework for e2e test implemented in golang

YaoZengzeng:
Signed-off-by: YaoZengzeng <yaozengzeng@huawei.com>

To fixes issue #875 

The purpose of this PR is implement the framework of e2e test and only have one test case which check the endpoint `/healthz` to run through the framework and prove it works fine.

Fixes #


tests/e2e.sh
```diff
@@ -164,6 +164,11 @@ set -e
 
 echo "kube-state-metrics is up and running"
 
+echo "start e2e test for kube-state-metrics"
+KSMURL='http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state-metrics:http-metrics/proxy'
+go test -v ./tests/e2e/ --ksmurl=${KSMURL}
+
+# TODO: re-implement the following test cases with golang and remove them.
```
LiliC: Nit: 
```suggestion
# TODO: re-implement the following test cases in Go with the goal of removing this file.
```

tests/e2e/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+	}
+	u.Path = strings.TrimRight(u.Path, "/")
+
+	return &KSMClient{
+		endpoint: u,
+		client:   &http.Client{},
+	}, nil
+}
+
+func (k *KSMClient) Healthz() (bool, error) {
```
LiliC: I would call this maybe `isHelathz()` for lack of better ideas, also this is not used outside of this file so lets not expose it.

tests/e2e/main_test.go
```diff
@@ -0,0 +1,123 @@
...
+
+	resp, err := k.client.Do(req)
+	if err != nil || resp.StatusCode != 200 {
+		return false, err
+	}
+
+	return true, nil
+}
+
+func TestMain(m *testing.M) {
```
LiliC: Can we replace the order and have this as the first function in the main_test.go file.

YaoZengzeng:
@LiliC Updated.

tests/e2e/framework.go
```diff
@@ -0,0 +1,85 @@
...
+	}
+	u.Path = strings.TrimRight(u.Path, "/")
+
+	return &KSMClient{
+		endpoint: u,
+		client:   &http.Client{},
+	}, nil
+}
+
+func (k *KSMClient) isHealthz() (bool, error) {
```
LiliC: I would move this to the main_test.go as well. Thanks

YaoZengzeng: `KSMClient` is defined in `framework.go` , but its method `isHealthz()` defined in `main_test.go`. This looks a bit strange. I think define all setup related work in `main_test.go` would be more reasonable. WDYT?

LiliC: But the `isHealthz ` is used in  `TestIsHealthz ` only? Anyways its a nit really.

LiliC:
Ping @tariq1890 to have a look as well, but overall it looks good to me! Thanks again!

tests/e2e/framework.go
YaoZengzeng: We will define other methods of `KSMClient`, if only put `isHealthz` in `framework.go`. It's weird for me 😂 . Maybe it would be better if `xxx_test.go` only contained `TestXXX` functions.

Anyway, we can change the layout of code to make it more appropriate any time :)

tests/e2e/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"log"
+	"os"
+	"testing"
+)
+
+func TestMain(m *testing.M) {
+	ksmurl := flag.String(
+		"ksmurl",
+		"",
+		"url to access access the kube-state-metrics",
```
tariq1890:
```suggestion
		"url to access the kube-state-metrics service",
```

tests/e2e/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+
+	exitCode = m.Run()
+
+	os.Exit(exitCode)
+}
+
+func TestIsHealthz(t *testing.T) {
+	ok, err := framework.KsmClient.isHealthz()
+	if err != nil {
+		t.Fatal(err)
```
tariq1890: I would prefer something like this
```suggestion
		t.Fatalf("kube-state healthz check failed: %v", err)
```

YaoZengzeng:
@tariq1890 Updated.

In addition, @tariq1890 @LiliC do we have slack channel or something like that ? Maybe it would be more convenient to talk there :)

tariq1890:
https://kubernetes.slack.com/messages/CJJ529RUY

tests/e2e/main_test.go
```diff
@@ -0,0 +1,58 @@
...
+	)
+	flag.Parse()
+
+	var (
+		err      error
+		exitCode int
+	)
+
+	if framework, err = NewFramework(*ksmurl); err != nil {
+		log.Printf("failed to setup framework: %v\n", err)
```
tariq1890: Let's use log.Fatalf instead since it calls `os.Exit(1)` anyway.

YaoZengzeng: Make sense, updated!

YaoZengzeng:
Any new comments? @LiliC @tariq1890

LiliC:
LGTM, but letting @tariq1890 have a final look. Thanks again!

tariq1890:
Thanks @YaoZengZeng!

PR #354, opened 2018-01-20
increase log verbosity

andyxning:
Increase log verbosity to `V(4)` to decrease useless log in kube-state-metrics common use. This has been discussed in slack sig-instrumentation.

/cc @brancz @DirectXMan12

---
This change is [Reviewable](https://reviewable.kubernetes.io/reviews/kubernetes/kube-state-metrics/354)

PR #902, opened 2019-09-26
no log level set for kube-state-metrics

like-inspur:
pull #354 has merged and set log level to 4, so now there is no method to set log level


like-inspur:
try

like-inspur:
I signed it

like-inspur:
I signed it

like-inspur:
> Hello, thanks for you PR, but these docs are generated, so changing it here, won't make a difference. https://github.com/kubernetes/kube-state-metrics/blob/master/Makefile#L130-L133
> I don't know how to change it and also think there is no need to change Makefile, because it can apply to most scenarios, junt when I change cli argument doc, it becomes inappropriate 
> If I understand you want to modify the verbosity level?
I don't want to modify log level, just found that according to pull #354 , the log level is set to 4. So there is no method to change log level by config -v or --v, and this argument descripition should not be printd in cli-arguments.md file

PR #908, opened 2019-10-01
*: Generate all manifests

### LINKED ISSUE #903: Autogenerate manifests
With the introduction of sharding and autosharding (#613) we now have a number of manifests that are duplicated, and they are 95% the same. I'd like to have these manifests generated using jsonnet, which would not only make our lives easier here but also allow using the definitions by using [jsonnet-bundler](https://github.com/jsonnet-bundler/jsonnet-bundler). This will also make our lives easier in releases as we can have a canonical version bump and have it propagate through the repo.

Just to be clear, this is not actually proposing to change anything, the manifests generated will be identical to the ones we have and kustomize will be possible just as it is today.

@LiliC @tariq1890

brancz:

Sharding and autosharding introduced a number of manifests that are
very similar, but not identical to the standard manifests. This
introduces generating all manifests using jsonnet and moves them to
the explicit `/examples` directory.

Fixes #903


README.md
```diff
@@ -214,7 +214,7 @@ service account token that has read-only access to the Kubernetes cluster.
 
 #### Kubernetes Deployment
 
-To deploy this project, you can simply run `kubectl apply -f kubernetes` and a
+To deploy this project, you can simply run `kubectl apply -f kubernetes/standard` and a
```
LiliC: standard? Hmm we can come up with a better name I think here.

kubernetes/autosharding/service.yaml
```diff
@@ -1,21 +1,19 @@
 apiVersion: v1
 kind: Service
 metadata:
+  labels:
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: v1.8.0-rc.0
```
LiliC: Should not all be `v1.8.0-rc.1`?

kubernetes/autosharding/statefulset.yaml
```diff
@@ -1,53 +1,58 @@
 apiVersion: apps/v1
 kind: StatefulSet
 metadata:
+  labels:
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: v1.8.0-rc.0
```
LiliC: v1.8.0-rc.1?

scripts/jsonnetfile.json
```diff
@@ -0,0 +1,13 @@
+{
+    "dependencies": [
+        {
+            "name": "kube-state-metrics",
+            "source": {
+                "local": {
+                    "directory": "../jsonnet/kube-state-metrics"
```
LiliC: Where is this directory? 🤔

scripts/jsonnetfile.json
```diff
@@ -0,0 +1,13 @@
...
+                }
+            },
+            "version": ""
```
LiliC: Should we pin this to master?

README.md
brancz: Any suggestions? I haven't been able to come up with anything better, and this is the standard way to deploy kube-state-metrics.

kubernetes/autosharding/service.yaml
brancz: It's whatever is in the `VERSION` file

scripts/jsonnetfile.json
brancz: `/jsonnet` is in the root of the repo

scripts/jsonnetfile.json
brancz: it's a directory based package so it will always do whatever is currently in the directory

kubernetes/autosharding/service.yaml
LiliC: Not true, in the file above its v1.8.0-rc.1.

README.md
LiliC: We said we want to also since we moving it, rename the directory to `examples`.

brancz:
Added a note to the release doc, and I believe I addressed all comments, if you have a suggestion for the name of the "standard" deployment, then I'm happy to adapt it.

kustomization.yaml
```diff
@@ -1,11 +1,8 @@
...
-
 namespace: kube-system
 
 resources:
-- kubernetes/kube-state-metrics-cluster-role-binding.yaml
-- kubernetes/kube-state-metrics-cluster-role.yaml
-- kubernetes/kube-state-metrics-deployment.yaml
-- kubernetes/kube-state-metrics-service-account.yaml
-- kubernetes/kube-state-metrics-service.yaml
+- kubernetes/cluster-role-binding.yaml
```
LiliC: Should this be examples?

tests/e2e.sh
```diff
@@ -117,20 +117,20 @@ KUBE_STATE_METRICS_IMAGE_TAG=$(docker images -a|grep 'quay.io/coreos/kube-state-
...
-sed -i.bak "s|${KUBE_STATE_METRICS_IMAGE_NAME}:v.*|${KUBE_STATE_METRICS_IMAGE_NAME}:${KUBE_STATE_METRICS_IMAGE_TAG}|g" ./kubernetes/kube-state-metrics-deployment.yaml
-cat ./kubernetes/kube-state-metrics-deployment.yaml
+sed -i.bak "s|${KUBE_STATE_METRICS_IMAGE_NAME}:v.*|${KUBE_STATE_METRICS_IMAGE_NAME}:${KUBE_STATE_METRICS_IMAGE_TAG}|g" ./kubernetes/standard/deployment.yaml
+cat ./kubernetes/standard/kube-state-metrics-deployment.yaml
 
 trap finish EXIT
 
 # set up kube-state-metrics manifests
-kubectl create -f ./kubernetes/kube-state-metrics-service-account.yaml
+kubectl create -f ./kubernetes/standard/service-account.yaml
```
LiliC: This should be examples now here as well.

kustomization.yaml
brancz: good catch, forgot about this one

PR #915, opened 2019-10-01
Update cli-arguments.md

gtorre:
Probably unnecessary in the first place since it requires human updates after each release, but figured it would be best to update it to the current version.


brancz:
This doc is auto generated, it should always be correct, are you seeing something different?

stigok:
@brancz Yes, the cli-arguments.md is not updating. At least `--resources` has been removed from the CLI, while `--collectors` is not documented.

PR #927, opened 2019-10-02
Add tools as go modules

### LINKED ISSUE #918: Import tool dependencies in go modules instead of go get
As per the [best practices](https://github.com/golang/go/wiki/Modules#how-can-i-track-tool-dependencies-for-a-module) recommended by the Go team, it is better to add tool dependencies like embedmd, benchcmp etc as module dependencies instead of `go get` as done over [here](https://github.com/kubernetes/kube-state-metrics/blob/master/Makefile#L136).

You can follow the steps mentioned here: https://github.com/kubernetes/kube-state-metrics/blob/master/Makefile#L136

Jukie:
Follows go best practices. Imports tool dependencies as go modules instead of with "go get" 

Fixes #918


brancz:
I have seen this practice lately, but I'm unsure how I feel about it. What benefits does this actually have?

Jukie:
One thing I see is that it pins module versions between different machines so it removes another avenue of the “works on my machine” issue.

Jukie:
I’m still new at this but it might also help keep your build cleaner according to this - https://stackoverflow.com/questions/56842385/using-go-get-to-download-binaries-without-adding-them-to-go-mod/57313319#57313319

brancz:
hmm .. especially with #903, we just introduced two more rather large tools, I'm not sure that's something that I always want to have checked into the repo. Wdyt @tariq1890 @LiliC ?

Jukie:
It's a separate package so it wouldn't be included in the builds if that's one of the concerns, but if you're checking out the repo is it not also expected that you'd want to include the build tools? I'm trying to understand what the negatives would be, doesn't the current local workflow include a `go get X` (or via the Makefile)?

brancz:
We check in the vendor directory which is what concerns me in terms of checking in the dependency code as well blowing up the repo size.

Jukie:
This is highly debated topic (https://github.com/golang/go/issues/25922) but where else should those dependencies be stored?
I imagine a user would be running `go get` on the build tools after checkout anyways, is that correct?

tariq1890:
@brancz I am in favour of tracking tool dependencies this way since we also have better control over which version we'd like to import. Since it's a recommended practice by the Go team, we can follow this.

We can also look into removing the `vendor` directory altogether. (similar to what `prometheus` does).

brancz:
Ok. Not too strong of an opinion, so I'm ok with it. This does need a rebase and add the new tools added in #903 .

Jukie:
@brancz This should be ready now, could you please take a look?

Makefile
```diff
@@ -152,15 +152,15 @@ scripts/vendor: jb scripts/jsonnetfile.json scripts/jsonnetfile.lock.json
 	cd scripts && jb install
 
 jsonnet:
-	GO111MODULE=off go get github.com/google/go-jsonnet/cmd/jsonnet
+	GO111MODULE=on go install github.com/google/go-jsonnet/cmd/jsonnet
```
tariq1890: Since we are using go 1.13, we don't need to set the `GO111MODULE` env var.
```suggestion
	go install github.com/google/go-jsonnet/cmd/jsonnet
```

Jukie:
I'm not yet sure why this is segfaulting
```
$ make validate-manifests
go install github.com/jsonnet-bundler/jsonnet-bundler/cmd/jb
cd scripts && jb install
panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x18 pc=0x540b42]

goroutine 1 [running]:
github.com/jsonnet-bundler/jsonnet-bundler/pkg.Install(0x6c0180, 0xc000086008, 0x101, 0xc0000ba2a0, 0x15, 0xc00007e0e0, 0x2, 0x4, 0x672a2c, 0x6, ...)
	/go/pkg/mod/github.com/jsonnet-bundler/jsonnet-bundler@v0.1.0/pkg/packages.go:58 +0x352
main.installCommand(0x672a2c, 0x6, 0x0, 0x0, 0x0, 0x7)
	/go/pkg/mod/github.com/jsonnet-bundler/jsonnet-bundler@v0.1.0/cmd/jb/main.go:295 +0x9ff
main.Main(0xc000072058)
	/go/pkg/mod/github.com/jsonnet-bundler/jsonnet-bundler@v0.1.0/cmd/jb/main.go:93 +0x712
main.main()
	/go/pkg/mod/github.com/jsonnet-bundler/jsonnet-bundler@v0.1.0/cmd/jb/main.go:61 +0x22
make: *** [Makefile:152: scripts/vendor] Error 2
```

Makefile
Jukie: Alright I'll remove them

brancz:
The `jb` version must be master, not v0.1.0.

Jukie:
Alright it's actually ready this time, sorry about that.

tariq1890:
Thank you for your PR @jukie. Just one review comments needs addressed. Once that is done, please squash the commits.

Jukie:
Which review comment is that? I applied your suggestion of removing `GO111MODULE=on` from the `go install` lines in the Makefile and that appears to be the only one on this PR. Was there something else I need to change?

tariq1890:
The copyright header in tools.go has the wrong year

Issue #899, opened 2019-09-25
Enabling storageclasses causing container crash


liangrog:
**Is this a BUG REPORT or FEATURE REQUEST?**:

 /kind bug

**What happened**:
Metrics container crashed at start caused by enabling "storageclasses". However sometimes it succeeded. 

kube-metrics-kube-state-metrics-679cb65777-rv47l   0/1     CrashLoopBackOff   6          8m28s

I0925 01:05:16.248532       1 main.go:90] Using collectors configmaps,cronjobs,daemonsets,deployments,endpoints,ingresses,jobs,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses
I0925 01:05:16.248634       1 main.go:99] Using all namespace
I0925 01:05:16.248649       1 main.go:140] metric white-blacklisting: blacklisting the following items:
W0925 01:05:16.248683       1 client_config.go:541] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0925 01:05:16.251069       1 main.go:185] Testing communication with server
I0925 01:05:16.258568       1 main.go:190] Running with Kubernetes cluster version: v1.9. git version: v1.9.3. git tree state: clean. commit: d2835416544f298c919e2ead3be3d0864b52323b. platform: linux/amd64
I0925 01:05:16.258631       1 main.go:192] Communication with server successful
I0925 01:05:16.258925       1 main.go:201] Starting kube-state-metrics self metrics server: 0.0.0.0:8081
I0925 01:05:16.259273       1 builder.go:126] Active collectors: configmaps,cronjobs,daemonsets,deployments,endpoints,ingresses,jobs,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses
I0925 01:05:16.259296       1 main.go:226] Starting metrics server: 0.0.0.0:8080
E0925 01:05:16.266880       1 runtime.go:73] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 74 [running]:
k8s.io/kube-state-metrics/vendor/k8s.io/apimachinery/pkg/util/runtime.logPanic(0x129ff40, 0x21c2590)
[...]

PR #929, opened 2019-10-03
Add check for ReclaimPolicy, VolumeBindingMode storageclass metrics.

clickyotomy:
**What this PR does / why we need it**: Checks for `nil` `*StorageClass.{ReclaimPolicy, VolumeBindingMode}` fields and add default values for them. This should prevent segmentation faults from occurring.

**Which issue(s) this PR fixes**: Fixes https://github.com/kubernetes/kube-state-metrics/issues/899.


internal/store/storageclass.go
```diff
@@ -35,8 +36,20 @@ var (
 			Type: metric.Gauge,
 			Help: "Information about storageclass.",
 			GenerateFunc: wrapStorageClassFunc(func(s *storagev1.StorageClass) *metric.Family {
+
+				// Add default values if missing.
+				if s.ReclaimPolicy == nil {
+					s.ReclaimPolicy = new(v1.PersistentVolumeReclaimPolicy)
```
tariq1890: Let's define a variable named `defaultReclaimPolicy` which will hold this value. You can then assign `s.ReclaimPolicy` the pointer to the `defaultReclaimPolicy` variable.

And let's do the same for `VolumeBindingMode`

tariq1890: We won't need to use `new` either.

tariq1890:
Thanks for your PR @clickyotomy! I've provided my comments. Please make sure to squash your commits as well.

clickyotomy:
Thanks, @tariq1890! I've made the changes you suggested.

internal/store/storageclass.go
```diff
@@ -28,15 +29,27 @@ var (
...
+					s.ReclaimPolicy = &defaultReclaimPolicy
+				}
+
+				if s.VolumeBindingMode == nil {
+					s.VolumeBindingMode = &defaultVolumeBindingMode
+				}
+
 				m := metric.Metric{
-					LabelKeys:   []string{"provisioner", "reclaimPolicy", "volumeBindingMode"},
+					LabelKeys:   []string{"provisioner", "reclaim_policy", "volume_binding_mode"},
```
tariq1890: We shouldn't be changing Labelkeys. This becomes a breaking change. Can we revert this please?

clickyotomy: I changed it because `make e2e` failed locally for me. `promtool` suggested that label names should be in snake case.

```
--- cut ---
check metrics format with promtool
kube_storageclass_info label names should be written in 'snake_case' not 'camelCase'
kube_storageclass_info label names should be written in 'snake_case' not 'camelCase'
--- cut ---
make: *** [e2e] Error 3
```

tariq1890: Looks like this new rule was added in the more recent versions of prometheus. This is something we  definitely have to look into. Can you revert this change for now? It will pass in the travis e2e since we use an older version of prometheus that doesn't include this check.

PR #934, opened 2019-10-04
Enable unconvert linter

bvwells:
**What this PR does / why we need it**:
This PR enables the unconvert linter and fixes various issues across the code base where type conversion is unnecessary. The PR aims to improve code readability.

Hope this is useful!

PR #935, opened 2019-10-06
Remove unnecessary type conversion

bvwells:
This PR is a follow up to PR #934 to remove some more unnecessary type conversions. The aim is to improve readability.


bvwells:
Thanks @tariq1890!

PR #936, opened 2019-10-06
Pre-allocation of slices

bvwells:
This PR cleans up collector/metric set code to make sure slices are pre-allocated when conversion from maps occurs. It also cleans up the IsEmpty functions to check size of underlying map rather than doing a conversion to a slice.

The PR also cleans up whitelist/blacklist code to make sure that slices are pre-allocated.

### DIFF
--- a/pkg/options/types.go
+++ b/pkg/options/types.go
@@ -48,16 +48,11 @@ func (ms *MetricSet) Set(value string) error {
 
 // asSlice returns the MetricSet in the form of plain string slice.
 func (ms MetricSet) asSlice() []string {
-	metrics := []string{}
+	metrics := make([]string, 0, len(ms))
 	for metric := range ms {
 		metrics = append(metrics, metric)
 	}
 	return metrics
-}
-
-// IsEmpty returns true if the length of the MetricSet is zero.
-func (ms MetricSet) IsEmpty() bool {
-	return len(ms.asSlice()) == 0
 }
 
 // Type returns a descriptive string about the MetricSet type.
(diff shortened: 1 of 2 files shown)


pkg/options/types.go
```diff
@@ -99,7 +99,7 @@ func (c CollectorSet) AsSlice() []string {
 
 // isEmpty() returns true if the length of the CollectorSet is zero.
 func (c CollectorSet) isEmpty() bool {
```
tariq1890: `isEmpty()` is unused. Let's remove it for now :).

pkg/options/types.go
```diff
@@ -57,7 +57,7 @@ func (ms MetricSet) asSlice() []string {
 
 // IsEmpty returns true if the length of the MetricSet is zero.
 func (ms MetricSet) IsEmpty() bool {
```
tariq1890: Let's remove this method as it's unused.

pkg/options/types.go
bvwells: Sounds good. I noticed this was unused, but wasn't sure whether to remove in case this was work in progress.

pkg/options/types.go
bvwells: I had a feeling that this was used. Let me double check...

pkg/options/types.go
bvwells: Removed!

pkg/options/types.go
bvwells: Removed!

bvwells:
Thanks for the review @tariq1890!

PR #939, opened 2019-10-08
e2e.sh: Update to prometheus 2.13.0, kubernetes 1.16.1

mrueg:
Updates versions for e2e tests.


mrueg:
Looks like minikube update caused some issues, I'll drop it from the update.

tariq1890:
@LiliC @mrueg We can't use the latest versions of prometheus because of the new `promlint` rules that have been added. Some of the label keys in `storageclass` violate those rules.

`kube_storageclass_info label names should be written in 'snake_case' not 'camelCase'`

If we were to address this, we would need to change the label keys and that would become a breaking change.

Can you revert the prometheus update? Also, let's use Kubernetes 1.16.x .

mrueg:
@tariq1890 happy to make the change to storageclass as well. As it's a breaking change, is there a plan to change to snake case for everything? This is how I would suggest to rename for now:

```
-                                       LabelKeys:   []string{"provisioner", "reclaimPolicy", "volumeBindingMode"},
+                                       LabelKeys:   []string{"provisioner", "reclaimpolicy", "volumebindingmode"},
```

tariq1890:
Renaming labelKeys would be a breaking change. We can only do it in the next major release.

brancz:
Added this to the breaking release umbrella issue.

mrueg:
Let me know if this PR is still needed, I can rebase.

PR #946, opened 2019-10-11
add VolumeAttachment collector

### LINKED ISSUE #872: Feature request: export VolumeAttachment metrics

/kind feature

**What happened**:

Kubernetes has a new resource type: [`VolumeAttachment`s](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.15/#volumeattachment-v1-storage-k8s-io). They provide helpful information on where a volume is attached and to alert on unexpected attachment status (for example, differences between information scraped from `node-exporter` and `kube-state-metrics`.

**Environment**:

JensErat:

Kubernetes has a new resource type: `VolumeAttachments`. They provide helpful information on where a volume is attached and to alert on unexpected attachment status (for example, differences between information scraped from node-exporter and kube-state-metrics).

The collector adds a bunch of new metrics. Each VolumeAttachment (ie., each CSI-attached volume) will have one of each, so we do not overly pollute the metrics space. Most metrics are rather unsurprising.

- `kube_volumeattachment_status_attachment_metadata`: provides a  label-like export of the attachment metadata map. Generalizing the   label-conversion function slightly helps at providing this metric.
- `kube_volumeattachment_created`: as VolumeAttachments are   automatically created and we already suffered from duplicate  `VolumeAttachments`, this can be invaluable for debugging
  misattachments.
- `kube_volumeattachment_spec_source_persistentvolume`: will only be  generated when the volume source is of `PersistentVolume` type. The  other type `inlineVolumeSpec` is still alpha-level and hard to map to metrics.

No end-to-end test manifest was added, as `VolumeAttachment`s are automatically generated when mounting volumes.

Since kube-state-metrics 1.8.0 wants to support Kubernetes 1.11 onwards, I went for storagev1beta1; the API is GA since Kubernetes 1.13 ( so still some time to go, now matter what version this gets released under).

**Important note**: I'm on vacation soon. My coworker @sbueringer will take over regarding any issues as soon as I'm gone.

Fixes #872 -- mostly follows the ideas discussed in this feature request; I had to deviate in some metrics to follow the other collectors and also properly take specifics of `VolumeAttachment`s into account.

### DIFF
--- a/docs/volumeattachment-metrics.md
+++ b/docs/volumeattachment-metrics.md
@@ -1,0 +1,10 @@
+# PersistentVolume Metrics
+
+| Metric name| Metric type | Labels/tags | Status |
+| ---------- | ----------- | ----------- | ----------- |
+| kube_volumeattachment_info | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; <br> `attacher`=&lt;attacher-name&gt; <br> `nodeName`=&lt;node-name&gt; | EXPERIMENTAL |
+| kube_volumeattachment_created | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; | EXPERIMENTAL |
+| kube_volumeattachment_labels | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; <br> `label_VOLUMEATTACHMENT_LABEL`=&lt;VOLUMEATTACHMENT_LABEL&gt;  | EXPERIMENTAL |
+| kube_volumeattachment_spec_source_persistentvolume | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; <br> `volumename`=&lt;persistentvolume-name&gt; | EXPERIMENTAL |
+| kube_volumeattachment_status_attached | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; | EXPERIMENTAL |
+| kube_volumeattachment_status_attachment_metadata | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; <br> `metadata_METADATA_KEY`=&lt;METADATA_VALUE&gt;  | EXPERIMENTAL |
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -34,6 +34,7 @@ import (
 	extensions "k8s.io/api/extensions/v1beta1"
 	policy "k8s.io/api/policy/v1beta1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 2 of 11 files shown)


docs/volumeattachment-metrics.md
```diff
@@ -0,0 +1,9 @@
+# PersistentVolume Metrics
+
+| Metric name| Metric type | Labels/tags | Status |
+| ---------- | ----------- | ----------- | ----------- |
+| kube_volumeattachment_info | Gauge | `volumeattachment`=&lt;volumeattachment-name&gt; <br> `attacher`=&lt;attacher-name&gt; <br> `nodeName`=&lt;node-name&gt; | STABLE |
```
tariq1890: These metrics should be EXPERIMENTAL

internal/store/builder.go
```diff
@@ -34,6 +34,7 @@ import (
 	extensions "k8s.io/api/extensions/v1beta1"
 	policy "k8s.io/api/policy/v1beta1"
 	storagev1 "k8s.io/api/storage/v1"
+	storagev1beta1 "k8s.io/api/storage/v1beta1"
```
tariq1890: I think we can use `storagev1` for VolumeAttachment. It has been promoted to v1 since 1.13 if I am not mistaken.

tariq1890: It should even support the watching and listing of VolumeAttachments created under `v1beta1`

JensErat: Yes, it graduated to GA in 1.13 (as also stated in the PR). I started with `storagev1` originally, but it failed in a Kubernetes 1.14 cluster -- kube-state-metrics got v1beta1 objects returned and paniced.

On the other hand, following the compatibility chart from the README we need to support Kubernetes 1.11? In this case, only `storagev1beta1` is available.

docs/volumeattachment-metrics.md
JensErat: I'll switch to EXPERIMENTAL; just to understand the semantics better: I thought it was about stability of the Kubernetes API, or is it also about "being new to kube-state-metrics"?

JensErat:
Fixed documentation (`STABLE` to `EXPERIMENTAL`, missing created metrics documentation), added manifest for e2e-tests and squashed.

As already mentioned, I'm not sure about going to storage/v1; it didn't work out for me against Kubernetes 1.14 and kube-state-metrics even wants to support Kubernetes 1.11 (or 1.12 with the next release) as I understand, and those generally only support storage/v1beta1 for `VolumeAttachment`s.

brancz:
lgtm

/hold cancel

PR #948, opened 2019-10-13
Add PostStartHookError to kube_pod_container_status_waiting_reason

### LINKED ISSUE #849: Add lifecycle failed reason  to pod waiting reason

/kind feature

I think we should add a reason for lifecycle hook when it fails,as below:
```
state:
    waiting:
    message: 'Exec lifecycle hook ([/bin/sh -c /data/poststart.sh])
        for Container "retail-stock-scm" in Pod "test-78f7dcbf75-47f28_daily(7940c259-af33-11e9-9622-525400df97ed)"
        failed - error: command '/bin/sh -c /data/preload.sh'
        exited with 124: , message: ""'
    reason: 'PostStartHookError: command '/bin/sh -c /data/poststart.sh'
        exited with 124: '
```

pigletfly:

Add PostStartHookError to kube_pod_container_status_waiting_reason

Fixes #849


internal/store/pod.go
```diff
@@ -1071,7 +1073,7 @@ func waitingReason(cs v1.ContainerStatus, reason string) bool {
 	if cs.State.Waiting == nil {
 		return false
 	}
-	return cs.State.Waiting.Reason == reason
+	return cs.State.Waiting.Reason == reason || strings.HasPrefix(cs.State.Waiting.Reason, postStartHookError)
```
LiliC: hmm can you explain why this change is needed as well?

pigletfly: because in `poststarthook` case,the reason varies, as described in https://github.com/kubernetes/kube-state-metrics/issues/849,so the waiting reason either equals to any of `{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff", "CreateContainerError", "InvalidImageName"}` or starts with `PostStartHookError`.

LiliC: I see now I think, so we want to just create a label `reason="PostStartHookError"` if any reason is prefixed with "PostStartHookError......", correct?

pigletfly: exactly

pigletfly:
kindly ping @LiliC

LiliC:
@pigletfly Sorry about this, conference season :) 

cc @brancz as you discussed this initially, does that look right to you?

.travis.yml
```diff
@@ -3,7 +3,7 @@ dist: xenial
 language: go
 
 go:
-    - "1.13.x"
+    - "1.13"
```
tariq1890: Let's drop this change. If we need to do this, we can do this in another PR.

internal/store/pod.go
```diff
@@ -31,10 +32,11 @@ import (
 )
 
 const nodeUnreachablePodReason = "NodeLost"
+const postStartHookError = "PostStartHookError"
 
 var (
 	descPodLabelsDefaultLabels = []string{"namespace", "pod"}
-	containerWaitingReasons    = []string{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff", "CreateContainerError", "InvalidImageName"}
+	containerWaitingReasons    = []string{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff", "CreateContainerError", "InvalidImageName", postStartHookError}
```
tariq1890: Let's not have a variable just for this reason string.
```suggestion
	containerWaitingReasons    = []string{"ContainerCreating", "CrashLoopBackOff", "CreateContainerConfigError", "ErrImagePull", "ImagePullBackOff", "CreateContainerError", "InvalidImageName", "PostStartHookError"}
```

internal/store/pod.go
tariq1890: I think we could use this instead https://golang.org/pkg/strings/#Contains

pigletfly:
PTAL @LiliC @tariq1890 @brancz

internal/store/pod.go
```diff
@@ -1071,7 +1072,7 @@ func waitingReason(cs v1.ContainerStatus, reason string) bool {
 	if cs.State.Waiting == nil {
 		return false
 	}
-	return cs.State.Waiting.Reason == reason
+	return cs.State.Waiting.Reason == reason || strings.Contains(cs.State.Waiting.Reason, "PostStartHookError")
```
tariq1890: @pigletfly I am not very comfortable with this logic since we're hardcoding `PostStartHookError` here. Thoughts @brancz and @LiliC ?

brancz: Yeah I'm not a fan of this either, it leaks into some weird hybrid of something like log analysis I feel. I don't think this belongs in kube state metrics unless it becomes a standardized API.

PR #950, opened 2019-10-15
[e2e] check all default collector metrics are available

YaoZengzeng:
Signed-off-by: YaoZengzeng <yaozengzeng@zju.edu.cn>

fixes part of #875 

Fixes #


YaoZengzeng:
@brancz @tariq1890 @LiliC PTAL.

YaoZengzeng:
Updated

YaoZengzeng:
@brancz @tariq1890 @LiliC Any new comments ?

tests/e2e.sh
```diff
@@ -31,8 +31,6 @@ SUDO=${SUDO:-}
 OS=$(uname -s | awk '{print tolower($0)}')
 OS=${OS:-linux}
 
-EXCLUDED_RESOURCE_REGEX="verticalpodautoscaler"
```
LiliC: Hmm wonder why we excluded this? cc @brancz

tariq1890: I believe it's because of verticalpodautoscaler compatibility issues with minikube. If we move to KIND, we might not have this anymore.

YaoZengzeng: It's not easy to create `verticalpodautoscaler` in test environment and we can't get the metrics of it. So it should be excluded :)

YaoZengzeng:
@brancz PTAL :)

tests/e2e/main_test.go
```diff
@@ -77,3 +82,49 @@ func TestLintMetrics(t *testing.T) {
...
+func TestDefaultCollectorMetricsAvailable(t *testing.T) {
+	buf := &bytes.Buffer{}
+
+	err := framework.KsmClient.metrics(buf)
+	if err != nil {
+		t.Fatalf("failed to get metrics from kube-state-metrics: %v", err)
+	}
+
+	resources := map[string]struct{}{}
+	files, err := ioutil.ReadDir("../../internal/store/")
```
LiliC: Maybe instead of this we could use this list? To avoid the reading of directories.

https://github.com/kubernetes/kube-state-metrics/blob/0bc19488277fb94adbfc9b6f64611ccd4cea1e78/pkg/options/collector.go#L28-L55

YaoZengzeng: Actually I tried at first :0

But it could not work. Because the resource name in this list is plural form and that in metric name is singular. Also the resource name is not exactly corresponding.

If we want to use this list directly, we have to change the command line config as well which is a break change:) So I still use the name of directories.

tests/e2e/main_test.go
```diff
@@ -77,3 +82,49 @@ func TestLintMetrics(t *testing.T) {
 		t.Fatalf("the problems encountered in Lint are: %v", problems)
 	}
 }
+
+func TestDefaultCollectorMetricsAvailable(t *testing.T) {
```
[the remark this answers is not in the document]
YaoZengzeng: Actually I don't have local environment.

But AFASK, if it works in CI, then it should work in local as well?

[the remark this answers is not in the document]
LiliC: Does not work for me locally, fails to find resources.

cc @tariq1890 do you mind running it locally as well, thanks!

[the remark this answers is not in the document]
YaoZengzeng: @LiliC The code has been updated to check the scan error.

Maybe it could give some clue why the test failed in local environment :)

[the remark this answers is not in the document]
olivierlemasle: @lilic When testing locally, rebased on my PR #1065 (because without it I'm not able to run e2e locally), these tests pass:

```
start e2e test for kube-state-metrics
=== RUN   TestIsHealthz
--- PASS: TestIsHealthz (0.00s)
=== RUN   TestLintMetrics
--- PASS: TestLintMetrics (0.01s)
=== RUN   TestDefaultCollectorMetricsAvailable
--- PASS: TestDefaultCollectorMetricsAvailable (0.01s)
PASS
ok  	k8s.io/kube-state-metrics/tests/e2e	0.031s
```
I can provide you with the full log if needed.

tests/e2e/main_test.go
tariq1890: I think the main Kubernetes repository maintains a list of the resource names along with its plurals. We could, perhaps, leverage that. WDYT @YaoZengzeng?

olivierlemasle: @tariq1890 I found this [ResourceSingularizer](https://github.com/kubernetes/apimachinery/blob/v0.17.3/pkg/api/meta/restmapper.go#L149) function, but it needs a `DefaultRESTMapper` that has already been initialized with all possible resource kinds, so I don't think this could be used. Did you have another list in mind?

olivierlemasle:
Hi @YaoZengzeng, would you mind rebasing your PR please?
As I'm responsible for the recent changes on e2e causing conflicts :sweat_smile:,  feel free to ask me if I can give you a hand.

lilic:
@YaoZengzeng do you want to pick this up, otherwise @olivierlemasle are you interested in taking over? No worries if neither are interested :)

olivierlemasle:
I'm ok to work on this!

lilic:
@olivierlemasle Thanks! 🎉 do you mind just cherry-picking commits with credit to @YaoZengzeng and just fixing and rebasing, thanks! :)

fpetkovski:
/remove-lifecycle stale

still relevant

PR #957, opened 2019-10-23
Add prometheus.io annotations to example services

vandrijevik:
The [`README` section `Kubernetes Deployment`](https://github.com/kubernetes/kube-state-metrics#kubernetes-deployment) states:

> The service already has a prometheus.io/scrape: 'true' annotation and if you added the recommended Prometheus service-endpoint scraping configuration, Prometheus will pick it up automatically and you can start using the generated metrics right away.

However, neither https://github.com/kubernetes/kube-state-metrics/blob/master/examples/autosharding/service.yaml nor https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/service.yaml have these annotations in place, which is confusing when debugging why Prometheus service discovery does not work right away, as advertised.

This PR adds the necessary annotations for Prometheus to be able to discover the service, and so one can start using the generated metrics right away.

### DIFF
--- a/examples/autosharding/service.yaml
+++ b/examples/autosharding/service.yaml
@@ -4,6 +4,9 @@ metadata:
   labels:
     app.kubernetes.io/name: kube-state-metrics
     app.kubernetes.io/version: v1.8.0
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "8080"
   name: kube-state-metrics
   namespace: kube-system
 spec:
--- a/examples/standard/service.yaml
+++ b/examples/standard/service.yaml
@@ -4,6 +4,9 @@ metadata:
   labels:
     app.kubernetes.io/name: kube-state-metrics
     app.kubernetes.io/version: v1.8.0
+  annotations:
+    prometheus.io/scrape: "true"
+    prometheus.io/port: "8080"
   name: kube-state-metrics
   namespace: kube-system
 spec:


vandrijevik:
Hello folks! I followed the instructions above to create a Linux Foundation profile and sign the CLA, but unfortunately the agreement form requires filling in a mailing address and telephone number, which I am not willing to provide to the Linux Foundation.

If this makes my documentation fix ineligible for merging, I understand and I regret that you require the above data in order to contribute. I hope that someone ends up fixing this in either case, because as-is, the README is wrong.

Thank you,
Vlad

tariq1890:
> but unfortunately the agreement form requires filling in a mailing address and telephone number, 
I have signed the CLA and I've never had to do this. Can you make sure that you've followed the right steps? 

https://github.com/kubernetes/community/blob/master/CLA.md

vandrijevik:
Hello @tariq1890,

Yes, as I mentioned I’ve followed exactly those steps, and the document that HelloSign presents for me to fill in and sign looks like this:

Screen Shot 2019-10-24 at 08 51 20

Without filling in the mailing address and telephone number fields, I am unable to sign the document, as mentioned above.

brancz:
kube state metrics is actually a classic example of where this annotation based approach fails entirely. How do we actually monitor kube state metrics itself on the other port? It’s impossible. How do we make sure the metrics endpoint is authenticated and authorized? Unless there is a common scheme across all scrape targets, also impossible.

This is precisely why we (as in the Prometheus team) have started to advise against the annotation based approach for a while now. :)

brancz:
A distinct scrape config just for kube-state-metrics, that simply selects all pods/endpoints directly.

vandrijevik:
Thanks, @brancz. Per the comment above, I’m closing this PR so that it doesn’t update the examples with a practice that the Prometheus team now recommends against. I hope the README gets updated not to include the claim that the examples contain Prometheus annotations.

PR #961, opened 2019-10-29
add hpa stats for current utilization and average value

ashleykasim:
When debugging issues such as https://github.com/kubernetes/kubernetes/issues/82195, it was necessary to export these stats in order to understand what the hpa sees at the time of the overflow and other such bugs. 

Fixes #

### DIFF
--- a/docs/horizontalpodautoscaler-metrics.md
+++ b/docs/horizontalpodautoscaler-metrics.md
@@ -2,10 +2,12 @@
 
 | Metric name                       | Metric type | Labels/tags                                                   | Status |
 | --------------------------------  | ----------- | ------------------------------------------------------------- | ------ |
+| kube_hpa_labels                   | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_metadata_generation      | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_spec_max_replicas        | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_spec_min_replicas        | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
-| kube_hpa_status_current_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
-| kube_hpa_status_desired_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_status_condition         | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; <br> `condition`=&lt;hpa-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; | STABLE |
-| kube_hpa_labels                   | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
+| kube_hpa_status_current_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
+| kube_hpa_status_currentmetrics_average_utilization | Gauge     | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | EXPERIMENTAL |
+| kube_hpa_status_currentmetrics_average_value | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | EXPERIMENTAL |
+| kube_hpa_status_desired_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
--- a/internal/store/hpa.go
+++ b/internal/store/hpa.go
@@ -20,6 +20,7 @@ import (
 	"k8s.io/kube-state-metrics/pkg/metric"
 
 	autoscaling "k8s.io/api/autoscaling/v2beta1"
+	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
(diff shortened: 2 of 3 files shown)


internal/store/hpa.go
```diff
@@ -132,8 +132,8 @@ var (
 
 					for j, m := range metrics {
 						metric := m
-						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						metric.LabelKeys = []string{"condition", "status", "reason"}
+						metric.LabelValues = append(metric.LabelValues, string(c.Type), c.Reason)
```
tariq1890: For controlling cardinality, It's better to have a known set of Reasons to iterate over. You can find a similar implementation in [pod.go](https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/pod.go#L37).

internal/store/hpa.go
```diff
@@ -143,6 +143,56 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_hpa_status_average_value",
```
tariq1890:
```suggestion
			Name: "kube_hpa_status_current_average_value",
```

internal/store/hpa.go
```diff
@@ -143,6 +143,56 @@ var (
...
+						Value: float64(value),
+					}
+				}
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
+			Name: "kube_hpa_status_average_utilization",
```
tariq1890:
```suggestion
			Name: "kube_hpa_statuscurrent_average_utilization",
```

docs/horizontalpodautoscaler-metrics.md
```diff
@@ -7,5 +7,7 @@
 | kube_hpa_spec_min_replicas        | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_status_current_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_status_desired_replicas  | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
-| kube_hpa_status_condition         | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; <br> `condition`=&lt;hpa-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; | STABLE |
+| kube_hpa_status_condition         | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; <br> `condition`=&lt;hpa-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; <br> `reason`=&lt;reason&gt; | STABLE |
```
tariq1890: Can you sort the metrics in these docs in lexical order? (I know that it's not necessarily in this PR's scope, but this change would be great as the PR is already touching this file).

tariq1890:
Thanks for your PR @ashleykasim. I've provided my comments. Will wait on @LiliC for her comment as well.

internal/store/hpa.go
```diff
@@ -143,6 +143,56 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_hpa_status_average_value",
+			Type: metric.Gauge,
+			Help: "Average metric value observed by the autoscaler.",
```
LiliC: Are these metrics not already exposed by k8s?

internal/store/hpa.go
```diff
@@ -143,6 +143,56 @@ var (
...
+			GenerateFunc: wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				ms := make([]*metric.Metric, len(a.Status.CurrentMetrics))
+				for i, c := range a.Status.CurrentMetrics {
+					var value int64
+					if c.Type == "Resource" {
```
LiliC: Nit: can we use the constants of `MetricSourceType` -> e.g. `ObjectMetricSourceType `, `PodsMetricSourceType `

internal/store/hpa.go
```diff
@@ -143,6 +143,56 @@ var (
...
+			Type: metric.Gauge,
+			Help: "Average metric utilization observed by the autoscaler.",
+			GenerateFunc: wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				ms := make([]*metric.Metric, len(a.Status.CurrentMetrics))
+				for i, c := range a.Status.CurrentMetrics {
+					if c.Type == "Resource" {
+						ms[i] = &metric.Metric{
+							Value: float64(*c.Resource.CurrentAverageUtilization),
+						}
+					} else {
```
LiliC: We don't need this else.

internal/store/hpa.go
ashleykasim: You can view them by querying metrics server or by viewing the hpa object. However, these metrics are not exported anywhere that is consumable by an observability stack.

docs/horizontalpodautoscaler-metrics.md
```diff
@@ -2,10 +2,12 @@
 
 | Metric name                       | Metric type | Labels/tags                                                   | Status |
 | --------------------------------  | ----------- | ------------------------------------------------------------- | ------ |
+| kube_hpa_labels                   | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_metadata_generation      | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_spec_max_replicas        | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
 | kube_hpa_spec_min_replicas        | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
+| kube_hpa_status_condition         | Gauge       | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; <br> `condition`=&lt;hpa-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; <br> `reason`=&lt;reason&gt; | STABLE |
+| kube_hpa_status_current_average_utilization | Gauge     | `hpa`=&lt;hpa-name&gt; <br> `namespace`=&lt;hpa-namespace&gt; | STABLE |
```
tariq1890: @ashleykasim The newly added metrics would need to be marked as `EXPERIMENTAL`.

internal/store/hpa.go
ashleykasim: I am going to remove the addition of `reason` and defer this to a separate PR since this is not related to utilization metrics anyway.

internal/store/hpa.go
```diff
@@ -143,6 +143,53 @@ var (
...
+		{
+			Name: "kube_hpa_status_current_average_value",
+			Type: metric.Gauge,
+			Help: "Average metric value observed by the autoscaler.",
+			GenerateFunc: wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				ms := make([]*metric.Metric, len(a.Status.CurrentMetrics))
+				for i, c := range a.Status.CurrentMetrics {
+					var value int64
+					if c.Type == autoscaling.ResourceMetricSourceType {
+						value, _ = c.Resource.CurrentAverageValue.AsInt64()
```
tariq1890: I'd like to change the approach here. Can we `switch-case` statements instead of the `if-else-if`? Also, let's grab the `*resource.Quantity` type first from the switch-case block, Call `.AsInt64()` on the fetched `*resource.Quantity`,  check the 2nd return param of `AsInt64()` and then add it to the metric family only if the 2nd return param is true.

internal/store/hpa.go
```diff
@@ -143,6 +143,53 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_hpa_status_current_average_value",
```
tariq1890: @ashleykasim I didn't take a closer look at your change before. I think this would be a better name `kube_hpa_status_currentmetrics_average_value`.

internal/store/hpa.go
```diff
@@ -143,6 +143,53 @@ var (
...
+						Value: float64(value),
+					}
+				}
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
+			Name: "kube_hpa_status_current_average_utilization",
```
tariq1890:
```suggestion
			Name: "kube_hpa_status_currentmetrics_average_utilization",
```

LiliC:
@ashleykasim is this ready for another review?

ashleykasim:
@LiliC Yes please!

internal/store/hpa.go
```diff
@@ -143,6 +144,57 @@ var (
...
+					case autoscaling.ObjectMetricSourceType:
+						value = c.Object.AverageValue
+					case autoscaling.ExternalMetricSourceType:
+						value = c.External.CurrentAverageValue
+					default:
+						// Skip unsupported metric type
+						continue
+					}
+					intVal, canFastConvert := value.AsInt64()
+					if canFastConvert {
```
tariq1890: It can be further condensed to this.
```suggestion
					if intVal, canFastConvert := value.AsInt64(); canFastConvert {
```

internal/store/hpa.go
```diff
@@ -143,6 +144,57 @@ var (
...
+						value = &c.Pods.CurrentAverageValue
+					case autoscaling.ObjectMetricSourceType:
+						value = c.Object.AverageValue
+					case autoscaling.ExternalMetricSourceType:
+						value = c.External.CurrentAverageValue
+					default:
+						// Skip unsupported metric type
+						continue
+					}
+					intVal, canFastConvert := value.AsInt64()
```
tariq1890: This line can be removed.

ashleykasim: Done,

LiliC:
Thanks for the contribution! 🎉

PR #963, opened 2019-10-30
expose the PodCIDR in node info

jacobstr:

It can be helpful to know the PodCIDR of nodes in order to predict IP
space utilization. In GKE for example, one typically uses a Secondary
Alias IP range that doles out PodCIDRs dynamically. When this is
exhausted, nodes will fail to boot with messages such as:

```
Instance 'gke-test-ingre-43fd70d6-9v77' creation failed: IP space of 'projects/network-project/regions/us-central1/subnetworks/pod-alias-b85f82b0bca36afa' is exhausted.
```

Although not comprehensive, on could use this label to construct a query
such as:

```
count(kube_node_info(pod_cidr~="/24")) * 256
```

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -51,6 +51,7 @@ var (
 								"kubelet_version",
 								"kubeproxy_version",
 								"provider_id",
+								"pod_cidr",
 							},
 							LabelValues: []string{
 								n.Status.NodeInfo.KernelVersion,
@@ -59,6 +60,7 @@ var (
 								n.Status.NodeInfo.KubeletVersion,
 								n.Status.NodeInfo.KubeProxyVersion,
 								n.Spec.ProviderID,
+								n.Spec.PodCIDR,
 							},
 							Value: 1,
 						},
--- a/internal/store/node_test.go
+++ b/internal/store/node_test.go
@@ -62,6 +62,20 @@ func TestNodeStore(t *testing.T) {
 			`,
 			MetricNames: []string{"kube_node_spec_unschedulable", "kube_node_labels", "kube_node_info"},
 		},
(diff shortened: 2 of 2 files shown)


internal/store/node.go
```diff
@@ -59,6 +60,7 @@ var (
 								n.Status.NodeInfo.KubeletVersion,
 								n.Status.NodeInfo.KubeProxyVersion,
 								n.Spec.ProviderID,
+								n.Spec.PodCIDR,
```
LiliC: This is an optional label, not sure how we usually handle optional labels cc @brancz @tariq1890

internal/store/node_test.go
```diff
@@ -46,6 +46,7 @@ func TestNodeStore(t *testing.T) {
 				},
 				Spec: v1.NodeSpec{
 					ProviderID: "provider://i-uniqueid",
+					PodCIDR:    "172.24.10.0/24",
```
LiliC: Since this can be empty or not set, lets have one test where this is not set.

internal/store/node.go
brancz: It’s not a pointer so at worst this is the empty string in which case Prometheus just drops the label so it’s fine as is.

internal/store/node_test.go
jacobstr: I'm adding a general test case with an empty `v1.Node` and verifying all empty strings. Not sure if you have preferences over the ordering of test cases but I'm happy to change that.

jacobstr:
Note: I'm not committing a spurious ```
+	github.com/gogo/protobuf v1.1.1 // indirect
```

That show up in my local diff. I don't expect any changes to `go.mod` contents given I have not changed added / updated deps.

brancz:
Thanks!

/lgtm
/approve

PR #965, opened 2019-11-01
document the new pod-cidr dimension

jacobstr:

https://github.com/kubernetes/kube-state-metrics/pull/963 was recently merged. I neglected to document this new dimension.


tariq1890:
Thanks @jacobstr :)!

PR #966, opened 2019-11-01
Added a kube_hpa_target_metric metric that publishes all of a HPA's target metrics.

### LINKED ISSUE #768: Add target and current cpu utilization in hpa metrics
Want to monitor the current and target CPU utilization for deployments via Prometheus. However, it does not seem to available in the current list of HPA metrics (https://github.com/kubernetes/kube-state-metrics/blob/1dfe6681e9/docs/horizontalpodautoscaler-metrics.md). 

Is there any other way to get these values in Prometheus? If not, can they be added to the HPA metrics?

janzantinge:
kube-state-metrics currently publishes only part of a Horizontal Pod Autoscaler's spec, specifically the Minimum and Maximum Replicas values. This PR aims to add the HPA's target metric values to the list metrics to be published.

Fixes #768

### DIFF
--- a/internal/store/hpa.go
+++ b/internal/store/hpa.go
@@ -28,6 +28,20 @@ import (
 	"k8s.io/client-go/tools/cache"
 )
 
+type MetricTargetType int
+
+const (
+	Value MetricTargetType = iota
+	Utilization
+	Average
+
+	MetricTargetTypeCount // Used as a length argument to arrays
+)
+
+func (m MetricTargetType) String() string {
+	return [...]string{"value", "utilization", "average"}[m]
+}
+
 var (
 	descHorizontalPodAutoscalerLabelsName          = "kube_hpa_labels"
 	descHorizontalPodAutoscalerLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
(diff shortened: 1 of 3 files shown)


internal/store/hpa_test.go
```diff
@@ -111,3 +169,11 @@ func TestHPAStore(t *testing.T) {
...
 	}
 }
+
+func int32ptr(value int32) *int32 {
+	return &value
+}
+
+func resourcePtr(quantity resource.Quantity) *resource.Quantity {
+	return &quantity
+}
```
janzantinge: Let me know if these helper functions exist somewhere else, I didn't find any at first glance.

tariq1890: I don't think we need these helper functions right now. Not a strong opinion though.

janzantinge:
Is there anything you all need from me to get this PR reviewed? Just more time?

internal/store/hpa.go
```diff
@@ -76,6 +78,91 @@ var (
...
+							averageValue, ok := m.Object.AverageValue.AsInt64()
+							if ok {
+								targets = append(targets, &metric.Metric{
+									LabelKeys:   targetMetricLabels,
+									LabelValues: []string{m.Object.MetricName, "average"},
+									Value:       float64(averageValue),
+								})
+							}
+						}
+					case autoscaling.PodsMetricSourceType:
```
tariq1890: We can condense this switch-case block further. Let's just assign the label values and metric values based on the hpa target metric type. After that, we can have just one instance of the target metric creation and the append to the metrics slice.

internal/store/hpa.go
```diff
@@ -76,6 +78,91 @@ var (
...
+						})
+
+						if m.Resource.TargetAverageValue != nil {
+							value, ok := m.Resource.TargetAverageValue.AsInt64()
+							if !ok {
+								continue
+							}
+							targets = append(targets, &metric.Metric{
+								LabelKeys:   targetMetricLabels,
+								LabelValues: []string{string(m.Resource.Name), "value"},
```
tariq1890: Don't forget this change when you address the previous comment.
```suggestion
								LabelValues: []string{string(m.Resource.Name), "average"},
```

tariq1890:
Very sorry for the delay @janzantinge ! I've provided my review comments.

internal/store/hpa.go
```diff
@@ -76,6 +78,73 @@ var (
...
+		{
+			Name: "kube_hpa_spec_target_metric",
+			Type: metric.Gauge,
+			Help: "The metric specifications used by this autoscaler when calculating the desired replica count.",
+			GenerateFunc: wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				targets := make([]*metric.Metric, 0, len(a.Spec.Metrics))
+				for _, m := range a.Spec.Metrics {
+					var metricName string
+					var value, average int64
+					valueLabel := "value"
```
tariq1890: We don’t need this variable. Let’s just use the string directly. I think we can miss out on some string optimization for readability’s sake.

internal/store/hpa.go
```diff
@@ -76,6 +78,73 @@ var (
...
+					case autoscaling.PodsMetricSourceType:
+						metricName = m.Pods.MetricName
+
+						average, averageOk = m.Pods.TargetAverageValue.AsInt64()
+					case autoscaling.ResourceMetricSourceType:
+						metricName = string(m.Resource.Name)
+
+						value = int64(*m.Resource.TargetAverageUtilization)
+						valueOk = true
+						valueLabel = "utilization"
```
janzantinge: I use the valueLabel var here, should I drop the "utilization" label value?

internal/store/hpa.go
janzantinge: I update this variable in one of the cases (highlighted in another comment), I can avoid that and remove the variable

internal/store/hpa.go
```diff
@@ -76,6 +78,71 @@ var (
...
+							average, averageOk = m.Object.AverageValue.AsInt64()
+						}
+					case autoscaling.PodsMetricSourceType:
+						metricName = m.Pods.MetricName
+
+						average, averageOk = m.Pods.TargetAverageValue.AsInt64()
+					case autoscaling.ResourceMetricSourceType:
+						metricName = string(m.Resource.Name)
+
+						value = int64(*m.Resource.TargetAverageUtilization)
```
tariq1890: Why aren't we nil checking here?

tariq1890: Also, can we not use `.AsInt64()` here?

internal/store/hpa.go
tariq1890: Oh, I see. I was a bit confused by that.

internal/store/hpa.go
tariq1890: Hmmmmm, I think having "utilization" as the metric value is important. It would be misleading otherwise.
You can probably do either of the two
i) Have a new pair of vars `utilization` and `utilizationOk` (could shorten it to `utilz` and `utilzOk`.

OR

ii) Have an custom enum Type to represent the three types of metric vals - `value`, `average`, and `utilization`.

Your call :).

internal/store/hpa.go
janzantinge: I should definitely be checking `nil` here. 

`m.Resource.TargetAverageUtilization` is an `int32` so it just needs to be cast to an `int64`
https://github.com/kubernetes/api/blob/master/autoscaling/v2beta1/types.go#L173

internal/store/hpa.go
janzantinge: Went the enum route 😄

tariq1890: Awesome!

tariq1890:
/lgtm
/approve

Thank you so much for your contribution @janzantinge ! Sorry once again for the delay :)

PR #949, opened 2019-10-14
Namespace status conditions

wozniakjan:
**What this PR does / why we need it**:
Add metrics for namespace status conditions - https://github.com/kubernetes/community/pull/4045, https://github.com/kubernetes/kubernetes/pull/73405

~NOTE: this requires kubernetes v1.16 among dependencies. There is a tracker which updates the dependencies https://github.com/kubernetes/kube-state-metrics/pull/945 and this PR should wait and rebase when https://github.com/kubernetes/kube-state-metrics/pull/945 merges. As a makeshift solution for testing, this uses fork of autoscaler https://github.com/wozniakjan/autoscaler with swapped `go dep` to `go mod` and includes necessary changes to bring kubernetes v1.16.1 as a dependency. I will try to upstream the changes to autoscaler to unblock dependency mismatch.~ https://github.com/kubernetes/kube-state-metrics/commit/7a75a4a3040720bc7ba15b41f90655a45b551262 contains the dependency update mentioned in https://github.com/kubernetes/kube-state-metrics/pull/945

**Sample output**:
```
$ k get ns
NAME              STATUS        AGE
default           Active        3d5h
kube-node-lease   Active        3d5h
kube-public       Active        3d5h
kube-system       Active        3d5h
test              Terminating   3d5h

$ curl localhost:8080/metrics | grep namespace_status
# HELP kube_namespace_status_phase kubernetes namespace status phase.
# TYPE kube_namespace_status_phase gauge
kube_namespace_status_phase{namespace="kube-node-lease",phase="Active"} 1
kube_namespace_status_phase{namespace="kube-node-lease",phase="Terminating"} 0
kube_namespace_status_phase{namespace="default",phase="Active"} 1
kube_namespace_status_phase{namespace="default",phase="Terminating"} 0
kube_namespace_status_phase{namespace="test",phase="Active"} 0
kube_namespace_status_phase{namespace="test",phase="Terminating"} 1
kube_namespace_status_phase{namespace="kube-system",phase="Active"} 1
kube_namespace_status_phase{namespace="kube-system",phase="Terminating"} 0
kube_namespace_status_phase{namespace="kube-public",phase="Active"} 1
kube_namespace_status_phase{namespace="kube-public",phase="Terminating"} 0
# HELP kube_namespace_status_condition The condition of a namespace.
# TYPE kube_namespace_status_condition gauge
kube_namespace_status_condition{namespace="test",condition="NamespaceDeletionDiscoveryFailure",status="true"} 1
[...]

PR #969, opened 2019-11-13
Multi list watch

### LINKED ISSUE #942: Observing More Than One Ns Leads to Inconsistent Metrics Being Exported
/kind bug

**What happened:**

Note: I'm going to name only the kube_pod_container_resource_requests metric in my following explanation. The problem affects all exported metrics though, just with different counts and numbers than the ones I wrote down for kube_pod_container_resource_requests. To verify the amount of metrics being exported I directly fetch the kube-state-metrics metrics endpoint with a browser and count the number of series occurrences using the find feature.

I'm running kube-state-metrics trying to observe a total of 148 PODs over 17 namespaces running 175 containers. Therefore I'm expecting to see >at least< 175x2=350 kube_pod_container_resource_requests metrics being exported. Two for each container - resource="cpu" and resource="memory". I'm saying "at least", since kube-state-metrics also exports completed POD resource metrics while I'm only interested and counting running PODs. 
However I only see around 20-25 of those metrics when observing all 17 namespaces. As soon as I start lowering the amount of observed namespaces to the first 8 of those 17, I see more metrics. In those first 8 NS 129 containers are running - and I can see 218 of the expected 129x2=258 resource metrics. 

As long as I observe only one of those namespaces everything seems fine and I can find both memory and cpu metrics for each container in the metrics endpoint. The same applies for all the other exported metrics.

Also, over time the amount of exported metrics seems to decline. Right after start-up the kube-state-metrics instance might export the above stated 218 resource series. However when I let it sit for an hour or so that amount might have dropped to e.g. 83. Then a couple minutes later it might go up again to 102 or so but never reach the expected amount relating to the actual container count. 

What I also find quite weird is the fact that when I keep reloading the metrics endpoint over and over again I may always see say only 10 metrics when I might expect 20. But every time I reload those 10 metrics being exported seem to randomly window over an internally larger known subset of metrics as I always get a different set of 10. 

There is nothing in terms of errors in the logs. Only the to be expected watch expiration warnings. 

**What you expected to happen:**

Always export all metrics of all observed resources. 

**How to reproduce it (as minimally and precisely as possible):**

IDK. Come to my desk? o_0 
After reading on a couple of other issues here I cannot believe kube-state-metrics is hitting a resource barrier here, especially since there is nothing in the logs pointing to any sort of failure. Also I don't think I'm doing anything fancy. Just trying to consistently observe the PODs of more than one namespace. 

**Anything else we need to know?:**

Since we're running this metrics instance specifically for a sub-project it is part of the specific application monitoring infrastructure and not deployed at the cluster level. So I set it up like noted here: https://github.com/kubernetes/kube-state-metrics#limited-privileges-environment 

When testing I have all PODs under my control. So all of this cannot be caused by POD life-cycling caused by someone else. 

**Environment:**

Args:

- 	--collectors=pods
- 	--v=5
- 	--namespace=[1..17]

Tried:

- 	quay.io/coreos/kube-state-metrics:v1.7.2
- 	quay.io/coreos/kube-state-metrics:v1.8.0

Kubernetes version:

- 	openshift v3.11.135
- 	kubernetes v1.11.0+d4cacc0

Requests: 

-  1core/800M (I used the recommended defaults initially)

	
Limits: 
- 2cores/2G
		

Thank you for your time! I appreciate any help. Maybe some of you have ideas how to debug this, where to look and what to check.

brancz:

Previously kube-state-metrics created multiple ListerWatchers, with
multiple reflectors for a single store. This caused that on "replace"
events, the entire store was replaced with the content of a single
reflector instead of the union of all reflectors acting on a single
store.

This patch changes the architecture to have a single reflector that is
fed by a MultiListerWatcher acting on a single store, which as a result
fixes the replace actions.

This needs a little bit more testing on my side, but I wanted to open this already to give others the chance to validate as well.

Fixes #942

@LiliC @tariq1890 

@paxbit as the original reporter, could you try this out on your setup?

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -41,6 +41,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog"
 
+	"k8s.io/kube-state-metrics/pkg/listwatch"
 	"k8s.io/kube-state-metrics/pkg/metric"
 	metricsstore "k8s.io/kube-state-metrics/pkg/metrics_store"
 	"k8s.io/kube-state-metrics/pkg/options"
@@ -329,10 +330,9 @@ func (b *Builder) reflectorPerNamespace(
 	store cache.Store,
 	listWatchFunc func(kubeClient clientset.Interface, ns string) cache.ListerWatcher,
 ) {
-	for _, ns := range b.namespaces {
-		lw := listWatchFunc(b.kubeClient, ns)
-		instrumentedListWatch := watch.NewInstrumentedListerWatcher(lw, b.metrics, reflect.TypeOf(expectedType).String())
-		reflector := cache.NewReflector(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, 0)
-		go reflector.Run(b.ctx.Done())
-	}
+	lwf := func(ns string) cache.ListerWatcher { return listWatchFunc(b.kubeClient, ns) }
+	lw := listwatch.MultiNamespaceListerWatcher(b.namespaces, nil, lwf)
+	instrumentedListWatch := watch.NewInstrumentedListerWatcher(lw, b.metrics, reflect.TypeOf(expectedType).String())
+	reflector := cache.NewReflector(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, 0)
+	go reflector.Run(b.ctx.Done())
 }
--- a/pkg/listwatch/listwatch.go
+++ b/pkg/listwatch/listwatch.go
@@ -1,0 +1,265 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package listwatch
+
+import (
+	"fmt"
+	"strings"
+	"sync"
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/fields"
+	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/watch"
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/tools/cache"
+)
+
+// NewUnprivilegedNamespaceListWatchFromClient mimics
+// cache.NewListWatchFromClient.
+// It allows for the creation of a cache.ListWatch for namespaces from a client
+// that does not have `List` privileges. If the slice of namespaces contains
+// only v1.NamespaceAll, then this func assumes that the client has List and
+// Watch privileges and returns a regular cache.ListWatch, since there is no
+// other way to get all namespaces.
+//
+// The allowed namespaces and denied namespaces are mutually exclusive.
+// See NewFilteredUnprivilegedNamespaceListWatchFromClient for a description on how they are applied.
+func NewUnprivilegedNamespaceListWatchFromClient(c cache.Getter, allowedNamespaces, deniedNamespaces []string, fieldSelector fields.Selector) cache.ListerWatcher {
+	optionsModifier := func(options *metav1.ListOptions) {
+		options.FieldSelector = fieldSelector.String()
+	}
+	return NewFilteredUnprivilegedNamespaceListWatchFromClient(c, allowedNamespaces, deniedNamespaces, optionsModifier)
+}
+
+// NewFilteredUnprivilegedNamespaceListWatchFromClient mimics
+// cache.NewUnprivilegedNamespaceListWatchFromClient.
+// It allows for the creation of a cache.ListWatch for allowed or denied namespaces
+// from a client that does not have `List` privileges.
+//
+// If the given allowed namespaces contain only v1.NamespaceAll,
+// then this function assumes that the client has List and
+// Watch privileges and returns a regular cache.ListWatch, since there is no
+// other way to get all namespaces.
+//
+// The given allowed and denied namespaces are mutually exclusive.
+// If allowed namespaces contain multiple items, the given denied namespaces have no effect.
+// If the allowed namespaces includes exactly one entry with the value v1.NamespaceAll (empty string),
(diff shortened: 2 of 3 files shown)


brancz:
My apologies, somehow this worked yesterday, but now I'm getting mod errors that look the same as in https://github.com/kubernetes/kube-state-metrics/pull/949.

paxbit:
@brancz Ok, thanks! I'll try as soon as you're happy regarding #949.
Also atm. I seem unable to build ksm from source using `make container` as it fails with:
`go: k8s.io/autoscaler/vertical-pod-autoscaler@v0.0.0-20191114093134-0e82aa045fa1 requires
        k8s.io/api@v0.0.0: reading k8s.io/api/go.mod at revision v0.0.0: unknown revision v0.0.0
` 
I'm not much of a go guy and cannot spend time digging into whats causing this. Do you have any suggestions?

tariq1890:
@brancz Can you rebase this?

brancz:
Rebased, and now everything should build fine @paxbit @tariq1890 .

brancz:
Putting on hold until we verify that it works as expected.

/hold

go.mod
```diff
@@ -3,7 +3,9 @@ module k8s.io/kube-state-metrics
 require (
 	github.com/brancz/gojsontoyaml v0.0.0-20190425155809-e8bd32d46b3d
 	github.com/campoy/embedmd v1.0.0
+	github.com/coreos/prometheus-operator v0.34.0
```
tariq1890: This is a rather heavy dependency. I'd actually prefer either of the two approaches below:

i) Move the MultiListWatcher to its own project/module.
OR
ii) We could copy the implementation code over to this project. If we choose this approach, we can tweak it a bit to use `klog` instead of a NoOp logger implementation of a foreign (in this project's context) log project.

go.mod
```diff
@@ -3,7 +3,9 @@ module k8s.io/kube-state-metrics
...
 	github.com/dgryski/go-jump v0.0.0-20170409065014-e1f439676b57
+	github.com/go-kit/kit v0.9.0
```
tariq1890: I would recommend not importing this.

go.mod
```diff
@@ -15,8 +17,17 @@ require (
 	k8s.io/api v0.0.0-20191112020540-7f9008e52f64
 	k8s.io/apimachinery v0.0.0-20191111054156-6eb29fdf75dc
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.0.0-20191115143342-4cf961056038
-	k8s.io/client-go v0.0.0-20191109102209-3c0d1af94be5
+	k8s.io/client-go v12.0.0+incompatible
 	k8s.io/klog v1.0.0
 )
 
+replace (
```
tariq1890: I believe we don't need these anymore.

internal/store/builder.go
```diff
@@ -329,10 +331,9 @@ func (b *Builder) reflectorPerNamespace(
...
 	listWatchFunc func(kubeClient clientset.Interface, ns string) cache.ListerWatcher,
 ) {
-	for _, ns := range b.namespaces {
-		lw := listWatchFunc(b.kubeClient, ns)
-		instrumentedListWatch := watch.NewInstrumentedListerWatcher(lw, b.metrics, reflect.TypeOf(expectedType).String())
-		reflector := cache.NewReflector(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, 0)
-		go reflector.Run(b.ctx.Done())
-	}
+	lwf := func(ns string) cache.ListerWatcher { return listWatchFunc(b.kubeClient, ns) }
+	lw := listwatch.MultiNamespaceListerWatcher(log.NewNopLogger(), b.namespaces, nil, lwf)
```
tariq1890: I am not too sure about injecting a Noop logger. There definitely is value in logging the ops here. This goes back to my first comment.

go.mod
tariq1890: Preference for approach 2.

LiliC: i) I don't think we would move this into its own project from the prometheus-operator as we in prometheus-operator would find it hard to maintain that and bump deps there etc.

ii) so someone, I think istio did this :D with this exact pkg, they copied it over to their project a while ago, but never kept up with our bug fixes and they ended up with a huge bug we fixed. What I mean is, we would need to maintain both pkgs then.

While I agree being dependent on prom-operator is a huge thing, I would like to see how much code we would have to maintain in that pkg in here? @brancz how much work do you think this would be to move it to ksm?

brancz: It’s already a well encapsulated package there so copying it over is trivial.

paxbit:
@brancz @LiliC Built it today and it is running for 20m without showing the old behavior. Will observe it over time today till some time after lunch. If it's still ok at 4pm CET today I'll post a LGTM later.

LiliC:
Thanks @paxbit for testing!

paxbit:
@LiliC @brancz Still getting coherent results after ~3.5h runtime. LGTM. Thanks :) Appreciate the support!

pkg/listwatch/listwatch.go
```diff
@@ -0,0 +1,269 @@
+// This was originally copied from the Prometheus Operator project. Due to
```
tariq1890: Would there be a license breach if we just add this as 2019 Copyright of Kube-state-metrics? Considering my earlier review comments, we would be modifying the signature of ListWatch method to avoid the `go-kit` dependency and other changes should trickle down. 

Thoughts? cc @LiliC

LiliC: I would keep this, or like suggested above just vendor this in. We want this in ASAP to fix the bug, so I think just merging it ASAP and fixing things in a future release would be best, wdyt @brancz @tariq1890

brancz:
Fixed all comments I believe. @LiliC @tariq1890 can you give this a re-review?

tariq1890:
@brancz I pulled down your branch and confirmed that you don't need the replace directives. Providing my last set of review comments.

go.mod
```diff
@@ -15,8 +16,17 @@ require (
 	k8s.io/api v0.0.0-20191112020540-7f9008e52f64
 	k8s.io/apimachinery v0.0.0-20191111054156-6eb29fdf75dc
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.0.0-20191115143342-4cf961056038
-	k8s.io/client-go v0.0.0-20191109102209-3c0d1af94be5
+	k8s.io/client-go v12.0.0+incompatible
```
tariq1890: k8s.io/client-go v0.0.0-20191114101535-6c5935290e33

go.mod
```diff
@@ -15,8 +16,17 @@ require (
 	k8s.io/api v0.0.0-20191112020540-7f9008e52f64
```
tariq1890:
```suggestion
	k8s.io/api v0.0.0-20191114100352-16d7abae0d2a
```

tariq1890:
```suggestion
	k8s.io/api v0.0.0-20191114100352-16d7abae0d2a
```

go.mod
tariq1890: Confirmed that we can remove these. Once you add the revision hashes I'd mentioned earlier, run `go mod tidy` and `go mod vendor`, it should work.

brancz:
Reverted the mod updates. Please take a look.

PR #989, opened 2019-12-05
Add the possibility to provide “custom” metric-store to the builder

### LINKED ISSUE #973: Add the possibility to provide “custom” metric-store to the builder
/kind feature

The “kube-state-metrics” project is well organized in terms of Golang packages and functionalities. It also uses the concept of “internal” packages that allow seamless refactoring for the users.

I started a discussion with some of you during the last Kubecon Europe in Barcelona (I know it’s been a while), about how a third-party project can reuse the “K.S.M. collectors” to generate other kinds of metric directly (other than Prometheus metrics). 

Few months ago, I worked on a POC to highlight all the blocking points that I see to achieve my goal. The main idea was to allow third-party-projects to provide a “custom” MetrisStore to the `Builder`.

In summary, I saw the following blocking points:
* the “NewBuilder()” function is in an internal package preventing use it in a third-party project. 
  **Solution**: To keep all the logic in the internal package, I created a public package that implements a kind of “bridge” pattern and also an interface for the Builder.
  You can see the code here: 
https://github.com/clamoriniere/kube-state-metrics/pull/1/commits/3f6d9d9f6e938e3ae0288f8dec11d9fd09ab7848
* The interface `FamilyByteSlicer` is preventing access the metrics' raw information. For the current usage, it completely makes sense to have it, but in the scope of the POC, it is a blocking point.
  **Solution**: Remove the `FamilyByteSlicer` interface and provide access to “metrics.Family” directly. Also, I have to split into two packages “metrics” package (creation of “metrics generator”) package in order to solve a cyclic import issue.
https://github.com/clamoriniere/kube-state-metrics/pull/1/commits/8e1d67c7278598fff0fc905ca313452cb0710c8a
* The last blocking point to allow the usage of “custom” MetricsStore that the internal package uses the `metricsstore.MetricStore` struct instead of `cache.Store` interface (k8s.io/client-go/tools/cache).
  **Solution**: use `cache.Store` interface in “Builder”. It is a lot of code changes, but it is mainly a one to one replacement.
https://github.com/clamoriniere/kube-state-metrics/pull/1/commits/cfd9e78aef60f3938fbf94739f5f23cee93022eb

After resolving the previous points, I was able to add a new configuration function to the Builder allowing the registration of a “Store” generator function (https://github.com/clamoriniere/kube-state-metrics/pull/1/commits/cfd9e78aef60f3938fbf94739f5f23cee93022eb#diff-a20aeff27f3302b071a3e4fe4d66001eR138). It allows to inject custom cache.Store implementation in the Builder.
The sum of all the changes is in this draft PR: https://github.com/clamoriniere/kube-state-metrics/pull/1

To validate that it is working as expected, I created this “simple” project: https://github.com/clamoriniere/ddksm this is using previous changes and generates directly Datadog metrics.

I will be more than happy to share with you more implementation details if someone is interested.

I'm also interested to know if people think it is a relevant use-case, and if I can move forward and propose an official pull-request.

clamoriniere:

This PullRequest allows a third-party project to reuse the kube-state-metrics `collectors` by adding a way to configure a custom `cache.Store` generator function in the `store.Builder`.

In summary, the modifications are:
* Use `cache.Store` interface in “Builder”. It is a lot of code changes, but it is mainly a one to one replacement.
* Remove the FamilyByteSlicer interface and provide access to “metrics.Family” directly. Also, I have to split into two packages “metrics” package (creation of “metrics generator”) package to solve a cyclic import issue.
* To keep all the logic in the internal package, I created a public package that implements a kind of “bridge” pattern and also an interface for the Builder.

Fixes #973

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -1,0 +1,95 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package builder
+
+import (
+	"context"
+
+	"github.com/prometheus/client_golang/prometheus"
+	vpaclientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
+	clientset "k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/tools/cache"
+
+	internalstore "k8s.io/kube-state-metrics/internal/store"
+	ksmtypes "k8s.io/kube-state-metrics/pkg/builder/types"
+	"k8s.io/kube-state-metrics/pkg/options"
+)
+
+// Builder helps to build store. It follows the builder pattern
+// (https://en.wikipedia.org/wiki/Builder_pattern).
+type Builder struct {
+	internal ksmtypes.BuilderInterface
+}
+
+// NewBuilder returns a new builder.
+func NewBuilder() *Builder {
+	b := &Builder{
+		internal: internalstore.NewBuilder(),
+	}
+	return b
+}
+
+// WithMetrics sets the metrics property of a Builder.
+func (b *Builder) WithMetrics(r *prometheus.Registry) {
+	b.internal.WithMetrics(r)
+}
+
+// WithEnabledResources sets the enabledResources property of a Builder.
+func (b *Builder) WithEnabledResources(c []string) error {
+	return b.internal.WithEnabledResources(c)
+}
+
+// WithNamespaces sets the namespaces property of a Builder.
+func (b *Builder) WithNamespaces(n options.NamespaceList) {
+	b.internal.WithNamespaces(n)
+}
+
+// WithSharding sets the shard and totalShards property of a Builder.
+func (b *Builder) WithSharding(shard int32, totalShards int) {
+	b.internal.WithSharding(shard, totalShards)
+}
+
+// WithContext sets the ctx property of a Builder.
+func (b *Builder) WithContext(ctx context.Context) {
+	b.internal.WithContext(ctx)
+}
+
+// WithKubeClient sets the kubeClient property of a Builder.
+func (b *Builder) WithKubeClient(c clientset.Interface) {
+	b.internal.WithKubeClient(c)
+}
+
+// WithVPAClient sets the vpaClient property of a Builder so that the verticalpodautoscaler collector can query VPA objects.
+func (b *Builder) WithVPAClient(c vpaclientset.Interface) {
+	b.internal.WithVPAClient(c)
+}
+
+// WithWhiteBlackList configures the white or blacklisted metric to be exposed
+// by the store build by the Builder.
+func (b *Builder) WithWhiteBlackList(l ksmtypes.WhiteBlackLister) {
+	b.internal.WithWhiteBlackList(l)
+}
+
+// WithCustomGenerateStoreFunc configures a constom generate store function
+func (b *Builder) WithCustomGenerateStoreFunc(f ksmtypes.BuildStoreFunc) {
+	b.internal.WithCustomGenerateStoreFunc(f)
+}
+
+// Build initializes and registers all enabled stores.
+func (b *Builder) Build() []cache.Store {
+	return b.internal.Build()
+}
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -84,9 +84,14 @@ func (b *Builder) WithWhiteBlackList(l ksmtypes.WhiteBlackLister) {
 	b.internal.WithWhiteBlackList(l)
 }
 
-// WithCustomGenerateStoreFunc configures a constom generate store function
-func (b *Builder) WithCustomGenerateStoreFunc(f ksmtypes.BuildStoreFunc) {
-	b.internal.WithCustomGenerateStoreFunc(f)
+// WithGenerateStoreFunc configures a constom generate store function
+func (b *Builder) WithGenerateStoreFunc(f ksmtypes.BuildStoreFunc) {
+	b.internal.WithGenerateStoreFunc(f)
+}
+
+// DefaultGenerateStoreFunc returns default buildStore function
+func (b *Builder) DefaultGenerateStoreFunc() ksmtypes.BuildStoreFunc {
+	return b.internal.DefaultGenerateStoreFunc()
 }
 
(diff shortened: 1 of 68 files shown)


brancz:
hmm it looks like this unfortunately has a rather big hit on our benchmarks (see the benchmark job on CI) .. 10% increase in /metrics response is something I'm not entirely sure we want.

cc @tariq1890 @LiliC what do you think?

clamoriniere:
I'm agree with you, I will look at decreasing this overhead.

LiliC:
Anything that impacts the performance in a negative way is not something we want, while I understand it would be good to have this. 10% is a lot IMO.

clamoriniere:
Hi @brancz @LiliC @tariq1890,

So I checked the benchmark results. In fact the 10% increase on certain benchmark tests is when we compare my PR branch with the `release-1.8 branch` which makes sens since, I rebased by branch with master yesterday and so we bench other changes with mine. :)

If we look the benchmark comparison with `master` branch, the results are way better! 

```
### Result
old=master new=HEAD
BenchmarkKubeStateMetrics/GenerateMetrics-2     1000992426      1000926748      -0.01%
BenchmarkKubeStateMetrics/MakeRequests-2        14959897766     14214421176     -4.98%
BenchmarkPodStore-2                             37307           37964           +1.76%
BenchmarkMetricWrite/value-1-2                  649             597             -8.01%
BenchmarkMetricWrite/value-35.7-2               775             722             -6.84%
BenchmarkKubeStateMetrics/MakeRequests-2     694.01       730.41       1.05x
BenchmarkKubeStateMetrics/GenerateMetrics-2     849198         849204         +0.00%
BenchmarkKubeStateMetrics/MakeRequests-2        453935         453908         -0.01%
BenchmarkPodStore-2                             523            523            +0.00%
BenchmarkMetricWrite/value-1-2                  7              7              +0.00%
BenchmarkMetricWrite/value-35.7-2               7              7              +0.00%
BenchmarkKubeStateMetrics/GenerateMetrics-2     87319544        87705328        +0.44%
BenchmarkKubeStateMetrics/MakeRequests-2        26518620176     26518936496     +0.00%
BenchmarkPodStore-2                             25984           26288           +1.17%
BenchmarkMetricWrite/value-1-2                  536             536             +0.00%
BenchmarkMetricWrite/value-35.7-2               536             536             +0.00%
```

clamoriniere:
Hi @LiliC @brancz @zouyee,

Please let me know if you need more information or context about his PR.

pkg/metric/family.go
```diff
@@ -20,12 +20,24 @@ import (
 	"strings"
 )
 
+// FamilityInterface interface for a family
+type FamilityInterface interface {
```
brancz: typo `FamilityInterface` -> `FamilyInterface`

brancz:
Generally this looks fine, but I'm still a little confused about the consumption model. Could you draft a bit of code showing what you're intending to do?

pkg/builder/types/interfaces.go
```diff
@@ -0,0 +1,55 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
```
zouyee: s/2018/2019

pkg/builder/builder.go
```diff
@@ -0,0 +1,95 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
```
zouyee: same

clamoriniere:
Hi @brancz 
sorry for the late reply.

to illustrate the usage of this modification, I built this POC: https://github.com/clamoriniere/ddksm
some key entries in the code:
- the custom metrics store that stores `metric.Family` by resources and generates Datadog metrics: https://github.com/clamoriniere/ddksm/blob/master/pkg/store/store.go
- the new `builder` method `WithCustomGenerateStoreFunc()` is used in addition to the common `kube-state-metrics` configuration: https://github.com/clamoriniere/ddksm/blob/master/main.go#L125
- to trigger metrics generation and export to the `Dogstatd` server, the main process is looping over the `metricStore` and call the`.push()` method: https://github.com/clamoriniere/ddksm/blob/master/main.go#L133

pkg/builder/builder.go
```diff
@@ -0,0 +1,95 @@
...
+}
+
+// WithWhiteBlackList configures the white or blacklisted metric to be exposed
+// by the store build by the Builder.
+func (b *Builder) WithWhiteBlackList(l ksmtypes.WhiteBlackLister) {
+	b.internal.WithWhiteBlackList(l)
+}
+
+// WithCustomGenerateStoreFunc configures a constom generate store function
+func (b *Builder) WithCustomGenerateStoreFunc(f ksmtypes.BuildStoreFunc) {
```
brancz: Why not make the code always call this? And dependency inject even the "normal" store builder? Also seems less likely that we would mess with this and translate to a bit better stability for you :)

pkg/builder/types/interfaces.go
clamoriniere: updated

pkg/builder/builder.go
clamoriniere: updated

pkg/builder/builder.go
clamoriniere: I tried to follow the same "Builder" configuration pattern that is used for `WithWhiteBlackList()` and other options. Also I tried to not add extra builder configuration step for the "default" use case.

currently the "normal" store builder is set in the `NewBuilder()` function:
https://github.com/kubernetes/kube-state-metrics/blob/ae822f3a2c7678e009e2cc9f32eef687762c5380/internal/store/builder.go#L71

Do you mean that I should remove this default configuration, and expose new function: `DefaultStoreBuilder() ksmtypes.BuildStoreFunc`. 
Then add the step for everyone
```golang
builder.WithCustomGenerateStoreFunc(DefaultStoreBuilder())
```

brancz: Yes I think just like `WithWhiteBlackList` function it should always be called.

clamoriniere: Make sense, I have updated the PR accordingly to your comment.

pkg/builder/builder.go
```diff
@@ -0,0 +1,100 @@
...
+	b.internal.WithWhiteBlackList(l)
+}
+
+// WithGenerateStoreFunc configures a constom generate store function
+func (b *Builder) WithGenerateStoreFunc(f ksmtypes.BuildStoreFunc) {
+	b.internal.WithGenerateStoreFunc(f)
+}
+
+// DefaultGenerateStoreFunc returns default buildStore function
+func (b *Builder) DefaultGenerateStoreFunc() ksmtypes.BuildStoreFunc {
```
brancz: This doesn’t seem like something that needs to be bound to the builder strict.

clamoriniere: Unfortunately, it is needed due to the `(b *Builder) buildStore(...)` that accesses Builder fields. 
https://github.com/kubernetes/kube-state-metrics/pull/989/files#diff-a20aeff27f3302b071a3e4fe4d66001eR331

If you think it is needed, I can look at how we can solve it.

brancz: Ah yes got it. I think this is ok like this then.

clamoriniere:
@brancz thank you so much for your time on this PR. 👍

PR #993, opened 2019-12-09
fix nil panics parsing hpa cpu current metrics

tariq1890:
This PR fixes a nil de-referencing issue [here](https://github.com/kubernetes/kube-state-metrics/blob/4e4a49fe935318adff0c6ed0c3f044e5e7686875/internal/store/hpa.go#L285).

The root cause of this issue is a nil metric element being added to the metric family slice. These nil metrics are added when parsing `kube_hpa_status_current_metrics_average_value` CPU metrics. This is because CPU metrics need to be handled differently owing to the fact that they are scaled metrics.

### DIFF
--- a/internal/store/hpa.go
+++ b/internal/store/hpa.go
@@ -17,15 +17,16 @@ limitations under the License.
 package store
 
 import (
-	"k8s.io/kube-state-metrics/pkg/metric"
-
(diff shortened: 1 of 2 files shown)


brancz:
/

Issue #982, opened 2019-12-04
Remove resource version metrics


lilic:
kube_secret_metadata_resource_version, kube_configmap_metadata_resource_version and kube_ingress_metadata_resource_version expose the resource version as a string in its set of labels. This value can change often and would therefore create huge cardinality. This should be a number or not existing at all.

This is a breaking change but we should remove it as its currently causing a cardinality explosion.

@tariq1890 we discussed with @brancz and we should really do this in 1.x already. SGTY

brancz:
I think this should already be fixed in 1.x yes.

tariq1890:
Sounds good to me!

lilic:
I think this was already implemented in my PR.
/close

PR #997, opened 2019-12-11
Move resource version from metric label to metric number value

LiliC:
These metrics expose the resource version as a string in the label. This value can change often and would therefore create huge cardinality. This PR changes it to be the number value of the metric itself instead.

I checked with API team and while formally it is not guaranteed that this is always a numerical value, majority of the ecosystem assumes this is the case.

cc @tariq1890 @brancz 

Fixes https://github.com/kubernetes/kube-state-metrics/issues/982

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 ## v1.9.0-rc.0 / 2019-12-04
 
+* [BUGFIX] Move resource version from metric label to metric number value #997
+
+## v1.9.0-rc.0 / 2019-12-04
+
 * [CHANGE] Add tools as go modules #927
 * [FEATURE] Add `kube_hpa_spec_target_metric` metric. #966
 * [FEATURE] Add hpa stats for current utilization and average value. #961
--- a/docs/configmap-metrics.md
+++ b/docs/configmap-metrics.md
@@ -4,4 +4,4 @@
 | ---------- | ----------- | ----------- | ----------- |
 | kube_configmap_info | Gauge | `configmap`=&lt;configmap-name&gt; <br> `namespace`=&lt;configmap-namespace&gt; | STABLE |
 | kube_configmap_created  | Gauge | `configmap`=&lt;configmap-name&gt; <br> `namespace`=&lt;configmap-namespace&gt; | STABLE |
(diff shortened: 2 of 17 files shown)


LiliC:
Benchmark results:
https://travis-ci.org/kubernetes/kube-state-metrics/jobs/623616433#L541-L566

```
### Result
old=master new=HEAD
BenchmarkKubeStateMetrics/GenerateMetrics-2     1001245093      1000977482      -0.03%
BenchmarkKubeStateMetrics/MakeRequests-2        19276868580     17907123355     -7.11%
BenchmarkPodStore-2                             52782           50941           -3.49%
BenchmarkMetricWrite/value-1-2                  919             880             -4.24%
BenchmarkMetricWrite/value-35.7-2               1080            1052            -2.59%

BenchmarkKubeStateMetrics/MakeRequests-2     538.59       578.61       1.07x
BenchmarkKubeStateMetrics/GenerateMetrics-2     849192         844208         -0.59%
BenchmarkKubeStateMetrics/MakeRequests-2        453942         454000         +0.01%
BenchmarkPodStore-2                             523            523            +0.00%
BenchmarkMetricWrite/value-1-2                  7              7              +0.00%
BenchmarkMetricWrite/value-35.7-2               7              7              +0.00%

BenchmarkKubeStateMetrics/GenerateMetrics-2     87317896        86970760        -0.40%
BenchmarkKubeStateMetrics/MakeRequests-2        26519171184     26518128840     -0.00%
BenchmarkPodStore-2                             25984           25984           +0.00%
BenchmarkMetricWrite/value-1-2                  536             536             +0.00%
BenchmarkMetricWrite/value-35.7-2               536             536             +0.00%
```

brancz:
Very nice!

/lgtm
/hold

Putting on hold and I'll let @tariq1890 decide whether to include this as a next release candidate or whether we want to create a follow up 1.9.1 release.

LiliC:
I think I need to update the docs maybe? 🤔 

Also should we mark it experiential?

brancz:
Right the docs do need updating. It's a little weird to change them to experimental now, but realistically we're dependent on kubernetes/etcd behavior here, so if that would break this metric would break, so I'm not sure STABLE was ever the right thing/possible.

tariq1890:
My preference is to have this in the rc itself. Let’s target this to release-1.9 branch.

We need to update the CHANGELOG and squash the commits.

brancz:
That sounds perfect to me.

PR #1009, opened 2019-12-19
*: Remove non-identifying labels from pod metrics

### LINKED ISSUE #979: v2: Remove non-identifying labels from pod metrics
Remove non-identifying labels from pod metrics (eg. node labels) https://github.com/kubernetes/kube-state-metrics/blob/aed94850b839d111d15308c97e1db7ada8017c6f/internal/collector/pod.go#L511

cc @brancz do you mind taking care of this one for v2, as I believe it was your suggestion?

PhilipGough:

Fixes #979

@LiliC I made a start on this and I can't see anything else that I would consider "non-identifying" labels.

I also wondered about the removal of node name from metrics re cpu/mem usage, perhaps it is still valid there?

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -787,7 +787,7 @@ var (
 						switch resourceName {
 						case v1.ResourceCPU:
 							ms = append(ms, &metric.Metric{
-								LabelValues: []string{c.Name, p.Spec.NodeName, sanitizeLabelName(string(resourceName)), string(constant.UnitCore)},
+								LabelValues: []string{c.Name, sanitizeLabelName(string(resourceName)), string(constant.UnitCore)},
 								Value:       float64(val.MilliValue()) / 1000,
 							})
 						case v1.ResourceStorage:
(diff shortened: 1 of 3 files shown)


internal/store/pod.go
```diff
@@ -57,8 +57,8 @@ var (
 
 				m := metric.Metric{
 
-					LabelKeys:   []string{"host_ip", "pod_ip", "uid", "node", "created_by_kind", "created_by_name", "priority_class"},
-					LabelValues: []string{p.Status.HostIP, p.Status.PodIP, string(p.UID), p.Spec.NodeName, createdByKind, createdByName, p.Spec.PriorityClassName},
+					LabelKeys:   []string{"host_ip", "pod_ip", "uid", "created_by_kind", "created_by_name", "priority_class"},
```
brancz: This “info” type metric is ok to have it, just not all metrics. The idea is that any other metric can be easily joined with this one to add metadata when needed.

PhilipGough: @brancz thanks makes sense. So replacing here and removing from all other Pod metrics is sufficient?

brancz: Yes I think so. I'll try and see if I can find any other cases in my review, but I believe this was all of them.

LiliC:
/hold

until the current release is out

brancz:
/lgtm

Nice! :)

LiliC:
@PhilipGough  you need to fixup manually as bot is complaining it seems, do you mind merging all into one single commit, thanks!

> do-not-merge/invalid-commit-message

PhilipGough:
@LiliC, sorry, this needed a rebase by the time I got around to it so might need another review.

Issue #348, opened 2018-01-17
Is there or will there ever be metric for pods in terminating state?


cullepl:
I see there are metrics for if a pod is terminated `kube_pod_container_status_terminated` however, we sometimes observe pods in terminating state. 

I was wondering if there is a way via kube-state-metrics to find this state?  I couldnt see anything and I'm running v1.2.0

All I see in our prometheus are these:

![image](https://user-images.githubusercontent.com/13000925/35053646-9d18807e-fba2-11e7-85b9-ed0c76f9e276.png)

If pods are stuck in terminating state then we usually have to take actions against the hosts they are running on.
We'd ideally like a solution via prometheus and alertmanager using kube-state-metrics metrics to alert us rather than having to create something home grown.

brancz:
How are you determining this today?

If kube-state-metrics can find what you are looking for in a Pod object then yes, that's something we could add a metric for.

cullepl:
If I query pods using `kubectl get pods -n my-namespace` today, I see this:

```
my-namespace  my-pod-1   6/6       Terminating   1          11d       <IP>   1.2.3.4
my-namespace   my-pod-2   1/6       Terminating   22         10d       <none>           1.2.3.5
```

If I describe one of the PODs, the status is
```
Status:				Terminating (expires Sat, 13 Jan 2018 16:35:28 +0000)
```

brancz:
When you view the yaml of this pod, can you see it there as well, or is this only available through events?

cullepl:
Looks like it's only through describing the pod.
If i do `kubectl get pod -n <ns> <pod-name> -o yaml` there is no reference to terminating.  At the end of the yaml output there is 

```
  phase: Running
  podIP: 1.2.3.4
  startTime: 2018-01-06T06:21:34Z
```

brancz:
Yes that's what I thought. The tricky thing here is that we can only extract this information from Kubernetes Events. The problem with events is their cardinality, it is most likely going to explode. Something we could do is aggregate events and expose lower cardinality aggregations. This would definitely be a very different pattern from all the other collectors today, where the objects we create metrics from are in memory. With events we would need to listen to events and garbage collect them if they're older than some period.

We could try this as an experiment, but my general feeling is that this is going to be somewhat unstable data.

nucl3arj4zz:
Will this issue be re-opened? /reopen

mxinden:
[...]

PR #1013, opened 2019-12-24
Add "Terminating" status in kube_pod_status_phase metrics

clamoriniere:
**What this PR does**:

Aim of this PullRequest is to add several new pod metrics:
- kube_pod_deleted: Unix deletion timestamp
- kube_pod_status_reason: The pod status reasons (NodeLost, Evicted)
These new metrics can be used to determine the `pod.status.phase`
displayed by `kubectl`: Running, Terminated, Unknown…
It will allow removing the `kubectl` display logic added previously
to compute the “Unknown” phase.

https://github.com/kubernetes/kube-state-metrics/blob/d35e7ba89478779515fea87d2daf7ec369ff80fc/internal/store/pod.go#L288

**why we need it**:

In some cases, a Pod can be stuck in the "Terminating" phase due to a "Kubelet" issue; for example: the Kubelet is not able to communicate with the container runtime, or the container runtime is not able to delete the associated container.

And so, it can be interesting to have a way to create an alert on this kind of "bad" Pod state, which is currently not possible to do since the pod is flag by `kube_pod_status_phase` metrics as "Running".
 

Fixes #

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -40,3 +40,32 @@
 | kube_pod_status_reason | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;NodeLost\|Evicted\&gt; | EXPERIMENTAL |
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_status_unschedulable | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+
+## Useful metrics queries
+
+### How to retrieve none standard Pod state
+
+It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknow" since it is not stored behind a field in the `Pod.status`.
+
+So to get them, you need to compose multiple metrics (like it is done in the `kubectl` command line).
+
+For example:
+
+* To get the list the Pods that are in an `Unknow` state you can to the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
+
+* For Pods in `Terminated` state: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_deleted) by (namespace, pod) * count(kube_pod_status_reason{reason!="NodeLost"})) by (namespace, pod)`
+
+A useful "Rule" can be to be alerted when a Pod is blocked in `Terminated` state for more than `5m`.
+
+```yaml
+groups:
+- name: Pod state
+  rules:
+  - alert: PodsBlockInTerminatingState
+    expr: count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_deleted) by (namespace, pod) * count(kube_pod_status_reason{reason!="NodeLost"})) by (namespace, pod) > 0
+    for: 5m
+    labels:
+      severity: page
+    annotations:
+      summary: Pod {{labels.namespace}}/{{labels.pod}} block in terminating state.
+```
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -44,17 +44,17 @@
 
 ### How to retrieve none standard Pod state
 
-It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknow" since it is not stored behind a field in the `Pod.status`.
+It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknown" since it is not stored behind a field in the `Pod.Status`.
 
-So to get them, you need to compose multiple metrics (like it is done in the `kubectl` command line).
+So to get them, you will need to compose multiple metrics (like it is done in the `kubectl` command line code).
 
 For example:
 
-* To get the list the Pods that are in an `Unknow` state you can to the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
+* To get the list of pods that are in the `Unknown` state, you can run the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
 
 * For Pods in `Terminated` state: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_deleted) by (namespace, pod) * count(kube_pod_status_reason{reason!="NodeLost"})) by (namespace, pod)`
 
-A useful "Rule" can be to be alerted when a Pod is blocked in `Terminated` state for more than `5m`.
+Here is an example of a Prometheus rule that can be used to alert on a Pod that has been in the `Terminated` state for more than `5m`.
 
 ```yaml
 groups:
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -21,7 +21,6 @@
 | kube_pod_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
 | kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
 | kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
-| kube_pod_container_resource_limits_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | DEPRECATED |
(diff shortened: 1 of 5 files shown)


jinnovation:
This PR seems like it addresses #348 as well, which was closed w/o resolution.

clamoriniere:
> This PR seems like it addresses #348 as well, which was closed w/o resolution.

@jinnovation it is exactly the use case that we want to monitor with this fix: Pod blocked in Terminated state that require specific operation on the node.

internal/store/pod.go
```diff
@@ -274,6 +278,9 @@ var (
 					}
 				}
 
+				// compute the "Terminating" phase in order to reflect what a user can see with kubectl: https://github.com/kubernetes/kubernetes/blob/6a4216ba59ce1d09c8ac1c6229649bb71b7f6c85/pkg/printers/internalversion/printers.go#L737-L741
+				isPodTerminating := p.DeletionTimestamp != nil && p.Status.Reason != nodeUnreachablePodReason
```
brancz: This is not really how kube state metrics is implemented otherwise. We usually just expose the raw data and leave computations like this to query time.

clamoriniere: I completely agree, but some logic was already added to compute the `Unknow status`. 
https://github.com/kubernetes/kube-state-metrics/pull/1013/files#diff-791ee1d484ea70de1ecd1d8875f8f5c0L286

Here we want to reflect what is displayed by "kubectl" to avoid confusion, and avoid having a Pod in a Running state when it is not.

brancz: I think that was a mistake and since we're in the process of cleaning up kube-state-metrics and releasing preparing a breaking release, the time is right to change this. @LiliC @tariq1890 what do you think? We should definitely make sure that for these cases we have the required raw data available so we _can_ compute these things at runtime, but we should strive to remove any of these types of things from kube-state-metrics, just like we state [in our docs](https://github.com/kubernetes/kube-state-metrics#overview) :) .

LiliC: Yes agreed, we should expose just raw metrics, users can compute whatever they want afterwards, and/or create recording rules for this.

lbernail: Is there a way with the current information to know how many pods have the deletionTimestamp set?
With this info, we could indeed easily compute the "Terminating" status

clamoriniere: @brancz and @LiliC
I can rework this PR to add missing metrics for "raw" information to allow computing the `Terminating` and `Unknown` pod state from metrics.
Does that sound good?

brancz: Yes that sounds perfect! :)

clamoriniere: PR updated with 2 new metrics:
- kube_pod_deleted
- kube_pod_status_reason

I can create another PR to remove the current logic for the `Unknown` pod phase.

clamoriniere:
Hi @LiliC 
sorry about the failing tests.

I updated the PR with proper documentation and also flagged new metrics as "EXPERIMENTAL" in the documentation.

docs/pod-metrics.md
```diff
@@ -41,3 +42,4 @@
 | kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_status_unschedulable | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+| kube_pod_status_reason | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;NodeLost\|Evicted\&gt; | EXPERIMENTAL |
```
tariq1890: Let's maintain the lexical order. This will need to be before `kube_pod_status_unscheduled_time`.

clamoriniere: sure

tariq1890: @clamoriniere That was a typo from me. I meant before `kube_pod_status_scheduled_time`. Sorry about that :(

internal/store/pod_test.go
```diff
@@ -863,6 +863,33 @@ kube_pod_container_status_last_terminated_reason{container="container7",namespac
...
+					NodeName:          "node1",
+					PriorityClassName: "system-node-critical",
+				},
+				Status: v1.PodStatus{
+					HostIP:    "1.1.1.1",
+					PodIP:     "1.2.3.4",
+					StartTime: &metav1StartTime,
+				},
+			},
+			// TODO: Should it be '1501569018' instead?
```
tariq1890: This is fine as it's the convention we follow for timestamp values in metrics in kube-state-metrics.

docs/pod-metrics.md
clamoriniere: no problem :) updated

clamoriniere:
Hi @tariq1890 
I'm wondering if you may have some time to review my last code update?
thanks

tariq1890:
@clamoriniere Sorry for the delay and Thank you so much for your patience with this PR! It's greatly appreciated :).

The code looks good, my last request is for you to provide a few examples (doesn;'t have to be complete) of using these raw basic metrics together for asserting on particular pod states (such as Terminating) in the `README`. This will be of great help to people who are transitioning over to the next version.

clamoriniere:
Ok no problem.

tariq1890:
@clamoriniere Just following up on this :)

clamoriniere:
> @clamoriniere Just following up on this :)

Hi @tariq1890,
Sorry I was in vacation last week. 

* I added the promQL queries in the `docs/pod-metrics.md` file. I thought it makes more sens in this file than in the `README.md`. 
* I also needed to update a bit the `make doccheck` to support extra documentation in the `docs/pod-metrics.md` file.

Let me know if you prefer that I move the doc in the README.md

docs/pod-metrics.md
```diff
@@ -35,5 +36,35 @@
...
+
+So to get them, you need to compose multiple metrics (like it is done in the `kubectl` command line).
+
+For example:
+
+* To get the list the Pods that are in an `Unknow` state you can to the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
+
+* For Pods in `Terminated` state: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_deleted) by (namespace, pod) * count(kube_pod_status_reason{reason!="NodeLost"})) by (namespace, pod)`
+
+A useful "Rule" can be to be alerted when a Pod is blocked in `Terminated` state for more than `5m`.
```
tariq1890:
```suggestion
Here is an example of a Prometheus rule that can be used to alert on a Pod that has been in the `Terminated` state for more than `5m`.
```

docs/pod-metrics.md
```diff
@@ -35,5 +36,35 @@
...
+
+### How to retrieve none standard Pod state
+
+It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknow" since it is not stored behind a field in the `Pod.status`.
+
+So to get them, you need to compose multiple metrics (like it is done in the `kubectl` command line).
+
+For example:
+
+* To get the list the Pods that are in an `Unknow` state you can to the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
```
tariq1890:
```suggestion
* To get the list of pods that are in the `Unknown` state, you can run the following promQL query: `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`
```

docs/pod-metrics.md
```diff
@@ -35,5 +36,35 @@
...
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_status_unschedulable | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+
+## Useful metrics queries
+
+### How to retrieve none standard Pod state
+
+It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknow" since it is not stored behind a field in the `Pod.status`.
+
+So to get them, you need to compose multiple metrics (like it is done in the `kubectl` command line).
```
tariq1890:
```suggestion
So to get them, you will need to compose multiple metrics (like it is done in the `kubectl` command line code).
```

docs/pod-metrics.md
```diff
@@ -35,5 +36,35 @@
...
 | kube_pod_spec_volumes_persistentvolumeclaims_readonly | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt;  <br> `volume`=&lt;volume-name&gt;  <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-claimname&gt; | STABLE |
+| kube_pod_status_reason | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;NodeLost\|Evicted\&gt; | EXPERIMENTAL |
 | kube_pod_status_scheduled_time | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_status_unschedulable | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+
+## Useful metrics queries
+
+### How to retrieve none standard Pod state
+
+It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknow" since it is not stored behind a field in the `Pod.status`.
```
tariq1890:
```suggestion
It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknown" since it is not stored behind a field in the `Pod.Status`.
```

clamoriniere:
@tariq1890 
thanks for the review and suggestions. PR updated

tariq1890:
Sure! That sounds good to me :)

buttonbuilder:
Which release will this be included in ?

clamoriniere:
> Which release will this be included in ?

I think in the next major release: `v2.0.0`

vc43:
I have a pod in status Terminating but whith kube-state-metrics:v2.7.0 can not see `kube_pod_status_phase{phase="Terminating"}`

Issue #1012, opened 2019-12-23
Segfault with 1.9.0


lawliet89:
/kind bug

**What happened**:

Kube state metrics pod crashlops with a segfault.

```
I 2019-12-23T01:32:57.901573Z Using default collectors 
I 2019-12-23T01:32:57.901633Z Using all namespace 
I 2019-12-23T01:32:57.901643Z metric white-blacklisting: blacklisting the following items:  
W 2019-12-23T01:32:57.901666Z Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work. 
I 2019-12-23T01:32:57.903415Z Testing communication with server 
I 2019-12-23T01:32:57.912850Z Running with Kubernetes cluster version: v1.15+. git version: v1.15.4-gke.22. git tree state: clean. commit: a6ba43f5a24ac29e631bb627c9b2a719c4e93638. platform: linux/amd64 
I 2019-12-23T01:32:57.912877Z Communication with server successful 
I 2019-12-23T01:32:57.913028Z Starting metrics server: 0.0.0.0:8080 
I 2019-12-23T01:32:57.913135Z Starting kube-state-metrics self metrics server: 0.0.0.0:8081 
I 2019-12-23T01:32:57.913254Z Autosharding disabled 
I 2019-12-23T01:32:57.914977Z Active collectors: certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments 
E 2019-12-23T01:32:57.917328Z pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1beta1.MutatingWebhookConfiguration: mutatingwebhookconfigurations.admissionregistration.k8s.io is forbidden: User "system:serviceaccount:core:prometheus-kube-state-metrics" cannot list resource "mutatingwebhookconfigurations" in API group "admissionregistration.k8s.io" at the cluster scope 
[...]

PR #1019, opened 2020-01-09
internal/store/hpa: Fix segfault in hpa metrics

LiliC:
This fixes https://github.com/kubernetes/kube-state-metrics/issues/1012, also adds a test case based on the example @lawliet89 provided. Thanks for that!

@lawliet89 Do you mind verifying this if you have time in your cluster, if this fix works for you?

cc @tariq1890 @brancz


brancz:
I'm happy with this as a fix to the segfault, and we'll need another fix to actually also make the external metric status to appear in a metric, which it currently doesn't.

/lgtm
/approve

PR #1020, opened 2020-01-09
internal/store/hpa.go: Set kube_hpa_status_current_metrics_average_value metric for the External metric source

lilic:
We were not setting the kube_hpa_status_current_metrics_average_value when the metric source was External.

@tariq1890 @brancz

### DIFF
--- a/internal/store/hpa.go
+++ b/internal/store/hpa.go
@@ -255,6 +255,8 @@ var (
 						}
 					} else if intVal, canFastConvert := value.AsInt64(); canFastConvert {
 						metricValue = float64(intVal)
+					} else if c.Type == autoscaling.ExternalMetricSourceType {
+						metricValue = float64(value.MilliValue()) / 1000.0
 					} else {
 						// Skip unsupported metric value format
 						continue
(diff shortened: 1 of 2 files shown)


internal/store/hpa.go
```diff
@@ -255,6 +255,8 @@ var (
 						}
 					} else if intVal, canFastConvert := value.AsInt64(); canFastConvert {
 						metricValue = float64(intVal)
+					} else if c.Type == autoscaling.ExternalMetricSourceType {
```
brancz: That would mean external metrics are always going to be specified in millis? Do we have docs for this?

LiliC: No, haven't found any docs yet, just made the assumption the same way we make the assumptions for other metrics. We could find if there is an `m` and only then do the calculation, otherwise leave the number as is?

tariq1890: Are there other known representations of `resource.Quantity`? Other than plain intValues and `m`.

LiliC: Yes, there are many:

https://github.com/kubernetes/kube-state-metrics/blob/6b5f9ef06c61a94a27a93110fba7bc8ea84678e8/vendor/k8s.io/apimachinery/pkg/api/resource/quantity.go#L34-L48

LiliC: Update according to some k8s folks I asked, we should be using the ScalledValue, full msg:
> calling ScaledValue(0) - so first you get it unscaled, detect the order and see if scaling it will fit int64

Will give that a try.

lilic:
/close

Will come back to this at a later point.

PR #1028, opened 2020-01-18
Change metrics store to use sync.Map

### LINKED ISSUE #995: kube-state-metrics API scraping timeout

>
/kind bug
> /kind feature

**What happened**:
We are using kube-state-metrics (1.6.0) from prometheus-operator. Our cluster periodically creates hundreds of pods and deletes them within a 15 minutes window for bursting traffic. When this happens, kube-state-metrics target in Prometheus will often see API timeout (default 10 seconds). Suspecting this is due to the mutex locking mechanism used for `MetricsStore` that when heavy writing of new entries into the map happens it'll block reading for extended time. 

More consistent API response time to not cause Prometheus scraping timeout. 

Our cluster normally has 30k metrics in total, with roughly 25k metrics are from pod collectors (seems there are 27 pod_info_ metrics, we have close to 1000 pods). When the bursting traffic comes, the cluster create a few hundreds of pods to push total metrics to around 60k, most added in pod_info metrics. 

Response size doesn't seem to be an issue, with 60k metrics, the total response size is roughly 10MB. When the scraping call doesn't time out, it can response in 100 ~ 200ms. 

Single kube-state-metrics pod.

Tried modifying kube-state-metrics code by using a `sync.Map` for `MetricsStore` in hope of better concurrency, it seems relieving the problem a little bit, but when the flood comes in it can still take seconds for kube-state-metrics to respond, as `sync.Map` is really good for infrequently changed map entries but in our case there are always a ton of new entries being created. 

I think padding the gap in Prometheus with previous sample should be fine, not sure if it makes sense for kube-state-metrics to provide this behavior, something like keeping a copy of previous response, and if the collector reading takes too long simply using the copy.

**Environment**:
1.13
1.6.0

zhengl7:
The cache map for metrics cannot be concurrently accessed in the current version, so there is a mutex for map reading/writing. This caused a problem that heavy metrics writing to the map will cause a contention with reading, and can potentially cause Prometheus scrape call to timeout. Using sync.Map can remove the need of the mutex.

The non-concurrent map implementation led to issues reported in #995, there is constant Prometheus scraping timeouts when the cluster reaches certain busyness of creating new pods. Applied a version of kube-state-metrics with this change to the same cluster, timeouts are completely gone. 

Fixes #995


zhengl7:
Results of running `make test-benchmark-compare` locally,

```
### Result
old=release-1.8 new=syncmap_metrics_store
BenchmarkKubeStateMetrics/GenerateMetrics-8     1014742988      1005576671      -0.90%
BenchmarkKubeStateMetrics/MakeRequests-8        10689884448     21356121974     +99.78%
BenchmarkPodStore-8                             28217           31236           +10.70%
BenchmarkMetricWrite/value-1-8                  565             551             -2.48%
BenchmarkMetricWrite/value-35.7-8               704             687             -2.41%

BenchmarkKubeStateMetrics/MakeRequests-8     925.12       970.94       1.05x

BenchmarkKubeStateMetrics/GenerateMetrics-8     824843         864269         +4.78%
BenchmarkKubeStateMetrics/MakeRequests-8        410539         455643         +10.99%
BenchmarkPodStore-8                             488            523            +7.17%
BenchmarkMetricWrite/value-1-8                  7              7              +0.00%
BenchmarkMetricWrite/value-35.7-8               7              7              +0.00%

BenchmarkKubeStateMetrics/GenerateMetrics-8     83766392        87583568        +4.56%
BenchmarkKubeStateMetrics/MakeRequests-8        26515301456     52930885600     +99.62%
BenchmarkPodStore-8                             24304           25984           +6.91%
BenchmarkMetricWrite/value-1-8                  536             536             +0.00%
BenchmarkMetricWrite/value-35.7-8               536             536             +0.00%
```
 

```

zhengl7:
Force pushing after rebasing

zhengl7:
ok, tests failure, need to understand it better. Should I close the PR and come back later when it's ready?

LiliC:
hmm those do seem like very huge performance degradation. 🤔

PR #1030, opened 2020-01-21
internal/store/: Remove hpa metric kube_hpa_status_current_metrics_av…

LiliC:
…erage_value

due to unhandled edge cases and precision loss. Because we are not
handling nil cases correctly we often experience segfault in various edge
case scenarios.

cc @tariq1890 @brancz


brancz:
Looks like the metric wasn't entirely removed, as documentation CI failed expecting it to be documented.

LiliC:
@brancz PTAL thanks!

LiliC:
@brancz PTAL, thanks!

tariq1890:
Thanks @LiliC !

PR #1035, opened 2020-01-22
Merge 1.9.3 back into master

brancz:
Follow up from the release.

/cc @LiliC @tariq1890


tariq1890:
44 commits?

tariq1890:
Do we need to update the branch?

brancz:
I think we did a mistake when we merged the release branch previously. As the merge itself looks clean I think we should move ahead with this.

PR #1041, opened 2020-01-31
add annotation metrics and ability to exclude.

### LINKED ISSUE #941: annotation metrics, what is the future direction?

/kind feature

**What happened**:

We make extensive use of `kube_namespace_annotations` which has been removed. This PR https://github.com/kubernetes/kube-state-metrics/pull/859 says "rethink the annotation metrics in a separate PR".   Reading around the PRs and Issues, the main concern seems to be that annotations can be numerous/large and security concerns. Would it be acceptable to reinstate annotation metrics, but with them off by default and enabled in the startup args?   Making it optional would allow users that don't have security concerns and do not have a large number of annotations to use it. 

If metrics carrying annotations will not be returning, then we will have to find another way internally so that we can upgrade beyond v1.6.0. For example, write an annotations exporter to run alongside kube-state.  Before starting on that, i thought I would check to see if there are any plans/timescales that you can share.  

Thanks

**Environment**:
- Kubernetes version (use `kubectl version`): v1.15.3
- Kube-state-metrics image version: > v1.8.0+

jw-s:
* Provides a metric which exposes annotations for that particular API kind.
* Whitelist on all annotation related metrics, users will have to specify the annotation labels they would like to expose.
Fixes #941


pkg/options/options.go
```diff
@@ -88,6 +90,7 @@ func (o *Options) AddFlags() {
 	o.flags.Var(&o.Namespaces, "namespace", fmt.Sprintf("Comma-separated list of namespaces to be enabled. Defaults to %q", &DefaultNamespaces))
 	o.flags.Var(&o.MetricWhitelist, "metric-whitelist", "Comma-separated list of metrics to be exposed. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.")
 	o.flags.Var(&o.MetricBlacklist, "metric-blacklist", "Comma-separated list of metrics not to be enabled. This list comprises of exact metric names and/or regex patterns. The whitelist and blacklist are mutually exclusive.")
+	o.flags.Var(&o.AnnotationLabelWhitelist, "annotation-label-whitelist", "Comma-separated list of annotation labels to be exposed. This list comprises of exact annotation label names and/or regex patterns.")
```
LiliC: This needs a warning for users enabling this:

> Warning: annotations might include sensitive information and produce unbound high cardinality.

cc @brancz @tariq1890 sgty ^
```suggestion
	o.flags.Var(&o.AnnotationLabelWhitelist, "annotation-label-whitelist", "Comma-separated list of annotation labels to be exposed. This list comprises of exact annotation label names and/or regex patterns. Warning: annotations might include sensitive information and produce unbound high cardinality.")
```

jw-s: Hi @LiliC, I will add the metrics which are currently not documented and failing the tests.

jw-s: @lilic ready to review again :)

docs/certificatessigningrequest-metrics.md
```diff
@@ -6,3 +6,4 @@
 | kube_certificatesigningrequest_condition | Gauge | `certificatesigningrequest`=&lt;certificatesigningrequest-name&gt; <br> `condition`=&lt;approved\|denied&gt; | STABLE |
 | kube_certificatesigningrequest_labels | Gauge | `certificatesigningrequest`=&lt;certificatesigningrequest-name&gt;| STABLE |
 | kube_certificatesigningrequest_cert_length | Gauge | `certificatesigningrequest`=&lt;certificatesigningrequest-name&gt;| STABLE |
+| kube_certificatesigningrequest_annotations | Gauge | `certificatesigningrequest`=&lt;certificatesigningrequest-name&gt;| EXPERIMENTAL |
```
olivierlemasle: Why don't you document the labels for all new metrics?
E.g. for ConfigMaps, you documented label `annotation_CONFIGMAP_ANNOTATION`=&lt;CONFIGMAP_ANNOTATION&gt;, but for CertificateSigningRequest, you don't document all labels.

docs/persistentvolumeclaim-metrics.md
```diff
@@ -8,7 +8,7 @@
 | kube_persistentvolumeclaim_resource_requests_storage_bytes | Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; | STABLE |
 | kube_persistentvolumeclaim_status_condition | Gauge | `namespace` =&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `type`=&lt;persistentvolumeclaim-condition-type&gt; <br> `status`=&lt;true\|false\|unknown&gt;  | EXPERIMENTAL |
 | kube_persistentvolumeclaim_status_phase | Gauge | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `phase`=&lt;Pending\|Bound\|Lost&gt; | STABLE |
-
+| kube_persistentvolumeclaim_annotations | Gauge | `annotation_PERSISTENTVOLUMECLAIM_ANNOTATION`=&lt;PERSISTENTVOLUMECLAIM_ANNOTATION&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `namespace`=&lt;persistentvolumeclaim-namespace&gt; | EXPERIMENTAL |
 Note:
```
olivierlemasle: Missing an empty line between the table and the line _Note_.

jw-s:
> Two other remarks:
> 
> * the `--annotation-allowlist` is not intuitive to use because filtering is done on annotations already transformed to Prometheus labels.
>   For example, to see the annotation `volumes.kubernetes.io/controller-managed-attach-detach`, you need to pass `--annotation-allowlist="volumes_kubernetes_io_controller_managed_attach_detach"`. That's neither the real K8s annotation, nor the effective prometheus label (which is `annotation_volumes_kubernetes_io_controller_managed_attach_detach`).
> * When annotations are filtered out, we have this kind of results:
>   ```
>    # HELP kube_namespace_annotations Kubernetes annotations converted to Prometheus labels.
>    # TYPE kube_namespace_annotations gauge
>    kube_namespace_annotations 1
>    kube_namespace_annotations 1
>    kube_namespace_annotations 1
>    kube_namespace_annotations 1
>    kube_namespace_annotations 1
>    kube_namespace_annotations 1
>   ```
>   
>   
>   I guess we should have either the other labels (`namespace` in that case) or no metric at all...

Completely agree, wanted to discuss this with everyone before doing the implementation. I feel we should allow (whitelist) default annotations i.e namespace or should we allow all metrics default labels?

docs/certificatessigningrequest-metrics.md
jw-s: Good catch, this should be consistent.

lilic:
> Completely agree, wanted to discuss this with everyone before doing the implementation. I feel we should allow (whitelist) default annotations i.e namespace or should we allow all metrics default labels?

Having namespace by default included in the metric makes sense, users who don't want this can exclude this metric entirely, but not having the namespace label by default doesn't make the metric useful.

cc @brancz @tariq1890 any objections?

jw-s:
> > Completely agree, wanted to discuss this with everyone before doing the implementation. I feel we should allow (whitelist) default annotations i.e namespace or should we allow all metrics default labels?
> 
> Having namespace by default included in the metric makes sense, users who don't want this can exclude this metric entirely, but not having the namespace label by default doesn't make the metric useful.
> 
> cc @brancz @tariq1890 any objections?

Only problem with this approach is it will only work for namespace scoped objects , my feeling was we include all metrics default labels.

jw-s:
Just pushed the changes to allow each resource's Labels by default. What do you guys think? Will also make the cli argument name more intuitive.

lilic:
@tariq1890 @brancz please also have a look

internal/store/metriclabels.go
```diff
@@ -0,0 +1,40 @@
+package store
```
tariq1890: This needs a license header

tariq1890: Can we change the name to `default_labels.go`?

docs/cli-arguments.md
```diff
@@ -27,6 +27,7 @@ $ kube-state-metrics -h
 Usage of ./kube-state-metrics:
       --add_dir_header                   If true, adds the file directory to the header
       --alsologtostderr                  log to standard error as well as files
+      --annotation-allowlist string      Comma-separated list of annotation labels to be exposed. This list comprises of exact annotation label names and/or regex patterns.
```
tariq1890: I prefer leaving out `allowlist` as a part of the param name since there is allow/deny dichotomy here.

Perhaps something along the lines of `--annotation-filter`?

Thoughts @brancz @lilic ?

internal/store/endpoint_test.go
```diff
@@ -51,6 +53,10 @@ func TestEndpointStore(t *testing.T) {
 					Labels: map[string]string{
 						"app": "foobar",
 					},
+					Annotations: map[string]string{
+						"whitelisted":     "true",
```
tariq1890: The usage of whitelist and blacklist has some rather problematic undertones. As we are trying to be rid of them for the next major release, we should stick to `allowed/denied`.

tariq1890:
@jw-s Thank you so much for your hard work with this PR. It is really appreciated :). I've provided my first round of comments.

Sorry for the delay.

internal/store/endpoint_test.go
lilic: Agreed!

dgrisonnet:
@jw-s Thank you for the hard work!
I have been thinking about a way to do the same with `_labels` metrics recently (#1047) and tried to create a mechanism to be able to allow any kind of denied by default metrics.
I like the way you tackled the problem with annotations but I don't think we can apply the exact same solution to labels. Thus, I made a proposal to introduce a generic solution that could be applied to both annotations and labels (#1102) and I would love to have your feedback on it! :smiley:

jw-s:
I am going to look at this over the weekend and resolve conflicts and address your comments. Regarding #1102 do we continue with the way i've tackled the annotation allow/deny or wait until we have an outcome? We want a good user experience and i don't think we will get that if we're introducing a new cli flag here to then be removed shortly after... thoughts?

lilic:
>We want a good user experience and i don't think we will get that if we're introducing a new cli flag here to then be removed shortly after... thoughts?
Agreed. 

@jw-s I believe the #1102 discussion is over and a decision was made. We should proceed with that. I do suggest to not put too much into one PR, this already adds whitelist and annotation metrics and makes it hard to review. If you could or if @dgrisonnet wants to implement #1102. After that is merged we can come back to annotation metrics, hope that make sense? :)

jw-s:
@lilic I'm happy to get more involved in this project if needed. @dgrisonnet also  happy to implement if time doesn't permit on your side or pair or whatever 😄

jw-s:
And to answer, yeah makes sense! let's get #1102 merged in before continuing.

dgrisonnet:
@jw-s You can go ahead and tackle this issue. If you need any help feel free to ask, I will be more than happy to give you a hand. :)

mikonse:
Any update on this? Would be a huge help for us as we are currently stuck with an old kube-state-metrics version shipping with openshift and we're unable to upgrade due to the missing metrics.

lilic:
@jw-s are you willing to work on this, its okay if not, I will take over otherwise. Thanks! :)

dgrisonnet:
@jw-s @lilic If needed, I can also take over it.

jw-s:
@lilic @dgrisonnet Hi guys, sorry been busy with a few other things. Ready to start working on the feature gate approach 

> https://github.com/kubernetes/kube-state-metrics/issues/1102#issuecomment-603406929

@dgrisonnet If i understood correctly we would have a feature-gate per metric with an empty list as the value `--kube-namespace-annotations=[]` for example?

Do you have a particular plan to handle default labels in metrics? Do we want to provide a way to include these? A feature flag for whether or not to use default labels? `--include-default-labels=true|false`

dgrisonnet:
Hi @jw-s, thank you for taking care of this :). To answer your questions, I opened a proper issue #1124.

Note that, by concertation with @lilic we chose to descope the metrics feature gate part of the previous issue in order to create an issue more fitted to this PR particular needs.

If anything is unclear feel free to ask questions, I will be more than happy to answer them :).

jw-s:
@tariq1890 @lilic I guess I can continue on this, now that we have the ability to select which labels we want?

lilic:
@jw-s sounds good to me 👍 Thanks!

jw-s:
@lilic @tariq1890 How do we want to handle the annotation metrics and I guess also label metrics? Should we include the default labels in these? As currently we don't allow any labels on these metrics.

jw-s:
Also @lilic was hoping this could be in the initial 2.0 release.

lilic:
> Also @lilic was hoping this could be in the initial 2.0 release.

We can still add it to the 2.0 release, no worries, sorry for delay. Can you recap what its left to do here? Thanks!

jw-s:
@lilic We need to decide if we include default labels such as namespace and name of resource by default in these metrics. Though technically this can be a new feature. Currently as you know... no labels are allowed on label and annotation metrics.

lilic:
@jw-s I believe namespace is a good one, not sure about name we had an issue with this last time. @brancz @tariq1890 anything against including namespace as the default label here? This will avoid confusion I believe to new users.

jw-s:
@lilic I see the project already sets default labels of each metric so it's just a matter of reusing this and making a default labels list by merging all into one?

lilic:
@jw-s lets go with both name and namespace as default labels! Thanks! 🎉

jw-s:
> @jw-s lets go with both name and namespace as default labels! Thanks! 🎉

So just those labels and not reusing for example: https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/configmap.go#L34

lilic:
Yes if its possible to reuse them then yes! To make it clear we want to end up with the following by default:

> fooresource_labels{fooresource=“fooresourcename”, namespace=“fooresourcenamespace”}

Hope that makes more sense now?

lilic:
@jw-s is this ready for another review?

jw-s:
@lilic Yeah, though e2e tests are failing and no useful output in the logs, need to check why...

jw-s:
@lilic Sorry i didn't have time to check this. Would be nice to get this in before the cut. Any way i can test the e2e locally?

lilic:
Yes you should be able to run them with `make e2e` and running docker https://github.com/kubernetes/kube-state-metrics/tree/master/tests

lilic:
@jw-s could you possibly split the adding default labels into a separate PR and we can merge as bugfix in 2.0, as those should land in 2.0 release and annotation label metrics can land in a future release. Sounds good?

Mepemotrarial:
@jw-s Any progress on this?

PR #1043, opened 2020-02-03
internal/store/builder.go: Fix K8s API version

mjura:
(cherry picked from commit 7e6ee51012887301e9f62836566ecd1c18a146c9)

**What this PR does / why we need it**:

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes # https://github.com/kubernetes/kube-state-metrics/issues/1039

PR #1052, opened 2020-02-07
internal/store/validatingwebhookconfiguration.go: Switch to v1

lilic:
Initial fix was done https://github.com/kubernetes/kube-state-metrics/pull/1043/files but it seems like it didn't adjust all of it

As we are expecting v1 we should also create a watch on v1. Otherwise
see the following error in the logs:

> expected type *v1.ValidatingWebhookConfiguration, but watch event object had type *v1beta1.ValidatingWebhookConfiguration

cc @brancz @tariq1890


tariq1890:
Thanks @LiliC let’s go with v1 completely

tariq1890:
@lilic just following up on this :)

LiliC:
yeah tests failed looking at this.

lilic:
So strange, I ran this locally and did not seen any problems 🤔

lilic:
Figured out the problem, seems like we had some manifests v1beta1 and also old k8s cluster.

lilic:
@tariq1890 PTAL, tests passed now, thanks!

tests/e2e.sh
```diff
@@ -17,7 +17,7 @@
 set -e
 set -o pipefail
 
-KUBERNETES_VERSION=v1.15.3
+KUBERNETES_VERSION=v1.17.0
```
tariq1890:
```suggestion
KUBERNETES_VERSION=v1.17.3
```

tariq1890:
Looks good. Just one comment

tariq1890:
@lilic Also, don't forget to squash the commit history :)

lilic:
Done, PTAL @tariq1890

lawliet89:
On Kubernetes `1.15.8-gke.3`, after upgrading, I get endless `E0221 04:47:38.447162       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1.ValidatingWebhookConfiguration: the server could not find the requested resource`.

filip-havlicek:
@lawliet89 I have the same problem
it's because there is no admissionregistration.k8s.io/v1 in Kubernetes 1.15, there is only admissionregistration.k8s.io/v1beta1 (ValidatingWebhookConfiguration is part of admissionregistration.k8s.io api group)

lilic:
@lawliet89 @filip-havlicek you can disable that resource via args to kube-state-metrics or switch to using older version. We will be updating our compatibility matrix to reflect this, thanks for reporting.

Note that this change did not introduce this switch, this happened in an earlier PR, but only partially. So this just fixed that bug.

PR #1053, opened 2020-02-07
pod: add gauges for pod overhead

egernst:

This adds two metrics for observability of the PodOverhead feature:
kube_pod_overhead_mcpu_cores
kube_pod_overhead_memory_bytes

Using these metrics allows end-users to quickly verify that the PodOverhead feature is being utilized.

When combined with the kube_pod_status_phase, end-users can use tools like grafana to quickly assess how pods running with overhead are behaving compared to non-overhead pods.

This addresses https://github.com/kubernetes/kubernetes/issues/87259

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -21,6 +21,7 @@
 | kube_pod_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
 | kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
 | kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
+| kube_pod_overhead | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
 | kube_pod_created | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_deleted | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
 | kube_pod_restart_policy | Gauge | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `type`=&lt;Always|Never|OnFailure&gt; | STABLE |
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -991,6 +991,41 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_overhead",
+			Type: metric.Gauge,
+			Help: "The pod overhead associated with running a pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if p.Spec.Overhead != nil {
+					for resourceName, val := range p.Spec.Overhead {
+						switch resourceName {
+						case v1.ResourceCPU:
+							ms = append(ms, &metric.Metric{
+								LabelValues: []string{sanitizeLabelName(string(resourceName)), string(constant.UnitCore)},
+								Value:       float64(val.MilliValue()) / 1000,
+							})
+
+						case v1.ResourceMemory:
+							ms = append(ms, &metric.Metric{
+								LabelValues: []string{sanitizeLabelName(string(resourceName)), string(constant.UnitByte)},
+								Value:       float64(val.Value()),
+							})
+						}
+					}
+
+				}
+
+				for _, metric := range ms {
+					metric.LabelKeys = []string{"resource", "unit"}
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
 	}
 )
 
--- a/internal/store/pod_test.go
+++ b/internal/store/pod_test.go
@@ -1605,7 +1605,7 @@ func BenchmarkPodStore(b *testing.B) {
 		},
 	}
 
-	expectedFamilies := 37
+	expectedFamilies := 38
 	for n := 0; n < b.N; n++ {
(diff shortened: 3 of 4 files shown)


internal/store/pod.go
```diff
@@ -1034,6 +1034,50 @@ var (
...
+					return &metric.Family{
+						Metrics: ms,
+					}
+				}
+			}),
+		},
+		{
+			Name: "kube_pod_overhead_memory_bytes",
+			Type: metric.Gauge,
+			Help: "The pod overead in bytes of memory for running a pod.",
```
dashpole: s/overead/overhead

internal/store/pod.go
```diff
@@ -1034,6 +1034,50 @@ var (
...
+			Help: "The pod overhead in mCPU cores for running a pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if p.Spec.Overhead != nil {
+					if cpu, ok := p.Spec.Overhead[v1.ResourceCPU]; ok {
+						ms = append(ms, &metric.Metric{
+							LabelKeys:   []string{"pod", "node"},
+							LabelValues: []string{p.Name, p.Spec.NodeName},
+							Value:       cpu.MilliValue(),
```
dashpole: it is probably best to stay consistent with other CPU metrics if we can, and report it in cores, rather than millicores.  Does that end up reporting a double?  Or does it report an int (which is presumably floored).

internal/store/pod.go
```diff
@@ -1034,6 +1034,50 @@ var (
...
+			Name: "kube_pod_overhead_memory_bytes",
+			Type: metric.Gauge,
+			Help: "The pod overead in bytes of memory for running a pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if p.Spec.Overhead != nil {
+					if mem, ok := p.Spec.Overhead[v1.ResourceMemory]; ok {
+						ms = append(ms, &metric.Metric{
+							LabelKeys:   []string{"container", "node"},
```
dashpole: I think you want to use the "pod" label here.

internal/store/pod.go
```diff
@@ -1034,6 +1034,50 @@ var (
...
+							LabelValues: []string{c.Name, p.Spec.NodeName},
```
dashpole: I don't think "c" is defined here.  I think you are looking for `p.Name`.

internal/store/pod.go
egernst: Yes, yikes -- didn't commit the latest changes before pushing.  Thanks.

internal/store/pod.go
egernst: It returns a double -- I updated to be consistent.

egernst:
@dashpole - thanks for the (quick) feedback.  Updated - PTAL

tariq1890:
Hello :), Thanks for your PR.

Also, I think only ksm designated code-reviewers should `/lgtm` or `/approve`.

/lgtm cancel

egernst:
@tariq1890 - you're welcome.  Please let me know if you have feedback; it'd be very useful to have this in to help users of the PodOvehead feature.  Thanks!

internal/store/pod.go
```diff
@@ -986,6 +986,42 @@ var (
...
+								LabelValues: []string{p.Name, p.Spec.NodeName, sanitizeLabelName(string(resourceName)), string(constant.UnitByte)},
+								Value:       float64(val.Value()),
+							})
+						}
+					}
+
+				}
+
+				for _, metric := range ms {
+					metric.LabelKeys = []string{"pod", "node", "resource", "unit"}
```
tariq1890: `pod` is a default label that will be added to all pod metrics.
```suggestion
					metric.LabelKeys = []string{"node", "resource", "unit"}
```

tariq1890: Please remove the instances of pod label values in your new metric as well.

internal/store/pod.go
```diff
@@ -986,6 +986,42 @@ var (
...
+					for resourceName, val := range p.Spec.Overhead {
+
+						switch resourceName {
+						case v1.ResourceCPU:
+							ms = append(ms, &metric.Metric{
+								LabelValues: []string{p.Name, p.Spec.NodeName, sanitizeLabelName(string(resourceName)), string(constant.UnitCore)},
+								Value:       float64(val.MilliValue()) / 1000,
+							})
+
+						case v1.ResourceMemory:
```
tariq1890: Let's handle the cases for Storage and Ephemeral storage as well
```
						case v1.ResourceStorage:
							fallthrough
						case v1.ResourceEphemeralStorage:
							fallthrough
```

As seen in the `kube_pod_container_resource_limits` metric.

internal/store/pod.go
```diff
@@ -986,6 +986,42 @@ var (
...
+		{
+			Name: "kube_pod_overhead",
+			Type: metric.Gauge,
+			Help: "The pod overhead associated with running a pod.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if p.Spec.Overhead != nil {
+					for resourceName, val := range p.Spec.Overhead {
+
```
tariq1890: extraneous new line

docs/pod-metrics.md
```diff
@@ -21,6 +21,7 @@
 | kube_pod_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
 | kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
 | kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
+| kube_pod_overhead | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | EXPERIMENTAL |
```
tariq1890: Let's drop the node metric

tariq1890:
Please update the expected list of metrics in the [unit test](https://github.com/kubernetes/kube-state-metrics/blob/master/main_test.go#L158)

tariq1890:
Benchmark test in `pod_test.go` needs to be updated as well.

internal/store/pod.go
egernst: I don't expect these to be relevant in the context of PodOverhead, since these are usually resources actually utilized by the workload itself. I don't foresee a scenario/use-case where this would be used, but could add "in case"...

internal/store/pod.go
egernst: ACK - sounds good.

docs/pod-metrics.md
egernst: ACK

tariq1890:
/lgtm

Thanks once again for you contribution @egernst :)!

PR #1006, opened 2019-12-14
v2: Rename collector to resource

Alexander-Carpenter:
**What this PR does / why we need it**: Rename collector to resource since the collet type pattern is no longer used.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #980

PR #1059, opened 2020-02-14
docs: --resources replaced with --collectors in the CLI

stigok:
This is related to #915

It corrects invalid arguments used in the docs.

### DIFF
--- a/README.md
+++ b/README.md
@@ -259,7 +259,7 @@ subjects:
     namespace: your-namespace-where-kube-state-metrics-will-deployed
 ```
 
-- then specify a set of namespaces (using the `--namespace` option) and a set of kubernetes objects (using the `--resources`) that your serviceaccount has access to in the `kube-state-metrics` deployment configuration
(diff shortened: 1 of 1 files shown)


tariq1890:
Thanks @stigok!

olivierlemasle:
@tariq1890 PR #1080 and its twin #915 made me look again at this PR...
It's should be the opposite, right? Collectors have been renamed to resources with #1006, and documentation had already been updated to reflect this change. This PR actually reverts documentation to the old name.

tariq1890:
Oh yes, thanks so much for catching this mistake @olivierlemasle. I will make the corrections.

stigok:
@olivierlemasle lol. Now I understand. git blame sent me here https://github.com/kubernetes/kube-state-metrics/pull/1006

So how did I come about creating this PR? It was indeed because `--resources` didn't work, but `--collectors` did. My confusion stems from you merging v2 changes into master without bumping version. It could be nice to put up a warning about master being used to builld v2 in the readme :)

tariq1890:
@stigok That’s a fair ask. We should update the README about breaking changes in master

PR #1073, opened 2020-02-21
Fix example PromQL queries related to Pod status

### LINKED ISSUE #1063: Documentation on pod status queries is incorrect
Example PromQL queries relative to pod status have been added to Pod metrics documentation [here](https://github.com/kubernetes/kube-state-metrics/blob/c1895cd/docs/pod-metrics.md#useful-metrics-queries), as part of #1013.

However, these queries seem to be incorrect.

For example, `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod)` is always 1, even for pods in phases e.g. PENDING, because in that case the value of `kube_pod_status_phase{phase="Running"}` is 0, but the count is 1.

That's why, when I test `count(kube_pod_status_phase{phase="Running"}) by (namespace, pod) * count(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod)`, I get **all** pods, not PENDING ones.

NB: I've also proposed typo fixes (including on PromQL) in that document with PR #1062

cc @clamoriniere

/kind documentation
/kind bug

olivierlemasle:

This PR fixes the example PromQL queries in Pod metrics documentation, used to mimic `kubectl` logic on pod statuses.

Fixes #1063

cc @clamoriniere

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -47,13 +47,13 @@
 
 It is not straightforward to get the Pod states for certain cases like "Terminating" and "Unknown" since it is not stored behind a field in the `Pod.Status`.
 
-So to get them, you will need to compose multiple metrics (like it is done in the `kubectl` command line code).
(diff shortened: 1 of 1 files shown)


brancz:
Very awesome!

Thank you! :tada: 

/lgtm

clamoriniere:
Thank you! 🥇

PR #1087, opened 2020-03-06
Added kube_pod_status_condition and kube_pod_status_condition_last_transition_time metric

vinicyusmacedo:

- ~~Provides `kube_pod_status_ready_time`, which tells when a pod has become Ready, allowing us to see how much time it took to spin up and accept connections.~~
- Provides kube_pod_status_condition, which shows a pod's conditions. 
- Provides kube_pod_status_condition_last_transition_time, which shows the transition time of each condition.

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -91,6 +91,8 @@ func podMetricFamilies(allowLabelsList []string) []generator.FamilyGenerator {
 		createPodStatusScheduledFamilyGenerator(),
 		createPodStatusScheduledTimeFamilyGenerator(),
 		createPodStatusUnschedulableFamilyGenerator(),
+		createPodStatusConditionFamilyGenerator(),
+		createPodStatusConditionLastTransitionTimeFamilyGenerator(),
 	}
 }
 
@@ -1406,6 +1408,66 @@ func createPodStatusPhaseFamilyGenerator() generator.FamilyGenerator {
 					LabelKeys:   []string{"phase"},
 					LabelValues: []string{p.n},
 					Value:       boolFloat64(p.v),
+				}
+			}
+
+			return &metric.Family{
+				Metrics: ms,
+			}
+		}),
+	)
+}
+
+func createPodStatusConditionFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGenerator(
+		"kube_pod_status_condition",
+		"The pods conditions.",
+		metric.Gauge,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			conditions := p.Status.Conditions
+
+			ms := make([]*metric.Metric, len(conditions))
+
+			for i, condition := range conditions {
+				conditionType := string(condition.Type)
+				conditionStatusStr := string(condition.Status)
+				conditionStatus, err := strconv.ParseBool(conditionStatusStr)
+				if err != nil {
+					return nil
+				}
+				ms[i] = &metric.Metric{
+
+					LabelKeys:   []string{"condition_type"},
+					LabelValues: []string{conditionType},
+					Value:       boolFloat64(conditionStatus),
+				}
+			}
+
+			return &metric.Family{
+				Metrics: ms,
+			}
+		}),
+	)
+}
+
+func createPodStatusConditionLastTransitionTimeFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGenerator(
+		"kube_pod_status_condition_last_transition_time",
+		"The last transition time of each pod condition.",
+		metric.Gauge,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			conditions := p.Status.Conditions
+
+			ms := make([]*metric.Metric, len(conditions))
+
+			for i, condition := range conditions {
+				ms[i] = &metric.Metric{
+
+					LabelKeys:   []string{"condition_type"},
+					LabelValues: []string{string(condition.Type)},
+					Value:       float64((condition.LastTransitionTime).Unix()),
 				}
 			}
 
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -91,7 +91,6 @@ func podMetricFamilies(allowLabelsList []string) []generator.FamilyGenerator {
 		createPodStatusScheduledFamilyGenerator(),
 		createPodStatusScheduledTimeFamilyGenerator(),
 		createPodStatusUnschedulableFamilyGenerator(),
-		createPodStatusConditionFamilyGenerator(),
 		createPodStatusConditionLastTransitionTimeFamilyGenerator(),
 	}
 }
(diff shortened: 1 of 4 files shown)


internal/store/pod.go
```diff
@@ -87,6 +87,34 @@ var (
...
+			Type: metric.Gauge,
+			Help: "Unix timestamp for the moment a pod becomes ready.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				var readyTime *metav1.Time
+
+				for _, condition := range p.Status.Conditions {
+					if condition.Type == "Ready" {
+						readyTime = &condition.LastTransitionTime
```
lilic: Hmm this is very approximate value to me, as we don't directly map like we do in start time metric where we can directly look at the value of StartTime. I would if we want to introduce this mark this metric as experimental. 

cc @brancz @tariq1890

internal/store/pod_test.go
```diff
@@ -842,6 +844,13 @@ kube_pod_container_status_last_terminated_reason{container="container7",namespac
 					HostIP:    "1.1.1.1",
 					PodIP:     "1.2.3.4",
 					StartTime: &metav1StartTime,
+					Conditions: []v1.PodCondition{
+						{
+							Type:               "Ready",
+							Status:             "True",
+							LastTransitionTime: metav1ReadyTime,
```
lilic: Can we add a test case when condition is not ready and set the status and LastTransitionTime. e.g. when the metric should not be set then.

vinicyusmacedo: Sure, I can do it.

internal/store/pod.go
```diff
@@ -87,6 +87,34 @@ var (
...
+			Name: "kube_pod_ready_time",
+			Type: metric.Gauge,
+			Help: "Unix timestamp for the moment a pod becomes ready.",
+			GenerateFunc: wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				var readyTime *metav1.Time
+
+				for _, condition := range p.Status.Conditions {
+					if condition.Type == "Ready" {
```
vinicyusmacedo: I forgot to check the condition status. The type might report as ready, but the pod might be not ready (thus reporting the status "False")

internal/store/pod.go
vinicyusmacedo: @lilic should I just add the EXPERIMENTAL flag to the docs? I also wouldn't trust this metric so much right now.

lilic: Yes go ahead, add EXPERIMENTAL, still waiting for one of the other two maintainers if they agree on adding this.

tariq1890: So the issue here is that we are adding custom logic to create a metric. We are trying to the stick to the pattern where kube-state-metrics acts as a mirror of the Kubernetes API objects. Any special logic that needs to be executed should be done via promQL. 

In light of this, I am leaning towards creating a metric like `kube_pod_status_condition_last_transition_time`. This could include the pod status condition type as a label.

vinicyusmacedo: That could be even better, making the kube-state-metrics side simpler. I'll rewrite it later today.

vinicyusmacedo: @tariq1890 I removed the other metric and added `kube_pod_status_condition` and `kube_pod_status_condition_last_transition_time`.

brancz:
Pod metrics already have the highest cardinality in kube-state-metrics, this adds 2 types of metrics which each have cardinality in themselves. I don't think this is going to work out well. Can you maybe elaborate on the concrete use case you have with these?

vinicyusmacedo:
@brancz about `kube_pod_status_condition_last_transition_time`: say you want to see how much does it take for a pod to become ready. You may want to do a delta between `kube_pod_start_time` and `kube_pod_status_condition_last_transition_time`. This may be useful for identifying node issues (like the Docker daemon taking too long to respond) or measuring the impact of image caching.

About `kube_pod_status_condition`: I really just created it because I was working on `kube_pod_status_condition_last_transition_time` and found that maybe it made sense to create a `kube_pod_status_condition` that doesn't return a timestamp, but simply if a pod had some transition (I may remove this one if it doesn't seem useful).

jlhuilier-1a:
Looking at the inital need (having the status ready transition time) why don't you create a new gauge kube_pod_status_ready_last_transition_time reflecting the last transition time of the status ready?
It seem more in line with how the other status are already exported.

vinicyusmacedo:
@jlhuilier-1a I was considering the approach of @tariq1890 to push all the extra logic to the promQL. By having a `kube_pod_status_condition_last_transition_time` instead of `kube_pod_status_ready_last_transition_time`, we would avoid doing many checks within kube-state-metrics (for example, looping through all the conditions to get the one with status Ready) and it would be easier to get other conditions as well (Scheduled, Initialized, ContainersReady)

jlhuilier-1a:
Yes for sure it simplifies the collection part but my feeling is that it is better to store and use simple values (As already done for Scheduled status). Simple loop/if logic should be fine in kube-state-metrics as long as we keep reflecting values coming from etcd. 
Let's wait for feedback of the maintainers.

lilic:
> @jlhuilier-1a I was considering the approach of @tariq1890 to push all the extra logic to the promQL. By having a kube_pod_status_condition_last_transition_time instead of kube_pod_status_ready_last_transition_time, we would avoid doing many checks within kube-state-metrics 

Yes, that is the approach we want in kube-state-metrics, if you can do it via promql or recording rules, lets keep it simple then.

jlhuilier-1a:
> > @jlhuilier-1a I was considering the approach of @tariq1890 to push all the extra logic to the promQL. By having a kube_pod_status_condition_last_transition_time instead of kube_pod_status_ready_last_transition_time, we would avoid doing many checks within kube-state-metrics
> 
> Yes, that is the approach we want in kube-state-metrics, if you can do it via promql or recording rules, lets keep it simple then.

Ok so basically what you mean is that the existing status metrics ([kube_pod_status_ready](https://github.com/kubernetes/kube-state-metrics/blob/1d132e7e728face5ba35be3f3310139c98f3a001/internal/store/pod.go#L325),[kube_pod_status_scheduled](https://github.com/kubernetes/kube-state-metrics/blob/1d132e7e728face5ba35be3f3310139c98f3a001/internal/store/pod.go#L350) and associated times) will be decommissioned and replaced by this new all in one kube_pod_status_condition ?

vinicyusmacedo:
Hey @lilic @jlhuilier-1a, are you still considering this PR?

internal/store/pod.go
```diff
@@ -343,6 +343,58 @@ var (
 				}
 			}),
 		},
+		{
+			Name: "kube_pod_status_condition",
```
lilic, line 347: So we already have kube_pod_status_unschedulable, kube_pod_status_ready, kube_pod_status_scheduled. What other conditions would there be? This would be duplicating those metrics no?

jlhuilier-1a: That was my point, the initial need was adding a new gauge kube_pod_status_ready_time with the exact same code than the existing [kube_pod_status_scheduled_time](https://github.com/kubernetes/kube-state-metrics/blob/1d132e7e728face5ba35be3f3310139c98f3a001/internal/store/pod.go#L236).

jlhuilier-1a: Reworking the overall way status are exposed could definitely be discussed but I think it is out of scope of this PR.

jlhuilier-1a:
> Not sure if the kube_pod_status_condition is needed as per my comment, sorry for not catching sooner.

That was my point, the initial need was adding a new gauge kube_pod_status_ready_time with the exact same code than the existing [kube_pod_status_scheduled_time](https://github.com/kubernetes/kube-state-metrics/blob/1d132e7e728face5ba35be3f3310139c98f3a001/internal/store/pod.go#L236).

Reworking the overall way status are exposed could definitely be discussed but I think it is out of scope of this PR.

jlhuilier-1a:
@lilic @brancz @tariq1890 It seem they was a kind of misunderstanding during the analysis of this PR. It was initially just adding a new gauge kube_pod_status_ready_time reusing existing logic and after discussions started to move to reworking the way status conditions are exposed. 

Do we all agree that for the initial need we should go back to the simple (in line with existing status) implementation?

lilic:
Agreed, we can do one or the other but not both. I would say since the pod metrics are the most used out of all of kube-state-metrics, I would rather add new condition and not remote/rework the existing ones. cc @tariq1890 @brancz wdyt?

lilic:
@vinicyusmacedo do you mind having a look at my PR comment, thanks!

vinicyusmacedo:
Hey there, sorry for not catching up. @lilic will take a look at the comments and rebase.

internal/store/pod.go
vinicyusmacedo: @lilic yes, the scope of this PR was basically adding `kube_pod_status_condition_last_transition_time`.

You may be interested in `ContainersReady` and `Initialized`, but these can be done the same way as `kube_pod_status_ready` and I think it's out of scope. I will remove `kube_pod_status_condition`.

PR #1091, opened 2020-03-09
Controller labels

gnosek:

This PR adds `kube_*_controller` metrics. They are similar to `kube_*_owner` (effectively expose the same data) except for two crucial differences:

1. They only report the controller owner
2. The owner kind is returned in the label name, not in its own owner_kind label. This makes PromQL joins significantly easier.

Example: get the sum of memory limits for all deployments with the label app=nginx

**`owner` labels**

You need to use `label_replace` to make the label names line up to join on them:

```
sum(
kube_pod_container_resource_limits_memory_bytes and on(namespace, pod)
(label_replace(kube_pod_owner{owner_kind="ReplicaSet"}, "replicaset", "$1", "owner_name", "(.+)") and on (namespace, replicaset) 
(label_replace(kube_replicaset_owner{owner_kind="Deployment"}, "deployment", "$1", "owner_name", "(.+)") and on (namespace, deployment) 
kube_deployment_labels{label_app="nginx"})))
```

**`controller` labels**

No relabelling needed:

```
sum(
kube_pod_container_resource_limits_memory_bytes and on(namespace, pod) 
(kube_pod_controller and on (namespace, replicaset)
(kube_replicaset_controller and on (namespace, deployment) 
kube_deployment_labels{label_app="nginx"})))
```

### DIFF
--- a/docs/job-metrics.md
+++ b/docs/job-metrics.md
@@ -5,6 +5,7 @@
 | kube_job_info | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_labels | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `label_JOB_LABEL`=&lt;JOB_LABEL&gt;  | STABLE |
 | kube_job_owner | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `owner_kind`=&lt;owner kind&gt; <br> `owner_name`=&lt;owner name&gt; <br> `owner_is_controller`=&lt;whether owner is controller&gt;  | STABLE |
+| kube_job_controller | Gauge | `job`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> &lt;lowercase owner kind&gt;=&lt;owner name&gt; | STABLE |
 | kube_job_spec_parallelism | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_spec_completions | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_spec_active_deadline_seconds | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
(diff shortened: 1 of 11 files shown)


gnosek:
I expected we'd want to keep *_owner for backward compatibility reasons. If you prefer that we dropped *_owner (e.g. if the 2.0 release is backwards incompatible anyway), I can do that as well.

The only assumption made is that we limit ourselves to report the controlling owner, not the other (non-controller) ones. Apart from that, it's exactly the same data as in *_owner and I would argue it's just as direct but easier to query. I suppose I could make a time series for each owner and add a label like owner_is_controller=0|1. Then, the only discrepancy between *_owner and *_controller would be the sanitization of owner kind names.

How does that sound?

tariq1890:
Hi @gnosek . Thanks for this PR and for explaining your use case; it definitely makes sense. However, my main concern is that we are heavily deviating from the "k8s API to metrics" idiom here. I am not entirely sure how I feel about adding an object field value as a label key( the value of `owner.Kind` in this case).

In our transition to kube-state-metrics 2.0, we are trying to adhere more to kube-state-metrics's role as a 1:1 exporter of Kubernetes API Objects.

brian-brazil:
> However, my main concern is that we are heavily deviating from the "k8s API to metrics" idiom here. 

The problem is that idiom doesn't work when you try to do joins in PromQL, metrics shouldn't merely be a dump of a database but have meaningful labelnames too. This PR is still translating the API to metrics, but in a much more useful way in terms of labels. I'd actually suggest going further and see if there's anywhere else in ksm that metrics could also be changed to be more like what this PR does.

brancz:
This is more about preferring consistency over unmaintainable differences in opinions, as many different ones exist. We've experienced such things in the past with kube-state-metrics which is why we did adopted this scheme in kube-state-metrics.

@brian-brazil you yourself like to argue with "it's hard but possible", and that exactly what this is. And the fact that projects like the kubernetes-mixin [clearly are able to work with this](https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/6f1b1e1e530a5547418e894039dc31d8a62c3132/rules/apps.libsonnet#L101-L154), shows that it's not a show stopper.

I'm actually trying to understand what the problem is here, because it was shown in the original message already that it is already possible to do what is needed with PromQL, just requiring 2 `label_replace` calls more than with this approach. Why is that a problem?

brian-brazil:
> Why is that a problem?

Any use of label_replace is a smell, and in this case there's a reasonable and better way to represent the data.

lilic:
> Any use of label_replace is a smell, and in this case there's a reasonable and better way to represent the data.

These were choices made when kube-state-metrics first started and this is how we continue as we have a lot of metrics that we need to maintain. Changes in the k8s api are frequent and adopting to them makes this much easier. As Frederic said and its noted in the description of this PR a workaround exists, even if it is a "smell" as you say :) Is there any reason you can't use that? 
See docs -> https://github.com/kubernetes/kube-state-metrics#overview

gnosek:
@lilic, we want to expose PromQL to our end users (technical but not necessarily experts) and having to figure out extra `label_replace` calls does make a difference in the expertise required to write the queries.

Recording rules with specific joins don't really work for us either as we cannot enumerate all possible queries in advance.

lilic:
@gnosek thanks for explaining! Would these that frederic linked to would not be enough, they are for us at least, we haven't had any problems or further requests so far:
https://github.com/kubernetes-monitoring/kubernetes-mixin/blob/6f1b1e1e530a5547418e894039dc31d8a62c3132/rules/apps.libsonnet#L101-L154

gnosek:
@lilic, not really, we cannot enumerate all the queries that are possible with our current system and PromQL will only enable more. We cannot use a static solution.

To clarify, we're not using KSM to monitor our own infrastructure but provide the building blocks for our end users to do so (currently using a proprietary tool in place of KSM but looking to switch).

Issue #1090, opened 2020-03-09
Wrong Version of admissionregistration API for older k8s versions


AlexMichaelJonesNC:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
 /kind bug
> /kind feature

**What happened**:
Deploying Kube State Metrics 1.9.5 into a 1.14 k8s cluster, we get logs similar to the following:
```
E0309 14:57:02.411880       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1.ValidatingWebhookConfiguration: the server could not find the requested resource
E0309 14:57:03.413702       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1.ValidatingWebhookConfiguration: the server could not find the requested resource
E0309 14:57:04.415417       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1.ValidatingWebhookConfiguration: the server could not find the requested resource
```
**What you expected to happen**:
Not get logs related to using v1 rather than v1beta1 for the admissionsregistrationAPI

**How to reproduce it (as minimally and precisely as possible)**:
Deploy kube-state-metrics into a 1.14 kubernetes cluster

**Anything else we need to know?**:
This seems to have been caused by https://github.com/kubernetes/kube-state-metrics/issues/1039
That patch changed the api to v1 to support new versions. I believe an appropiate fix would require dynamically selecting the api version based on kubernetes version. 

**Environment**:
- Kubernetes version (use `kubectl version`): 1.14
```
Server Version: version.Info{Major:"1", Minor:"14+", GitVersion:"v1.14.9-eks-502bfb", GitCommit:"502bfb383169b124d87848f89e17a04b9fc1f6f0", GitTreeState:"clean", BuildDate:"2020-02-07T01:31:02Z", GoVersion:"go1.12.12", Compiler:"gc", Platform:"linux/amd64"}
```
- Kube-state-metrics image version
1.9.5

lilic:
>  I believe an appropiate fix would require dynamically selecting the api version based on kubernetes version.

[...]

PR #1092, opened 2020-03-09
Update README.md

AlexMichaelJonesNC:
This update is to address https://github.com/kubernetes/kube-state-metrics/issues/1090. Since K8s 1.9 - 1.15 uses the v1beta1 version of admissionregistration, where as 1.16 and newer use v1 of the admissionregistration API, the 1.9.5 release of kube-state-metrics is not backwards compatible with kubernetes 1.15 or older. 

These versions are from https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#prerequisites

Fixes #


AlexMichaelJonesNC:
Sorry for the delay on that, it was after hours in my timezone, and I didn't get back around to looking at this until now. I've now signed it.

PR #1094, opened 2020-03-11
[e2e] Check errors in kube-state-metrics own metrics

### LINKED ISSUE #1084: Add test to check that no kube-state-metrics own alerts are firing in the e2e test
We have [alerts](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/prometheus-alerting-rules/alerts.yaml) relavant to kube-state-metrics, KubeStateMetricsListErrors and KubeStateMetricsWatchErrors, we need to add a test to our e2e tests that checks the results of the alerting expressions. The expression should be evaluated during the entire run of the e2e tests.

Adding this check in our e2e tests will allow us to make sure that when adding a new metric the resource we are trying to add a list/watch to are not erroring out.

/kind feature

olivierlemasle:

This PR adds e2e tests checking that:
- metrics `kube_state_metrics_list_total` and `kube_state_metrics_watch_total` are well present on `/metrics` on telemetry port;
- these metrics do not report any errors.

This PR also moves the "e2e framework" to its own package.

Fixes #1084

@lilic In #1084, you mentioned "The expression should be evaluated during the entire run of the e2e tests."
In this PR, I currently check telemetry metrics only once. Would you prefer if I add this check at the end of each other check for example?


tests/e2e/main_test.go
```diff
@@ -191,3 +193,53 @@ func getLabelsDocumentation() (map[string][]string, error) {
...
 	return documentedMetrics, nil
 }
+
+func TestTelemetryMetrics(t *testing.T) {
+	metricFamilies, err := framework.ParseMetrics(framework.KsmClient.TelemetryMetrics)
+	if err != nil {
+		t.Fatal("Failed to get or decode telemetry metrics", err)
+	}
+
+	// This map'keys are the metrics expected in kube-state-metrics telemetry.
```
tariq1890:
```suggestion
	// This map's keys are the metrics expected in kube-state-metrics' telemetry.
```

tests/e2e/main_test.go
```diff
@@ -191,3 +193,53 @@ func getLabelsDocumentation() (map[string][]string, error) {
...
+	// Its values are a boolean, which is set to true when the metric is found.
```
tariq1890:
```suggestion
	// Its values are boolean, which is set to true when the metric is found.
```

tariq1890:
Thanks @olivierlemasle . PR looks great. Very very minor comments from me.

tests/e2e/main_test.go
```diff
@@ -191,3 +193,53 @@ func getLabelsDocumentation() (map[string][]string, error) {
 	}
 	return documentedMetrics, nil
 }
+
+func TestTelemetryMetrics(t *testing.T) {
```
lilic: I would rename this to something like "TestKubeStateMetricsErrorMetrics", not exactly that but something close.

lilic: When is this test called, are we sure this is called right after everything is run? 🤔

olivierlemasle: @lilic No, it is not!
That's what I explained in the PR's first message, but I was perhaps not clear 😊. I was asking you if would prefer if I called this test after each other test...

olivierlemasle: Ok!

tests/e2e/main_test.go
olivierlemasle: Done

tests/e2e/main_test.go
lilic: Sorry missed that! At the end would be good, as this metrics only increases we should be able to see any error at the end. 

We seem to run the tests not at the very end, so I guess we need to run the go tests at the end or rethink the bash script? @tariq1890 
https://github.com/kubernetes/kube-state-metrics/blob/master/tests/e2e.sh#L180

olivierlemasle: Yes, we could move go tests execution later. However, as e2e tests are re-written to Go (for example with #950), the bash tests currently executed after the Go ones will disappear.

lilic: yep maybe we should merge that first, good point!

brancz:
Looks good but leaving last call up to @lilic. 

/lgtm
/hold

olivierlemasle:
@lilic Yes, of course!
Done.

tariq1890:
/lgtm

Thanks a lot @olivierlemasle :)

lilic:
Thanks!! 🎉

PR #1117, opened 2020-04-08
Fix various CI issues

brancz:

CI caught a violation of the label name being `nodeName` (camelCase) rather than snake_case, however, this is actually a case of another violation, which is the kubernetes instrumentation guideline says, that this label should have always been `node` to begin with, to allow for joining without label_replace. This metric is experimental anyways, so we're free to change this.

~We're not sure yet why PR CI didn't catch this.~

Travis CI not reporting seems to have magically fixed itself (possibly by either @lilic or myself logging in to Travis refreshing something?).

To get CI into a working state, I fixed the remaining issues as well.

@lilic @tariq1890


lilic:
> We're not sure yet why PR CI didn't catch this.

This was most likely due to an issue on travis side, things are back to normal now.

Prabhakar-17:
Thanks @brancz

PR #1119, opened 2020-04-10
fix e2e test run when the minikube_driver is not none

Prabhakar-17:
Fixes the e2e test run when the minikube_driver is not none.
This was introduced here: https://github.com/kubernetes/kube-state-metrics/pull/1117/files#diff-44bdf2bf033618abcd6345950e4ed565R71 see line 71
The reason last PR build (https://travis-ci.org/github/kubernetes/kube-state-metrics/jobs/672429529) was successful because it was only running for the case when minikube_driver is none see line 186 and 250 in the logs.

@lilic @tariq1890 @brancz It might be a good idea to cover both the cases while running the e2e tests in future so that we do not miss these cases?

Fixes #

### DIFF
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -68,7 +68,7 @@ touch "${HOME}"/.kube/config
 export KUBECONFIG=$HOME/.kube/config
 
 if [[ "$MINIKUBE_DRIVER" != "none" ]]; then 
-  export MINIKUBE_PROFILE_ARG="--profile ${MINIKUBE_PROFILE}"
+  export MINIKUBE_PROFILE_ARG=" --profile ${MINIKUBE_PROFILE}"
 else
   export MINIKUBE_PROFILE_ARG=''
 fi


lilic:
> might be a good idea to cover both the cases while running the e2e tests in future so that we do not miss these cases?

We can never run anything but none in CI, this is only for local development so we should not have another test case for this.

Prabhakar-17:
> > might be a good idea to cover both the cases while running the e2e tests in future so that we do not miss these cases?
> 
> We can never run anything but none in CI, this is only for local development so we should not have another test case for this.

Ok, got it. We will need this fix to be merged, right?

lilic:
/assign @brancz 

as he introduced those changes

brancz:
Where exactly is the absence of the space causing a problem? I tend to rather fix things in those places rather than in the variable.

> It might be a good idea to cover both the cases while running the e2e tests in future so that we do not miss these cases?

Our inability to do this is exactly why we are using the driver that we are in e2e tests :)

olivierlemasle:
@brancz  Actually, with 

```
export MINIKUBE_PROFILE_ARG="--profile ${MINIKUBE_PROFILE}"
${SUDO} minikube start --vm-driver="${MINIKUBE_DRIVER}" "${MINIKUBE_PROFILE_ARG}" --kubernetes-version=${KUBERNETES_VERSION} --logtostderr
```

there's indeed a failure:
```
./tests/e2e.sh 
minikube version: v1.8.2
commit: eb13446e786c9ef70cb0a9f85a633194e62396a1
Error: unknown flag: --profile ksm-e2e
See 'minikube start --help' for usage.
```

It works with `MINIKUBE_PROFILE_ARG=" --profile ${MINIKUBE_PROFILE}"` (with an extra space,  which is not very elegant, I agree).

It also works without the extra space if we remove the quotes: 

```
export MINIKUBE_PROFILE_ARG="--profile ${MINIKUBE_PROFILE}"
${SUDO} minikube start --vm-driver="${MINIKUBE_DRIVER}" ${MINIKUBE_PROFILE_ARG} --kubernetes-version=${KUBERNETES_VERSION} --logtostderr
```
... but you added these quotes (in f2cf68958d) because shellcheck complains without them:

```
In ./tests/e2e.sh line 76:
${SUDO} minikube start --vm-driver="${MINIKUBE_DRIVER}" ${MINIKUBE_PROFILE_ARG} --kubernetes-version=${KUBERNETES_VERSION} --logtostderr
                                                        ^---------------------^ SC2086: Double quote to prevent globbing and word splitting.


```

brancz:
Ok makes sense. I guess this is our best thing to do right now.

/lgtm

tests/e2e.sh
```diff
@@ -68,7 +68,7 @@ touch "${HOME}"/.kube/config
 export KUBECONFIG=$HOME/.kube/config
 
 if [[ "$MINIKUBE_DRIVER" != "none" ]]; then 
-  export MINIKUBE_PROFILE_ARG="--profile ${MINIKUBE_PROFILE}"
+  export MINIKUBE_PROFILE_ARG=" --profile ${MINIKUBE_PROFILE}"
```
dgrisonnet, line 71: I believe you can solve this issue by setting the variable the following way:
```suggestion
  export MINIKUBE_PROFILE_ARG="--profile=${MINIKUBE_PROFILE}"
```
What happened previously was that the variable was replaced literally by: `'--profile ksm-e2e'` since it contains space separated strings.

Note: adding a space before the profile option here will actually have a really different behaviour from what you expect. Bash will process it the following way: ```' --profile ksm-e2e'```. Since this option is not starting by either `-` or `--` minikube will from my understanding skip it. In order to test this behaviour, you can set `MINIKUBE_PROFILE_ARG` to `this option will be skipped`.

brancz: Maybe the best option then is to make the profile itself the variable and always set it? How would minikube react to an empty string for that flag?

dgrisonnet: I don't know, my guess would be that they replace it by the [default value](https://github.com/kubernetes/minikube/blob/4f704271d7f8802bba7ec6c72b80bba2e4d4f021/site/content/en/docs/commands/config.md#options-inherited-from-parent-commands-5). Otherwise, we can set it to its default value `minikube`. What do you think?

brancz: Sounds good to me. Could you do a PR to change the strategy to that?

dgrisonnet: Sure

PR #1120, opened 2020-04-10
bump kubernetes version in tests

### LINKED ISSUE #1118: Bump kubernetes version in tests to release v1.18
We are currently still testing against 1.17.x in CI, we should switch to 1.18.x by launching a kubernetes cluster 1.18 with minikube.

Prabhakar-17:
Bump kubernetes version to 1.8.0 in tests

Fixes #1118


tests/e2e.sh
```diff
@@ -17,7 +17,7 @@
 set -e
 set -o pipefail
 
-KUBERNETES_VERSION=v1.17.3
+KUBERNETES_VERSION=v1.18.0
```
lilic, line 20: There was a newer k8s version do you mind bumping to the latest here, thanks!

Prabhakar-17: Yes, sure. v1.18.1? the other releases are in alpha or beta stage.

lilic: Yes that is that what I meant.

README.md
```diff
@@ -66,13 +66,13 @@ All additional compatibility is only best effort, or happens to still/already be
 #### Compatibility matrix
 At most, 5 kube-state-metrics and 5 [kubernetes releases](https://github.com/kubernetes/kubernetes/releases) will be recorded below.
```
lilic, line 67: As it states here, we at most show 5, so we need to remove 1.13 from the list.

Prabhakar-17: ok, updated it. sorry, missed it.

tariq1890:
IMO, we need to update the kubernetes go dependencies as well before we change the k8s support matrix.

Prabhakar-17:
> IMO, we need to update the kubernetes go dependencies as well before we change the k8s support matrix.

You mean in the e2e.sh file, right? We have updated it to v1.8.1 please see line 20 in the diff.

tariq1890:
I mean the dependencies here: https://github.com/kubernetes/kube-state-metrics/blob/fb24c40b7dfd1401e284df2dbfd766994acafaa1/go.mod#L17

Prabhakar-17:
> I mean the dependencies here:
> 
> https://github.com/kubernetes/kube-state-metrics/blob/fb24c40b7dfd1401e284df2dbfd766994acafaa1/go.mod#L17

Yes, you are right, I think we should update it to v0.18.1
@lilic what do you think?

lilic:
Okay, we can then just bump the tested k8s version and not update the supported versions.

Prabhakar-17:
> Okay, we can then just bump the tested k8s version and not update the supported versions. I would bump k8s dep version in a separate PR.

Ok, so if understand correctly - 
1. We update the k8s version only in the e2e.sh file (which we have already did in this PR)
2. Reverse the compatibility matrix change from this PR?
3. Create a new issue to bump the k8s dep version and then update the compatibility matrix as well?

lilic:
> Create a new issue to bump the k8s dep version

I believe we have an issue, you assigned yourself to it :) 

Yes otherwise sgtm @tariq1890 any objections?

Prabhakar-17:
> > Create a new issue to bump the k8s dep version
> 
> I believe we have an issue, you assigned yourself to it :)
> 
> Yes otherwise sgtm @tariq1890 any objections?

Ok, I should have checked the issue. I will do it after this PR :)
I will wait for @tariq1890 to confirm.

tariq1890:
Sounds good to me :)

PR #1121, opened 2020-04-12
Fix `internal.Builder` to use `buildStoreFunc` in very buildXXXStore() method

clamoriniere:

This PR is fixing an issue introduced in a previous contribution when new Resources were added. 
To allows the usage of other "BuildStore" function, the method `Builder.buildStoreFunc` should be used everywhere instead of the using `builder.buildStore` directly.

Fixes #

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -249,11 +249,11 @@ func (b *Builder) buildNamespaceStore() cache.Store {
 }
 
 func (b *Builder) buildNetworkPolicyStore() cache.Store {
-	return b.buildStore(networkpolicyMetricFamilies, &networkingv1.NetworkPolicy{}, createNetworkPolicyListWatch)
+	return b.buildStoreFunc(networkpolicyMetricFamilies, &networkingv1.NetworkPolicy{}, createNetworkPolicyListWatch)
 }
 
 func (b *Builder) buildNodeStore() cache.Store {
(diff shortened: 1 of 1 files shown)


lilic:
We have been seeing some problems with Travis not reporting the status of CI, I restarted the job, it was green but guess was not reported back. If it doesn't merge guess you have to force push :/ 

thanks for the PR!

internal/store/builder.go
```diff
@@ -309,15 +309,15 @@ func (b *Builder) buildValidatingWebhookConfigurationStore() cache.Store {
...
+	return b.buildStoreFunc(volumeAttachmentMetricFamilies, &storagev1.VolumeAttachment{}, createVolumeAttachmentListWatch)
 }
 
 func (b *Builder) buildVPAStore() cache.Store {
-	return b.buildStore(vpaMetricFamilies, &vpaautoscaling.VerticalPodAutoscaler{}, createVPAListWatchFunc(b.vpaClient))
+	return b.buildStoreFunc(vpaMetricFamilies, &vpaautoscaling.VerticalPodAutoscaler{}, createVPAListWatchFunc(b.vpaClient))
 }
 
 func (b *Builder) buildLeases() cache.Store {
-	return b.buildStore(leaseMetricFamilies, &coordinationv1.Lease{}, createLeaseListWatch)
```
lilic, line 320: Do we use buildStore anywhere else or can it be removed?

PR #1122, opened 2020-04-14
Fix minikube profile in e2e tests

dgrisonnet:
This PR is a follow-up to the following discussion: https://github.com/kubernetes/kube-state-metrics/pull/1119#discussion_r408271195

cc @brancz

### DIFF
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -67,15 +67,16 @@ touch "${HOME}"/.kube/config
 
 export KUBECONFIG=$HOME/.kube/config
 
-if [[ "$MINIKUBE_DRIVER" != "none" ]]; then 
-  export MINIKUBE_PROFILE_ARG=" --profile ${MINIKUBE_PROFILE}"
-else
-  export MINIKUBE_PROFILE_ARG=''
+# use profile default value if driver is none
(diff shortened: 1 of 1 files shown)


tests/e2e.sh
```diff
@@ -67,13 +67,14 @@ touch "${HOME}"/.kube/config
 
 export KUBECONFIG=$HOME/.kube/config
 
-if [[ "$MINIKUBE_DRIVER" != "none" ]]; then 
-  export MINIKUBE_PROFILE_ARG=" --profile ${MINIKUBE_PROFILE}"
+if [[ "$MINIKUBE_DRIVER" != "none" ]]; then
```
tariq1890, line 70: Let's set the default value to the variable first. This eliminates the need for the else statement.

Once the setting is done, we can then `export` the var as opposed to `export`ing in every conditional path

tariq1890:
Thanks for the PR @dgrisonnet :)

dgrisonnet:
@olivierlemasle good catch, I completely forgot about that. It should be fixed now :).

brancz:
Thanks a lot for the quick follow up! lgtm :tada: 

Also congrats on your first contribution to the Kubernetes project! :tada: :partying_face: 

/lgtm
/approve

PR #1125, opened 2020-04-18
apply boundaries to metrics and allow via flag what labels to include

### LINKED ISSUE #1124: Bound the number of labels of all metrics

/kind feature

This proposal is originating from this discussion #1102.

**Motivation:**

More and more users want to expose high cardinality metrics; however, it comes with great costs for the default configuration of kube-state-metrics.

#941
#1047

Moreover, some of these metrics (for example: Kubernetes labels or annotations) can generate an unbound amount of labels. Thus, it is possible for these metrics to produce an unlimited number of series leading to some severe security issues.

**Goals:**

Make it possible to **safely** register metrics having an unbound amount of labels.

**Non-Goals:**

Set the sane defaults

**Proposal:**

In order to safely register metrics having an unbound amount of labels, a feature gates mechanism where labels could be allowed per metrics can be used.

In order to bound the number of labels of all metrics, this gate will only allow labels instead of allowing and denying. The reason behind this is that a deny mechanism would mean that everything is allowed except what is specifically denied. In the case of metrics with an unbound number of labels that would still result in an unbound number of labels.

By default, sensitive metrics will have all their labels denied and it will be left to the users to specify which labels they want to use. This will allow to only expose metrics having a limited number of labels and thus prevent some security issues.

The gates will have the following pattern:

| Metric          | Default value |
|-----------------|---------------|
| "_labels$"      | []            |
| "_annotations$" | []            |
| others          | [*]           |

Users will then be able to allow specific labels the following way:

```
--labels-allow-list="kube_job_labels=[label1,label2],kube_node_labels=[label1]"
```

jw-s:

Make it possible to safely register metrics having an unbound amount of labels.
Fixes #1124

### DIFF
--- a/pkg/allow/allow_labels.go
+++ b/pkg/allow/allow_labels.go
@@ -1,0 +1,72 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package allow
+
+import (
+	"regexp"
+)
+
+var defaultMetricLabels = map[*regexp.Regexp][]string{
+	regexp.MustCompile(".*_labels$"):      {},
+	regexp.MustCompile(".*_annotations$"): {},
+}
+
+// Labels provide a way to allow only specified labels for metrics.
+// Falls back to default labels, if your metric doesn't have defaults
+// then all labels are allowed.
+type Labels map[string][]string
+
+// Allowed returns allowed metric keys and values for the metric.
+func (a Labels) Allowed(metric string, labels, values []string) ([]string, []string) {
+	allowedLabels, ok := a[metric]
+	if !ok {
+		var defaultsPresent bool
+		for metricReg, lbls := range defaultMetricLabels {
+			if metricReg.MatchString(metric) {
+				allowedLabels = append(allowedLabels, lbls...)
+				defaultsPresent = true
+			}
+		}
+
+		if !defaultsPresent {
+			return labels, values
+		}
+	}
+
+	var finalLabels, finalValues []string
+	labelSet := labelSet(allowedLabels)
+	for allowedLabel := range labelSet {
+		for i, label := range labels {
+			if label == allowedLabel {
+				finalLabels = append(finalLabels, label)
+				finalValues = append(finalValues, values[i])
+			}
+		}
+	}
+
+	return finalLabels, finalValues
+}
+
+func labelSet(lists ...[]string) map[string]interface{} {
+	m := make(map[string]interface{})
+	for _, list := range lists {
+		for _, e := range list {
+			m[e] = struct{}{}
+		}
+	}
+	return m
+}
--- a/pkg/metric_generator/generator.go
+++ b/pkg/metric_generator/generator.go
@@ -120,3 +120,30 @@ func FilterMetricFamilies(l allowDenyLister, families []FamilyGenerator) []Famil
 
 	return filtered
 }
+
+type allowLabels interface {
+	Allowed(metric string, labels, values []string) (allowedLabels, allowedValues []string)
+}
+
+// FilterMetricFamiliesLabels takes a AllowLabels and a slice of metric
+// families and returns a filtered slice with metric families with their functions adapted to include only allowed labels for that metric Family.
+func FilterMetricFamiliesLabels(allowLabels allowLabels, families []FamilyGenerator) []FamilyGenerator {
+	var filtered []FamilyGenerator
+	for _, f := range families {
+		generateFunc := func(familyGenerator FamilyGenerator) func(obj interface{}) *metric.Family {
+			return func(obj interface{}) *metric.Family {
+				metricFamily := familyGenerator.GenerateFunc(obj)
+
+				for _, m := range metricFamily.Metrics {
+					m.LabelKeys, m.LabelValues = allowLabels.Allowed(f.Name, m.LabelKeys, m.LabelValues)
+				}
+				return metricFamily
+			}
+		}(f)
+
+		f.GenerateFunc = generateFunc
+		filtered = append(filtered, f)
+	}
+
+	return filtered
+}
--- a/pkg/metric_generator/generator_test.go
+++ b/pkg/metric_generator/generator_test.go
@@ -1,0 +1,258 @@
+/*
+Copyright 2020 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package generator
+
+import (
+	"reflect"
+	"testing"
+
+	"k8s.io/kube-state-metrics/pkg/allow"
+	"k8s.io/kube-state-metrics/pkg/metric"
+)
+
+func TestFilterMetricFamiliesLabels(t *testing.T) {
+	tests := []struct {
+		name             string
+		allowLabels      allow.Labels
+		familyGenerators []FamilyGenerator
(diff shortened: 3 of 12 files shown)


jw-s:
@dgrisonnet Thoughts? If happy with current direction then I'll carry on.

pkg/metric_generator/generator.go
```diff
@@ -101,3 +101,26 @@ func FilterMetricFamilies(l allowDenyLister, families []FamilyGenerator) []Famil
...
 }
+
+// FilterMetricFamiliesLabels takes a AllowLabels and a slice of metric
+// families and returns a filtered slice with metric families with their functions adapted to include only allowed labels for that metric Family.
+func FilterMetricFamiliesLabels(allowLabels AllowLabels, families []FamilyGenerator) []FamilyGenerator {
+	var filtered []FamilyGenerator
+	for _, f := range families {
+		generateFunc := func(familyGenerator FamilyGenerator) func(obj interface{}) *metric.Family {
+			return func(obj interface{}) *metric.Family {
+				metricFamily := familyGenerator.GenerateFunc(obj)
```
dgrisonnet, line 112: With the current implementation if we call GenerateFunc on an object having the following label: `app.kubernetes.io/name` it will be converted to the Prometheus format: `label_app_kubernetes_io_name`. This means that if we have the following option passed to ksm: `--labels-allow-list="example_labels=[app.kubernetes.io/name]"`, the label will not be allowed.

Your approach is correct; however, if we want to be able to handle Kubernetes annotations and labels some rework needs to be done around the GenerateFunc and more precisely the `kubeLabelsToPrometheusLabels` calls.

pkg/metric_generator/generator.go
```diff
@@ -101,3 +101,26 @@ func FilterMetricFamilies(l allowDenyLister, families []FamilyGenerator) []Famil
 
 	return filtered
 }
+
+// FilterMetricFamiliesLabels takes a AllowLabels and a slice of metric
+// families and returns a filtered slice with metric families with their functions adapted to include only allowed labels for that metric Family.
+func FilterMetricFamiliesLabels(allowLabels AllowLabels, families []FamilyGenerator) []FamilyGenerator {
```
dgrisonnet, line 107: It feels like you are doing what `ComposeMetricGenFuncs` is doing + filtering labels here. How about calling this function with the metric families composed by `ComposeMetricGenFuncs` instead?

lilic:
@jw-s are you planning on continuing the work here, otherwise happy to take what you have and finish it up. Let me know :) 🎉

jw-s:
@lilic Hi, sorry for the delay, it's been crazy the last few months...

pkg/metric_generator/generator.go
jw-s: Hey sorry for the late response. The thing is we need to exclude certain labels and AFAIK we can only inspect metric labels by wrapping the generateFunc?

lilic:
No worries, understandable! :) We just want to cut the 2.0 release of the project in the next week or so. I am happy to take over the rest of the PR if you don't have the time? Up to you :)

jw-s:
I have a few open questions that i've commented, if you already know the direction and answers then maybe it would be better for you to continue as it'll be faster

pkg/metric_generator/generator.go
dgrisonnet: Indeed, you are right.

pkg/metric_generator/generator.go
jw-s: WDYT about a quick win and just pass the labels allow list to `kubeLabelsToPrometheusLabels` function?

lilic:
cc @brancz PTAL thanks!

pkg/metric_generator/generator.go
brancz: > WDYT about a quick win and just pass the labels allow list to kubeLabelsToPrometheusLabels function?

I think this makes sense, it will be the behavior users are most likely to expect.

pkg/metric_generator/allow_labels.go
```diff
@@ -0,0 +1,40 @@
...
+	return &allowLabels{
+		allow: allow,
+	}
+}
+
+func (a *allowLabels) Allowed(metric string, labels, values []string) ([]string, []string) {
+	var allowedLabels, allowedValues []string
+	allowed, ok := a.allow[metric]
+	if !ok{
+		if denyAllByDefaultMetrics.MatchString(metric){
```
brancz, line 25: I'm a bit worried about the consequence this might have on performance if we do this on every metric. Let's make sure that this doesn't impact the benchmarks very much.

brancz:
I wonder if we're maybe going a step too far by doing this for _any_ metric, but on the other hand I can definitely see this being useful as a potential protection mechanism. If it doesn't affect benchmarks too much then I think I'm ok with this strategy. Nice work!

brancz:
CI still seems to have some failures.

Were there specific questions that you still need answers to?

pkg/metric_generator/generator.go
jw-s: The problem with this approach is that we prefix labels coming from label related metrics.

I have a few open questions/thoughts...

Should we check if the metric is of a certain kind and add the prefix in the filtering stage/allow list stage rather than in the resource generate functions?

Or do we expect users to provide this type of value to the allowList flag ex: `label.labelname`  which technically would be just `labelname` label in k8s terms.

jw-s: @brancz

pkg/allow/allow_labels.go
```diff
@@ -0,0 +1,52 @@
+package allow
```
tariq1890, line 1: Add the Copyright License header here

pkg/metric_generator/generator_test.go
```diff
@@ -0,0 +1,241 @@
+package generator
```
tariq1890, line 1: Add License header here

pkg/metric_generator/generator_test.go
```diff
@@ -0,0 +1,241 @@
...
+
+import (
+	"k8s.io/kube-state-metrics/pkg/allow"
```
tariq1890, line 4: Need to apply goimports here. Let's separate the stdlib and third-party imports into different import groups.

Consider using this tool: https://github.com/incu6us/goimports-reviser

jw-s: Done

pkg/allow/allow_labels.go
```diff
@@ -0,0 +1,68 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
```
tariq1890, line 2:
```suggestion
Copyright 2020 The Kubernetes Authors All rights reserved.
```

jw-s: Fixed

brancz:
Linting and docs check CI jobs are still failing, but otherwise I think we're closing in on this! :)

I'll have another look when CI is all green

jw-s:
All green now, we need to discuss how we want to handle label/annotation metrics as this is slightly different to normal metrics for allowing atleast.

As you know these metrics are prefixed and with the current implementation users need to provide this also i.e metric_label{mylabel="whatever"} would mean --labels-allow-list=metric_label=[label_mylabel]

sky-philipalmeida:
Hi, can the scope of this PR also include annotations?
The main reason is that Labels have mask constrains they cannot have spaces - and other rules.
If we want to share some fancy string to work on we must use annotations.

jw-s:
> Hi, can the scope of this PR also include annotations?
> The main reason is that Labels have mask constrains they cannot have spaces - and other rules.
> If we want to share some fancy string to work on we must use annotations.

When we say labels, we mean Prometheus labels - the flag is used to restrict these. Meaning it will work for any metric, annotation related metrics also. Once this is merged in, we can rework #1041

jw-s:
@brancz @tariq1890 What's happening with this?

brancz:
The regex appears to add ~10-15% performance regression. I think this is probably ok, but I'd like @lilic's and @tariq1890's opinion on this as well.

tariq1890:
Given that several users have requested this features, I am okay with the performance regression.

It's an ok frome

jw-s:
I won't be available next week just as an FYI if we need to do anything to this PR.

jw-s:
Any update on this?

tariq1890:
@jw-s The PR looks good to me. Can you rebase so that the latest CI changes are pulled into your branch?

I'll give @lilic or @brancz a chance to look at this PR before merging this.

jw-s:
> All green now, we need to discuss how we want to handle label/annotation metrics as this is slightly different to normal metrics for allowing atleast.
> 
> As you know these metrics are prefixed and with the current implementation users need to provide this also i.e metric_label{mylabel="whatever"} would mean --labels-allow-list=metric_label=[label_mylabel]

@tariq1890 What do you think of this?

I will rebase as mentioned.

brancz:
/lgtm
/hold

Giving @tariq1890 and or @lilic last call on this.

tariq1890:
@jw-s I would prefer not having to specify the prefix. For the sake of better UX. But I think this can be tackled in another PR :).

It's an LGTM from me. 

/hold cancel

tariq1890:
Thanks for your hard work and patience on this @jw-s. It's really appreciated :)

PR #1131, opened 2020-04-29
add new service metric, kube_service_spec_selectors, to export servic…

jimmyseto:
…e selector info

This PR extends the current set of metrics exported for a service.  More specifically, it introduces a new metric (kube_service_spec_selectors) to export service spec selector information.  Service selectors facilitate the connection of services to pods (or vice-versa).  Without this information, one cannot make such connections with the current set of metrics.

Fixes #

### DIFF
--- a/docs/service-metrics.md
+++ b/docs/service-metrics.md
@@ -5,6 +5,7 @@
 | kube_service_info | Gauge | `service`=&lt;service-name&gt; <br> `namespace`=&lt;service-namespace&gt; <br> `cluster_ip`=&lt;service cluster ip&gt; <br> `external_name`=&lt;service external name&gt; <btr> `load_balancer_ip`=&lt;service load balancer ip&gt; | STABLE |
 | kube_service_labels | Gauge | `service`=&lt;service-name&gt; <br> `namespace`=&lt;service-namespace&gt; <br> `label_SERVICE_LABEL`=&lt;SERVICE_LABEL&gt;  | STABLE |
 | kube_service_created | Gauge | `service`=&lt;service-name&gt; <br> `namespace`=&lt;service-namespace&gt; | STABLE |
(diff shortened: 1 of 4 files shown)


brancz:
We have a lot of problems with unbound cardinality metrics like this one, I would prefer to avoid something like this. You can already attach service metadata to your targets discovered by Prometheus, why is that not enough?

jimmyseto:
Thanks for the feedback @brancz 

RE: unbound cardinality
This is the same situation as with labels (e.g. kube_pod_labels), right?

RE: attach service metadata
Are you to referring to users manually labeling their targets (i.e. pods) to indicate where their services may be running?  Or, are you talking about something else?

brancz:
> This is the same situation as with labels (e.g. kube_pod_labels), right?

Correct, which we're actively trying to remove/replace with a different mechanism for a 2.0 version.

jimmyseto:
> > This is the same situation as with labels (e.g. kube_pod_labels), right?
> 
> Correct, which we're actively trying to remove/replace with a different mechanism for a 2.0 version.
> 
I see.  Can that mechanism for 2.0 be used for selectors as well?  If so, can we go down this road for 1.x, and then switch over to the new mechanism for 2.0 (e.g. like we plan to do with labels)?

> > Are you to referring to users manually labeling their targets (i.e. pods) to indicate where their services may be running? Or, are you talking about something else?
> 
> Exactly. Except "manually" is relative as you only need a single relabeling config for that.
I was hoping for an out-of-the-box solution, where users would not need to edit any configs.  If a user is using a scraper on the endpoint, they would need to edit each of their service configs and insert labels to expose the selector information.  This does not seem very user-friendly.

brancz:
The proposed mechanism is to specifically allow certain labels as flags on kube-state-metrics. I think you're looking for something more dynamic than that. If the cardinality is worth it to you, why don't you build a controller, that watches endpoints objects and produces a metric like `service_pod`, that has the `service` and `pod` label? Then you really only have to do one join and you reduce the cardinality greatly and it works generically.

lilic:
@jimmyseto was Frederics answer above enough for you and we can close this PR?

Issue #1135, opened 2020-05-06
VolumeAttachment API version mismatch: expected v1 but watching v1beta1


benjaminhuo:
/kind bug

**What happened**:
Deploy ksm v1.9.5 via kube-prometheus in k8s 1.17.3 cluster, got errors in kube-state-metrics log like below:
```
I0506 02:13:29.155880       1 builder.go:146] Active collectors: certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments
E0506 02:13:37.787456       1 reflector.go:382] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: expected type *v1.VolumeAttachment, but watch event object had type *v1beta1.VolumeAttachment
E0506 02:14:55.589302       1 reflector.go:382] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: expected type *v1.VolumeAttachment, but watch event object had 
```
I can see that VolumeAttachment has been available in storage.k8s.io/v1 since kubernetes v1.13, maybe we can update kube-state-metrics ?

**What you expected to happen**:
no such errors

**Environment**:
- Kubernetes version (use `kubectl version`):
```
kubectl version
Client Version: version.Info{Major:"1", Minor:"17", GitVersion:"v1.17.3", GitCommit:"06ad960bfd03b39c8310aaf92d1e7c12ce618213", GitTreeState:"clean", BuildDate:"2020-02-11T18:14:22Z", GoVersion:"go1.13.6", Compiler:"gc", Platform:"linux/amd64"}
```
- Kube-state-metrics image version
quay.io/coreos/kube-state-metrics:v1.9.5

PR #1136, opened 2020-05-06
Fix VolumeAttachment API version mismatch: expected v1 but watching v1beta1

benjaminhuo:
Signed-off-by: Benjamin <benjamin@yunify.com>

Fix for issue https://github.com/kubernetes/kube-state-metrics/issues/1135


tariq1890:
Thanks for the PR @benjaminhuo :)

PR #1141, opened 2020-05-14
Avoid conflicts when mapping Kubernetes labels to Prometheus labels

### LINKED ISSUE #1115: Unexpected label is not unique error
/kind bug

**What happened**:
We had a pod get created with `ops-managed` and `ops_managed` labels in our cluster. In this scenario, all metrics from kube-state-metrics are getting rejected by Prometheus with label is not unique error message. 

```
level=warn ts=2020-04-06T15:07:36.476Z caller=scrape.go:972 component="scrape manager" scrape_pool=kubernetes-service-endpoints target=http://10.4.263.218:8080/metrics msg="append failed" err="label name \"label_ops_managed\" is not unique: invalid sample"
```

Metrics to be ingested... with either one or two separate labels in this scenario. 

Add `ops-managed` and `ops_managed` labels to a single pod. 

I could not find a duplicate issue, my apologies if I overlooked one. Thank you for the help and the great work on this project! 

**Environment:**
v1.9.5 for ksm
v2.16.0 for prometheus

laggyluke:

Prometheus 2.16.0 introduced a breaking change that prevents _any_ metrics from being ingested from a target in case _even a single_ metric has duplicate labels.

See https://github.com/prometheus/prometheus/pull/6664

This becomes a problem for ksm as Kubernetes label naming is less restrictive than Prometheus label naming, so we have a "sanitization" process that replaces unsupported characters with an underscore.

A side effect of this process is that it can produce duplicate labels, e.g. Kubernetes labels "foo.bar", "foo-bar" and "foo_bar" are mapped to the same Prometheus label "foo_bar".

To work around this, we're only adding the first label (sorted by `sort.Strings()`) and discarding the rest of duplicates. In the example above, the value of "foo-bar" will be used because hyphen (45) comes before dot (46) and underscore (95) in ASCII.

If you'd like to have more control over which value is chosen, you should consider addressing this on a different level of the stack, e.g. by standardizing Kubernetes labels using Admission Webhook.

Fixes #1115


brancz:
I'm happy with this solution, but @tariq1890 and @lilic should be at least aware of these changes and review as well.

internal/store/utils.go
```diff
@@ -78,15 +78,22 @@ func kubeLabelsToPrometheusLabels(labels map[string]string) ([]string, []string)
...
+	sort.Strings(sortedKeys)
+
+	seen := make(map[string]struct{})
+	for _, k := range sortedKeys {
+		labelKey := prefix + "_" + sanitizeLabelName(k)
+		if _, ok := seen[labelKey]; !ok {
+			labelKeys = append(labelKeys, labelKey)
+			labelValues = append(labelValues, labels[k])
+			seen[labelKey] = struct{}{}
+		}
```
lilic, line 96: I would like to add a log line here for debugging purposes that prints which ones are thrown away. WDYT?

tariq1890:
Thank you for the PR :)

If we are going to drop conflicting metrics, then we should document this and provide the caveats to users. 

I am not sure I agree with providing the label key with hyphen the highest order of precedence.ASCII order should not be factored in. I think the order of preference should start with underscore since that is closest to the Prometheus metrics format

internal/store/utils.go
laggyluke: I was worried that it might get too spammy as this code triggers multiple times per second.
So I'll let you folks decide, just let me how should I approach it.

laggyluke:
@tariq1890 hey, thanks for a review!

Just for some context, this commit came directly from our internal fork.
If the part about Admission Webhook sounds too prescriptive, just let me know and I'll drop / reword it.

We were in a bit of a firefighting mode and decided that our primary goal is to prevent ksm from breaking when used with Prometheus 2.16.0+ . We did not really care which label to pick as long as the rules are consistent. In our case the values were actually identical, so there wasn't even a real conflict.

> If we are going to drop conflicting metrics, then we should document this and provide the caveats to users.

Sure, I can add those docs once we agree on the desired behavior (see below).

> I think the order of preference should start with underscore since that is closest to the Prometheus metrics format

So how would you like the priority to look like? Underscore first, then everything else ordered alphabetically?

IMO as the source of data is a Kubernetes object, giving preference to a Kubernetes label naming convention of `pod-template-hash` etc. might actually be more intuitive, but I'll leave the final decision to you.

internal/store/utils.go
lilic: I was thinking of log level debug, so folks can enable it if they need to debug something.

lilic:
> IMO as the source of data is a Kubernetes object, giving preference to a Kubernetes label naming convention of pod-template-hash etc. might actually be more intuitive, but I'll leave the final decision to you.

Agreed, that sounds reasonable to me. @tariq1890 @brancz

tariq1890:
@laggyluke Whoops my bad, you're definitely right there. We definitely want hyphen/dash to be the preferred label delimiter. Sorry for the confusion.

However, I don't agree with sorting for achieving this order of precedence, as the ASCII order is just incidental and it would be misleading to codify that. We should specify the fallback order from dash to underscore to dot explicitly IMO.

laggyluke:
@tariq1890 Oh, sure, I agree that if we want to give priority to a specific character, we shouldn't rely on ASCII order alone. And giving priority to dashes makes sense indeed.

But we'd still need to have some sort of ordering (or another way to achieve deterministic behavior) for the rest of the characters. I included dots and underscores as an example, but I believe slashes are also allowed in k8s label names.

So instead of defining a priority of individual characters, maybe we can have some sort of generic rule, e.g. dashes go first, but if there are no dashes - then whatever comes first in ASCII order?

brancz:
I tend to agree and sorting the strings essentially and taking the first seems good enough. FWIW if someone runs into this, nothing is going to be intuitive, the user is always going to wonder what happened. We could also have other conflict resolution that still retains all labels like appending “_conflict1”, “_conflict2“ which would make it clear to the user _that_ a conflict did happen and was resolved. Kind of similar to the prometheus “_exported” conflict resolution.

tariq1890:
If we were to go by this document - https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/, one would be led to believe that `.` is the preferred label key delimiter. Now I feel that sorting or even my suggestion of preferring `-` over others is us imposing a rather arbitrary opinion on our users.

Maybe to quote @dgrisonnet 

>Since allowed labels would be defined via the arguments, ksm would exit due to conflicting allowed labels. What do you think?
This does seem like the best approach to me at this point; we just leave it up to the users to fix this?

brancz:
Isn't the most important thing to make people know that there is a conflict? At that point it doesn't matter anymore what has precedence, as long as it's consistent.

spencergilbert:
I ran into this internally, user reported kube-state-metrics no longer returning from queries - after some digging found that a user leveraged https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=71013666 to create some pods with `kubernetes_pod_operator=True,kubernetes-pod-operator=True`.

After tracking down the source of the data and realizing the cause it was easy enough to find the offending pods with `kubectl get pods --all-namespaces --show-labels | rg kubernetes-pod-operator` (which was nice that the Prometheus UI reported `label_kubernetes_pod_operator` as the non-unique label).

From a user POV just coming into this... It would have been nice to not lose _all_ metrics due to the labels of three pods, but I'm also not sure we would have noticed the missing metrics for those pods if they had been "silently dropped"

internal/store/utils.go
laggyluke: @lilic, I've added basic logging, but it might have some issues:

1. This is the first time klog is being used inside `internal`, so I wasn't sure if that's okay. I also wasn't sure about the exact verbosity level, so I've set it at 4 for now.
2. The context is limited to what's available inside `mapToPrometheusLabels`, so currently I can only log the label itself before and after sanitization. I could also log the value, but I don't think that's useful. We could add more context by passing extra data about resource type and resource name inside `mapToPrometheusLabels ` / `kubeLabelsToPrometheusLabels` to make logs more useful.

Please let me know if I should change anything.

laggyluke:
Hey folks, sorry for a delay.

> FWIW if someone runs into this, nothing is going to be intuitive, the user is always going to wonder what happened.

I completely agree and I don't think logging will help much as it's basically opt-in. But it's better than nothing :)

On the other hand, I'm pretty sure that 99% of the users will never experience this problem at all as they won't have any duplicate labels.
Of those who will experience it, I imagine a significant portion will have a small misconfiguration where duplicate labels have the same value, so it doesn't really matter which one "wins" as the result is the same. I believe all the reports we've seen so far (including my own) fall into this bucket.

The remaining fraction of users might end up confused if the value they see in Prometheus is different from the value they set in Kubernetes.
Once they realize there is a discrepancy, I assume they will start troubleshooting by inspecting the resource in Kubernetes.
And once they see all the labels in Kubernetes, the issue should become pretty obvious to them.

If wear my Kubernetes operator hat, once I discover this issue, I wouldn't blame ksm for keeping a single value and discarding the rest.
Fundamentally ksm is in a bad position to make any sort of authoritative decisions here.
So that small fraction of users who are affected by this issue should seek a solution on a different level of the stack, e.g. by introducing an Admission Webhook that will reject labels that do not confirm to a certain format that is ksm-friendly.

@dgrisonnet, I believe you mentioned this behavior can be used as an attack. I'm not sure how much damage you can do by overriding the value of a label in ksm alone, but do you think Admission Webhook could also protect from this?

laggyluke:
@brancz

> We could also have other conflict resolution that still retains all labels like appending “_conflict1”, “_conflict2“ which would make it clear to the user that a conflict did happen and was resolved. Kind of similar to the prometheus “_exported” conflict resolution.

That's an interesting approach that I haven't considered!

Unfortunately we'd still have to make a decision: if we have three labels that are in conflict, do we add suffixes to all of them or just to two "unlucky" ones?

For example, given labels:

```
foo-bar
foo.bar
foo_bar
```

do we turn them into

```
label_foo_bar
label_foo_bar_conflict1
label_foo_bar_conflict2
```

or

```
label_foo_bar_conflict1
label_foo_bar_conflict2
label_foo_bar_conflict3
```
?

First approach means we still have to define who "wins" and users won't immediately notice the problem, making suffixes less useful.
Second approach means users won't be able to find what they're looking for using PromQL unless they inspect all labels of a metric and realize there are conflicts.

WDYT?

dgrisonnet:
@laggyluke The security issue I mentioned before is related to another problem we have in ksm, which is that annotations and labels metrics can take an unbound amount of labels. Whilst this problem is not related to this particular issue, the way we are trying to solve it might bring a way to detect conflicting labels before starting ksm.

To put some context behind that, the solution we are trying to come with to solve the cardinality problem is to disable the labels and annotations metrics by default. Then it will up to the users to allow them by opt-in the exact list of labels that these metrics can take. (#1124)

With this approach, the only way for a user to introduce conflicting labels would be to specify both `foo-bar` and `foo_bar` as allowed labels. This is very unlikely but in the event that a user wants to do so, we can always log an error telling that conflicting labels have been allowed and exit ksm.

dgrisonnet:
From what I understand of the current approach, we want to relabel the labels based on the order we get them to avoid enforcing precedence. But with this approach how would a user be able to tell which Prometheus label belongs to which Kubernetes one?

Also, as @laggyluke pointed out, there are some issues with the current relabeling approach. In one case, we are always relabeling as if there was a conflict, which might be misleading to users. Whilst the other one is assuming that the first label is the correct one and all the others are conflicting with it. This one might lead to huge data inconsistencies as users might no detect the conflict right away and use the `label_foo_bar` as if it was the correct label whilst being the one conflicting.

laggyluke:
@dgrisonnet I'm afraid we can't reason about certain label being "first" because they are stored in an unordered map. So if we want to order them, we have to define those rules first, e.g. I sorted them alphabetically.

But again, I don't think the ordering rules themselves are that important.
I'm happy to change them to whatever rules we deem more appropriate, just let me know what to do.

I just want to prevent ksm from breaking completely when used with Prometheus 2.16.0+ the moment someone sets a conflicting label by accident.

lilic:
> I just want to prevent ksm from breaking completely when used with Prometheus 2.16.0+ the moment someone sets a conflicting label by accident.

Agreed! 

> We could also have other conflict resolution that still retains all labels like appending “_conflict1”, “_conflict2“ which would make it clear to the user that a conflict did happen and was resolved. Kind of similar to the prometheus “_exported” conflict resolution.

Despite being part of kube, we are after all a prometheus metrics exporter so we should keep in line with prometheus practices. I vote for the above @tariq1890 any objections? In a follow up PR **if** this produces too many labels or any other problems, we can also have a flag of just dropping all conflicts and keeping first one, but I would first keep it simple for now.

tariq1890:
> Despite being part of kube, we are after all a prometheus metrics exporter so we should keep in line with prometheus practices. I vote for the above @tariq1890 any objections? In a follow up PR if this produces too many labels or any other problems, we can also have a flag of just dropping all conflicts and keeping first one, but I would first keep it simple for now.

Sure, let's move forward with this approach 👍

laggyluke:
Thanks folks!

I'd just like to get some input on https://github.com/kubernetes/kube-state-metrics/pull/1141#issuecomment-636845591 before I change the implementation to use `conflictX` suffixes.

brancz:
I think we should sort the original keys, and then go with:

```
label_foo_bar_conflict1
label_foo_bar_conflict2
label_foo_bar_conflict3
```

This way it's at least obvious that a conflict is indeed occurring. If we relatively randomly choose one to not have the suffix I fear that will lead to people thinking everything is ok, when it's not.

laggyluke:
The new implementation is ready, so let me close this PR in favor of #1156.

PR #1144, opened 2020-05-16
internal/store/mutatingwebhookconfiguration.go: Switch to v1

### LINKED ISSUE #1142: MutatingWebhookConfiguration v1 vs v1beta1 mismatch

/kind bug

**What happened**:
Installed kube-state-metrics into k8s cluster using prometheus-operator helm chart 8.13.7 and getting the following errors:
`E0515 10:14:07.385907       1 reflector.go:382] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: expected type *v1.MutatingWebhookConfiguration, but watch event object had type *v1beta1.MutatingWebhookConfiguration

E0515 10:19:54.430864       1 reflector.go:382] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: expected type *v1.VolumeAttachment, but watch event object had type *v1beta1.VolumeAttachment`

AFAIK the second log line represents a bug which is already fixed but fix version is not released yet.

No errors mentioned above. 

Install kube-state-metrics v 1.9.5 into k8s 1.17.5

**Environment**:
- Kubernetes version (use `kubectl version`): 1.17.5
- Kube-state-metrics image version 1.9.5

qw1mb0:
Switch mutatingWebhookConfiguration to use v1 api

Fixes #1142

### DIFF
--- a/internal/store/mutatingwebhookconfiguration.go
+++ b/internal/store/mutatingwebhookconfiguration.go
@@ -17,7 +17,7 @@ limitations under the License.
 package store
 
 import (
-	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
+	admissionregistrationv1 "k8s.io/api/admissionregistration/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/watch"
@@ -37,7 +37,7 @@ var (
 			Name: "kube_mutatingwebhookconfiguration_info",
 			Type: metric.Gauge,
 			Help: "Information about the MutatingWebhookConfiguration.",
-			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistration.MutatingWebhookConfiguration) *metric.Family {
+			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistrationv1.MutatingWebhookConfiguration) *metric.Family {
 				return &metric.Family{
 					Metrics: []*metric.Metric{
 						{
@@ -51,7 +51,7 @@ var (
 			Name: "kube_mutatingwebhookconfiguration_created",
 			Type: metric.Gauge,
 			Help: "Unix creation timestamp.",
-			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistration.MutatingWebhookConfiguration) *metric.Family {
+			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistrationv1.MutatingWebhookConfiguration) *metric.Family {
 				ms := []*metric.Metric{}
 
 				if !mwc.CreationTimestamp.IsZero() {
@@ -68,7 +68,7 @@ var (
 			Name: "kube_mutatingwebhookconfiguration_metadata_resource_version",
 			Type: metric.Gauge,
 			Help: "Resource version representing a specific version of the MutatingWebhookConfiguration.",
-			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistration.MutatingWebhookConfiguration) *metric.Family {
+			GenerateFunc: wrapMutatingWebhookConfigurationFunc(func(mwc *admissionregistrationv1.MutatingWebhookConfiguration) *metric.Family {
 				return &metric.Family{
 					Metrics: resourceVersionMetric(mwc.ObjectMeta.ResourceVersion),
 				}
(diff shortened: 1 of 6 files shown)


tariq1890:
Oh dear 🤦 , another apiVersion fumble. Thanks for fixing this @qw1mb0!

Can you rename the import aliases to `admissionregistrationv1`? This way it's more suggestive of the version we expect to use. Don't forget to make the change in `builder.go` as well :)

tariq1890:
Looks like `v1` has some breaking changes, I don't think we can incorporate these breaking changes since we have many consumers using `v1beta1`. Let's change the expected Types and Client-go reflectors to point to `v1beta1` instead.

internal/store/mutatingwebhookconfiguration.go
```diff
@@ -17,7 +17,7 @@ limitations under the License.
 package store
 
 import (
-	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
+	admissionregistration "k8s.io/api/admissionregistration/v1"
```
tariq1890, line 20:
```suggestion
	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
```

internal/store/mutatingwebhookconfiguration.go
```diff
@@ -80,10 +80,10 @@ var (
 func createMutatingWebhookConfigurationListWatch(kubeClient clientset.Interface, ns string) cache.ListerWatcher {
 	return &cache.ListWatch{
 		ListFunc: func(opts metav1.ListOptions) (runtime.Object, error) {
-			return kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().List(opts)
+			return kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().List(opts)
```
tariq1890, line 83:
```suggestion
			return kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().List(opts)
```

internal/store/mutatingwebhookconfiguration.go
```diff
@@ -80,10 +80,10 @@ var (
...
 		},
 		WatchFunc: func(opts metav1.ListOptions) (watch.Interface, error) {
-			return kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Watch(opts)
+			return kubeClient.AdmissionregistrationV1().MutatingWebhookConfigurations().Watch(opts)
```
tariq1890, line 86:
```suggestion
			return kubeClient.AdmissionregistrationV1beta1().MutatingWebhookConfigurations().Watch(opts)
```

internal/store/mutatingwebhookconfiguration_test.go
```diff
@@ -19,7 +19,7 @@ package store
 import (
 	"testing"
 
-	admissionregistration "k8s.io/api/admissionregistration/v1beta1"
+	admissionregistration "k8s.io/api/admissionregistration/v1"
```
tariq1890, line 22:
```suggestion
	admissionregistrationv1beta1 "k8s.io/api/admissionregistration/v1beta1"
```

tests/manifests/mutatingwebhookconfiguration.yaml
```diff
@@ -1,4 +1,4 @@
-apiVersion: admissionregistration.k8s.io/v1beta1
+apiVersion: admissionregistration.k8s.io/v1
```
tariq1890, line 1:
```suggestion
apiVersion: admissionregistration.k8s.io/v1beta1
```

qw1mb0:
So we have same problems with: https://github.com/kubernetes/kube-state-metrics/commit/97678462aa86b7d10f4272ca6757c0977fab514d

tariq1890:
Yes, I believe we may have fixed this from the wrong end. We should have changed the expected type to `v1beta1` in `builder.go`. The question now is whether to move forward with `v1beta1` or `v1`.

We could do either of the two things

i) Just expect type `v1beta1`, we would need to change the expectedType in `builder.go`

OR

ii) Switch completely to `v1`, this would mean changing the sample `mutatingwebhookconfiguration` manifest to `v1` as well so that e2e tests will pass

cc @lilic

qw1mb0:
I think the second option is better.
I pushed the manifest changes and the e2e tests passed

tariq1890:
Can you change the import aliases to `admisstionregistrationv1`?

qw1mb0:
Of course, I also corrected it for validationwebhookconfiguration

tariq1890:
Thanks so much @qw1mb0 :)!

/lgtm
/hold

Waiting on @lilic's approval too

AndrewDryga:
Guys, can you please release a new version with this fix? Our logs are heavily spammed with this message: `expected type *v1.MutatingWebhookConfiguration, but watch event object had type *v1beta1.MutatingWebhookConfiguration`.

tariq1890:
Thanks for the ping. We are currently working on getting the latest patch release out by Monday.

In the meantime, you can also disable the `mutatingwebhookconfiguration` collector if you're not monitoring them. Sorry for the inconvenience.

Issue #1089, opened 2020-03-07
Use image promoter release process 


serathius:
Using k8s release process would allow to migrate to official k8s repositories without depending on Googlers. It also supports automated build process which will make it more auditable and secure. 

We should also stop depending on quay.io registry if official option can be used.

Staging repo was already setup by @brancz https://github.com/kubernetes/k8s.io/pull/384

/kind feature
Tasks based on https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io

Migrate to promotion process:
* [x] Create google group image release (@dims)
* [x] Create staging repo (@justinsb)
* [x] Add image promotion config in [test-infra](https://github.com/kubernetes/test-infra/tree/master/config/jobs/image-pushing)
* [X] Add image promotion instructions in [RELEASE.md](https://github.com/kubernetes/kube-state-metrics/blob/master/RELEASE.md)
* [X] Replace quay.io repository with k8s.gcr.io as default - we decided to support both

Automate builds:
* [x] Create cloudbuild.yaml [instructions](https://github.com/kubernetes/test-infra/blob/master/config/jobs/image-pushing/README.md) [example](https://github.com/kubernetes-sigs/metrics-server/pull/460)
* [x] Add image build [job in prow](https://github.com/kubernetes/test-infra/tree/2657fccb7bd5af5ff2d6fb6d7f1349cf14d2cc1e/config/jobs/kubernetes-sigs)
* [x] Update image build instructions in [RELEASE.md](https://github.com/kubernetes/kube-state-metrics/blob/master/RELEASE.md)

/cc @lilic please confirm that plan makes sense for you.

serathius:
@dims Can you create k8s-infra-staging-kube-state-metrics@kubernetes.io google group?

serathius:
/cc @justinsb 
Can you create staging repo for kube-state-metrics?

lilic:
Last time there was still an issue with the tool, is this solved now? https://github.com/kubernetes/k8s.io/pull/384#issuecomment-536551561

serathius:
I think yes, because gcr.io/google-containers is deprecated. Googler no longer can release new images to it.
https://github.com/kubernetes/k8s.io/tree/master/k8s.gcr.io/images/DO-NOT-MODIFY-legacy-backfill

I will confirm it with someone from sig-release

serathius:
@bartsmykla Can you confirm that new pipeline is fully functional?

[...]

PR #1155, opened 2020-06-11
cloudbuild.yaml: Add builds for k8s.gcr.io

lilic:
I do not know if I am doing correctly, we have push make target, which should do the trick but not sure about the name part?

@serathius Marek do you mind having a look, thanks so much!

Background -> https://github.com/kubernetes/kube-state-metrics/issues/1089


lilic:
PR in test-infra https://github.com/kubernetes/test-infra/pull/17916

tariq1890:
/hold until the test-infra changes are merged

Thanks @lilic !

PR #1160, opened 2020-06-25
Add DeprecatedVersion to struct FamilyGenerator and func NewFamilyGenerator

YuikoTakada:

Currently, in kube-state-metrics, deprecated metrics is managed just in document. And also there are many docs about metrics(node-metrics.md, pod-metrics.md, ...)
This PR adds a new field "DeprecatedVersion" to struct FamilyGenerator and new func "NewFamilyGenerator" which generate struct FamilyGenerator. In NewFamilyGenerator, if the metrics is deprecated, "(Deprecated since x.x)" will be added at the begining of help string. It makes easier to manage deprecated metrics.

Also see:
https://github.com/kubernetes/kubernetes/blob/master/staging/src/k8s.io/component-base/metrics/opts.go#L198
```
type SummaryOpts struct {
	Namespace         string
	Subsystem         string
	Name              string
	Help              string
	ConstLabels       map[string]string
	Objectives        map[float64]float64
	MaxAge            time.Duration
	AgeBuckets        uint32
	BufCap            uint32
	DeprecatedVersion string <---
	deprecateOnce     sync.Once
	annotateOnce      sync.Once
	StabilityLevel    StabilityLevel
}
```

Fixes #

### DIFF
--- a/pkg/metric_generator/generator.go
+++ b/pkg/metric_generator/generator.go
@@ -17,6 +17,7 @@ limitations under the License.
 package generator
 
 import (
+	"fmt"
 	"strings"
 
 	"k8s.io/kube-state-metrics/pkg/metric"
@@ -24,11 +25,29 @@ import (
 
 // FamilyGenerator provides everything needed to generate a metric family with a
 // Kubernetes object.
+// DeprecatedVersion is defined only if the metric for which this options applies is,
+// in fact, deprecated.
 type FamilyGenerator struct {
-	Name         string
-	Help         string
-	Type         metric.Type
-	GenerateFunc func(obj interface{}) *metric.Family
+	Name              string
+	Help              string
+	Type              metric.Type
+	DeprecatedVersion string
+	GenerateFunc      func(obj interface{}) *metric.Family
+}
+
+// NewFamilyGenerator creates new FamilyGenerator instances.
+func NewFamilyGenerator(name string, help string, metricType metric.Type, deprecatedVersion string, generateFunc func(obj interface{}) *metric.Family) *FamilyGenerator {
+	f := &FamilyGenerator{
+		Name:              name,
+		Type:              metricType,
+		Help:              help,
+		DeprecatedVersion: deprecatedVersion,
+		GenerateFunc:      generateFunc,
+	}
+	if deprecatedVersion != "" {
+		f.Help = fmt.Sprintf("(Deprecated since %s) %s", deprecatedVersion, help)
+	}
+	return f
 }
 
 // Generate calls the FamilyGenerator.GenerateFunc and gives the family its
--- a/internal/store/certificatesigningrequest.go
+++ b/internal/store/certificatesigningrequest.go
@@ -36,11 +36,12 @@ var (
 	descCSRLabelsDefaultLabels = []string{"certificatesigningrequest"}
 
 	csrMetricFamilies = []generator.FamilyGenerator{
-		{
-			Name: descCSRLabelsName,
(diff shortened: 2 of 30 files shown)


brancz:
I like this a lot! Similar to the metrics stability framework, could you add that the description is prefixed with the deprecated version?

YuikoTakada:
@brancz Thank you for your comments. Sorry I don't understand what you say. You mean DeprecatedVersion will be set like "DeprecatedVersion: 1.19"? It would be written as just "1.19". I'm sorry if I misunderstand what you say.

brancz:
In Kubernetes, when a metric is deprecated, it will say in the help text for example:

> kube_pod_info (Deprecated since 1.19) <description>

I would like that we also do this in kube-state-metrics.

YuikoTakada:
@brancz Thanks, I understand. In kubernetes, in case of kube-scheduler, the string "Deprecated since x.xx.x" in #HELP is written manually like https://github.com/kubernetes/kubernetes/pull/92160/files (This is my PR, approved but not merged yet😅)

brancz:
In kubernetes/kubernetes a metric is annotated with it's deprecated version (if it's deprecated), and since which version. https://github.com/kubernetes/component-base/blob/d18546f5347cab6a8690e9b863d9ab977f771079/metrics/desc.go#L196

YuikoTakada:
@brancz oh, I didn't know it. Thank you for the advice. I will try to fix this PR.

YuikoTakada:
I've added new func NewFamilyGenerator also. If it looks good, I will fix all callers of FamilyGenerator to use NewFamilyGenerator. Thanks.

brancz:
The strategy looks good, but there are some failures that need fixing.

lilic:
Seems like the failures is:

```
ERRO Running error: golint: analysis skipped: errors in package: [/home/travis/gopath/src/k8s.io/kube-state-metrics/pkg/metric_generator/generator.go:47:35: undeclared name: obj /home/travis/gopath/src/k8s.io/kube-state-metrics/pkg/metric_generator/generator.go:47:22: cannot use generateFunc(obj) (value of type *metric.Family) as func(obj interface{}) *metric.Family value in struct literal] 
```

pkg/metric_generator/generator.go
```diff
@@ -17,20 +17,41 @@ limitations under the License.
...
 
 // FamilyGenerator provides everything needed to generate a metric family with a
 // Kubernetes object.
+// DeprecatedVersion is defined only if the metric for which this options applies is,
+// in fact, deprecated.
 type FamilyGenerator struct {
 	Name         string
 	Help         string
 	Type         metric.Type
+	DeprecatedVersion string
```
tariq1890, line 34: Looks like this needs `gofmt`

pkg/metric_generator/generator.go
```diff
@@ -17,20 +17,41 @@ limitations under the License.
...
 	GenerateFunc func(obj interface{}) *metric.Family
 }
 
+// NewFamilyGenerator creates new FamilyGenerator instances.
+// The deprecatedVersion represents in which version this Metric be deprecated.
```
tariq1890, line 39: I don't think we need to add the details of `DeprecatedVersion` here

brancz:
/lgtm

Leaving approval up to @tariq1890 or @lilic

brancz:
sorry I didn't realize that it would also automatically approve the PR. @tariq1890 @lilic please feel free to comment anyways! :)

PR #1168, opened 2020-07-15
Rework resource metrics

brancz:

Rework resource metrics and return named metrics for well known resources, and keep metrics for extension metrics.

@tariq1890 @lilic

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -20,15 +20,15 @@
 | kube_pod_container_status_ready | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
 | kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
-| kube_pod_container_resource_requests_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
-| kube_pod_container_resource_requests_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
-| kube_pod_container_resource_requests_storage_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
-| kube_pod_container_resource_requests_ephemeral_storage_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
+| kube_pod_container_resource_requests_cpu_cores | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
+| kube_pod_container_resource_requests_memory_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
+| kube_pod_container_resource_requests_storage_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
+| kube_pod_container_resource_requests_ephemeral_storage_bytes | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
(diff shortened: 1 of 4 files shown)


brancz:
This is already pretty huge, so I didn't include the pod overhead metrics yet.

lilic:
You need to update the expected number of pods in the tests > pod_test.go:1678: expected 39 but got 55

brancz:
Fixed.

docs/pod-metrics.md
```diff
@@ -19,8 +19,16 @@
...
 | kube_pod_container_status_ready | Gauge | `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | STABLE |
 | kube_pod_container_status_restarts_total | Counter | `container`=&lt;container-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `pod`=&lt;pod-name&gt; | STABLE |
-| kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
-| kube_pod_container_resource_limits | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; | STABLE |
+| kube_pod_container_resource_requests | Gauge | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; | EXPERIMENTAL |
```
lilic, line 22: Why is this one EXPERIMENTAL but the other are STABLE?

brancz: Primarily because I want us to keep the possibility to move metrics from this to a named metric at any point in time.

brancz:
I marked all resource metrics as experimental as there is an effort in sig instrumentation to have the scheduler expose these so the metrics represent what the scheduler sees and acts on (and how), as opposed to all the individual fairly high cardinality metrics plus having to maintain the "same" view of the scheduler in PromQL separately.

internal/store/pod.go
```diff
@@ -769,6 +769,110 @@ var (
...
+
+				for _, c := range p.Spec.Containers {
+					req := c.Resources.Requests
+
+					for resourceName, val := range req {
+						if resourceName == v1.ResourceCPU {
+							ms = append(ms, &metric.Metric{
+								LabelKeys:   []string{"container"},
+								LabelValues: []string{c.Name},
+								Value:       float64(val.MilliValue()) / 1000,
```
lilic, line 787: Why not just use the ScaledValue with the wanted Scale? IIRC with MilliValue we could potentially lose some data if its not the thing we want in the end. https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/amount.go#L36-L46

internal/store/pod.go
```diff
@@ -780,40 +884,23 @@ var (
 					req := c.Resources.Requests
 
 					for resourceName, val := range req {
-						switch resourceName {
-						case v1.ResourceCPU:
+						if isHugePageResourceName(resourceName) {
```
lilic, line 887: Can you explain the below change a bit more, you wanted to just do the breaking changes around the units, why all these changes?

internal/store/pod.go
```diff
@@ -827,6 +914,110 @@ var (
...
+
+				for _, c := range p.Spec.Containers {
+					req := c.Resources.Limits
+
+					for resourceName, val := range req {
+						if resourceName == v1.ResourceCPU {
+							ms = append(ms, &metric.Metric{
+								LabelKeys:   []string{"container"},
+								LabelValues: []string{c.Name},
+								Value:       float64(val.MilliValue()) / 1000,
```
lilic, line 932: Again here like above can't we just use scale value?

internal/store/pod.go
brancz: This is what we've always been doing with CPU resources, and have never had a problem with. My understanding is that CPU request always gets parsed in millis, so that's why there is no precision loss in this case. I agree the API of these types is very confusing, but this appears to be correct.

internal/store/pod.go
brancz: This code doesn't change anything, this just extracted the `default` path from the `switch` statement, as all the other cases now have explicit metrics again.

internal/store/pod.go
brancz: Same answer. Extented resources in kubernetes, can only be [whole amounts](https://kubernetes.io/docs/tasks/administer-cluster/extended-resource-node/#discussion), which is why the scaled value is correct for all other cases, but this one, as CPU is always specify-able in millis, and just gets parsed as such. Scaled value can't represent this. The types are arguably super confusing.

internal/store/pod.go
lilic: Not judging by the code https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L702

internal/store/pod.go
lilic: Same answer as above in the API code itself https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go#L702 it seems like MilliValue function uses the `q.ScaledValue(Milli)`

internal/store/pod.go
brancz: I'm fine with revisiting this, just this has worked for ~3 years, I'd be hesitant to mix these two changes.

lilic: Sounds good to me! Lets revisit if we get problems then :)

lilic:
/lgtm
/hold

cc @tariq1890 can you please have a look as well, thanks!

docs/pod-metrics.md
tariq1890: Let's move this to maintain lexical order in the docs

tariq1890:
Since we are touching pod.md and pod.go, let's use this opportunity to maintain lexical order of the metrics.  This comes in handy when you want to look up metrics when going through the source.

LGTM otherwise

brancz:
Makes sense, I'll do that as a separate commit, so reviews are hopefully easier.

Issue #1086, opened 2020-03-05
v1.9.5 image from GCR does not work


riconnon:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
I pulled the latest image from both Quay and GCR

**What you expected to happen**:
I expected them to be the same

**How to reproduce it (as minimally and precisely as possible)**:
```
$ docker image pull quay.io/coreos/kube-state-metrics:v1.9.5
$ docker image pull k8s.gcr.io/kube-state-metrics:v1.9.5
```
These two images ought to be the same image but I see two different images:
```
$ docker image ls
k8s.gcr.io/kube-state-metrics                                                                                                 v1.9.5                                             303a82af2ebf        12 days ago         32.8MB
quay.io/coreos/kube-state-metrics                                                                                             v1.9.5                                             43cd3e9ae81e        2 weeks ago         32.8MB
```
The image from k8s.gcr.io will not execute, but that from quay will:
```
$ docker run -it --rm quay.io/coreos/kube-state-metrics:v1.9.5
I0305 18:00:59.869735       1 main.go:86] Using default collectors
I0305 18:00:59.869887       1 main.go:98] Using all namespace
I0305 18:00:59.869898       1 main.go:139] metric white-blacklisting: blacklisting the following items: 
W0305 18:00:59.869915       1 client_config.go:543] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W0305 18:00:59.869924       1 client_config.go:548] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
F0305 18:00:59.870113       1 main.go:147] Failed to create client: invalid configuration: no configuration has been provided
$ docker run -it --rm k8s.gcr.io/kube-state-metrics:v1.9.5    
standard_init_linux.go:190: exec user process caused "permission denied"
```

**Anything else we need to know?**:

**Environment**:
- Kubernetes version (use `kubectl version`): v1.14.8
- Kube-state-metrics image version: v1.9.5

serathius:
Ups, looks like my fault. 
/assign

serathius:
Hmm I rebuild the image locally. I run
```
git clone git@github.com:kubernetes/kube-state-metrics.git
cd kube-state-metrics
[...]

PR #1171, opened 2020-07-21
Build binary fully in container build

mrueg:
As a first step towards https://github.com/kubernetes/kube-state-metrics/issues/1089, this now builds the kube-state-metrics binary fully in Docker. I'm not sure how images are build via the image-promoter, in order to build correctly there, the image-promoter would need to set a build arg for the arch it needs to build for.

FYI @paulfantom @lilic @tariq1890

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,6 +1,13 @@
-FROM gcr.io/distroless/static
+FROM golang:1.14 as builder
+ARG GOARCH
+ENV GOARCH=${GOARCH}
+WORKDIR /go/src/k8s.io/kube-state-metrics/
+COPY . /go/src/k8s.io/kube-state-metrics/
 
-COPY kube-state-metrics /
+RUN make build-local
+
+FROM gcr.io/distroless/static:latest
+COPY --from=builder /go/src/k8s.io/kube-state-metrics/kube-state-metrics /
 
 USER nobody
 
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,0 @@
-*
-!kube-state-metrics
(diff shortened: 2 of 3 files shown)


Dockerfile
```diff
@@ -1,6 +1,13 @@
-FROM gcr.io/distroless/static
+FROM golang:1.14 as builder
+ARG GOARCH=amd64
+ENV GOARCH=${GOARCH}
+WORKDIR /go/src/k8s.io/kube-state-metrics/
+COPY . /go/src/k8s.io/kube-state-metrics/
```
tariq1890, line 5: Since you've already set the workdir, shouldn't this do?
```suggestion
COPY . .
```

tariq1890:
Thanks @mrueg :)! PR looks good; just one minor comment to address

Dockerfile
mrueg: Thanks for the feedback, added it!

brancz:
If I'm not mistaken this was done intentionally to support the image promotion process on k8s.gcr.io, but I don't remember exactly. Maybe @lilic or @serathius remember? (maybe I'm totally mixing up things though)

serathius:
Yes, using container multi stage build was my recommendation to make integration with image promotion process easier. Build process for k8s.gcr.io is already done in docker, creating lot of problems with path mounting and permissions. I found that using docker multi stage build solves all those problems. 

Secondly I found that using locally build binaries lead to broken images https://github.com/kubernetes/kube-state-metrics/issues/1086#issuecomment-596078150

brancz:
Just to make sure I understand correctly, you also approve of this change right? :)

serathius:
Yes, it's a great improvement.

brancz:
(just wanted to make sure there is not some specific reason why we didn't have this :) )

PR #1172, opened 2020-07-22
add internal_ip to node_info

jimmyseto:

Fixes #

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -42,28 +42,40 @@ var (
 			Type: metric.Gauge,
 			Help: "Information about a cluster node.",
 			GenerateFunc: wrapNodeFunc(func(n *v1.Node) *metric.Family {
+				labelKeys := []string{
+					"kernel_version",
+					"os_image",
+					"container_runtime_version",
+					"kubelet_version",
+					"kubeproxy_version",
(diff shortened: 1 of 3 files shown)


internal/store/node.go
```diff
@@ -42,28 +42,40 @@ var (
...
+					n.Status.NodeInfo.ContainerRuntimeVersion,
+					n.Status.NodeInfo.KubeletVersion,
+					n.Status.NodeInfo.KubeProxyVersion,
+					n.Spec.ProviderID,
+					n.Spec.PodCIDR,
+				}
+
+				internalIP := ""
+				for _, address := range n.Status.Addresses {
+					if address.Type == "InternalIP" {
```
brancz, line 66: as far as I can tell, nothing prevents this array to have multiple `InternalIP`s

jimmyseto: if i'm interpreting the code correctly, i think if multiple internal IPs are specified, only one will be reported in the node object.

https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodestatus/setters.go#L104
https://github.com/kubernetes/kubernetes/blob/master/pkg/kubelet/nodestatus/setters_test.go#L133

PR #1173, opened 2020-07-23
Dockerfile,Makefile: Fix building for k8s.gcr.io images

lilic:
Building fails as we run clean target afterwards as git does not exist.
There is not need for clean anyways.

See error here -> https://prow.k8s.io/view/gs/kubernetes-jenkins/logs/post-kube-state-metrics-push-images/1285961432038379522

cc @brancz @tariq1890 if you have other suggestions how to fix it, I thought this was the smoothest approach, the local should be kept for local building when you want cleanup to happen with git clean and removal of the binary.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,10 +59,10 @@ doccheck: generate
 	@cd docs; for doc in *.md; do if [ "$$doc" != "README.md" ] && ! grep -q "$$doc" *.md; then echo "ERROR: No link to documentation file $${doc} detected"; exit 1; fi; done
 	@echo OK
 
-build-local: clean
+build-local:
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -59,6 +59,9 @@ doccheck: generate
 	@cd docs; for doc in *.md; do if [ "$$doc" != "README.md" ] && ! grep -q "$$doc" *.md; then echo "ERROR: No link to documentation file $${doc} detected"; exit 1; fi; done
 	@echo OK
 
+build-docker:
```
brancz, line 62: can we just instead of duplicating this, make this a dependency of `build-local`?

PR #1175, opened 2020-07-23
Update docs to point to new gcr registry

lilic:
With the image building and push to `gcr.io/k8s-staging-kube-state-metrics/kube-state-metrics` working, I edited the docs.

cc @serathius can you double check this is correct now from your side, we no longer need to ping you so removing that part. :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -89,10 +89,6 @@ The latest container image can be found at:
 * `quay.io/coreos/kube-state-metrics:v1.9.7`
 * `k8s.gcr.io/kube-state-metrics:v1.9.7`
 
-**Note**:
-The recommended docker registry for kube-state-metrics is `quay.io`. kube-state-metrics on
-`gcr.io` is only maintained on best effort as it requires external help from Google employees.
-
 ### Metrics Documentation
 
 There are many more metrics we could report, but this first pass is focused on
--- a/RELEASE.md
+++ b/RELEASE.md
@@ -15,9 +15,9 @@ Maintaining the release branches for older minor releases happens on a best effo
 ## Prepare your release
 
 * Bump the version in the `VERSION` file in the root of the repository.
-* Run `make examples`, which will re-generate all example manifests to use the the new version.
+* Run `make examples`, which will re-generate all example manifests to use the new version.
 * Make a PR to update:
-  * kube-state-metrics image tag for both `quay.io` and `staging-k8s.gcr.io`.
+  * kube-state-metrics image tag for both `quay.io` and `gcr.io/k8s-staging-kube-state-metrics`.
   * [Compatibility matrix](README.md#compatibility-matrix)
   * Changelog entry
     * Only include user relevant changes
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -87,11 +87,7 @@ release.
 
 The latest container image can be found at:
 * `quay.io/coreos/kube-state-metrics:v1.9.7`
-* `k8s.gcr.io/kube-state-metrics:v1.9.7`
-
-**Note**:
-The recommended docker registry for kube-state-metrics is `quay.io`. kube-state-metrics on
-`gcr.io` is only maintained on best effort as it requires external help from Google employees.
+* `gcr.io/k8s-staging-kube-state-metrics/kube-state-metrics:v1.9.7`
```
serathius, line 90: I would not recommend using staging repo as official, images there can still be manually pushed. 

We should official promotion process which is done through a sending a PR to kubernetes/k8s.io, example: https://github.com/kubernetes/k8s.io/pull/761

Resulting image would be available in k8s.gcr.io/kube-state-metrics/kube-state-metrics:v1.9.7

RELEASE.md
```diff
@@ -32,8 +32,8 @@ Maintaining the release branches for older minor releases happens on a best effo
   * kube-state-metrics image tag used in Kubernetes deployment yaml config.
 * Cut the new release branch, i.e., `release-1.2`, or merge/cherry-pick changes onto the minor release branch you intend to tag the release on
 * Cut the new release tag, i.e., `v1.2.0-rc.0`
-* Create issue requesting for image release to `k8s.gcr.io` (or to `staging-k8s.gcr.io` in case of release candidates) and assign it to one of Googlers (@serathius, @loburm)
-* Build and push newest image to `quay.io`(@brancz)
+* New images are automatically built and pushed to `gcr.io/k8s-staging-kube-state-metrics/kube-state-metrics`
```
serathius, line 35:
```suggestion
* New images are automatically built and pushed to `gcr.io/k8s-staging-kube-state-metrics/kube-state-metrics`
* Promote image by sending a PR to kubernetes/k8s.io 
```

RELEASE.md
```diff
@@ -32,8 +32,8 @@ Maintaining the release branches for older minor releases happens on a best effo
...
+* Build and push newest image to `quay.io/coreos`(@brancz, @lilic)
```
serathius, line 36: I think it would be better to use image build by CI instead of having owners build it locally. We could add script to sync image from k8s.gcr.io/kube-state-metrics to quay.io/coreos

mrueg:
I'd suggest to backport the docker-multistage patch to the 1.9 release branch and cut a new release there?

README.md
lilic: Ah thanks for that piece of info!

lilic:
As mentioned before we won't do a new release as there were no user-facing changes in that release branch just image building is ready for next minor patch or 2.0 version, and since we have images already built for previous releases, this change will take effect on the next release.

Updated the PR to for now only mention the process in the release docs, after we have built the new images for 2.0, which should be soon, will update the README as well with the new image.

cc @brancz @tariq1890 @serathius PTAL, thanks!

RELEASE.md
lilic: Yes good idea, will open issue for this!

README.md
```diff
@@ -89,10 +89,6 @@ The latest container image can be found at:
 * `quay.io/coreos/kube-state-metrics:v1.9.7`
 * `k8s.gcr.io/kube-state-metrics:v1.9.7`
```
serathius, line 90: FYI, I found new image pipeline change image path forMetrics Server, now it is published on `k8s.gcr.io/metrics-server/metrics-server:v0.3.7`, I would expect that new kube-state-metrics image will be available on `k8s.gcr.io/kube-state-metrics/kube-state-metrics`

serathius:
LGTM

PR #1177, opened 2020-07-23
Backport Container Build Changes to Release 1.9

mrueg:
@lilic @serathius backporting the changes for https://github.com/kubernetes/kube-state-metrics/issues/1089 so a new release can be cut.

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,2 +1,0 @@
-*
-!kube-state-metrics
--- a/.travis.yml
+++ b/.travis.yml
@@ -47,4 +47,4 @@ jobs:
     - name: Build
       script: make build
     - name: End to end tests
-      script: make e2e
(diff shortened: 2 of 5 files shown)


mrueg:
@lilic apologies, looks like I missed one patch when backporting from main branch (DOCKER_CLI var was not set in the makefile). Adding it now.

lilic:
hmm on second thought we tend to pick only fixes to the release branches so @brancz @tariq1890 do we want to merge this as images are available for this on at least quay.io/coreos/kube-state-metrics and we want to release 2.0 soon anyways?

mrueg:
I can drop the e2e.sh changes and simply add the DOCKER_CLI var in the makefile as an alternative option, if we want to keep the changes minimal.

lilic:
Its about release process apart from anything else, we backport bug fixes only so 👍 from the other two would be good to get.

PR #1180, opened 2020-07-24
Use run groups

### LINKED ISSUE #1169: MetricsHandler errors are silently dropped

/kind bug

**What happened**:

When an error occurs in MetricsHandler, the error is never logged and do not trigger an unhealthy state. In my case, this happens when the ServiceAccount does not have enough permission for AutoSharding:
https://github.com/kubernetes/kube-state-metrics/blob/2415ca281c81039353adb90931acb8495b3df7f8/pkg/metricshandler/metrics_handler.go#L106

This means that with my invalid configuration (https://github.com/helm/charts/pull/23210), nothing happened. Kube-State-Metrics was "healthy", nothing in the logs but zero metrics were returned.

Kube-State-Metrics should have logged:
```
Detect failed with retrieve shard's StatefulSet: myns/prometheus-operator-kube-state-metrics: statefulsets.apps "prometheus-operator-kube-state-metrics" is forbidden: User "system:serviceaccount:myns:prometheus-operator-kube-state-metrics" cannot get resource "statefulsets" in API group "apps" in the namespace "myns"
```

And (IMHO), it should be in an unhealthy state. Autosharding is enabled but configuration failed.

- Deploy a kube-state-metrics with Autosharding enabled but no Kubernetes API permissions
- Start it (it will not print anything after `metrics_handler.go:101] Auto detecting sharding settings.`)
- Done, your kube-state-metrics is unhealthy, no metrics are coming and no errors were reported

From my understanding, this is because there is no code to handle reported errors when calling MetricsHandler: https://github.com/kubernetes/kube-state-metrics/blob/2415ca281c81039353adb90931acb8495b3df7f8/main.go#L224

**Environment**:
- Kubernetes version (use `kubectl version`): v1.16.8-gke.15
- Kube-state-metrics image version: 1.9.5, 1.9.7

Sayrus:

When an error occurs in a Goroutine, the error is ignored and does not trigger an unhealthy state. This MR tries to fix that by moving to oklog/run groups.

Fixes #1169

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -6,6 +6,7 @@ require (
 	github.com/dgryski/go-jump v0.0.0-20170409065014-e1f439676b57
 	github.com/google/go-jsonnet v0.14.0
 	github.com/jsonnet-bundler/jsonnet-bundler v0.1.1-0.20190930114713-10e24cb86976
+	github.com/oklog/run v1.1.0
 	github.com/pkg/errors v0.9.1
 	github.com/prometheus/client_golang v1.5.1
 	github.com/prometheus/client_model v0.2.0
--- a/main.go
+++ b/main.go
@@ -144,8 +144,8 @@ func main() {
 		opts.EnableGZIPEncoding,
 	)
 	// Run MetricsHandler
-	ctx, cancel := context.WithCancel(context.Background())
 	{
+		ctx, cancel := context.WithCancel(context.Background())
 		g.Add(func() error {
 			return m.Run(ctx)
 		}, func(error) {
--- a/main.go
+++ b/main.go
@@ -24,6 +24,7 @@ import (
 	"net/http/pprof"
 	"os"
 	"strconv"
+	"time"
 
 	"github.com/oklog/run"
 	"github.com/pkg/errors"
@@ -58,9 +59,6 @@ func (pl promLogger) Println(v ...interface{}) {
 func main() {
 	opts := options.NewOptions()
 	opts.AddFlags()
-
-	ctx, cancel := context.WithCancel(context.Background())
-	defer cancel()
 
 	err := opts.Parse()
 	if err != nil {
@@ -138,19 +136,20 @@ func main() {
 	)
 
 	var g run.Group
+
 	m := metricshandler.New(
 		opts,
 		kubeClient,
 		storeBuilder,
 		opts.EnableGZIPEncoding,
 	)
-
 	// Run MetricsHandler
+	ctx, cancel := context.WithCancel(context.Background())
 	{
 		g.Add(func() error {
 			return m.Run(ctx)
 		}, func(error) {
-			//cancel()
+			cancel()
 		})
 	}
 
@@ -175,7 +174,9 @@ func main() {
 			klog.Infof("Starting kube-state-metrics self metrics server: %s", telemetryListenAddress)
 			return telemetryServer.Serve(telemetryLn)
 		}, func(error) {
-			telemetryServer.Shutdown(ctx)
+			ctxShutDown, cancel := context.WithTimeout(context.Background(), 3 * time.Second)
+			defer cancel()
+			telemetryServer.Shutdown(ctxShutDown)
 		})
 	}
 	// Run Metrics server
@@ -184,7 +185,9 @@ func main() {
 			klog.Infof("Starting metrics server: %s", metricsServerListenAddress)
 			return metricsServer.Serve(metricsServerLn)
 		}, func(error) {
-			metricsServer.Shutdown(ctx)
+			ctxShutDown, cancel := context.WithTimeout(context.Background(), 3 * time.Second)
+			defer cancel()
+			metricsServer.Shutdown(ctxShutDown)
 		})
 	}
 
--- a/main.go
+++ b/main.go
@@ -19,13 +19,13 @@ package main
 import (
 	"context"
 	"fmt"
-	"log"
 	"net"
 	"net/http"
 	"net/http/pprof"
 	"os"
 	"strconv"
 
+	"github.com/oklog/run"
 	"github.com/pkg/errors"
 	"github.com/prometheus/client_golang/prometheus"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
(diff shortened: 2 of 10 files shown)


brancz:
The metrics handler just needs the [context to be cancelled](https://github.com/kubernetes/kube-state-metrics/blob/ee4353fd33c99dc56b176d8f450fa31acc77d27c/pkg/metricshandler/metrics_handler.go#L173). More generally a typical pattern within a run group is to create a distinct context for each group like here: https://github.com/thanos-io/thanos/blob/a64884b19dbe7a6c2613078a07409c89ce4e2508/cmd/thanos/receive.go#L380-L385. They all share the initial root context, but each have their own execution branch.

Sayrus:
@brancz Thanks for the tip, that was definitely what I missed.

main.go
```diff
@@ -138,19 +136,20 @@ func main() {
...
+
 	m := metricshandler.New(
 		opts,
 		kubeClient,
 		storeBuilder,
 		opts.EnableGZIPEncoding,
 	)
-
 	// Run MetricsHandler
+	ctx, cancel := context.WithCancel(context.Background())
```
brancz, line 147: the initial background context should still be shared as the root of all context's .. it's also likely that we will make it cancelable in the future when we add listening to OS signals

brancz: you also want to move this into the `{`, as that makes the ctx and cancel variables only accessible to the scope that run group is defined in, it prevents accidental shadowing that go attempts to do

Sayrus: I'm unsure of what this means. Isn't this just another background context? It is only used once for the MetricsHandler.

The shared root is currently the main context if I understand correctly. Can't we just create a new context for the signal handler and use the Run Groups to cancel everything if such a situation happen?

main.go
```diff
@@ -59,9 +60,6 @@ func main() {
 	opts := options.NewOptions()
 	opts.AddFlags()
 
-	ctx, cancel := context.WithCancel(context.Background())
```
brancz, line 62: this we want to keep, for now we can keep it non cancelable so this would just be

```
ctx := context.Background()
```

main.go
```diff
@@ -136,9 +134,67 @@ func main() {
...
+
+	m := metricshandler.New(
+		opts,
+		kubeClient,
+		storeBuilder,
+		opts.EnableGZIPEncoding,
+	)
+	// Run MetricsHandler
+	{
+		ctx, cancel := context.WithCancel(context.Background())
```
brancz, line 148: here all we want is to reuse the root context we create at the very top, and then this line becomes:

```
context.WithCancel(ctx)
```

same goes for the other context's that are created

main.go
brancz: They are currently distinct background contexts. Typically they are wired up so they all lead to the root context, this is so that if a context somewhere in the middle gets cancelled, all children get cancelled. I added a review to show what I mean.

Sayrus: Oh, that was a misunderstanding on my side. I thought that `context.Background()` was a singleton or something but it creates a different instance on each call.
I understand what you mean and will update with your patch.

go.mod
```diff
@@ -6,6 +6,7 @@ require (
 	github.com/dgryski/go-jump v0.0.0-20170409065014-e1f439676b57
 	github.com/google/go-jsonnet v0.14.0
 	github.com/jsonnet-bundler/jsonnet-bundler v0.1.1-0.20190930114713-10e24cb86976
+	github.com/oklog/run v1.1.0
```
tariq1890, line 9: Why not use `errgroup` instead? It's a part of the go stdlib

tariq1890:
@Sayrus Thanks for doing this. Might I suggest using `errgroups` with Context?

This is a helpful resource: https://bionic.fullstory.com/why-you-should-be-using-errgroup-withcontext-in-golang-server-handlers/

go.mod
Sayrus: To be fair, I'm quite new to developing in Go so I just followed the recommendation made on #1169 and how it is done on Thanos.
IMHO, it makes sense to avoid using an external dependency when there is an alternative in the stdlib. On the other hand, I have no idea what the (technical) differences are between oklog's run and the stdlib's errgroup.

@brancz Any reason to use oklog/run over errgroup?

brancz: errgroup is for goroutines of the same "type", rungroups are for orchestrating goroutines that have distinct purposes

Sayrus:
Updated the first message to reflect that this is not WIP anymore.

Feel free to ping me if I need to change anything.

main.go
```diff
@@ -136,9 +136,67 @@ func main() {
...
+		klog.Fatalf("Failed to create MetricsServer Listener: %v", err)
+	}
+
+	// Run Telemetry server
+	{
+		g.Add(func() error {
+			klog.Infof("Starting kube-state-metrics self metrics server: %s", telemetryListenAddress)
+			return telemetryServer.Serve(telemetryLn)
+		}, func(error) {
+			ctxShutDown, cancel := context.WithTimeout(ctx, 3*time.Second)
```
brancz, line 179: What are we trying to achieve with this?

main.go
```diff
@@ -136,9 +136,67 @@ func main() {
...
+			defer cancel()
+			telemetryServer.Shutdown(ctxShutDown)
```
brancz, line 181: Does this also close the net.Listener? If not then we need to explicitly close that one as well.

main.go
Sayrus: The purpose of my WithTimeout was to let some time to gracefully shutdown but force a timeout if there was a long running query.

main.go
Sayrus: Also linked to the previous comment:

> Shutdown gracefully shuts down the server without interrupting any active connections. Shutdown works by first closing all open listeners, then closing all idle connections, and then waiting indefinitely for connections to return to idle and then shut down. If the provided context expires before the shutdown is complete, Shutdown returns the context's error, otherwise it returns any error returned from closing the Server's underlying Listener(s). 

As I understand it, it closes the Listener first. The `waiting indefinitely` part is the reason I am using `.WithTimeout`.

main.go
brancz: that shouldn't be necessary, canceling allows the go routine to finish anyways

main.go
brancz: got it :+1:

main.go
Sayrus: Hmm, I understand.
Shutdown being a blocking call that can wait indefinitely, when would Cancel be called when using WithCancel?

brancz: cancel for all run groups is called when any run group exists

Sayrus: Oh ... then there is no reason to graceful shutdown and we can just Close it?

brancz: Sorry my bad I understand now what this is achieving, I think it's worth trying this yes.

brancz:
/lgtm
/approve

Awesome! :tada:

main.go
Sayrus: Still, I understand your concern but I'm unsure on how to test whether this really does a Graceful shutdown or not.

PR #1178, opened 2020-07-24
Add observedGeneration metric for DaemonSets

vishnukarthikl:
**What this PR does / why we need it**:
This adds a new metric from DaemonSet status

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1176

PR #1181, opened 2020-07-25
Update DaemonSet updated_number_scheduled metric name to be consistent

vishnukarthikl:
rename metric `kube_daemonset_updated_number_scheduled` to `kube_daemonset_status_updated_number_scheduled` so that it is consistent with the field path. This was discussed [here](https://github.com/kubernetes/kube-state-metrics/pull/1178#pullrequestreview-454717595)


tariq1890:
/lgtm
/approve

Thanks @vishnukarthikl :)

PR #1184, opened 2020-07-26
Added the ability to fetch the kube_job_status_failed_reason

### LINKED ISSUE #947: Need metric to return job failed reason

/kind feature

**What happened**:
Missing a metric like `kube_job_status_failed_reason` 
Add `kube_job_status_failed_reason` to display the failed reason, such as `Evicted`
Create a cronjob but limit the resource

K8S returns the following in the Evicted job
```
status:
  message: 'The node was low on resource: ephemeral-storage. Container backups was
    using 3405160Ki, which exceeds its request of 0. '
  phase: Failed
  reason: Evicted
  startTime: "2019-10-11T00:00:02Z"
```
We would like to filter out the Evicted jobs, even if it's failed.
**Environment**:
```
Client Version: version.Info{Major:"1", Minor:"16", GitVersion:"v1.16.1", GitCommit:"d647ddbd755faf07169599a625faf302ffc34458", GitTreeState:"clean", BuildDate:"2019-10-02T23:49:07Z", GoVersion:"go1.12.9", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"13+", GitVersion:"v1.13.7-gke.24", GitCommit:"2ce02ef1754a457ba464ab87dba9090d90cf0468", GitTreeState:"clean", BuildDate:"2019-08-12T22:05:28Z", GoVersion:"go1.11.5b4", Compiler:"gc", Platform:"linux/amd64"}
```
```
quay.io/coreos/kube-state-metrics:v1.7.2
```

AdityaMisra:
Added a metric `kube_job_status_failed_reason` for tracking failure reason for any given metric.

Fixes #947


internal/store/job.go
```diff
@@ -302,6 +302,35 @@ var (
...
+					if condition.Type != v1batch.JobFailed {
+						continue
+					}
+
+					conditionMetrics := addConditionMetrics(condition.Status)
+
+					for _, m := range conditionMetrics {
+						metric := m
+						metric.LabelKeys = []string{"condition", "reason", "status"}
+						metric.LabelValues = append([]string{string(condition.Type), condition.Reason}, metric.LabelValues...)
```
lilic, line 328: I would prefer this to be modeled on the pod reason we already have, this means we only allow the reasons
https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/pod.go#L38
https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/pod.go#L381

internal/store/job.go
```diff
@@ -302,6 +302,35 @@ var (
 					}
 				}
 
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		},
+		{
+			Name: "kube_job_status_failed_reason",
```
lilic, line 311: I am curious if we should instead add the reason label to the already existing metric `kube_job_status_failed ` cc @brancz @tariq1890 any preferences?

internal/store/job_test.go
```diff
@@ -207,6 +233,15 @@ func TestJobStore(t *testing.T) {
 				kube_job_status_failed{job_name="FailedJob1",namespace="ns1"} 1
 				kube_job_status_start_time{job_name="FailedJob1",namespace="ns1"} 1.495807207e+09
 				kube_job_status_succeeded{job_name="FailedJob1",namespace="ns1"} 0
+				kube_job_status_failed_reason{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="",status="false"} 0
+				kube_job_status_failed_reason{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="",status="true"} 1
+				kube_job_status_failed_reason{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="",status="unknown"} 0
+				kube_job_status_failed_reason{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="BackoffLimitExceeded",status="false"} 0
+		        kube_job_status_failed_reason{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="BackoffLimitExceeded",status="true"} 1
```
lilic, line 240: Seems like an indent problem here?

internal/store/job.go
AdityaMisra: @lilic - Should I remove the `condition, status` label keys from the metric? So, I should be only sending `reason` in the label keys, right? 

One more thing, I could only find `["BackoffLimitExceeded", "DeadLineExceeded"]` job failure reasons. `Evicted` seems to be a reason for pod failure. 
Could you please verify these findings?

internal/store/job.go
AdityaMisra: I was also thinking on the similar lines. @lilic, @brancz & @tariq1890 - Should I make the changes accordingly?

internal/store/job.go
lilic: Yes the ones that are for the job failure, the pod reasons were just an example.

AdityaMisra: @lilic I got your point. The problem is I'm not able to find the `Evicted` failure reason for the job in the Kubernetes job/controller code-base but there is an `Evicted` failure reason for the pod. In the [Issue 947](https://github.com/kubernetes/kube-state-metrics/issues/947) the reporter has given an example with `Evicted` reason. So, I just want to confirm whether there is an `Evicted` failure reason for the job or not.

internal/store/job.go
brancz: yeah let's collapse it into one I agree

AdityaMisra: Cool, I'll take care of it. 👍

internal/store/job.go
lilic: Let me have a look at the controller.

lilic: If user reported I would add all 3, worst case it will just never get that reason, but since we see it is possible it most likely reflects it from the Pod Reason. So lets add two you found plus Evicted. Makes sense?

AdityaMisra: Sounds good to me. I'll update the PR.

tariq1890:
Don't forget to squash your commits :)

tariq1890:
My ask is that you squash all of the commits to one. You wouldn't need to cherry pick them.

https://www.brianhan.co/2019-01-11-using-git-to-squash-commits-with-confidence/

PR #1186, opened 2020-07-27
Add repo SECURITY.md

joelsmith:
See https://github.com/kubernetes/security/issues/105

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,22 @@
+# Security Policy
+
+## Security Announcements
+
+Join the [kubernetes-security-announce] group for security and vulnerability announcements.
+
+You can also subscribe to an RSS feed of the above using [this link][kubernetes-security-announce-rss].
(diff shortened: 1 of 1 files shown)


tariq1890:
Thanks @joelsmith . I think it would be useful for the SECURITY.md to also reference the `SECURITY_CONTACTS` file.

https://github.com/kubernetes/kube-state-metrics/blob/master/SECURITY_CONTACTS

Or is there a plan to phase that out?

joelsmith:
@tariq1890  we're phasing out `SECURITY_CONTACTS` and will be adding an optional section to the owners files instead. If security contacts aren't mentioned in the owners files, the PSC will default to approvers.

tariq1890:
In that case, can we not club deleting `SECURITY_CONTACTS` with this PR. Should make your life easier too since you're doing this org-wide, no?

PR #1196, opened 2020-08-03
revise OWNERS list

tariq1890:
Updating the OWNERS list based on activity in the project


brancz:
if I'm not mistaken these should be moved emeritus no?

PR #1199, opened 2020-08-04
use github actions for CI checks

tariq1890:
This PR replaces TravisCI with GitHub Actions


.github/workflows/ci.yml
```diff
@@ -0,0 +1,64 @@
+name: Build
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
```
mrueg, line 7: does that mean CI won't run on backporting PR to the release branches?

tariq1890: I've fixed this. Thanks @mrueg !

tariq1890:
@lilic 

> 
> I am not too sure about re-running Github Actions. I think that should be available to us from the GH interface. Not sure If that needs to be taken care of via the workflow YAML
> 
> Addressed. Created a trigger for tag pushes
> 
> I've parallelised it. You'll see a lot of duplication, unfortunately GitHub Actions does not re-usable steps as of yet. Once they do, we can remove the code duplication

tariq1890:
Looks the GitHub Actions jobs (Benchmark and E2E) are faster than TravisCI by a minute. That's a win for us :)

brancz:
Giving @lilic last call, but this lgtm! :tada: 

/lgtm
/hold

lilic:
You seem to have a conflict, but otherwise lgtm, thanks Tariq!

PR #1200, opened 2020-08-07
add dev guide for squashing commit history

tariq1890:
Taken from [here](https://v1-16.docs.kubernetes.io/docs/contribute/intermediate/).

Since these correspond to an older k8s version and are no longer available in documentation's trunk, they will most likely be archived in the future. So I have chosen to copy the text from there to the ksm repo

### DIFF
--- a/docs/developer/guide.md
+++ b/docs/developer/guide.md
@@ -6,7 +6,7 @@ Any contribution to improving this documentation will be appreciated.
 ## Table of Contents
 
 - [Add New Kubernetes Resource Metric Collector](#add-new-kubernetes-resource-metric-collector)
-
+- [Squashing Commit History](#squash-commit-history)
 ### Add New Kubernetes Resource Metric Collector
 
 The following steps are needed to introduce a new resource and its respective resource metrics.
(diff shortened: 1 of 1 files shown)


docs/developer/guide.md
```diff
@@ -19,3 +19,40 @@ The following steps are needed to introduce a new resource and its respective re
 - Reference the new resource in [pkg/options/resource.go](https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/options/resource.go).
 - Add a sample Kubernetes manifest to be used by tests in the [tests/manifests/](https://github.com/kubernetes/kube-state-metrics/tree/master/tests/manifests) directory.
 - Lastly, and most importantly, actually implement your new resource(s) and its test binary in [internal/store](https://github.com/kubernetes/kube-state-metrics/tree/master/internal/store). Follow the formatting and structure of other resources.
+
+### Squash Commit History
+
+If your PR still has multiple commits after amending previous commits, you must squash multiple commits into a single commit before your PR can be merged. You can check the number of commits on your PR’s Commits tab or by running git log locally. Squashing commits is a form of rebasing.
```
brancz, line 25: I'm not a fan of this as dogmatic advice, I think encouraging logically splitting commits and squashing "addressed review" type commits makes perfect sense but always squashing is not universally good I believe.

brancz:
Sorry if I made it sound like that, I'm not against having guidelines encouraging clean history, but not necessarily be as strict as saying "you must always squash all changes".

tariq1890:
Not at all. I agree with your comment :)

I am not for dogmatic advice either. There is no one way to squash . I think it's better left an exercise to the contributor

PR #1214, opened 2020-08-28
Added the job failure reason in kube_job_status_failed metric

### LINKED ISSUE #947: Need metric to return job failed reason

AdityaMisra:
Fixed linting issues

Added the job failure reason in kube_job_status_failed metric
Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #947

### DIFF
--- a/docs/job-metrics.md
+++ b/docs/job-metrics.md
@@ -10,7 +10,7 @@
 | kube_job_spec_active_deadline_seconds | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_status_active | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_status_succeeded | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
-| kube_job_status_failed | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
+| kube_job_status_failed | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `reason`=&lt;failure reason&gt; | STABLE |
 | kube_job_status_start_time | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_status_completion_time | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_complete | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `condition`=&lt;true\|false\|unknown&gt; | STABLE |
--- a/internal/store/job.go
+++ b/internal/store/job.go
@@ -35,6 +35,7 @@ var (
 	descJobLabelsName          = "kube_job_labels"
 	descJobLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descJobLabelsDefaultLabels = []string{"namespace", "job_name"}
+	jobFailureReasons          = []string{"BackoffLimitExceeded", "DeadLineExceeded", "Evicted"}
 
 	jobMetricFamilies = []generator.FamilyGenerator{
 		*generator.NewFamilyGenerator(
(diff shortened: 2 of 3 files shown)


AdityaMisra:
You can refer to the historic conversation - [here](https://github.com/kubernetes/kube-state-metrics/pull/1184)

internal/store/job.go
```diff
@@ -163,16 +164,49 @@ var (
...
+					}
+				}
+
+				for _, condition := range j.Status.Conditions {
+
+					if condition.Type == v1batch.JobFailed {
+						reasonKnown := false
+						for _, reason := range jobFailureReasons {
+							reasonKnown = reasonKnown || failureReason(&condition, reason)
+
```
lilic, line 189: You are missing an if check here against `reasonKnown` no, similar to the unknown reasons?

lilic (commented):
Thanks! Few comments.

internal/store/job.go
AdityaMisra: Hey @lilic 
I don't think we need an `if` check here because we need to print for all the known reasons with `0` value. Suppose, we get a known reason `BackoffLimitExceeded` then we will be printing
```
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="BackoffLimitExceeded"} 1 
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="DeadLineExceeded"} 0
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="Evicted"} 0
```

and if we get an unknown reason then we will be printing 

```
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason=""} 1
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="BackoffLimitExceeded"} 0 
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="DeadLineExceeded"} 0
kube_job_status_failed{condition="Failed",job_name="FailedJob1",namespace="ns1",reason="Evicted"} 0
```

Let me know if we need to make any changes.

internal/store/job.go
```diff
@@ -163,16 +164,49 @@ var (
...
+							{
+								Value: float64(j.Status.Failed),
+							},
 						},
-					},
+					}
+				}
+
+				for _, condition := range j.Status.Conditions {
+
```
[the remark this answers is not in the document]
AdityaMisra: Sure.

internal/store/job.go
```diff
@@ -163,16 +164,49 @@ var (
...
+					if condition.Type == v1batch.JobFailed {
+						reasonKnown := false
+						for _, reason := range jobFailureReasons {
+							reasonKnown = reasonKnown || failureReason(&condition, reason)
+
+							// for known reasons
+							ms = append(ms, &metric.Metric{
+								LabelKeys:   []string{"condition", "reason"},
+								LabelValues: []string{string(condition.Type), reason},
```
dgrisonnet, line 193: The condition will always be `Failed` here, is that what you want?

Shouldn't this label contain the value of `condition.Status` instead? If so, you might want to have a look at the `addConditionMetrics` function.

dgrisonnet (commented):
Just a few comments, otherwise it looks good overall. Thanks!

docs/job-metrics.md
```diff
@@ -10,7 +10,7 @@
 | kube_job_spec_active_deadline_seconds | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_status_active | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
 | kube_job_status_succeeded | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
-| kube_job_status_failed | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; | STABLE |
+| kube_job_status_failed | Gauge | `job_name`=&lt;job-name&gt; <br> `namespace`=&lt;job-namespace&gt; <br> `condition`=&lt;condition&gt; <br> `reason`=&lt;failure reason&gt; <br> `status`=&lt;status&gt;  | STABLE |
```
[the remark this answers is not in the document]
AdityaMisra: It was added earlier. I've updated it now. Thanks.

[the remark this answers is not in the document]
lilic: @AdityaMisra seems like you did not push your changes? I can't see a recent push. Do you mind double checking, thanks!

internal/store/job.go
AdityaMisra: Yes, we want only the `Failed` conditions as this metric is `kube_job_status_failed`.
If required then we can change the metric to `kube_job_status` and include other statuses as well.

internal/store/job.go
AdityaMisra: @lilic - Should I resolve this conversation?

internal/store/job.go
dgrisonnet: Then, I don't think we need to add this dimension to the metric. We already have the information that the job failed in the metric name, it would be redundant to also have it in a label.

internal/store/job.go
```diff
@@ -350,3 +383,10 @@ func createJobListWatch(kubeClient clientset.Interface, ns string) cache.ListerW
 		},
 	}
 }
+
+func failureReason(jc *v1batch.JobCondition, reason string) bool {
```
dgrisonnet, line 387: I don't think the job condition is needed here

AdityaMisra: Basically, here we want to check if JobCondition's reason is one of the `jobFailureReasons` - `{"BackoffLimitExceeded", "DeadLineExceeded", "Evicted"}`. 
If it's true then we return the reason otherwise we return an empty string.

dgrisonnet: Indeed, I misread the code. I thought you were only checking if the condition was nil.

dgrisonnet:
This PR looks good from my side, @lilic can you please have another look at it?

lilic (commented):
/lgtm

Thanks!

Issue #1198, opened 2020-08-03
Version detection during build time


mrueg:
**Is this a BUG REPORT or FEATURE REQUEST?**:

/kind feature

**What happened**:
Docker images get build in prow and currently overwrite the existing tag in the staging repository
**What you expected to happen**:
Tags are a stable target and don't change over time

See also: https://github.com/kubernetes/kube-state-metrics/issues/1089

**Possible solution**
    * Read VERSION file (if there's a need to keep it and it can't be replaced with a build-time variable that ingests a git tag)
    * Check if current commit is tagged with that version
    * If it does, create a docker image with the version of VERSION -> kube-state-metrics:v1.9.8
    * If it does not, create a docker image with the version of VERSION and add git shortcommit (that creates a tag per commit) e.g. `kube-state-metrics:v1.9.8-aabbccdd`
    or add "-next" or similar (that's a moving target then) `kube-state-metrics:v1.9.8-next`

tariq1890:
This is a valuable feature to have 👍 

/help

lilic:
I believe we can close this now @mrueg, correct? :)

mrueg:
Yes I think the way it works for containers is fine

lilic:
Thanks!!

PR #1215, opened 2020-08-30
Generate version from git tag

mrueg:
This drop the need to maintain a separate version file.

Partially resolves: https://github.com/kubernetes/kube-state-metrics/issues/1198

Probably after thinking about a bit, it would be helpful to have a "release" target, that takes care of some of the steps.


brancz:
We actually used to do this but stopped because this was not always yielding the correct behavior. I would need to dig up the old issues when we changed it to what we have now to understand exactly what happened.

lilic:
@brancz did you find a reason for it?

mrueg:
before moving on with this, we could also look into a github action release step that would automate this. @tariq1890 would you be able to help out with more info here?

PR #1216, opened 2020-08-31
Label/Tags are shown in status tab

vinaykv1991:
Label/Tags for the first few metrics are shown in the status tab and status is not shown on UI.

Removed a wrong demarkation.

Correcting the docs of Vertical Pod Autoscaler.
Fixes #


brancz (approved):
Thank you for fixing!

/approve
/lgtm

PR #1217, opened 2020-09-02
Fix test as sometimes metrics keys and values are generated in differ…

jw-s:
This fixes the failed unit tests as the metrics generated sometimes have keys and values in a different order.

### DIFF
--- a/pkg/metric_generator/generator_test.go
+++ b/pkg/metric_generator/generator_test.go
@@ -18,6 +18,7 @@ package generator
 
 import (
 	"reflect"
+	"sort"
 	"testing"
 
 	"k8s.io/kube-state-metrics/pkg/allow"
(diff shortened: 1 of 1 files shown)


jw-s:
@tariq1890 This fixes the unit test issue which for some reason never happened locally or in the feature branch builds...

tariq1890:
Thanks @jw-s ! Can you move the sorting and checking routine into a private method? Just to reduce code duplication.

Looks like I will need to look into the CI issue as well

jw-s:
> Thanks @jw-s ! Can you move the sorting and checking routine into a private method? Just to reduce code duplication.
> 
> Looks like I will need to look into the CI issue as well

Done!

tariq1890:
Perfect!

/lgtm

PR #1224, opened 2020-09-14
Cut v2.0.0.-alpha

lilic:
This is work in progress, mainly to look at changelog entries. I want to test the release and update README a bit more. But CHANGELOG is ready for review.

I was using https://github.com/kubernetes/kube-state-metrics/compare/release-1.9...master to compile the changelog list. Would be good to double check to make sure it includes everything, as I feel like we are missing a few things 🤔 

cc @brancz @tariq1890 PTAL 🎉


brancz:
Compared quickly against

```
git log --pretty=format:%s --merges v1.9.7..HEAD
```

I don't see anything obvious that we missed. Let's move ahead?

PR #1233, opened 2020-09-17
Added Namespace to Rolebinding Jsonnet

### LINKED ISSUE #1232: Autosharding Rolebinding Missing Namespace - Jsonnet

/kind bug

**What happened**:
I created the autosharding role and rolebinding, the role was created in the correct namespace, but the rolebinding was created in default.

I expected the rolebinding to be created in the same namespace as the role

Apply autosharding jsonnet from master and check default namespace for rolebinding.

midnightconman:
Without this, autosharding won't work correctly... on startup pods will be unable to communicate with kubernetes api servers correctly and won't collect metrics.

Fixes #1232


tariq1890 (approved):
/lgtm

PR #1234, opened 2020-09-17
e2e.sh: Update kind to 0.9.0

mrueg:
Small updates to the versions used for e2e testing


tests/e2e.sh
```diff
@@ -23,12 +23,12 @@ case $(uname -m) in
 	*)		ARCH="$(uname -m)";;
 esac
 
-KUBERNETES_VERSION=v1.19.1
+KUBERNETES_VERSION=v1.19.2
```
tariq1890, line 26: unfortunately, kind's docker images don't have the same release cadence like that of kubernetes.

mrueg: Thanks, yes figured it after CI run and dropped the change now.

lilic (commented):
/lgtm

Thanks!

PR #1236, opened 2020-09-18
docs/pod-metrics.md: add missing condition label

dgrisonnet:
The condition label of the [kube_job_complete](https://github.com/kubernetes/kube-state-metrics/blob/315824ab81ff123c6c3bf016917c4d387f0d435f/internal/store/job.go#L194-L216) and [kube_job_failed](https://github.com/kubernetes/kube-state-metrics/blob/315824ab81ff123c6c3bf016917c4d387f0d435f/internal/store/job.go#L217-L240) metrics
were forgotten in the documentation.


brancz:
Nice find!

/lgtm

PR #1238, opened 2020-09-21
Update go module path to k8s.io/kube-state-metrics/v2

omegas27:
Signed-off-by: Quentin Bisson <quentin@giantswarm.io>

Updating go module path to v2 towards https://github.com/kubernetes/kube-state-metrics/issues/569


go.mod
```diff
@@ -1,20 +1,28 @@
-module k8s.io/kube-state-metrics
+module k8s.io/kube-state-metrics/v2
 
 require (
+	github.com/Azure/go-autorest/autorest v0.10.0 // indirect
```
lilic, line 4: Curious of all the changes that were added?

lilic (commented):
Thanks for the PR, have one question!

cc @brancz @tariq1890 PTAL

go.mod
omegas27: I just triggered `go mod tidy` but maybe I should have done it in a different way?

lilic: It's not a problem at all, mainly asking as we will be porting this to release 2.0 branch and did not want to introduce any new not needed changes, but don't think its a problem here.

PR #1244, opened 2020-10-03
feat(persistentvolume): claimRef info to labels

Duologic:

Adds a new metrics to provide claimRef information to Prometheus. In
tracking down lingering PVs, it would be of interest to know where the
PV came from. With the claimRef information, it is easier to track down
the resource owners and inform that they have unclaimed PVs lingering
around.

Main motivation is that PVs cost money at cloud providers, if PVs stick
around for long, the costs might increase significantly.

### DIFF
--- a/internal/store/persistentvolume.go
+++ b/internal/store/persistentvolume.go
@@ -32,11 +32,45 @@ import (
 )
 
 var (
+	descPersistentVolumeClaimRefName          = "kube_persistentvolume_claim_ref"
+	descPersistentVolumeClaimRefHelp          = "Information about the Persitant Volume Claim Reference."
+	descPersistentVolumeClaimRefDefaultLabels = []string{"persistentvolume"}
+
 	descPersistentVolumeLabelsName          = "kube_persistentvolume_labels"
 	descPersistentVolumeLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descPersistentVolumeLabelsDefaultLabels = []string{"persistentvolume"}
 
 	persistentVolumeMetricFamilies = []generator.FamilyGenerator{
+		*generator.NewFamilyGenerator(
+			descPersistentVolumeClaimRefName,
+			descPersistentVolumeClaimRefHelp,
+			metric.Gauge,
+			"",
+			wrapPersistentVolumeFunc(func(p *v1.PersistentVolume) *metric.Family {
+				claimRef := p.Spec.ClaimRef
+
+				if claimRef == nil {
+					return &metric.Family{
+						Metrics: []*metric.Metric{},
+					}
+				}
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys: []string{
+								"name",
+								"claim_namespace",
+							},
+							LabelValues: []string{
+								p.Spec.ClaimRef.Name,
+								p.Spec.ClaimRef.Namespace,
+							},
+							Value: 1,
+						},
+					},
+				}
+			}),
+		),
 		*generator.NewFamilyGenerator(
 			descPersistentVolumeLabelsName,
 			descPersistentVolumeLabelsHelp,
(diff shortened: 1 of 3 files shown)


Duologic:
I'm a bit unsure what to do best if there is no `spec.claimRef`. Feel free to put in some suggestions.

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,49 @@ import (
 )
 
 var (
+	descPersistentVolumeClaimRefName          = "kube_persistentvolume_claimref_info"
```
lilic, line 35: `kube_persistentvolume_claim_ref_info` might be better.

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,49 @@ import (
...
+	descPersistentVolumeClaimRefHelp          = "PersistentVolume spec.claimRef converted to Prometheus labels."
```
lilic, line 36: Let's add this to the metric itself instead of here.

```suggestion
	descPersistentVolumeClaimRefHelp          = "Information about the Persitant Volume Claim Reference."
```

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,49 @@ import (
...
+					}
+				}
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys: []string{
+								"apiversion",
+								"kind",
+								"name",
+								"namespace",
```
lilic, line 64: I would call this `claim_namespace`, but open to other suggestions, otherwise would be confusing if this is the namespace of the PV or the PVC.

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,49 @@ import (
...
+				if claimRef == nil {
+					return &metric.Family{
+						Metrics: []*metric.Metric{},
+					}
+				}
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys: []string{
+								"apiversion",
```
lilic, line 61: Can you explain the usecase for the apiserver value?

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,49 @@ import (
...
+								"kind",
```
lilic, line 62: Can this be anything else but PVC?

lilic (commented):
Thanks for your PR, couple of comments and you need to also generate the docs, hence the failing tets.

internal/store/persistentvolume.go
Duologic: I didn't want to make an assumption here, the same reason I kept the apiversion around. I have no hard arguments to keep it around.

brancz:
Can claimref change over the lifetime of a PV? If not then this would be a better fit for the general info metric I would say.

Duologic:
Considering a PV can exist without a PVC, a PV could be unclaimed and thus without a claimref, the claimref is added/updated by a process that binds a PV to a PVC. I do think that once the PV has a claimref, it is very unlikely that the name/namespace will change, but technically it could as it is part of the 'binding' process.

I'm up for adding it to the info metric, but I don't know what impact that change might have, feel free to tell me where to put it as I lack the confidence to make a decision. :blush: 

Upon reading a bit in random github issues to see how relevant the apiversion/kind is, they seem unlikely to contain anything else than v1/PVC, I think we can drop those labels because of that.

lilic:
> Upon reading a bit in random github issues to see how relevant the apiversion/kind is, they seem unlikely to contain anything else than v1/PVC, I think we can drop those labels because of that.

Yep my thinking exactly. 👍

brancz:
Taking all this information into consideration, I would say: let's drop the kind/apiversion, and have this be a separate metric and not part of the info metric since there is a chance this may change over the lifetime of the PV.

Duologic:
That should do it, shall I squash it or do we do that on merge?

internal/store/persistentvolume.go
```diff
@@ -32,11 +32,45 @@ import (
 )
 
 var (
+	descPersistentVolumeClaimRefName          = "kube_persistentvolume_claim_ref_info"
```
brancz, line 35: We tend to reserve the `_info` suffix for things that never change over the lifetime of an object, so this should just be `kube_persistentvolume_claim_ref`.

brancz (commented):
Small nit on naming, then should be good to go.

brancz:
Let's squash the changes I'd say.

Duologic:
Addressed comments, squashed and rebased on current master.

lilic (commented):
/lgtm

Thanks for the contribution!

PR #1247, opened 2020-10-04
Support regex for label allow list

### LINKED ISSUE #1246: --labels-allow-list flag does not support wildcards

>
> /kind bug
/kind feature

**What happened**:
--labels-allow-list flag does not support wildcards.  need to specify each and every label of interest.

--labels-allow-list flag to support wildcards.  should be able to configure kube-state-metrics to export labels without specifying each and every one.

for example:
--labels-allow-list=kube_pod_labels=[label_*] or --labels-allow-list=kube_pod_labels=[*]

install kube-state-metrics from https://github.com/kubernetes/kube-state-metrics/tree/master/examples/standard. enable --labels-allow-list flag using wildcard in argument.  observe invalid argument failure.  for example:

invalid argument "kube_pod_labels=[label_*]" for "--labels-allow-list" flag: invalid format, metric=[label1,label2,labeln...],metricN=[]

there are scenarios in which users will want to export all labels for all k8s resources, in which users will want to export all labels for a particular k8s resource, in which users will want to export all labels containing a certain substring, etc, etc.  the current implementation of the flag does not allow for this unless users (1) know exactly what flags they want to export a priori and (2) specify the complete list of all flags they want for each k8s resource.

**Environment**:
Server Version: version.Info{Major:"1", Minor:"11+", GitVersion:"v1.11.0+d4cacc0", GitCommit:"d4cacc0", GitTreeState:"clean", BuildDate:"2020-10-01T19:23:04Z", GoVersion:"go1.10.8", Compiler:"gc", Platform:"linux/amd64"}

quay.io/coreos/kube-state-metrics:v2.0.0-alpha

jw-s:

Supports wildcards for label allow list
Fixes #1246

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -32,7 +32,7 @@ Usage of ./kube-state-metrics:
   -h, --help                             Print Help text
       --host string                      Host to expose metrics on. (default "0.0.0.0")
       --kubeconfig string                Absolute path to the kubeconfig file
-      --labels-allow-list string         list of metric names and labels you would like to allow, metric_name=[label1,labeln...],metric_name[]...
+      --labels-allow-list string         list of metric names and labels you would like to allow, metric_name=[label1,^label2.*,labeln...],metric_name[]...
       --log_backtrace_at traceLocation   when logging hits line file:N, emit a stack trace (default :0)
       --log_dir string                   If non-empty, write log files in this directory
       --log_file string                  If non-empty, use this log file
--- a/pkg/allow/allow_labels.go
+++ b/pkg/allow/allow_labels.go
@@ -20,15 +20,32 @@ import (
 	"regexp"
 )
 
-var defaultMetricLabels = map[*regexp.Regexp][]string{
+var defaultMetricLabels = map[*regexp.Regexp][]*regexp.Regexp{
 	regexp.MustCompile(".*_labels$"):      {},
 	regexp.MustCompile(".*_annotations$"): {},
+}
+
+// ParseLabels parses and compiles all of the regexes for Labels
+func ParseLabels(l map[string][]string) (Labels, error) {
+	labels := make(map[string][]*regexp.Regexp)
+	for metric, lbls := range l {
+		var labelsRegex []*regexp.Regexp
+		for _, lbl := range lbls {
+			reg, err := regexp.Compile(lbl)
+			if err != nil {
+				return nil, err
+			}
+			labelsRegex = append(labelsRegex, reg)
+		}
+		labels[metric] = labelsRegex
+	}
+	return labels, nil
 }
 
 // Labels provide a way to allow only specified labels for metrics.
 // Falls back to default labels, if your metric doesn't have defaults
 // then all labels are allowed.
-type Labels map[string][]string
+type Labels map[string][]*regexp.Regexp
 
 // Allowed returns allowed metric keys and values for the metric.
 func (a Labels) Allowed(metric string, labels, values []string) ([]string, []string) {
@@ -49,11 +66,17 @@ func (a Labels) Allowed(metric string, labels, values []string) ([]string, []str
 
 	var finalLabels, finalValues []string
 	labelSet := labelSet(allowedLabels)
+	alreadyMatched := make(map[string]interface{}) // used to prevent duplicate labels
 	for _, allowedLabel := range labelSet {
 		for i, label := range labels {
-			if label == allowedLabel {
+			_, ok := alreadyMatched[label]
+			if ok {
+				continue
+			}
+			if allowedLabel.MatchString(label) {
 				finalLabels = append(finalLabels, label)
 				finalValues = append(finalValues, values[i])
+				alreadyMatched[label] = struct{}{}
 			}
 		}
 	}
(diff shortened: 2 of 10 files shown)


pkg/allow/allow_labels.go
```diff
@@ -49,11 +51,23 @@ func (a Labels) Allowed(metric string, labels, values []string) ([]string, []str
...
+			if ok {
+				continue
+			}
+			allowedLabelReg, err := regexp.Compile(allowedLabel)
+			if err != nil {
+				klog.Error(err)
+				continue
+			}
+
+			if allowedLabelReg.MatchString(label) {
```
jimmyseto, line 67: do we need to regulate this more?  for example, if the user specifies "label1" in the allow list, this would let "label1*" labels through (i.e. "label11", "label111", etc).

pkg/metric_generator/generator_test.go
```diff
@@ -238,6 +238,95 @@ func TestFilterMetricFamiliesLabels(t *testing.T) {
...
+									LabelKeys:   []string{"one", "myprefix-label", "myprefix-label2"},
+									LabelValues: []string{"value-one", "value-two", "value-three"},
+									Value:       1,
+								},
+							},
+						}
+					},
+				},
+			},
+		},
```
jimmyseto, line 329: can we add a test to make sure only specified labels are allowed?  for example, something like this...

                {
                        name: "Returns only specified labels",
                        allowLabels: allow.Labels(map[string][]string{
                                "node_labels": {
                                        "label1",
                                },
                        }),
                        familyGenerators: []FamilyGenerator{
                                {
                                        Name: "node_labels",
                                        Help: "some help",
                                        Type: metric.Gauge,
                                        GenerateFunc: func(obj interface{}) *metric.Family {
                                                return &metric.Family{
                                                        Metrics: []*metric.Metric{
                                                                {
                                                                        LabelKeys:   []string{"label1", "label11", "label111"},
                                                                        LabelValues: []string{"value-one", "value-two", "value-three"},
                                                                        Value:       1,
                                                                },
                                                        },
                                                }
                                        },
                                },
                        },
                        results: []FamilyGenerator{
                                {
                                        Name: "node_labels",
                                        Help: "some help",
                                        Type: metric.Gauge,
                                        GenerateFunc: func(obj interface{}) *metric.Family {
                                                return &metric.Family{
                                                        Metrics: []*metric.Metric{
                                                                {
                                                                        LabelKeys:   []string{"label1"},
                                                                        LabelValues: []string{"value-one"},
                                                                        Value:       1,
                                                                },
                                                        },
                                                }
                                        },
                                },
                        },
                },

pkg/allow/allow_labels.go
jw-s: Since it's just regex, one can specify `^label1$`. WDYT? I don't think we should support implicit `^<LABEL>$`

jimmyseto: works for me.  can we update the doc(s) to reflect this?

pkg/allow/allow_labels.go
```diff
@@ -49,11 +51,23 @@ func (a Labels) Allowed(metric string, labels, values []string) ([]string, []str
...
 	labelSet := labelSet(allowedLabels)
+	alreadyMatched := make(map[string]interface{}) // used to prevent duplicate labels
 	for _, allowedLabel := range labelSet {
 		for i, label := range labels {
-			if label == allowedLabel {
+			_, ok := alreadyMatched[label]
+			if ok {
+				continue
+			}
+			allowedLabelReg, err := regexp.Compile(allowedLabel)
```
dgrisonnet, line 61: regex should be compiled only once as this operation is expensive.

pkg/allow/allow_labels.go
```diff
@@ -49,11 +51,23 @@ func (a Labels) Allowed(metric string, labels, values []string) ([]string, []str
...
+			if err != nil {
+				klog.Error(err)
```
dgrisonnet, line 63: I think we should check if the regex are valid right away and exit if the configuration is invalid. This will avoid users running a misconfigured ksm without noticing. IIRC the `allow-deny-list` flag already has a similar mechanism.

dgrisonnet (commented):
Nice!

pkg/allow/allow_labels.go
jw-s: Perfect, i'll take a look!

lilic:
Hey just FYI the cutoff for any new features is tomorrow afternoon for release 2.0, after that we are calling it feature freeze. Otherwise we will include this in the next release, which is fine as well! :)

jimmyseto:
> Hey just FYI the cutoff for any new features is tomorrow afternoon for release 2.0, after that we are calling it feature freeze. Otherwise we will include this in the next release, which is fine as well! :)

it would be great if we could get this into 2.0!

jw-s:
@lilic @jimmyseto This is ready for review again

docs/cli-arguments.md
```diff
@@ -32,7 +32,7 @@ Usage of ./kube-state-metrics:
   -h, --help                             Print Help text
       --host string                      Host to expose metrics on. (default "0.0.0.0")
       --kubeconfig string                Absolute path to the kubeconfig file
-      --labels-allow-list string         list of metric names and labels you would like to allow, metric_name=[label1,labeln...],metric_name[]...
+      --labels-allow-list string         list of metric names and labels you would like to allow, metric_name=[label1,^label2.*,labeln...],metric_name[]...
```
jimmyseto, line 35: will this regex (^label2.*) get past labelsAllowListFormat.MatchString() in pkg/options/types.go?

should we update errLabelsAllowListFormat in pkg/options/types.go to be consistent?

pkg/allow/allow_labels.go
dsonck92: I do think this might need some more thought. The example `label1` actually also allows `some_label1`. Implicit `^<LABEL>$` would prevent that from happening but one can still specify `.*label1.*` for the same functionality as it does right now.

Having users require putting in `^label1$` for each exact label essentially creates the same issue as this one attempts to solve: adding more extra information that would have been necessary. Granted, it's probably not too much as most would probably want to use just a few well designed regex.

docs/cli-arguments.md
jw-s: This regex won't pass the command-line flag parsing based on the current approach. It will take some work to get it to work, we should definitely think about this more and not approach it lightly, cc @lilic

PR #1256, opened 2020-10-08
Fix test due to map iteration

### LINKED ISSUE #1242: Flake unit test
We have a flaky unit test which might be just the way we have setup the unit test or an actual problem with how we generate the series, where labels are expected to be in one order but we get them in reverse. In any case we should make it be a deterministic order.

Example: https://github.com/kubernetes/kube-state-metrics/runs/1181030112

> --- FAIL: TestFullScrapeCycle (1.07s)
>    main_test.go:379: expected:   
 >       kube_pod_labels{namespace="default",pod="pod0"} 1, but got:
  >      kube_pod_labels{pod="pod0",namespace="default"} 1

jw-s:

Fixes the "flaky" test.
Actually the reason for occasional failure is the expected behaviour of map iteration, as order is not guaranteed.
Fixes #1242

### DIFF
--- a/pkg/allow/allow_labels.go
+++ b/pkg/allow/allow_labels.go
@@ -49,7 +49,7 @@ func (a Labels) Allowed(metric string, labels, values []string) ([]string, []str
 
 	var finalLabels, finalValues []string
 	labelSet := labelSet(allowedLabels)
-	for allowedLabel := range labelSet {
+	for _, allowedLabel := range labelSet {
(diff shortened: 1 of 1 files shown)


lilic:
This should have been merged to the release-2.0 branch first as its a bug fix.

But now we already did it, @tariq1890 how do we proceed here? Or with newer merges, do we just merge to master and pull all to release-2.0 branch afterwards? cc @brancz

PR #1261, opened 2020-10-14
Update to golang-1.15.3

mrueg:
Move to latest version of golang (1.15)


Makefile
```diff
@@ -12,7 +12,7 @@ BuildDate = $(shell date -u +'%Y-%m-%dT%H:%M:%SZ')
 GIT_COMMIT ?= $(shell git rev-parse --short HEAD)
 ALL_ARCH = amd64 arm arm64 ppc64le s390x
 PKG = k8s.io/kube-state-metrics/v2/pkg
-GO_VERSION = 1.14.7
+GO_VERSION = 1.14.10
```
lilic, line 15: Would you mind making the change to 1.15.3 all over the repo? Planned on doing the switch myself but since you opened the PR. 

We have a few other places if you are willing to migrate to 1.15 :) 
https://github.com/kubernetes/kube-state-metrics/blob/f1166b40298a6846a329f567d0d509d32ee45a81/.github/workflows/ci.yml

mrueg: Sure, will do!

go.mod
```diff
@@ -28,4 +28,4 @@ require (
 	k8s.io/klog/v2 v2.2.0
 )
 
-go 1.14
+go 1.15
```
lilic, line 31: This is not needed as we are not using anything specific to go 1.15 so our users don't require to switch to 1.15, so should be compatible with 1.14.

lilic (commented):
One comment, otherwise 👍

lilic (approved):
/lgtm

PR #1262, opened 2020-10-14
Makefile: Override Commit in cloudbuild

mrueg:
```
docker run -it --rm "k8s.gcr.io/kube-state-metrics/kube-state-metrics:2.0.0-alpha.1" --version
version.Version{GitCommit:"", BuildDate:"2020-10-07T08:16:51Z", Release:"v2.0.0-alpha.1", GoVersion:"go1.14.7", Compiler:"gc", Platform:"linux/amd64"}
```
returns an empty GitCommit value, so we need to use the value set in cloudbuild.yaml

@lilic


lilic (commented):
/lgtm

Thanks!

PR #1264, opened 2020-10-15
Makefile: Set tag via cloudbuild

mrueg:
This allows to set the container tag via cloudbuild so we don't override an existing git tag anymore but instead use the _PULL_BASE_REF (e.g. 'master', 'release-1.9') on new commits. If the commit is tagged, it will use the tag name instead (e.g. 'v2.0.0-alpha2').

Also change variable style for BuildDate to match the style of the other variables.


lilic (commented):
/lgtm

Thanks!

PR #1265, opened 2020-10-16
fix html tag in service-metrics.md

jangdm:

fix html tag


brancz:
Thanks!

/lgtm
/approve

PR #1275, opened 2020-10-22
Remove ksonnet and move to absolute import paths

metalmatze:

In an effort to move kube-prometheus to use absolute import paths (for better discoverability) I needed to refactor these jsonnet bits.
Next to that, I also refactored the Kubernetes object generation to use plain Jsonnet objects and not ksonnet anymore.

Over the last months/years, we've come to the conclusion that readability is more important than the "type safety" that ksonnet promised.
Instead we're going for external tools that check the generated YAML after it has been generated like [kubeval](https://kubeval.com)

Fixes issues outside this repository.
I've talked to @lilic before making this PR though :innocent:

### DIFF
--- a/jsonnet/kube-state-metrics/jsonnetfile.json
+++ b/jsonnet/kube-state-metrics/jsonnetfile.json
@@ -1,14 +1,3 @@
 {
-    "dependencies": [
-        {
-            "name": "ksonnet",
-            "source": {
-                "git": {
-                    "remote": "https://github.com/ksonnet/ksonnet-lib",
-                    "subdir": ""
-                }
-            },
-            "version": "master"
-        }
-    ]
+    "dependencies": []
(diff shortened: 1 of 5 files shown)


jsonnet/kube-state-metrics/jsonnetfile.json
```diff
@@ -1,14 +1,3 @@
...
-            "source": {
-                "git": {
-                    "remote": "https://github.com/ksonnet/ksonnet-lib",
-                    "subdir": ""
-                }
-            },
-            "version": "master"
-        }
-    ]
+    "dependencies": []
```
lilic, line 2: 🎉 :)

lilic (commented):
Nice, thanks so much!!
 
/lgtm

Issue #1274, opened 2020-10-21
Add RuntimeClass metrics


tallclair:
We should add RuntimeClass metrics. Maybe as an additional label on `kube_pod_info`?

/cc @egernst @SergeyKanzhelev 

/kind feature

egernst:
Thanks @tallclair 

I think we want something similar to https://github.com/kubernetes/kube-state-metrics/pull/1053, but for runtimeclass itself.

lilic:
@egernst @egernst can you please explain the feature request in a bit more details so we can evaluate if we want this added or not, thanks!

egernst:
@lilic thanks.  As a user I would like to verify that a given RuntimeClass handler is being utilized.

When combined with kube_pod_status_phase, end-users can use tools like Grafana to quickly assess how pods running with various runtimeclass handlers compare. This is very similar to what was done for PodOverhead: https://github.com/kubernetes/kube-state-metrics/pull/1053

I would want to add something *like* kube_pod_runtimeclass_handler

lilic:
Thanks @egernst sounds good to me! Are you interested in contributing this, otherwise we can mark it as help wanted. Note we are in the release 2.0 cycle so might be some delay in getting PRs merged.

egernst:
Is there a cutoff for 2.0? I saw alpha from earlier this month (thanks!). 

In a perfect world, I would contribute this ASAP and it would be available in 2.0.  Do you think this is feasible?

PR #1276, opened 2020-10-23
pod: add gauge for runtimeclass handler

egernst:

This adds a metric for observability of the RuntimeClass feature:
kube_pod_runtimeclass_handler

Using this metrics allows end-users to quickly verify that the RuntimeClass feature is being utilized and is healthy.

When combined with the kube_pod_status_phase, end-users can use tools like grafana to quickly assess how pods running with a particular runtimeclass are behaving compared to other runtimeclasses and the default runtime. 

Fixes: https://github.com/kubernetes/kube-state-metrics/issues/1274

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1317,6 +1317,27 @@ var (
 				}
 			}),
 		),
+		*generator.NewFamilyGenerator(
+			"kube_pod_runtimeclass_name_info",
+			"The runtimeclass associated with the pod.",
+			metric.Gauge,
+			"",
+			wrapPodFunc(func(p *v1.Pod) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if p.Spec.RuntimeClassName != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"runtimeclass_name"},
+						LabelValues: []string{*p.Spec.RuntimeClassName},
+						Value:       1,
+					})
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		),
 	}
 )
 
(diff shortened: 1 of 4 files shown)


SergeyKanzhelev:
I have limited context here. From looking at the code around it looks consistent and seems to be taking the proper value for the dimension. As next step maybe we can add supported runtime handlers to the [kube_node_info](https://github.com/kubernetes/kube-state-metrics/blob/2e1ff1f529ea72fc857d1e595a85cfaab57a2c80/docs/node-metrics.md#user-content-node-metrics:~:text=kube_node_info)

/lgtm

egernst:
> I have limited context here. From looking at the code around it looks consistent and seems to be taking the proper value for the dimension. As next step maybe we can add supported runtime handlers to the [kube_node_info](https://github.com/kubernetes/kube-state-metrics/blob/2e1ff1f529ea72fc857d1e595a85cfaab57a2c80/docs/node-metrics.md#user-content-node-metrics:~:text=kube_node_info)
> 
> /lgtm

Interesting. I'm not sure there's an easy way for the node to know in a clean way what runtime handlers are available. RuntimeClass *can* optionally specify the nodeSelect, but that isn't always necessarily used.

internal/store/pod.go
```diff
@@ -1511,6 +1511,27 @@ var (
 					}
 				}
 
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		),
+		*generator.NewFamilyGenerator(
+			"kube_pod_runtimeclass_handler",
```
lilic, line 1520: Can we add a unit test for this, thanks! :)

lilic:
> As next step maybe we can add supported runtime handlers to the kube_node_info

What would be the use of that metric?

lilic:
To answer the issue question: If you manage to address the comment before our next release (which will be tomorrow most likely) we can merge this into 2.0.0-alpha.2 cut, otherwise it will have to wait until 2.0 is shipped as we are going into code freeze after that for 2.0.

internal/store/pod.go
lilic: As per best practices the name should be `kube_pod_runtimeclass_handler_info`, can you adjust that as well, thanks!

egernst:
Thanks for the quick feedback @lilic.  I updated the commit to include unit test (whoops sorry!) and to add `_info` to the metric name.

I also changed the metric name from runtimeclass_handler to runtimeclass_name.

lilic (approved):
/lgtm

Thanks!

Issue #1263, opened 2020-10-14
Stable metrics "kube_pod_container_resource_[limits|requests]" have cpu/memory units removed in v2


mcavoyk:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
In the latest v1 version (v1.9.7), `kube_pod_container_resource_[limits|requests]` is [marked as stable](https://github.com/kubernetes/kube-state-metrics/blob/v1.9.7/docs/pod-metrics.md#pod-metrics), and the duplicates `kube_pod_container_resource_[limits|requests]_[cpu_cores|memory_bytes]` are marked as deprecated. The deprecated metrics were removed with https://github.com/kubernetes/kube-state-metrics/pull/1004.

https://github.com/kubernetes/kube-state-metrics/pull/1168 then removes the memory/cpu units from the previously stable `kube_pod_container_resource_[limits|requests]` metrics and re-adds the deprecated and removed metrics.

**What you expected to happen**:
I had expected `kube_pod_container_resource_[limits|requests]` to retain the memory/cpu units in v2 and the `kube_pod_container_resource_[limits|requests]_[cpu_cores|memory_bytes]`  to be removed, instead this was flipped with what feels like little communication and no v1 version which marked these metrics as either not deprecated, experimental, or otherwise.

The changelog for v2 doesn't explain these changes particularly well. This change seems like it should need a v1.10 with more of these distinctions made clear, or v2 with a flag to re-enable the cpu/memory units on `kube_pod_container_resource_[limits|requests]`.

**Environment**:
- Kubernetes version (use `kubectl version`):  v1.17
- Kube-state-metrics image version: v2.0.0-alpha.1

mcavoyk:
Similiar to https://github.com/kubernetes/kube-state-metrics/issues/1245, but this is more targeted around the communication of these changes and how the latest v1 version seems to indicate the opposite of these changes.

lilic:
@mcavoyk hey thanks for the issue, yes the CHANGELOG is rather large and a lot PRs went in, we can help clarify it no problem. What would make most sense for you to understand the changes more clear, e.g. what should we write instead?

lilic:
@mcavoyk thanks for reporting the issue and bringing this issue to light, reason for closure is in the PR:
[...]

PR #1278, opened 2020-10-26
Revert "Rework resource metrics"

lilic:
After some talk with @brancz who initially did the work for reworking resource metrics we decided to revert the PR, as the KEP that will introduce more detailed resource metrics has been accepted now https://github.com/kubernetes/enhancements/pull/1916 this better aligns the resource metrics to the ones proposed in that KEP, this will prevent users from needing to do triple migration if they want to switch any alerting/recording rules to new metrics and avoid potential confusion. If we were to keep current metrics with the unit suffix the migration work would double, this way potentially only metric name would need to be adjusted.

Closes https://github.com/kubernetes/kube-state-metrics/issues/1263

cc @brancz as agreed before. @tariq1890 PTAL, thanks!


brancz:
If I'm reading the e2e test failure right, then looks like docs need to be adapted.

lilic:
@brancz yes :) PTAL, I updated the PR.

lilic:
@TomasKohout all the metrics in the docs should be up to date, in your case you are interested in pod metrics -> https://github.com/kubernetes/kube-state-metrics/blob/release-2.0/docs/pod-metrics.md. So that should be your source of truth always.

TomasKohout:
Well, I'm upgrading from 1.9.7 to 2.0.0. The main reason, why I'm pointing this out is because of this statement:

```
All deprecated stable metrics were removed as per the notice in the v1.9 release.
```

Link points [here](https://github.com/kubernetes/kube-state-metrics/tree/release-1.9/docs#metrics-deprecation) where metrics `kube_pod_container_resource_requests` and `kube_pod_container_resource_limits` are deprecated and according to announcement they should be removed in 2.0.0, but they're not.

TomasKohout:
I've found out that I was mistaken and that I should read more carefully! Sorry for that. :slightly_frowning_face:

PR #1279, opened 2020-10-26
Update README.md to reflect new --labels-allow-list

dsonck92:
This PR adds a new section to the README.md file specifying how the new `--labels-allow-list` can be used and why it's necessary.

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,6 +35,7 @@ are deleted they are no longer visible on the `/metrics` endpoint.
   - [Container Image](#container-image)
 - [Metrics Documentation](#metrics-documentation)
   - [Conflict resolution in label names](#conflict-resolution-in-label-names)
+  - [Exposing kubernetes labels](#enabling-kubernetes-labels)
   - [Enabling VerticalPodAutoscalers](#enabling-verticalpodautoscalers)
 - [Kube-state-metrics self metrics](#kube-state-metrics-self-metrics)
 - [Resource recommendation](#resource-recommendation)
@@ -51,6 +52,8 @@ are deleted they are no longer visible on the `/metrics` endpoint.
   - [Limited privileges environment](#limited-privileges-environment)
   - [Development](#development)
   - [Developer Contributions](#developer-contributions)
+
+### Versioning
 
 #### Kubernetes Version
 
@@ -114,6 +117,19 @@ you might want to consider addressing this issue on a different level of the sta
 e.g. by standardizing Kubernetes labels using an
 [Admission Webhook](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/)
 that ensures that there are no possible conflicts.
+
+#### Enabling Kubernetes labels
+
+In order to prevent high cardinality of the `*_labels` family of metrics, kube-state-metrics only exposes specifically allowed labels.
+The labels can be configured through the use of the `--labels-allow-list` flag. The flag accepts a comma separated
+list of `<metric>=[<labels>]`, where `<labels>` are again comma separated full label names. For example:
+
+```
+--labels-allow-list kube_deployment_labels[label_app,label_env,deployment,namespace],kube_pod_labels[label_app,label_env,pod,namespace]
+```
+
+Note that the current implementation requires the inclusion of all wanted labels, this includes the metrics own labels like `deployment`, `node`, `pod`, etc. and `namespace`.
+When `--labels-allow-list` is specified multiple times, only the last will be honoured.
 
 #### Enabling VerticalPodAutoscalers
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -115,6 +118,19 @@ e.g. by standardizing Kubernetes labels using an
...
 that ensures that there are no possible conflicts.
 
+#### Enabling Kubernetes labels
+
+In order to prevent high cardinality of the `*_labels` family of metrics, kube-state-metrics only exposes specifically allowed labels.
+The labels can be configured through the use of the `--labels-allow-list` flag. The flag accepts a comma separated
+list of `<metric>=[<labels>]`, where `<labels>` are again comma separated full label names. For example:
+
+```
+--labels-allow-list kube_deployment_labels[label_app,label_env,deployment,namespace],kube_pod_labels[label_app,label_env,pod,namespace]
```
gturbat, line 128:
```suggestion
--labels-allow-list kube_deployment_labels=[label_app,label_env,deployment,namespace],kube_pod_labels=[label_app,label_env,pod,namespace]
```

lilic: The namespace and pod name will be on by default, its currently a bug that is being fixed, rest lgtm.

README.md
```diff
@@ -115,6 +118,19 @@ e.g. by standardizing Kubernetes labels using an
...
+```
+
+Note that the current implementation requires the inclusion of all wanted labels, this includes the metrics own labels like `deployment`, `node`, `pod`, etc. and `namespace`.
```
lilic, line 131: This can be removed as mentioned above.

lilic (commented):
Thanks, few suggestion.

README.md
dsonck92: Yeah, I figured. I was wondering whether the fix for that bug would need more time, hence the "current implementation", as the bugfix could always patch the documentation. Regardless, I've updated the PR with the suggestions.

README.md
gturbat: My review was not about namespace an pod name, but about missing equals. They are now added so it's fine

dsonck92: @gturbat agreed and exactly why I submitted the last edit. But I have to admit that the comment below was throwing me off :sweat_smile:. But as I also read the mails from Github, it actually became clear what your intention was.

ntavares:
Hi all, may I suggest to update https://github.com/kubernetes/kube-state-metrics/blob/master/docs/node-metrics.md this too? We've recently tripped on a total kube_node_labels disappearance, and only with a bit of digging I found this (via #1270) and https://github.com/kubernetes/kube-state-metrics/pull/1125 left me confused what the status of this is...

dsonck92:
@ntavares I noticed this too when I went through the documentation, however, I was unsure how to properly add that. Because this would need to be done too on pod-metrics.md, ingress-metrics.md, etc. as node-metrics.md is not the only place _labels metrics are present.

I simply didn't really have an idea on how to update the table. Whether a footnote would be sufficient. But maybe you have some suggestions I can apply.

ntavares:
@dsonck92 indeed, I wasn't even aware of those other pages. I think a footnote on them all should suffice, once this change stabilizes...

Is this change definitive? This change is very radical as we have dozens of alerts that relied on some of these labels to work, and they now go silently innefective. I'm looking at https://github.com/kubernetes/kube-state-metrics/issues/1270#issuecomment-724023848 , and it seems very recent still.

If it helps: we're working on cluster upgrades, and the old KSM was flooding the logs with API changes. We figured out that the way to go would be to upgrade it to latest (despite still missing some metrics, AFAICU). So while chasing this down, and considering this is an important "breaking change" (in the sense that lots of platforms may have their important alerts now completely silenced by the inexistence of these labels), I really think all references to kube_*_labels should redirect the user as fast as possible to the paragraph being added in this PR, no matter how.

In fact, looking at node-metrics.md (for instance), it seems like kube_node_labels is available by default.... and it's not, anymore. So that should be emphasized as well, this time not so much as "a change" but that, in order to get that metric, you need to enable it via the command line (again, what this paragraph is about).

lilic:
Putting on hold on this, until the bug fix and rework PR is merged, after that we can continue with improving the docs around the new flag, thanks, see details here https://github.com/kubernetes/kube-state-metrics/issues/1270#issuecomment-724046106

/hold

lilic:
@dsonck92 hey, PR was merged, do you mind adjusting this PR to the new flag, thanks! And if its not too much, we wanted to include an example where its used in a Deployment manifest, see the number 5 case in this issue https://github.com/kubernetes/kube-state-metrics/issues/1313.

Let me know if you need any other info! thanks!!

dsonck92:
@lilic Great to hear, sure I will adjust the changes to reflect the new flag.

So considering the other issue, basically, arguments are handled in verbatim. There is no quote handling (I also cannot really see what quotes inside options would solve, label names cannot contain spaces or commas so it's never ambiguous AFAICS). I'll do this somewhere in the remainder of this week.

lilic:
Thanks! The issue is more about adding docs on how this flag can be used in a deployment, as the help text states "".

dsonck92:
/remove-lifecycle stale

I will pick up this again with the newly added info

PR #1280, opened 2020-10-27
Fix pod-metric missing reasons

jangdm:
Some metrics reason was missing. 
Adding missing reasons.


lilic:
Thanks for the PR, do you mind just squashing your commits into one? Otherwise lgtm, thanks so much for adjusting it!

jangdm:
@lilic Sorry. During squashing commits, there was a little problem with my forked repo..
I will close this PR and make again with no problem commits.

PR #1282, opened 2020-10-27
.github/: Update issue templates

lilic:
cc @tariq1890 @brancz I noticed we have on template only, with new feature requests we might want to get more info. WDYT?


lilic:
/hold

I created this via the UI, very fancy :D, just want to clean up existing templates.

brancz:
Very nice! One of those things that you don't notice when you primarily contribute :D

/lgtm
/approve

PR #1287, opened 2020-10-28
Fix pod-metric missing reasons

jangdm:
Some metrics reason was missing.
Adding missing reasons.


jangdm:
/assign @lilic 
reopened PR and squashed into 1 commit

lilic (commented):
Thanks again!!

/lgtm

PR #1301, opened 2020-11-12
Adjust and refactor allowing labels to work for kubernetes labels metrics

lilic:
During performance testing of the v2.0 release, I noticed the new filtering mechanism for labels always goes through all resources even in cases when flag is disabled, this introduces a performance regression. The use cases we have currently to filter labels are for Kubernetes resource labels metrics e.g. kube_pod_labels metrics, and in the future for annotation metrics. We heavily control other labels in metrics and make sure it does always contain a set list of labels and that set list is bound cardinality and does not contain any sensitive information. There are other ways of not exposing metrics, for example with the metric allow and deny list flags in kube-state-metrics as well as with resource allow list flag, as well as with Prometheus own mechanisms of metric relabeling.

In short, this PR introduces a flag called `--labels-metric-allow-list` which allows Kubernetes labels to be added to the specific resource labels metrics. By default all labels metrics will only have name of resource and namespace in them and all other labels are denied. This flag does not affect any other metrics and thus does not degrade performance. This same pattern and functions can be used for annotation metrics in the future.
Example:
`--labels-metric-allow-list=pods=["app","app.kubernetes.io/component"] `
will produce a metric the following:
```
kube_pod_labels{pod="my-pod-name", label_app="myfancy-label", label_app_kubernetes_io_component="myotherfancylabel"namespace="pod-namespace"}
```

Note PR is WIP, I run this idea already with @brancz and also proposed this in the various issues and PRs. I am looking for actual implementation feedback now.


docs/cli-arguments.md
```diff
@@ -25,34 +25,34 @@ spec:
...
-      --version                          kube-state-metrics build version information
-      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
+      --add_dir_header                    If true, adds the file directory to the header of the log messages
+      --alsologtostderr                   log to standard error as well as files
+      --apiserver string                  The URL of the apiserver to use as a master
+      --enable-gzip-encoding              Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+  -h, --help                              Print Help text
+      --host string                       Host to expose metrics on. (default "0.0.0.0")
+      --kubeconfig string                 Absolute path to the kubeconfig file
+      --labels-metric-allow-list string   By default all labels metric will only have resource name and namespace labels, to allow more labels pass in a list of resource names in their plural form and Kubernetes labels you would like to allow, =namespaces=[k8s-label1,k8s-labeln...],pods[app]...
```
lilic, line 35: Its hard to explain the `labels-metric-allow-list` flag due to the nature of the metric labels and k8s labels, if anyone has a more clear naming, open to suggestions! I have a similar idea for annotations where it would be --annotations-metric-allow-list and you pass the resource and annotations to allow.

internal/store/endpoint_test.go
```diff
@@ -93,8 +168,11 @@ func TestEndpointStore(t *testing.T) {
 		},
 	}
 	for i, c := range cases {
-		c.Func = generator.ComposeMetricGenFuncs(endpointMetricFamilies)
-		c.Headers = generator.ExtractMetricFamilyHeaders(endpointMetricFamilies)
+		allowLabels := []string{
```
lilic, line 171: Diff is strange here, so best expand, essentially this adds a Labels test, with labels we allow for the kube_endpoint_labels metric.

internal/store/mutatingwebhookconfiguration.go
```diff
@@ -31,7 +31,6 @@ import (
 )
 
 var (
-	descMutatingWebhookConfigurationHelp          = "Kubernetes labels converted to Prometheus labels."
```
lilic, line 34: We did not have this and a few other labels, but since we now filter out labels by default if users want we can add them for this and other resources like ConfigMaps for example.

lilic:
Note that I will adjust the unit tests and some other nits I noticed myself after initial thoughts for or against the approach. :)

docs/cli-arguments.md
dgrisonnet, line 35:
```suggestion
      --labels-metric-allow-list string   By default all labels metric will only have resource name and namespace labels, to allow more labels pass in a list of resource names in their plural form and Kubernetes labels you would like to allow, =namespaces=[k8s-label1,k8s-labeln...],pods=[app]...
```

internal/store/mutatingwebhookconfiguration.go
dgrisonnet: Agreed :+1:, it would prevent this feature from being half-baked.

internal/store/persistentvolumeclaim.go
```diff
@@ -34,15 +34,19 @@ var (
...
 		*generator.NewFamilyGenerator(
 			descPersistentVolumeClaimLabelsName,
 			descPersistentVolumeClaimLabelsHelp,
 			metric.Gauge,
 			"",
 			wrapPersistentVolumeClaimFunc(func(p *v1.PersistentVolumeClaim) *metric.Family {
-				labelKeys, labelValues := kubeLabelsToPrometheusLabels(p.Labels)
+				var labelKeys, labelValues []string
+				allowedKubeLabels := filterKubeLabes(p.Labels, allowLabelsList)
+				labelKeys, labelValues = kubeLabelsToPrometheusLabels(allowedKubeLabels)
```
dgrisonnet, line 49:
```suggestion
				labelKeys, labelValues = createLabelKeysValues(p.Labels, allowLabelsList)
```

internal/store/secret.go
```diff
@@ -74,7 +76,9 @@ var (
 			metric.Gauge,
 			"",
 			wrapSecretFunc(func(s *v1.Secret) *metric.Family {
-				labelKeys, labelValues := kubeLabelsToPrometheusLabels(s.Labels)
+				var labelKeys, labelValues []string
+				allowedKubeLabels := filterKubeLabes(s.Labels, allowLabelsList)
+				labelKeys, labelValues = kubeLabelsToPrometheusLabels(allowedKubeLabels)
```
dgrisonnet, line 81:
```suggestion
				labelKeys, labelValues = createLabelKeysValues(s.Labels, allowLabelsList)
```

internal/store/service.go
```diff
@@ -88,9 +90,10 @@ var (
 			metric.Gauge,
 			"",
 			wrapSvcFunc(func(s *v1.Service) *metric.Family {
-				labelKeys, labelValues := kubeLabelsToPrometheusLabels(s.Labels)
+				var labelKeys, labelValues []string
+				allowedKubeLabels := filterKubeLabes(s.Labels, allowLabelsList)
+				labelKeys, labelValues = kubeLabelsToPrometheusLabels(allowedKubeLabels)
```
dgrisonnet, line 95:
```suggestion
				labelKeys, labelValues = createLabelKeysValues(s.Labels, allowLabelsList)
```

internal/store/statefulset.go
```diff
@@ -170,7 +172,9 @@ var (
 			metric.Gauge,
 			"",
 			wrapStatefulSetFunc(func(s *v1.StatefulSet) *metric.Family {
-				labelKeys, labelValues := kubeLabelsToPrometheusLabels(s.Labels)
+				var labelKeys, labelValues []string
+				allowedKubeLabels := filterKubeLabes(s.Labels, allowLabelsList)
+				labelKeys, labelValues = kubeLabelsToPrometheusLabels(allowedKubeLabels)
```
dgrisonnet, line 177:
```suggestion
				labelKeys, labelValues = createLabelKeysValues(s.Labels, allowLabelsList)
```

pkg/options/options.go
```diff
@@ -97,7 +97,7 @@ func (o *Options) AddFlags() {
 	o.flags.StringVar(&o.Namespace, "pod-namespace", "", "Name of the namespace of the pod specified by --pod. "+autoshardingNotice)
 	o.flags.BoolVarP(&o.Version, "version", "", false, "kube-state-metrics build version information")
 	o.flags.BoolVar(&o.EnableGZIPEncoding, "enable-gzip-encoding", false, "Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.")
-	o.flags.Var(&o.LabelsAllowList, "labels-allow-list", "list of metric names and labels you would like to allow, metric_name=[label1,labeln...],metric_name[]...")
+	o.flags.Var(&o.LabelsAllowList, "labels-metric-allow-list", "By default all labels metric will only have resource name and namespace labels, to allow more labels pass in a list of resource names in their plural form and Kubernetes labels you would like to allow, =namespaces=[k8s-label1,k8s-labeln...],pods[app]...")
```
dgrisonnet, line 100:
```suggestion
	o.flags.Var(&o.LabelsAllowList, "labels-metric-allow-list", "By default all labels metric will only have resource name and namespace labels, to allow more labels pass in a list of resource names in their plural form and Kubernetes labels you would like to allow, =namespaces=[k8s-label1,k8s-labeln...],pods=[app]...")
```

dgrisonnet (commented):
I might have missed it, but are you planning to set the sane defaults in another PR?

lilic:
> I might have missed it, but are you planning to set the sane defaults in another PR?


Thanks for the review, what do you mean by sane defaults, defaults for?

brancz:
Strategy wise this looks like the perfect fit for the problem to me. I think we should go with this.

docs/cli-arguments.md
mrueg: I would suggest to change the order and explain first what the flag does, then its default. Maybe something like: `Allows to pass a list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them`

lilic:
@brancz @dgrisonnet @jw-s please have a look, should be ready for review the details are in the description of the PR, thanks!

lilic:
Since this includes a bug fix for the labels metrics name and namespace, I changed base to release-2.0.

jw-s:
Great job! Looks good :)

internal/store/builder.go
```diff
@@ -210,115 +212,124 @@ func availableResources() []string {
 }
 
 func (b *Builder) buildConfigMapStore() cache.Store {
+	// no labels
```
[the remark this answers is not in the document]
lilic: The idea was to make reviews easier and future changes to this function so it easily shows why this call is  different to other ones that do have label metrics. I was planning on removing it but maybe rephrasing it to something like:
```
// There are no labels metrics for this resource.
```
Would that be helpful, or should I simply remove the comment?

I would not make them to TODOs as if there is no user need to add those metrics. Just because we can, doesn't mean we should if no one is interested in them. With that said if someone adds them we will accept it. Hope that makes sense?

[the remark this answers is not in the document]
dgrisonnet: I might have misunderstood, I thought you wanted to add all the remaining labels metrics for completeness purposes.

Considering your previous statement, I agree that it's not really meaningful to have them as TODOs and I am fine with removing the comments completely.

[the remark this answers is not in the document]
lilic: Sorry was not clear, it makes sense to have them, but there is nothing broken if they are not there, and as soon as we get an actual use case for it we can add each one. For completeness sake I would not be adding it if folks won't make use of them. Some things like ConfigMaps make more sense than other resources.

.golangci.yml
```diff
@@ -18,6 +18,9 @@ linters:
 linters-settings:
   goimports:
     local-prefixes: k8s.io/kube-state-metrics
+  gocyclo:
```
lilic, line 21: I will open an issue for this after the PR is merged, to refactor the podMetricFamilies function, note this function did not change, it was just a var before.

internal/store/builder.go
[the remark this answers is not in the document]
lilic: Removed, thanks :)

internal/store/builder.go
```diff
@@ -318,7 +320,8 @@ func (b *Builder) buildVolumeAttachmentStore() cache.Store {
 }
 
 func (b *Builder) buildVPAStore() cache.Store {
-	return b.buildStoreFunc(vpaMetricFamilies, &vpaautoscaling.VerticalPodAutoscaler{}, createVPAListWatchFunc(b.vpaClient))
+	// no labels
```
brancz, line 323: no labels?

pkg/options/types.go
```diff
@@ -136,59 +134,68 @@ func (n *NamespaceList) Type() string {
...
 	)
-	for tok := s.Scan(); tok != scanner.EOF; tok = s.Scan() {
-		next = s.Peek()
-		switch tok {
+	firstWordPos = 0
+
+	for i, v := range value {
+		if i+1 == len(value) {
+			// We are at the last char so set next to EOF.
+			// TODO: create our own EOF const
```
brancz, line 154: seems like this is still todo

brancz (commented):
I think this is almost ready to merge. Just a few small things.

Side note: I love the parser for the label flag, and the ability to specify labels as they are in form of the Kubernetes label, I think this is much more ergonomic.

docs/cli-arguments.md
```diff
@@ -25,34 +25,34 @@ spec:
...
-      --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
+      --add_dir_header                    If true, adds the file directory to the header of the log messages
+      --alsologtostderr                   log to standard error as well as files
+      --apiserver string                  The URL of the apiserver to use as a master
+      --enable-gzip-encoding              Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+  -h, --help                              Print Help text
+      --host string                       Host to expose metrics on. (default "0.0.0.0")
+      --kubeconfig string                 Absolute path to the kubeconfig file
+      --labels-metric-allow-list string   Allows to pass a list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=["k8s-label-1","k8s-label-n",...],pods=["app"],...)'
```
lilic, line 35: Adjusted the help text, are we happy with flag name?
> --labels-metric-allow-list

My thinking for annotation it would be:
> --annotations-metric-label-allow-list

So might need a change here as well? 🤔  It gets confusing as its label metrics for label allow list. 🤯

mrueg: Flag name(s) sounds good to me.

I wonder if at some point, we want to consider reading arguments from a config file for all settings. I don't think we'll hit max cmdline input length, but let's assume someone is very verbose ;)
 It might also be easier to define it there in a structured way?
```
labels-metric-allow-list:
  - namespaces:
    - k8s-label1
    - k8s-label2
  - pods
    - app   
```

lilic:
Addressed all comments, let me know if I missed some, ready for another review! Thanks all!

docs/cli-arguments.md
lilic: Good point, I think if that happens we can add a flag called --labels-metric-allow-list-configfile that inputs a file, or one for both labels and annotations. Right now the parser makes certain assumptions (as it did before), I would say as we did not have a reading from a file until now I would not add it as part of this PR, but if there is something users face in the future, definitely!

brancz:
Amazing work! :100: 

/lgtm
/approve

docs/cli-arguments.md
mrueg: I agree, I was thinking  about a generic `--config` flag to do configuration for everything kube-state-metrics related.

lilic: I see, okay if that is what you need, and you have use case, please feel free to open a feature request w and we can discuss there :)

PR #1304, opened 2020-11-16
Add "uid" label to every pod metric

### LINKED ISSUE #1259: Add Pod UID to all "kube_pod_*" metrics

kind feature

**What happened**:

Currently, only `kube_pod_info` metric exposes pod `UID` value. 
It is assumed that `namespace/pod` combination should be always unique at any given point in time.
However, it does not accommodate for cases when pods can be created, deleted, and recreated with the same name over some period of time.
This is typically the case with `StatefulSet` pods and CRD's auxiliary pods (the pods created by custom resource controllers).
In those cases, it is not uncommon to recreate pods with the previously used names. 

This results in interesting and at times incorrect metric values, especially when used in "join" multiple metrics: `ON (namespace, pod)` prometheus queries.

I propose to add `uid` label to all `kube_pod_.*` metrics as it is done in `kube_pod_info`

The following graph demonstrates the incorrect durations (seconds) value between `kube_pod_created` and `container_start_time_seconds`

image

The green line represents the duration (seconds) and the lines below represent distinct pods according to `kube_pod_info` (which includes `uid` value).

We can observe that sometime after `12:28` the old pod was deleted and a new pod (w/ new `uid` but with the same name) was created shortly after `12:30`. 
However, the query join did not include `uid` value (since it does not exist on `kube_pod_created`) and resulted in incorrect duration computation where 
the newer pod's `container_start_time_seconds` was computed against the old pod's `kube_pod_created` value.

This issue would have been avoided if we could use join `ON (namespace,pod,uid)` instead of `ON(namespace,pod`).

**Environment**:

ichekrygin:
Adding UID label to all pod metrics allow us to differentiate between the pod metrics using unique identifier UID rather than not always unique "namespace/name".

Currently, only the kube_pod_info metric exposes the pod UID value.
It is assumed that the namespace/pod combination should be always unique at any given point in time.
However, it does not accommodate for cases when pods can be created, deleted, and recreated with the same name over some period of time.
This is typically the case with StatefulSet pods and CRD's auxiliary pods (the pods created by custom resource controllers).
In those cases, it is not uncommon to recreate pods with the previously used names.

This results in interesting and at times incorrect metric values, especially when used in "join" multiple metrics: ON (namespace, pod) prometheus queries.

Fixes #1259


lilic:
Nice thanks so much for the PR! 🎉  The PR looks good to me, we have a feature freeze right now for the release-2.0 but we can merge this in master and include in release after that. https://github.com/kubernetes/kube-state-metrics#overview

cc @brancz any objections to the above or this PR?

brancz:
This doesn't increase cardinality, so I'm ok with it. I would prefer to drop a note in the release notes that we may remove this label, just as a safety measure in case this makes the sheer bytes on the wire explode (in which case people should probably use gzip anyways).

lilic (commented):
/lgtm

Thanks for the contribution! Note this will be release in the next release, not the 2.0.0 as that was closed for features. 

@brancz sgtm, about note to the CHANGELOG

PR #1312, opened 2020-11-23
Change default image registry

lilic:
Due to only me having permissions to quay.io/coreos/kube-state-metrics, we want to slowly deprecate pushing to quay registry. Instead, now we have gcr images which all maintainers have access to promoting. This is the first step towards the deprecation, by changing the default in the example manifests.

cc @tariq1890 @brancz PTAL, thanks!


tariq1890:
Thanks for doing this @lilic !

/lgtm

PR #1315, opened 2020-11-27
internal/store: fix high cyclo complexity

### LINKED ISSUE #1306: Fix the high gocyclo  complexity setting
With the change to the podMetricFamilies and nodeMetricFamilies into a function, now our CI detected that cyclo complexity is too high, in case of the pod function it's 127. The temporary fix was that we just raised the allowed to 128 complexity.

See context in this PR https://github.com/kubernetes/kube-state-metrics/pull/1301/files#r526061712

We have a couple of options that I see:

1. Refactor the pod and node MetricFamilies functions to remove the complexity.
2. Disable testing the complexity as it might not apply to us.
3. Disable testing the complexity just for internal packages or specific files.

Other options or opinions more than welcome!

dgrisonnet:
To reduce the high cyclo complexity of the nodeMetricFamilies and podMetricFamilies, I moved the generators that don't need the labelsAllowList back into a global variable. This approach is the most straightforward that I could find and the one that requires fewer changes to the codebase.

Another option would be to define actual functions to create each metric family generator, but in my opinion, this would make the codebase less readable.

We could also subdivide the metric families based on the subsystem of the metric. For example, with Pods we could separate the families between: `status`, `container`, `init_container`, `spec`, ... However, I also think that this approach would needlessly complexity the codebase compare to the solution I went with here.

Once we come to an agreement, I'll apply this refactor to the other metric families.

Fixes #1306

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -40,205 +40,223 @@ var (
 
 func nodeMetricFamilies(allowLabelsList []string) []generator.FamilyGenerator {
 	return []generator.FamilyGenerator{
-		*generator.NewFamilyGenerator(
-			"kube_node_info",
-			"Information about a cluster node.",
-			metric.Gauge,
-			"",
-			wrapNodeFunc(func(n *v1.Node) *metric.Family {
-				labelKeys := []string{
-					"kernel_version",
-					"os_image",
-					"container_runtime_version",
-					"kubelet_version",
-					"kubeproxy_version",
-					"provider_id",
-					"pod_cidr",
-				}
-				labelValues := []string{
-					n.Status.NodeInfo.KernelVersion,
-					n.Status.NodeInfo.OSImage,
-					n.Status.NodeInfo.ContainerRuntimeVersion,
-					n.Status.NodeInfo.KubeletVersion,
-					n.Status.NodeInfo.KubeProxyVersion,
-					n.Spec.ProviderID,
-					n.Spec.PodCIDR,
-				}
+		createNodeCreatedFamilyGenerator(),
+		createNodeInfoFamilyGenerator(),
+		createNodeLabelsGenerator(allowLabelsList),
+		createNodeRoleFamilyGenerator(),
+		createNodeSpecTaintFamilyGenerator(),
+		createNodeSpecUnschedulableFamilyGenerator(),
+		createNodeStatusAllocatableFamilyGenerator(),
+		createNodeStatusCapacityFamilyGenerator(),
+		createNodeStatusConditionFamilyGenerator(),
+	}
+}
 
-				internalIP := ""
-				for _, address := range n.Status.Addresses {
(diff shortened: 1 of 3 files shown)


lilic (commented):
Thanks for the PR I would prefer if the approach was not, putting things back to a variable. As those are not tested by our golang CI linter. Last time when I did the move it caught some nice linter errors. This just hides the problems again.

Instead, ideally, we would want each of the vars to be generated by the function. So e.g. `kube_node_role` would be generated by a function generateNodeRole() for example and this would return a NewFamilyGenerator. Let me know if something is not clear, thanks again for working on this!

lilic:
> Another option would be to define actual functions to create each metric family generator, but in my opinion, this would make the codebase less readable.

Sorry only seen this now, why would you think that? A global function always makes things less readable vs multiple functions.

dgrisonnet:
@lilic I thought that splitting the code might make it harder to extend for new contributors. However, after thinking about it again, I agree with you that creating functions for each generator is a cleaner approach.
I updated this PR according to that and while I was at it, I also sorted the new functions to make it tidier hence the huge amount of changes introduced by this PR.

internal/store/node.go
```diff
@@ -40,341 +40,377 @@ var (
...
-				labelValues := []string{
-					n.Status.NodeInfo.KernelVersion,
-					n.Status.NodeInfo.OSImage,
-					n.Status.NodeInfo.ContainerRuntimeVersion,
-					n.Status.NodeInfo.KubeletVersion,
-					n.Status.NodeInfo.KubeProxyVersion,
-					n.Spec.ProviderID,
-					n.Spec.PodCIDR,
-				}
+		nodeCreatedGenerator(),
```
lilic, line 43: I assume you called the function Generator suffix due to it returning a FamilyGenerator type?

Still think might make sense to call them `createNodeCreateFamilyGenerator`, but not a blocker for me, wdyt?

lilic (commented):
Nice work, thanks so much!

I have two nits, one which is optional, let me know what you think about the naming.

We can merge this into master right after beta is cut, not sure its needed a backport to 2.0 release branch as it doesn't fix a bug.

internal/store/node.go
dgrisonnet: Agreed, I thought a getter like naming approach would be better here, but after giving it another thought, these functions are actual constructors so I prefer your suggestion.

lilic: Nice, thanks!

lilic:
@dgrisonnet sorry you have to rebase as we merged a PR that was opened before yours. After that its ready to be merged 🎉 :) Thanks again!

dgrisonnet:
@lilic I rebased :)

lilic (commented):
/lgtm

Thanks!! 🎉

PR #1322, opened 2020-12-04
Promote v2.0.0 to beta pre-release

lilic:
🎉  There were no bug reports around the functionality of the latest alpha releases, so we can now safely promote v2.0.0 to beta.

Next are scale and performance tests. If no major regressions are found, and no new bugs I would say we can cut a rc of v2.0.0. Unless any objections?

cc @brancz @tariq1890


brancz (approved):
Let's do this! :tada:

lilic:
Note we had one change since the last release but was not affecting users, not sure changing default image in examples makes sense to point out?

lilic:
@brancz I need an lgtm label :D

lilic:
@brancz I had to make some adjustments to our e2e tests due to recent changes in default image in examples, see last commit. PTAL, thanks!

PR #1325, opened 2020-12-16
add useExistingRole option - to support running in specific namespace…

KlavsKlavsen:
…s without needing clusteradmin privs

This is a repost of https://github.com/helm/charts/pull/23953

This adds support for running on kubernetes setups (such as is typical in openshift) where you only have access to your own namespaces.

I've gotten this support merged in both Prometheus and Grafana charts and
I'd really like to get it in this last chart I use for our Prometheus setup ;)

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - monitoring
 - prometheus
 - kubernetes
-version: 2.9.8
+version: 2.10.0
 appVersion: 1.9.7
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:
--- a/charts/kube-state-metrics/templates/clusterrolebinding.yaml
+++ b/charts/kube-state-metrics/templates/clusterrolebinding.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create -}}
+{{- if and .Values.rbac.create .Values.rbac.useClusterRole -}}
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
@@ -11,7 +11,11 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
+{{- if .Values.rbac.useExistingRole }}
+  name: {{ .Values.rbac.useExistingRole }}
+{{- else }}
   name: {{ template "kube-state-metrics.fullname" . }}
+{{- end }}
 subjects:
 - kind: ServiceAccount
   name: {{ template "kube-state-metrics.fullname" . }}
--- a/charts/kube-state-metrics/templates/role.yaml
+++ b/charts/kube-state-metrics/templates/role.yaml
@@ -1,180 +1,192 @@
-{{- if .Values.rbac.create -}}
+{{- if and (eq $.Values.rbac.create true) (not .Values.rbac.useExistingRole)  -}}
+{{- if eq .Values.rbac.useClusterRole false }}
+{{-   range (split "," $.Values.namespace) }}
+{{- end }}
+---
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if eq .Values.rbac.useClusterRole false }}
+kind: Role
+{{- else }}
 kind: ClusterRole
+{{- end }}
 metadata:
   labels:
-    app.kubernetes.io/name: {{ template "kube-state-metrics.name" . }}
-    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
-  name: {{ template "kube-state-metrics.fullname" . }}
+    app.kubernetes.io/name: {{ template "kube-state-metrics.name" $ }}
+    helm.sh/chart: {{ $.Chart.Name }}-{{ $.Chart.Version }}
+    app.kubernetes.io/managed-by: {{ $.Release.Service }}
+    app.kubernetes.io/instance: {{ $.Release.Name }}
+  name: {{ template "kube-state-metrics.fullname" $ }}
+{{- if eq .Values.rbac.useClusterRole false }}
+  namespace: {{ . }}
+{{- end }}
 rules:
-{{ if .Values.collectors.certificatesigningrequests }}
+{{ if $.Values.collectors.certificatesigningrequests }}
 - apiGroups: ["certificates.k8s.io"]
   resources:
   - certificatesigningrequests
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.configmaps }}
+{{ if $.Values.collectors.configmaps }}
 - apiGroups: [""]
   resources:
   - configmaps
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.cronjobs }}
+{{ if $.Values.collectors.cronjobs }}
 - apiGroups: ["batch"]
   resources:
   - cronjobs
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.daemonsets }}
+{{ if $.Values.collectors.daemonsets }}
 - apiGroups: ["extensions", "apps"]
   resources:
   - daemonsets
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.deployments }}
+{{ if $.Values.collectors.deployments }}
 - apiGroups: ["extensions", "apps"]
   resources:
   - deployments
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.endpoints }}
+{{ if $.Values.collectors.endpoints }}
 - apiGroups: [""]
   resources:
   - endpoints
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.horizontalpodautoscalers }}
+{{ if $.Values.collectors.horizontalpodautoscalers }}
 - apiGroups: ["autoscaling"]
   resources:
   - horizontalpodautoscalers
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.ingresses }}
+{{ if $.Values.collectors.ingresses }}
 - apiGroups: ["extensions", "networking.k8s.io"]
   resources:
   - ingresses
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.jobs }}
+{{ if $.Values.collectors.jobs }}
 - apiGroups: ["batch"]
   resources:
   - jobs
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.limitranges }}
+{{ if $.Values.collectors.limitranges }}
 - apiGroups: [""]
   resources:
   - limitranges
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.mutatingwebhookconfigurations }}
+{{ if $.Values.collectors.mutatingwebhookconfigurations }}
 - apiGroups: ["admissionregistration.k8s.io"]
   resources:
     - mutatingwebhookconfigurations
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.namespaces }}
+{{ if $.Values.collectors.namespaces }}
 - apiGroups: [""]
   resources:
   - namespaces
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.networkpolicies }}
+{{ if $.Values.collectors.networkpolicies }}
 - apiGroups: ["networking.k8s.io"]
   resources:
   - networkpolicies
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.nodes }}
+{{ if $.Values.collectors.nodes }}
 - apiGroups: [""]
   resources:
   - nodes
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.persistentvolumeclaims }}
+{{ if $.Values.collectors.persistentvolumeclaims }}
 - apiGroups: [""]
   resources:
   - persistentvolumeclaims
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.persistentvolumes }}
+{{ if $.Values.collectors.persistentvolumes }}
 - apiGroups: [""]
   resources:
   - persistentvolumes
   verbs: ["list", "watch"]
 {{ end -}}
-{{ if .Values.collectors.poddisruptionbudgets }}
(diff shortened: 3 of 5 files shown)


KlavsKlavsen:
CLA signed

lilic:
/assign @mrueg

@KlavsKlavsen you need to push an empty commit to validate the CLA again, I believe the command is gone now to revalidate. Thanks for the PR!

charts/kube-state-metrics/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
 - monitoring
 - prometheus
 - kubernetes
-version: 2.9.3
+version: 2.10.1
```
mrueg, line 9: can we go for 2.10.0 (since it adds a new feature) instead?

charts/kube-state-metrics/templates/clusterrolebinding.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.rbac.create -}}
+{{- if and .Values.rbac.create .Values.useClusterRole -}}
```
mrueg, line 1:
```suggestion
{{- if and .Values.rbac.create .Values.rbac.useClusterRole -}}
```

charts/kube-state-metrics/templates/role.yaml
```diff
@@ -1,180 +1,192 @@
-{{- if .Values.rbac.create -}}
+{{- if and (eq $.Values.rbac.create true) (not .Values.rbac.useExistingRole)  -}}
+{{- if eq .Values.rbac.useClusterRole false }}
+{{-   range (split "," $.Values.namespace) }}
+{{- end }}
+---
 apiVersion: rbac.authorization.k8s.io/v1
+{{- if eq .Values.rbac.useClusterRole false }}
```
mrueg, line 7: Can we switch branches here to avoid checking for false?

mrueg (commented):
Thanks for your contribution and apologies it takes so long to get this change in.

charts/kube-state-metrics/templates/role.yaml
KlavsKlavsen: How do you mean "switch branches" ?
I check for the same "exact thing" ( if eq .Values.rbac.useClusterRole false ) - in the entire file - to be certain it either "loops over namespaces" or goes "cluster-wide" way.
Do you mean I test for positive instead..  ie.( if .Values.rbac.useClusterrole )  .
Why do you dislike testing specificly for false? as I understand it, it follows these rules: https://godoc.org/text/template - ie. "false, 0, any nil pointer or interface value, and any array, slice, map, or string of length zero." is considered to match false.

IMHO doing it this way, removes ambiguity.

charts/kube-state-metrics/templates/clusterrolebinding.yaml
KlavsKlavsen: dooh. Good catch :)

charts/kube-state-metrics/Chart.yaml
KlavsKlavsen: agreed :)

charts/kube-state-metrics/templates/role.yaml
mrueg: Yes my idea was to check for true instead of false but you convinced me. Thanks for your clarification.

KlavsKlavsen:
Rebased - ready for merge.. ?

KlavsKlavsen:
@mrueg only missing an /lgtm label CI says ?

scottrigby:
/hold

Lets please wait for CI to be fully resolved before merging this.

1. Current PR we hope will fix it: https://github.com/kubernetes/test-infra/pull/20322
1. We will then be testing required checks on a low stakes PR (https://github.com/kubernetes/kube-state-metrics/pull/1335, which just adds a README)
v Once that succeeds, we will want to release the version of the chart that's now in `master` (before this PR,  otherwise end users won't have access to it)
1. From that point on we can merge PRs like this which make changes to the initial chart and create new releases

KlavsKlavsen:
The fix has been merged 6 days ago.. Hows the "simple PR" going? it seems to be stalled.

mrueg:
@KlavsKlavsen during this time of the year, please be aware that responses might take longer. 
I'm waiting for @scottrigby to remove the hold.

scottrigby:
@KlavsKlavsen the PR isn't the problem, it's test infra that needs to remove required checks from the gh-pages branch. I'm not a repo admin so the feedback loop is slower - and much slower right now because repo maintainers are away over the holidays (as they should be!) 😃 

If this PR were merged before that is resolved, it would not create a chart release. Backfilling that release would then have to be a manual process - more annoying and confusing to end users than just waiting to merge the PR.

Hope that helps explain it. Happy new year! Looking forward to resolving this soon.

KlavsKlavsen:
@scottrigby Hi Scott and a good new year to you :) - I see version of chart has been updated.. Is everything in place for merging this, or ?

scottrigby:
@KlavsKlavsen Not yet. @tariq1890 and I are trying to schedule a collab session. I would think we will find a time for this very soon. Feel free to also follow along or join in Slack chat: https://kubernetes.slack.com/archives/CJJ529RUY/p1609938365003800

scottrigby:
@KlavsKlavsen update: https://github.com/kubernetes/kube-state-metrics/issues/1342#issuecomment-757544796

charts/kube-state-metrics/Chart.yaml
```diff
@@ -6,7 +6,7 @@ keywords:
 - monitoring
 - prometheus
 - kubernetes
-version: 2.9.7
+version: 2.11.1
```
scottrigby, line 9: Assuming this should be `2.10.0` if we're bumping MINOR?

scottrigby (commented):
Chart CI is fixed, so chart PRs should be unblocked. I haven't reviewed this but only noticed one server question.

charts/kube-state-metrics/templates/role.yaml
```diff
@@ -1,180 +1,192 @@
...
+{{- end }}
 metadata:
   labels:
-    app.kubernetes.io/name: {{ template "kube-state-metrics.name" . }}
-    helm.sh/chart: {{ .Chart.Name }}-{{ .Chart.Version }}
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
-    app.kubernetes.io/instance: {{ .Release.Name }}
-  name: {{ template "kube-state-metrics.fullname" . }}
+    app.kubernetes.io/name: {{ template "kube-state-metrics.name" $ }}
+    helm.sh/chart: {{ $.Chart.Name }}-{{ $.Chart.Version }}
```
tariq1890, line 15: Hi @KlavsKlavsen, Can you elaborate on why we need to use the global scope explicitly here?

KlavsKlavsen: . does't work when inside a range loop. $ achieves the same thing and works inside a loop.

charts/kube-state-metrics/Chart.yaml
KlavsKlavsen: My bad - fixed :)

mrueg:
@scottrigby are we good to remove the hold on this one?

@tariq1890 are you okay with the response here https://github.com/kubernetes/kube-state-metrics/pull/1325#discussion_r559437483 ?

charts/kube-state-metrics/templates/role.yaml
tariq1890: Oh yes, I am aware of using the $ scope within range. Is it the range from Line 3? I didn't quite understand that. I see an end in Line 4?

charts/kube-state-metrics/templates/rolebinding.yaml
```diff
@@ -0,0 +1,27 @@
+{{- if and (eq  .Values.rbac.create true) (eq .Values.rbac.useClusterRole false) -}}
+{{- range (split "," $.Values.namespace) }}
```
tariq1890, line 2: the range makes sense over here. I see that the `{{- end -}}` is placed correctly. I am just confused by the range block used in role.yaml

tariq1890:
@KlavsKlavsen You have my sincere apologies for making you wait for this PR for so long. But I am afraid that we can only merge this PR once we go live with `kube-state-metrics v2.0` (correct me if I am mistaken @mrueg ). 

The reason is that `.Values.namespace` is intended to be a single value field. We need to use the `--namespaces` field and that change has been made only in v2.0. 

As a frequent user of helm charts myself, I feel it's in the best interest of open source users and maintainers to make the helm charts have the same cadence as that of the binary. In light of this, I think we should merge and publish this change only when kube-state-metrics v2 is released.

charts/kube-state-metrics/templates/role.yaml
```diff
@@ -1,180 +1,192 @@
-{{- if .Values.rbac.create -}}
+{{- if and (eq $.Values.rbac.create true) (not .Values.rbac.useExistingRole)  -}}
+{{- if eq .Values.rbac.useClusterRole false }}
+{{-   range (split "," $.Values.namespace) }}
```
tariq1890, line 3:
```suggestion
{{-   range (split "," $.Values.namespaces) }}
```

charts/kube-state-metrics/templates/rolebinding.yaml
tariq1890, line 2:
```suggestion
{{- range (split "," $.Values.namespaces) }}
```

KlavsKlavsen: The range block in role - is because a Clusterrole must be ONLY once.. But when operating on only a few namespaces - each must have the relevant role.. so there must be a role per namespace we operate on (as it's NOT a ClusterRole - it only affects the individual namespace and lives within it).
I hope that makes it clearer :)

KlavsKlavsen: and there USED to be a clusterrole.yaml and a role.yaml- but I was asked to merge the two - as they were 99% identical.

KlavsKlavsen:
> The reason is that `.Values.namespace` is intended to be a single value field. We need to use the `--namespaces` field and that change has been made only in v2.0.

In upstream 1.9 branch - the docs says: ``` --namespace string                            Comma-separated list of namespaces to be enabled. Defaults to ""```

So it has the exact same use for kube-state-metrics in 1.9.x as it has in 2.x - its just gotten an extra s for clarification.

So I'm not quite sure what you're referring to here ?

charts/kube-state-metrics/templates/role.yaml
KlavsKlavsen: I'm totally not understanding your question here.. sorry for being daft appearently :)
The end on line 4 - ends the IF - just before the range.. Its a bit peculiar - but the range is ONLY activated IF that if-clause right before is true. Its the only way I could make an optional range - seeing as I need to do either ClusterRole (once and no namespace(s) variable set) or otherwise a Role for each entry in namespace(s).

charts/kube-state-metrics/templates/rolebinding.yaml
mrueg: > @KlavsKlavsen You have my sincere apologies for making you wait for this PR for so long. But I am afraid that we can only merge this PR once we go live with `kube-state-metrics v2.0` (correct me if I am mistaken @mrueg ).
> 
> The reason is that `.Values.namespace` is intended to be a single value field. We need to use the `--namespaces` field and that change has been made only in v2.0.
> 
> As a frequent user of helm charts myself, I feel it's in the best interest of open source users and maintainers to make the helm charts have the same cadence as that of the binary. In light of this, I think we should merge and publish this change only when kube-state-metrics v2 is released.

Let's keep this change on ksm-v1.9.x to be able to move forward and integrate it. I'm tracking v2 changes in 
https://github.com/kubernetes/kube-state-metrics/pull/1358

mrueg:
@KlavsKlavsen Thanks for the quick turnaround, can you remove https://github.com/kubernetes/kube-state-metrics/pull/1325/commits/22057b4aaf6eb26b26368474e0fa64f6a221a979 and I'll make sure to get it merged immediately.

charts/kube-state-metrics/templates/rolebinding.yaml
KlavsKlavsen: In upstream 1.9 branch - the docs says: --namespace string Comma-separated list of namespaces to be enabled. Defaults to ""

So it has the exact same use for kube-state-metrics in 1.9.x as it has in 2.x - its just gotten an extra s for clarification.

So I'm not quite sure what you're referring to here, namespace also in 1.9.x supported multiple namespaces. ?

KlavsKlavsen:
reverted the change to support 2.x. I'll gladly submit a PR for 2.x with that small commit if you like.. I don't know where you keep the 2.x kube-state-metrics version of the chart though and how you want it.. :)

mrueg:
@KlavsKlavsen  I meant removing the commit, not reverting. While at it, can you stash the commits into a single one? Thanks!

KlavsKlavsen:
> @KlavsKlavsen I meant removing the commit, not reverting. While at it, can you stash the commits into a single one? Thanks!

Squashed all commits into one :)

go.mod
```diff
@@ -15,7 +15,10 @@ require (
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
+<<<<<<< HEAD
```
mrueg, line 18: @KlavsKlavsen can you take a look at this? It's causing CI to fail. Your changes should only touch files in the charts/ folder, right?

KlavsKlavsen: My bad - merge of upstream master.. rebased now.

examples/prometheus-alerting-rules/alerts.yaml
```diff
@@ -3,9 +3,7 @@ groups:
   rules:
   - alert: KubeStateMetricsListErrors
     annotations:
-      description: kube-state-metrics is experiencing errors at an elevated rate in
-        list operations. This is likely causing it to not be able to expose metrics
-        about Kubernetes objects correctly or at all.
+      description: kube-state-metrics is experiencing errors at an elevated rate in list operations. This is likely causing it to not be able to expose metrics about Kubernetes objects correctly or at all.
```
mrueg, line 6: @KlavsKlavsen  can you remove the changes to examples/ as well? CI is failing due to this I think.

KlavsKlavsen: My bad - sorry. Fixed :)

charts/kube-state-metrics/templates/role.yaml
KlavsKlavsen: You are absolutely correct Jianhao. (namespace was before 2.0 - its called namespaces in 2.0+). This chart has since been moved to https://github.com/prometheus-community/helm-charts/blob/d6ae19c6bb7089bd0080952ca6f98b68177e3dcb/charts/kube-state-metrics/templates/role.yaml#L3 - where it still lists namespace instead of namespaces - even though it has moved to 2.0.

PR #1327, opened 2020-12-16
Makefile: remove test-unit dependencies

dgrisonnet:
The test-unit Makefile rule had a dependency on the clean and build
rules. However, these rules aren't needed to run kube-state-metrics unit
tests.
The original setup can be problematic when running the unit tests in an
environment without docker installed.


lilic (commented):
Thanks!

/lgtm

PR #1237, opened 2020-09-18
Host stable/kube-state-metrics helm chart

scottrigby:
**What this PR does / why we need it**:

See #1153 

To-do:
- [x] import stable/kube-state-metrics helm chart history
- [x] clean up chart README
- [x] bump chart version (PATCH is fine, as this chart can upgrade in-place using the new repo info)
- [x] add chart-testing GitHub Action (for chart CI)
- [x] add chart-releaser GitHub Action (for helm repo CD)
    - [ ] configure a gh-pages branch for package releasing
- [ ] invite stable/kube-state-metrics chart maintainers to collaborate on this git repo
    - [ ] add an OWNERS file to match, if needed (I left the one in place from helm/charts. However please provide guidance on what you would like to do with this)

After PR merge:
- ~[ ] follow-up PR: import chart version package history, and update gh-pages index.yaml file (so that users can continue to access older chart package versions after the `stable` repo is deprecated and garbage collected)~ NO LONGER NEEDED now that GitHub has sponsored helm/charts package history. See https://helm.sh/blog/helm-turns-five/ 🎉 
- [ ] add repo to CNCF [Artifact Hub](https://artifacthub.io/)
    - [ ] add Artifact Hub repo metadata file to [verify repo](https://github.com/artifacthub/hub/issues/550) (gh-pages branch)
- ~[ ] [deprecate](https://github.com/helm/charts/blob/master/PROCESSES.md#deprecating-a-chart) stable/kube-state-metrics~ DONE as part of https://github.com/helm/charts/issues/24310
    - ~[ ] close any stable/kube-state-metrics issues, pointing users here ([there are currently none](https://github.com/helm/charts/issues?q=is%3Aissue+is%3Aopen+kube-state-metrics))~ This will be taken care of as part of the helm/charts repo deprecation follow-up.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1153

PR #1326, opened 2020-12-16
helm chart: Fix GitHub action

torstenwalter:
**What this PR does / why we need it**:

Chart releaser failed with the error message

`Unable to resolve action `helm/chart-releaser-action@v1.1.1`, unable to find version `v1.1.1``

https://github.com/kubernetes/kube-state-metrics/actions/runs/423148406

Latest version of chart-releaser action is [v1.1.0](https://github.com/helm/chart-releaser-action/releases/tag/v1.1.0)

@lilic / @brancz You also need to create an empty gh-pages branch for this to work. The command below should work:

```
git checkout --orphan gh-pages
git rm -rf .
git commit --allow-empty -m "root commit"
git push origin gh-pages
```

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1329, opened 2020-12-17
Chart releaser name template and release

scottrigby:

CD follow-up from #1237 

- Chart Releaser naming template variable added
- Chart PATCH version bumped to pass helm-ci workflow and trigger helm-cd workflow for a release

Requires:
- [x]  configure a gh-pages branch for package releasing (done with @lilic)

GPG signing for chart-releaser will be added in a separate follow-up PR.

### DIFF
--- a/.github/workflows/helm-cd.yaml
+++ b/.github/workflows/helm-cd.yaml
@@ -27,3 +27,4 @@ jobs:
           charts_dir: charts
         env:
           CR_TOKEN: "${{ secrets.GITHUB_TOKEN }}"
+          CR_RELEASE_NAME_TEMPLATE: "kube-state-metrics-helm-chart-{{ .Version }}"
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - monitoring
 - prometheus
 - kubernetes
-version: 2.9.3
+version: 2.9.4
 appVersion: 1.9.7
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:


lilic (commented):
Thanks!

/lgtm
/hold 

@tariq1890 @mrueg letting it up to you to remove the hold!

carlosescura:
Is there any ETA for merging this? This PR: https://github.com/kubernetes/kube-state-metrics/pull/1326 was merged but the release action was not triggered :( 

Kudos to all of you for maintaining this chart!! 🏅

mrueg:
Thanks for the work @lilic @scottrigby!
Changes look good to me
/hold cancel

scottrigby:
reindex

scottrigby:
@lilic @mrueg

The CD action failed because one of the checks was still running, and it appears gh-pages is a protected branch, requiring all checks to complete first: https://github.com/kubernetes/kube-state-metrics/runs/1574438846?check_suite_focus=true
```
Pushing to branch "gh-pages"
remote: error: GH006: Protected branch update failed for refs/heads/gh-pages.        
remote: error: 9 of 9 required status checks are expected.        
To https://github.com/kubernetes/kube-state-metrics
 ! [remote rejected]   HEAD -> gh-pages (protected branch hook declined)
error: failed to push some refs to 'https://github.com/kubernetes/kube-state-metrics'
Error: exit status 1
```

I do not have permission on this repo to re-run the job. But I suspect if someone is able to do that, it may succeed now that the other status checks have all passed.

Alternatively, you may want to change branch protection rules on `gh-pages`, since `ci-e2e-tests` appears to always take longer than this one, so this would be a recurring problem.

lilic:
I removed the branch protection on that branch. We should be good to go now 🎉

PR #1332, opened 2020-12-18
Add build info metric

### LINKED ISSUE #1132: Add Support for kube_state_metrics_build_info
This is a FEATURE REQUEST.

Please add support for kube_state_metrics_build_info as part of the metrics exposed on telemetry port 8081. 
kube_state_metrics_build_info shall include the version of the kube-state-metrics.
Most Prometheus exporters are currently supporting such metric.

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

This pr adds the kube_state_metrics_build_info metric as required by #1132

Fixes #1132


yeya24:
PTAL @lilic ^^

main.go
```diff
@@ -80,6 +81,7 @@ func main() {
 	storeBuilder := store.NewBuilder()
 
 	ksmMetricsRegistry := prometheus.NewRegistry()
+	registerBuildInfo(ksmMetricsRegistry)
```
kakkoyun, line 84: @yeya24 Any reason why you didn't consider the existing collector? https://github.com/prometheus/common/blob/master/version/info.go

lilic: Nice Kemal, agreed!

yeya24: @kakkoyun I was thinking about using that directly. But that works well with `promu` because `promu` injects all the info via build flags to that pkg. 

Right now the injected flags are added https://github.com/kubernetes/kube-state-metrics/blob/master/Makefile#L55 here, I still need to add more flags to make the `VersionCollector` usable. I feel it is not necessary.

kakkoyun: True. I have done exactly what you said in here https://github.com/kakkoyun/observable-remote-write/blob/master/Makefile 

Hard to judge just which way is simpler. I guess it's a matter of preference. My preference would be obviously reusing the package from the `prometheus/common`.

@lilic What do you think? Which code looks simpler to you?

lilic: Agreed, reusing Prometheus/common would be my preference here as well. 👍

yeya24: Got it! I will use that package. Thanks @kakkoyun for the example!

yeya24:
Updated. Please take a look.

Some screenshots:

Screenshot from 2021-01-05 01-29-05

Screenshot from 2021-01-05 01-30-01

yeya24:
Umm, not sure how to fix CI. I tried `make validate-manifests` locally and I didn't get any error output.

kakkoyun (approved):
LGTM. 

Concerning CI failure, I'd suggest making sure your tooling version is the same as the CI.

README.md
```diff
@@ -150,6 +150,12 @@ http_request_duration_seconds_sum{handler="metrics",method="get"} 0.021113919999
 http_request_duration_seconds_count{handler="metrics",method="get"} 30
 ```
 
+kube-state-metrics exposes the build info metric:
```
lilic, line 153: Nit: Would be good to add a few lines what those are useful for. Similiar to have we have them for the list and watch errors above.

lilic (commented):
Thanks for the changes! One suggestion only, but a nit.

The failure seems odd, but do you mind installing all the tools we use (some might need bumping so feel free to bump if you have the time) and try again.

I can also have a look tomorrow if you don't manage, thanks!!

README.md
yeya24: Fixed.

README.md
```diff
@@ -150,6 +150,15 @@ http_request_duration_seconds_sum{handler="metrics",method="get"} 0.021113919999
...
+```
+kube_state_metrics_build_info{branch="master",goversion="go1.15.3",revision="6c9d775d",version="v2.0.0-beta"} 1
+```
+
+`kube_state_metrics_build_info` is used to expose version and other build information. For more usage about the info pattern,
+please check the blog post [here](https://www.robustperception.io/exposing-the-software-version-to-prometheus).
```
lilic, line 159: I am not sure we are allowed to link to external companies blog posts? Otherwise, agreed this blog is very useful! @brancz wdyt?

lilic (commented):
Do you mind just squashing your commits as the bot will merge as is, thanks! Otherwise lgtm, thanks again!

README.md
brancz: I don't think there is a general rule here, if this was self-promotion I would be more critical, but here it's really just a helpful resource.

lilic: Yes sounds good, was on the edge. lgtm apart from the need to squash commits.

yeya24:
Updated. @lilic

lilic (commented):
/lgtm

Thanks for the contribution and addressing all the comments so quickly! :)

PR #1333, opened 2020-12-18
Remove vendor directory

### LINKED ISSUE #1111: Consider dropping vendor dir in favour of GOPROXY
In kube-state-metrics, we have managed to maintain a clean dependency tree. This puts us in a good position and we can consider using the `GOPROXY` feature available in go.13+.

Some advantages I can think of with `no vendor directory`:

- Decreased build times since the dependencies are cached in the `https://proxy.golang.org/` server. I have heard great things about the build times.
- Decrease the size of the repository

Disadvantage:
- This could cause issues in regions where the `GOPROXY` is accessible. One example would be China. But we can get around this by allowing for GOPROXY to be configurable.

@brancz @lilic 

/help

kakkoyun:
Signed-off-by: Kemal Akkoyun <kakkoyun@gmail.com>

This PR removes the `vendor` directory for the sake of new go module proxy usage.

Fixes #1111 

cc @lilic @brancz @tariq1890


lilic (approved):
/lgtm

Thanks so much!!

PR #1334, opened 2020-12-18
Retry releasing chart after gh-pages branch protection was removed

scottrigby:

Retrying #1329 

Fixes #

### DIFF
--- a/charts/kube-state-metrics/Chart.yaml
+++ b/charts/kube-state-metrics/Chart.yaml
@@ -6,7 +6,7 @@ keywords:
 - monitoring
 - prometheus
 - kubernetes
-version: 2.9.4
+version: 2.9.5
 appVersion: 1.9.7
 home: https://github.com/kubernetes/kube-state-metrics/
 sources:


lilic (commented):
/lgtm

scottrigby:
Update: @lilic had removed it from `gh-pages` but now just checked when this failed again and the protection is added back. It appears there must be some automation around adding branch protection. Perhaps this is k8s-wide. I will ask around k8s infra to try to track this down and find a solution.

For future ref, the issue with protection isn't that it's protected from force-pushing etc – it's that all checks need to pass before code can be pushed. However, the `ci-benchmark-tests` and `ci-e2e-tests` are longer running and never complete by the time the quicker chart-releaser workflow runs, so it fails with the message:
```
Pushing to branch "gh-pages"
remote: error: GH006: Protected branch update failed for refs/heads/gh-pages.        
remote: error: 9 of 9 required status checks are expected.        
To https://github.com/kubernetes/kube-state-metrics
 ! [remote rejected]   HEAD -> gh-pages (protected branch hook declined)
```

PR #1344, opened 2021-01-06
Expose sharding configuration via telemetry

Mikulas:

Kube-state-metrics component has sharding capabilities, but the sharding configuration is hidden in command arguments and cannot be easily tested after deployment. This may result in multiple forms of misconfiguration, such as:
* part of kubernetes objects may not be processed by any shard (0/3, 1/3, missing 2/3)
* some shards are duplicated, resulting in simple/naive queries yielding incorrect values (0/3, 0/3, 1/3, 2/3)
* variant of the previous state is a residue of old deployment (0/1, 0/3, 1/3, 2/3)

Auto-sharding does not by itself prevent any of those errors, but presumably makes them far less common.

This PR adds two new metrics to kube-state-metrics-self telemetry, namely:
```
# HELP kube_state_metrics_sharding_ordinal Current sharding ordinal/index of this instance
# TYPE kube_state_metrics_sharding_ordinal gauge
kube_state_metrics_sharding_ordinal 0
# HELP kube_state_metrics_sharding_total Number of total shards this instance is aware of
# TYPE kube_state_metrics_sharding_total gauge
kube_state_metrics_sharding_total 1
```

This works with explicit sharding configuration as well as autosharding.

Notes for reviewer:
* I'm not set on `ordinal` being the noun used. The code base uses `shard` which seems less descriptive, but it may be better to keep it consistent?
* Most of the new code is tests. I intentionally did not abstract any of the original tests to keep this separate. It could be eventually refactored, but since this adds new feature it is safer to not also alter the existing tests in the same patch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -150,14 +150,15 @@ http_request_duration_seconds_sum{handler="metrics",method="get"} 0.021113919999
 http_request_duration_seconds_count{handler="metrics",method="get"} 30
 ```
 
-kube-state-metrics exposes the build info metric:
+kube-state-metrics also exposes build and configuration metrics:
 ```
 kube_state_metrics_build_info{branch="master",goversion="go1.15.3",revision="6c9d775d",version="v2.0.0-beta"} 1
+kube_state_metrics_shard_ordinal 0
+kube_state_metrics_total_shards 1
 ```
 
 `kube_state_metrics_build_info` is used to expose version and other build information. For more usage about the info pattern,
 please check the blog post [here](https://www.robustperception.io/exposing-the-software-version-to-prometheus).
-
 
 ### Scaling kube-state-metrics
 
--- a/README.md
+++ b/README.md
@@ -153,12 +153,14 @@ http_request_duration_seconds_count{handler="metrics",method="get"} 30
 kube-state-metrics also exposes build and configuration metrics:
 ```
 kube_state_metrics_build_info{branch="master",goversion="go1.15.3",revision="6c9d775d",version="v2.0.0-beta"} 1
-kube_state_metrics_shard_ordinal 0
+kube_state_metrics_shard_ordinal{shard_ordinal="0"} 0
 kube_state_metrics_total_shards 1
 ```
 
 `kube_state_metrics_build_info` is used to expose version and other build information. For more usage about the info pattern,
 please check the blog post [here](https://www.robustperception.io/exposing-the-software-version-to-prometheus).
+Sharding metrics expose `--shard` and `--total-shards` flags and can be used to validate
+run-time configuration, see [`/examples/prometheus-alerting-rules`](./examples/prometheus-alerting-rules).
 
 ### Scaling kube-state-metrics
 
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -20,6 +20,7 @@ import (
 	"context"
 	"reflect"
 	"sort"
+	"strconv"
 	"strings"
 
 	"github.com/pkg/errors"
@@ -104,7 +105,9 @@ func (b *Builder) WithNamespaces(n options.NamespaceList) {
 // WithSharding sets the shard and totalShards property of a Builder.
 func (b *Builder) WithSharding(shard int32, totalShards int) {
 	b.shard = shard
-	b.shardingMetrics.Ordinal.Set(float64(shard))
+	labels := map[string]string{sharding.LabelOrdinal: strconv.Itoa(int(shard))}
+	b.shardingMetrics.Ordinal.Reset()
+	b.shardingMetrics.Ordinal.With(labels).Set(float64(shard))
 	b.totalShards = totalShards
 	b.shardingMetrics.Total.Set(float64(totalShards))
 }
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -59,7 +59,8 @@ type Builder struct {
 	ctx              context.Context
 	enabledResources []string
 	allowDenyList    ksmtypes.AllowDenyLister
-	metrics          *watch.ListWatchMetrics
+	listWatchMetrics *watch.ListWatchMetrics
+	shardingMetrics  *sharding.Metrics
 	shard            int32
 	totalShards      int
 	buildStoreFunc   ksmtypes.BuildStoreFunc
@@ -74,7 +75,8 @@ func NewBuilder() *Builder {
 
 // WithMetrics sets the metrics property of a Builder.
 func (b *Builder) WithMetrics(r prometheus.Registerer) {
-	b.metrics = watch.NewListWatchMetrics(r)
+	b.listWatchMetrics = watch.NewListWatchMetrics(r)
+	b.shardingMetrics = sharding.NewShardingMetrics(r)
 }
 
 // WithEnabledResources sets the enabledResources property of a Builder.
@@ -102,7 +104,9 @@ func (b *Builder) WithNamespaces(n options.NamespaceList) {
 // WithSharding sets the shard and totalShards property of a Builder.
 func (b *Builder) WithSharding(shard int32, totalShards int) {
 	b.shard = shard
+	b.shardingMetrics.Ordinal.Set(float64(shard))
 	b.totalShards = totalShards
+	b.shardingMetrics.Total.Set(float64(totalShards))
 }
 
 // WithContext sets the ctx property of a Builder.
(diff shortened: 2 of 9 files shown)


pkg/sharding/metrics.go
```diff
@@ -0,0 +1,48 @@
+/*
+Copyright 2019 The Kubernetes Authors All rights reserved.
```
lilic, line 2: This needs to be 2021 :)

pkg/sharding/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+)
+
+// Metrics stores the pointers of kube_state_metrics_sharding_(ordinal|total) metrics.
+type Metrics struct {
+	Ordinal prometheus.Gauge
+	Total   prometheus.Gauge
+}
+
+// NewShardingMetrics takes in a prometheus registry and initializes
+// and registers the kube_state_metrics_list_total and
```
lilic, line 31: There was some copy paste leftovers here? :)

internal/store/builder.go
```diff
@@ -74,7 +75,8 @@ func NewBuilder() *Builder {
 
 // WithMetrics sets the metrics property of a Builder.
 func (b *Builder) WithMetrics(r prometheus.Registerer) {
```
lilic, line 77: Hmm but if we don't use WithSharding this is still created, I would rather make it dependant on the Sharding enabled here.

lilic (commented):
Thanks very interesting! Left a few comments

cc @brancz can you have a look as you did the sharding feature, thanks!

internal/store/builder.go
Mikulas: Fair point! I should have mentioned that in the original description. My understanding is that KSM with no sharding configured behaves identically to `--shard 0 --total-shards 1`. In fact having the metric even for implicit (no-)sharding configuration is one of the benefits.

Is this a reasonable change https://github.com/kubernetes/kube-state-metrics/pull/1344/files#diff-8de36acb4313849c6ece61305374575fa2f18bdb6113d44022ee9749785d4af2R80 to resolve this?

Mikulas:
Hi @brancz, please take a look :)

pkg/sharding/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+}
+
+// NewShardingMetrics takes in a prometheus registry and initializes
+// and registers the kube_state_metrics_sharding_ordinal and
+// kube_state_metrics_sharding_total metrics. It returns those registered metrics.
+func NewShardingMetrics(r prometheus.Registerer) *Metrics {
+	return &Metrics{
+		Ordinal: promauto.With(r).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "kube_state_metrics_sharding_ordinal",
```
brancz, line 37: `kube_state_metrics_shard_ordinal` would be more aligned with naming best practices

pkg/sharding/metrics.go
```diff
@@ -0,0 +1,48 @@
...
+				Help: "Current sharding ordinal/index of this instance",
+			},
+		),
+		Total: promauto.With(r).NewGauge(
+			prometheus.GaugeOpts{
+				Name: "kube_state_metrics_sharding_total",
```
brancz, line 43: `kube_state_metrics_total_shards` would be more aligned with naming best practices. `_total` is reserved for counters.

brancz (commented):
Super nice! Just a small nit on naming of the metrics, otherwise this is amazing! :100:

internal/store/builder.go
brancz: Registering the metrics is fine I think, calling `.WithSharding(0, 1)` does seem a bit odd here, just in terms of placement of this call.

Mikulas: Ok, I've removed it again. Honestly I can't think of any way that would decouple sharding and metrics without overly complicating the code. The downside of current implementation (without the explicit `WithSharding(0,1)` call) is that if for whatever reason the `WithSharding` is not called later on, `kube_state_metrics_total_shards` will be set to `0` and incorrect. But hopefully this is rather academic; the builder is only used once in this specific file for this component.

Mikulas:
Wonderful, thanks :) I updated the metric names as well as the builder, PTAL

internal/store/builder.go
lilic: Yes there doesn't seem to be a smooth solution agreed. I am fine with this one, I believe its better than the alternative.

README.md
```diff
@@ -150,15 +150,16 @@ http_request_duration_seconds_sum{handler="metrics",method="get"} 0.021113919999
 http_request_duration_seconds_count{handler="metrics",method="get"} 30
 ```
 
-kube-state-metrics exposes the build info metric:
+kube-state-metrics also exposes build and configuration metrics:
 ```
 kube_state_metrics_build_info{branch="master",goversion="go1.15.3",revision="6c9d775d",version="v2.0.0-beta"} 1
+kube_state_metrics_shard_ordinal 0
```
lilic, line 156: Do you mind including a couple of use cases for these metrics, similar to how the `kube_state_metrics_build_info ` has below. Thanks!

(So either any alert suggestions or when users should use this)

lilic (approved):
One more thing left, otherwise this looks good!

README.md
Mikulas: Good point, added

Mikulas:
While adding the README documentation I thought of adding example alerts. It proved impossible to construct generic queries without also exposing the `shard_ordinal` as a label due to promql limitations. Without that the query would not be correct when there are multiple replicas for each shard.

I understand this is not a minor change and sorry for disrupting the process but the original proposal was not as useful for alerting as I hoped it to be.

jsonnet/kube-state-metrics-mixin/alerts.libsonnet
```diff
@@ -38,6 +38,43 @@
...
+            // Each shard ordinal is assigned a binary position (2^ordinal) and we compute a sum of those.
+            // This sum is compared to the expected number (2^total_shards - 1).
+            // Result of zero all shards are being scraped, anything else indicates an issue.
+            // A handy side effect of this computation is the result indicates what ordinals are missing.
+            // Eg. a result of "5" decimal, which translates to binary "101", means shards #0 and #2 are not available.
+            expr: |||
+              2^max(kube_state_metrics_total_shards{%(kubeStateMetricsSelector)s}) - 1
+                -
+              sum( 2 ^ max by (shard_ordinal) (kube_state_metrics_shard_ordinal{%(kubeStateMetricsSelector)s}) )
+              != 0
```
Mikulas, line 67: Without the `max by (shard_ordinal)` the best we can do is `max by (pod)` but that would only work if there is at most 1 replica for each shard.

jsonnet/kube-state-metrics-mixin/alerts.libsonnet
```diff
@@ -38,6 +38,43 @@
...
+              description: 'kube-state-metrics pods are running with different --total-shards configuration, some Kubernetes objects may be exposed multiple times or not exposed at all.',
+            },
+          },
+          {
+            alert: 'KubeStateMetricsShardsMissing',
+            // Each shard ordinal is assigned a binary position (2^ordinal) and we compute a sum of those.
+            // This sum is compared to the expected number (2^total_shards - 1).
+            // Result of zero all shards are being scraped, anything else indicates an issue.
+            // A handy side effect of this computation is the result indicates what ordinals are missing.
+            // Eg. a result of "5" decimal, which translates to binary "101", means shards #0 and #2 are not available.
```
Mikulas, line 62: There is an assumption that there are at most 53 shards, due to the _IEEE 754 double precision_. https://prometheus.io/docs/introduction/faq/#why-are-all-sample-values-64-bit-floats-i-want-integers

lilic:
> I understand this is not a minor change and sorry for disrupting the process but the original proposal was not as useful for alerting as I hoped it to be.

No worries, often this is how we find out which metrics make sense :)

lilic (commented):
Thank for the changes, would love to see the unit tests for these rules. From first glance looks okay to me otherwise.

Mikulas:
Thank you @lilic, very good point with the tests. I added promtool based unit tests for the new alerts and also a CI job, but that one may not be completely functional; I suspect that remains to be seen after a merge.

lilic:
Strange the CI is not reporting back, do you mind just force pushing to retrigger them? Thanks @Mikulas !

Mikulas:
Ah yes I broke CI by adding the new alert job and it never reported back. Fixed :)

PR #1346, opened 2021-01-08
options: change default host from 0.0.0.0 to ::

Frankkkkk:
Hi team,

This commit paves the way to plug-and-play IPv6-only or dual stack
clusters. Making the service listen on `::` instead of `0.0.0.0` is a no-brainer
as it still supports v4-only clusters but makes it ready for v6 ones.

Cheers !


tariq1890:
/lgtm

Thanks @Frankkkkk !!

PR #1354, opened 2021-01-16
feat: Add support for native TLS

### LINKED ISSUE #1353: Add native TLS support
Kube-State-Metrics does not support TLS, which recently got added to e.g. node_exporter
**Why is this needed**:
Secure and authenticated communication
There could be an easy way to use https://github.com/prometheus/exporter-toolkit to add support.

mayankshah1607:
Fixes #1353 

- uses [exporter-toolkit](https://github.com/prometheus/exporter-toolkit) library to add native TLS
- introduce new flag `tls-config` to pass the path to TLS config file

Signed-off-by: Mayank Shah <mayankshah1614@gmail.com>

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -51,6 +51,7 @@ Usage of ./kube-state-metrics:
       --stderrthreshold severity          logs at or above this threshold go to stderr (default 2)
       --telemetry-host string             Host to expose kube-state-metrics self metrics on. (default "::")
       --telemetry-port int                Port to expose kube-state-metrics self metrics on. (default 8081)
+      --tls-config string                 Path to the TLS configuration file
       --total-shards int                  The total number of shards. Sharding is disabled when total shards is set to 1. (default 1)
   -v, --v Level                           number for the log level verbosity
       --version                           kube-state-metrics build version information
--- a/go.mod
+++ b/go.mod
@@ -15,6 +15,7 @@ require (
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
+	github.com/prometheus/exporter-toolkit v0.5.1
 	github.com/prometheus/prometheus v2.5.0+incompatible
 	github.com/robfig/cron/v3 v3.0.0
 	github.com/spf13/pflag v1.0.5
--- a/main.go
+++ b/main.go
@@ -32,6 +32,7 @@ import (
 	"github.com/prometheus/client_golang/prometheus/promauto"
 	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
+	"github.com/prometheus/exporter-toolkit/web"
 	vpaclientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
 	clientset "k8s.io/client-go/kubernetes"
 	_ "k8s.io/client-go/plugin/pkg/client/auth"
@@ -57,9 +58,17 @@ func (pl promLogger) Println(v ...interface{}) {
 	klog.Error(v...)
 }
 
+// promLogger implements the Logger interface
+func (pl promLogger) Log(v ...interface{}) error {
+	klog.Info(v...)
+	return nil
+}
+
 func main() {
 	opts := options.NewOptions()
 	opts.AddFlags()
+
+	promLogger := promLogger{}
 
 	ctx := context.Background()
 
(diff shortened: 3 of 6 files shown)


mayankshah1607:
Hi @lilic @brancz 
If this approach looks good, I could go ahead and add relevant tests to ensure TLS is working as expected. I am not really sure why `ci-validate-manifests` is failing. I would really appreciate any help on that.

PTAL, thanks! :)

dgrisonnet:
Hi @mayankshah1607, it seems to be failing because of a conflict with the yaml package. Bumping gojsontoyaml to https://github.com/brancz/gojsontoyaml/commit/202f76bf8c1f8fb74941a845b349941064603185 should prevent that.

dgrisonnet:
If we want to go forward with the TLS configuration from the exporter-toolkit, we might want to document that the configuration can be found here: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

mayankshah1607:
> Hi @mayankshah1607, it seems to be failing because of a conflict with the yaml package. Bumping gojsontoyaml to [brancz/gojsontoyaml@202f76b](https://github.com/brancz/gojsontoyaml/commit/202f76bf8c1f8fb74941a845b349941064603185) should prevent that.

Thanks @dgrisonnet. I tried bumping it, but it did not seem to help. Could there be any other reason?

dgrisonnet:
Hum, yes after taking another look, it seems that I understood the error the other way around. The current `examples/prometheus-alerting-rules/alerts.yaml` manifests seems to have been incorrectly generated with the breaking version v2.3.0 of the yaml package. By introducing a new transitive-dependency on v2.4.0, you fixed this problem and made it possible to regenerate the manifest properly. Hence, regenerating the manifest and pushing the changes to this PR seems to be the right solution here.
Also, since the current gojsontoyaml version is not bound to v2.3.0, there is no need to bump it. Sorry for the inconvenience on that one.

mayankshah1607:
Thanks @dgrisonnet 

> If we want to go forward with the TLS configuration from the exporter-toolkit, we might want to document that the configuration can be found here: https://github.com/prometheus/exporter-toolkit/blob/master/docs/web-configuration.md

If everything looks fine in this PR, I could go ahead an add the required documentation.

docs/cli-arguments.md
```diff
@@ -51,6 +51,7 @@ Usage of ./kube-state-metrics:
       --stderrthreshold severity          logs at or above this threshold go to stderr (default 2)
       --telemetry-host string             Host to expose kube-state-metrics self metrics on. (default "::")
       --telemetry-port int                Port to expose kube-state-metrics self metrics on. (default 8081)
+      --tls-config string                 Path to TLS config file
```
tariq1890, line 54:
```suggestion
      --tls-config string                 Path to the TLS configuration file
```

go.mod
```diff
@@ -15,6 +15,7 @@ require (
 	github.com/prometheus/client_golang v1.7.1
 	github.com/prometheus/client_model v0.2.0
 	github.com/prometheus/common v0.15.0
+	github.com/prometheus/exporter-toolkit v0.5.1-0.20210113141400-89922ed1bafb
```
tariq1890, line 18: why do we have the timestamp and the commit hash?

Can we `go get @v0.5.1` again?

pkg/options/options.go
```diff
@@ -110,3 +112,8 @@ func (o *Options) Parse() error {
 func (o *Options) Usage() {
 	o.flags.Usage()
 }
+
+// GetTLSConfigPath returns the value of "tls-config" flag
+func (o *Options) GetTLSConfigPath() string {
```
tariq1890, line 117: I don't think the Get method is required

main.go
```diff
@@ -166,26 +168,23 @@ func main() {
 		})
 	}
 
+	tlsConfig := opts.GetTLSConfigPath()
```
tariq1890, line 171: Shouldn't this suffice?
```suggestion
	tlsConfig := opts.TLSConfig
```

main.go
```diff
@@ -166,26 +168,23 @@ func main() {
...
+	promlogConfig := &promlog.Config{}
```
tariq1890, line 172: We should stick to a unified logging mechanism. We already have a `promLogger` type that extends a Logging interface. We can make `promLogger` extend the new interface exposed in `web.ListenAndServe`

pkg/options/options.go
```diff
@@ -79,6 +80,7 @@ func (o *Options) AddFlags() {
 
 	o.flags.StringVar(&o.Apiserver, "apiserver", "", `The URL of the apiserver to use as a master`)
 	o.flags.StringVar(&o.Kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
+	o.flags.StringVar(&o.TLSConfig, "tls-config", "", "Path to TLS config file")
```
tariq1890, line 83: Can we also maintain lexical order of the flags now that we are making this change?
```suggestion
	o.flags.StringVar(&o.TLSConfig, "tls-config", "", "Path to the TLS config file")
```

tariq1890:
@mayankshah1607 Thanks so much for your PR. I have provided my review comments. Let me know if you need any more clarifications around them :)

mayankshah1607:
thanks @tariq1890 
I have made all the required changes. PTAL! :)

main.go
```diff
@@ -166,26 +173,21 @@ func main() {
...
-		klog.Fatalf("Failed to create MetricsServer Listener: %v", err)
-	}
+	metricsServer := http.Server{Handler: metricsMux, Addr: metricsServerListenAddress}
 
 	// Run Telemetry server
 	{
 		g.Add(func() error {
 			klog.Infof("Starting kube-state-metrics self metrics server: %s", telemetryListenAddress)
-			return telemetryServer.Serve(telemetryLn)
+			return web.ListenAndServe(&telemetryServer, tlsConfig, promLogger{})
```
tariq1890, line 190: Last comment. Let's use one instance of promLogger and assign that to a variable. We can reuse it across these methods

mayankshah1607: Done!

tariq1890:
So the PR e2e tests are failing because of this statement in the logs

`E0125 07:24:39.122249       1 main.go:63] levelinfomsgTLS is disabled.http2false`

That's a klog error. I think we might need use to klog.Info instead in the new interface implementing methd

tariq1890 (approved):

prasunsultania:
is there a sample --tls-config file ?

PR #1362, opened 2021-01-20
fix:--labels-allow-list argument in README.md

yosshi825:

If the label name is enclosed in quoted as currently written in the document, no error will be output and kms will start, but the label will not be output in the metrics.

The user wants to output the label metric and should have set it, so this document is a little difficult to understand.

Related issues and PR
https://github.com/kubernetes/kube-state-metrics/issues/1313
https://github.com/kubernetes/kube-state-metrics/pull/1279

Fixes #

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -25,35 +25,35 @@ spec:
 ```txt
 $ kube-state-metrics -h
 Usage of ./kube-state-metrics:
-      --add_dir_header                    If true, adds the file directory to the header of the log messages
-      --alsologtostderr                   log to standard error as well as files
-      --apiserver string                  The URL of the apiserver to use as a master
-      --enable-gzip-encoding              Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-  -h, --help                              Print Help text
-      --host string                       Host to expose metrics on. (default "0.0.0.0")
-      --kubeconfig string                 Absolute path to the kubeconfig file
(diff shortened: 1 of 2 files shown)


docs/cli-arguments.md
```diff
@@ -32,7 +32,7 @@ Usage of ./kube-state-metrics:
   -h, --help                              Print Help text
       --host string                       Host to expose metrics on. (default "::")
       --kubeconfig string                 Absolute path to the kubeconfig file
-      --labels-metric-allow-list string   Allows to pass a list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=["k8s-label-1","k8s-label-n",...],pods=["app"],...)'
+      --labels-metric-allow-list string   Allows to pass a list of additional Kubernetes label keys that will be used in the resource labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...')
```
lilic, line 35: Thanks for the PR! You need to modify this in the https://github.com/kubernetes/kube-state-metrics/blob/b4fb3d474bb4d1562bb11a62ae114b4c4291f582/pkg/options/options.go#L102 and afterwards run the `make generate` target, that will regenerate this file.

Do you mind also to aim this PR against release-2.0 branch, and then we can port the changes to master. As the issue originates there. Thank you! 🎉

lilic (commented):
Thanks!

docs/cli-arguments.md
yosshi825: Thank you for explaining how to do it in an easy-to-understand

lilic:
@yosshi825 I am so sorry for missing this, do you mind just rebasing then we can merge 🎉 Thank you!

PR #1378, opened 2021-02-08
internal/store/pod.go: Only create waiting_reason series if pods are in waiting state

lilic:
Due to the extremely high cardinality of this metric and with 2.0 release being breaking, this now changes that the waiting_reason metrics only include series that have state `Waiting`. We also now can include all the reasons as cardinality will only always be O(count of containers waiting). I made the adjustment for both containers and init containers.

The reduction of series can be seen in the unit tests, but I also tried it on my cluster and this is the result we only see the waiting containers:

```
# HELP kube_pod_container_status_waiting_reason Describes the reason the container is currently in waiting state.
# TYPE kube_pod_container_status_waiting_reason gauge
kube_pod_container_status_waiting_reason{namespace="default",pod="nginx-deployment-6b99d49f64-km96g",container="nginx",reason="ImagePullBackOff"} 1
kube_pod_container_status_waiting_reason{namespace="default",pod="nginx-deployment-6b99d49f64-2ghjb",container="nginx",reason="ImagePullBackOff"} 1
kube_pod_container_status_waiting_reason{namespace="default",pod="nginx-deployment-6b99d49f64-pgj2t",container="nginx",reason="ImagePullBackOff"} 1
```
Will open PRs to adjust the 0 values for other metrics here as well in another PR.

@brancz @smarterclayton @vsxen can you please have a look if this metric continues to work as you want.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1321

PR #1381, opened 2021-02-09
internal/store/pod.go: Only create terminated_reason series if pods are in terminated state

lilic:
A follow up on the https://github.com/kubernetes/kube-state-metrics/pull/1378

cc @brancz please take a look thanks!


lilic:
/hold

forgot to include the last_terminated_reason one

brancz:
Any reason why not to just do that in a follow up?

/lgtm

lilic:
Fair enough :D 

/hold cancel

PR #1383, opened 2021-02-11
chart: Add custom labels to pdb and servicemonitor

mrueg:
This adds custom labels to pdb and servicemonitor


tariq1890 (approved):
/lgtm

PR #1386, opened 2021-02-19
*: cut v1.9.8

tariq1890:


tests/e2e.sh
```diff
@@ -25,7 +25,7 @@ esac
 
 KUBERNETES_VERSION=v1.18.6
 KUBE_STATE_METRICS_LOG_DIR=./log
-KUBE_STATE_METRICS_IMAGE_NAME="quay.io/coreos/kube-state-metrics-${ARCH}"
+KUBE_STATE_METRICS_IMAGE_NAME="k8s.gcr.io/kube-state-metrics/kube-state-metrics-${ARCH}"
```
mrueg, line 28: small nitpick: k8s.gcr.io supports multi-arch with 1.9.8 so we could omit the -${ARCH} here

.github/workflows/ci.yml
```diff
@@ -194,4 +194,4 @@ jobs:
 
     - name: End-to-end tests
       run: |
-        REGISTRY="quay.io/coreos" make e2e
+        REGISTRY="k8s.gcr.io/kube-state-metrics" make e2e
```
mrueg, line 197: Does this one still have an effect at all or can we simply remove it?

mrueg:
```suggestion
        make e2e
```

tests/e2e.sh
mrueg, line 28:
```suggestion
KUBE_STATE_METRICS_IMAGE_NAME="k8s.gcr.io/kube-state-metrics/kube-state-metrics"
```

mrueg (approved):

.github/workflows/ci.yml
tariq1890: Yeah, I don't think it has an effect

tests/e2e.sh
tariq1890: I am thinking of going back to quay as the default image

mrueg (approved):

mrueg:
lgtm! Thanks @tariq1890

Issue #694, opened 2019-03-04
kube-state-metrics keep serving stale metrics after extended apiserver outage


bergerx:
**Is this a BUG REPORT or FEATURE REQUEST?**:
/kind bug

**What happened**:
We wanted to check the metrics for some pods and realised we don't have metrics for the pod in question, then check the deployment and other resources, no metrics are there, but some other pod metrics were available.
We checked the metrics exposed from the /metrics endpoint and we saw metrics exposed for pods which are deleted 2+ weeks ago.
So kube-state-metrics were reporting stale metrics for last 3 weeks.

Here are some relevant logs from kube-state-metrics pod:

```
2019-02-10T21:56:52.294138617Z E0210 21:56:52.293444       1 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=124961, ErrCode=NO_ERROR, debug=""
2019-02-10T21:56:52.295193557Z E0210 21:56:52.293956       1 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=124961, ErrCode=NO_ERROR, debug=""
2019-02-10T21:56:52.295465367Z E0210 21:56:52.294137       1 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=124961, ErrCode=NO_ERROR, debug=""
2019-02-10T21:56:52.29578788Z E0210 21:56:52.294170       1 streamwatcher.go:109] Unable to decode an event from the watch stream: http2: server sent GOAWAY and closed the connection; LastStreamID=124961, ErrCode=NO_ERROR, debug=""
... There are total of 17 `Unable to decode an event` logs within same second

2019-02-10T21:56:52.321674569Z W0210 21:56:52.321518       1 reflector.go:341] k8s.io/kube-state-metrics/pkg/collectors/collectors.go:91: watch of *v1.Endpoints ended with: very short watch: k8s.io/kube-state-metrics/pkg/collectors/collectors.go:91: Unexpected watch close - watch lasted less than a second and no items received
[...]

PR #1390, opened 2021-02-22
Propagate resource version when sharded

andreihagiescu-db:
Addresses a bug that causes a gap between `list` and `watch` when kube-state-metrics is sharded 

Partially fix for:#694 (https://github.com/kubernetes/kube-state-metrics/issues/694)
    
Kube-state-metrics does a `list` and then enters a `watch` loop. The intention is to `watch` **all** events after the initial list. The k8s API takes an optional `resource version` parameter which is returned as part of the `list` call and can be forwarded to the `watch` call, in order to fetch all events after the initial `list`.

In its sharded version, kube-state-metrics intercepts the returned `list` in order to filter out the events for other shards. It reconstructs the response, but it does not propagate the `resource version` to the modified response. The subsequent `watch` call does not refer to a resource version.

When `watch` is called without a `resource version`, it will provide a view consistent with the **most recent** resource version of the `watch` call, missing the events between the `resource version` at `list` call and the most recent one. The k8s documentation captures this as follows: _Get State and Start at Most Recent: Start a watch at the most recent resource version, which must be consistent (i.e. served from etcd via a quorum read). To establish initial state, the watch begins with synthetic "Added" events of all resources instances that exist at the starting resource version. All following watch events are for all changes that occurred after the resource version the watch started at._

Testing: Reproduced the original bug report deterministically by introducing an artificial delay (120s) in list, prior to returning the response, and terminating some pods. Unless the bug is fixed, the terminated pods continue to be reported as running for an indefinite amount of time by kube-state-metrics

### DIFF
--- a/pkg/sharding/listwatch.go
+++ b/pkg/sharding/listwatch.go
@@ -67,7 +67,7 @@ func (s *shardedListWatch) List(options metav1.ListOptions) (runtime.Object, err
 	}
 	
 	metaObj, err := meta.ListAccessor(list)
-        res.ListMeta.ResourceVersion = metaObj.GetResourceVersion()
+	res.ListMeta.ResourceVersion = metaObj.GetResourceVersion()
 
 	return res, nil
 }
--- a/pkg/sharding/listwatch.go
+++ b/pkg/sharding/listwatch.go
@@ -66,7 +66,7 @@ func (s *shardedListWatch) List(options metav1.ListOptions) (runtime.Object, err
 		}
 	}
 	
-	metaObj, err := meta.ListAccessor(list)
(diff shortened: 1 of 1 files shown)


brancz:
Nice find! This is definitely a problem, but I don't think it solves all of #694, as you mentioned in https://github.com/kubernetes/kube-state-metrics/issues/694#issuecomment-783557137, and I understand multiple reporters have seen this with non-sharded environments as well. Nonetheless this is an awesome fix for the sharded case!

We'll need you to sign the CLA, but otherwise this looks good to go! (could you also change the target branch to be `release-2.0`?)

andreihagiescu-db:
Signed the CLA

PR #1397, opened 2021-02-24
internal/store/pod.go: Only create last_terminated containers series if containers are terminated state

lilic:
See previous PRs for backgrounds. https://github.com/kubernetes/kube-state-metrics/pull/1381 and https://github.com/kubernetes/kube-state-metrics/pull/1378

`kube_pod_init_container_status_last_terminated_reason` and `kube_pod_container_status_last_terminated_reason` metrics are changed.

cc @brancz as you reviewed the others, this is the last one 🎉


jutley:
@lilic I added the `kube_pod_container_status_last_terminated_reason` metric in #535 because there was no good way to alert when a container was running out of memory. Since the container restarts automatically, it is usually not scraped when it is in the terminated state. Using the last terminated reason fixed this, because the metric continues to show that an OOM happened even when the container starts back up.

If this metric is only going to be reported when the container is in a terminated state, it seems like this use case will no longer be supported. Is there another way that we will be able to reliable alert on OOMs on kube-state-metrics:v2.0?

Issue #1309, opened 2020-11-20
Add feature to disable labels-allow-list


hoboroten:
**What would you like to be added**:
The possibility to disable --labels-allow-list
**Why is this needed**:
There are situations, when users will want to have all the labels in metrics (even keeping in mind the cardinality issues). 
**Describe the solution you'd like**
It would be great to have something like --disable-labels-allow-list, set to "false" by default.

lilic:
Hey! Thanks for the issue, we recently redesigned the labels-allow-list, it will be included in the next 2.0 release, see PR for more info https://github.com/kubernetes/kube-state-metrics/pull/1301

It now only affects, as it did before, the labels metrics e.g. kube_pod_labels{}. All other metrics have all labels allowed by default. The problem is not just unbound cardinality, but security as well, labels can contain sensitive information sometimes, so adding them to an allow list is a better approach than just allowing all. 

Can you explain why you cannot keep an allow list? Do you have too many labels on your resources? What are some use cases around k8s label metrics, etc. Thanks!

hoboroten:
Thanks for the reply.

hoboroten:
Thanks for the reply. In my case we have several different projects, located in different infrastructures (aws, eks, on-premise servers with k8s). So far, we had rather identical monitoring system thus simplifying deployment and maintenance. 
Most of monitoring filters are based on labels from kube_*_labels metrics. Such an approach gives great versatility, because:
- different environments generate different type of labels. Even k8s and software upgrades, sometimes. lead to such situation. Having all set of labels in one place (kube_*_labels metrics) allows to quickly adapt the monitoring on grafana dashboards level.
- labelling pod, nodes and other resources on the run is another useful instrument, because in most cases it's not causing restarts and do not need additional configuration of the infrastructure. 

And everything can be easily controlled and automated, because you know that all labels will be in one specific metric.
So the need to enable label list explicitly nullifies all the profits of such approach. Of course, we can try and find another metrics, which will be "stable" in different environment, having all the labels for the pod or node resource, for example. 
But in fact - it would become another "kube_*_labels@ metrics, with just different name and same cardinality and security issues.

And speaking of security issues - having the ability to give a user rights only to label resources and operate on dashboard level, IMHO, gives more control and stabilize the infrastructure, rather than having the need to configure that label list manually.
All the sensitive information in labels (is such exist) can be easily dropped on scrape/relabelling level.

lilic:
[...]

PR #1289, opened 2020-10-28
allow metric labels by default

jw-s:
Signed-off-by: Joel Whittaker-Smith <jdws.dev@gmail.com>

**What this PR does / why we need it**:
It adds each metric's default labels to be allowed for label/annotation related metrics
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1270

PR #1403, opened 2021-03-03
Add wildcard option to metric-labels-allowlist

Mikulas:

https://github.com/kubernetes/kube-state-metrics/pull/1301 introduced backwards compatibility issue by dropping all labels not defined in `--metric-labels-allowlist`. That feature is great and should arguably be enabled for all new deployments. However, constructing such a list may prove quite hard for users migrating from previous versions, especially if kube-state-metrics is managed by a different team than the ones consuming eg pod label metrics.

This patch adds a new wildcard option that allows all labels on given resources. The assumption is that using this is a last resort when migrating to 2.0 (well, 2.1) from previous version. The suggested feature is as simple as possible to allow users to migrate. Generic pattern matching (regex) are explicitly a non-goal for this PR.

**Prior art:**
https://github.com/kubernetes/kube-state-metrics/pull/1289 allow metric labels by default
https://github.com/kubernetes/kube-state-metrics/pull/1247 PR with regex support

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1309
Relates to https://github.com/kubernetes/kube-state-metrics/issues/1246


lilic:
@Mikulas Do you mind rebasing the PR, and we can have a look, thanks!

internal/store/pod_test.go
```diff
@@ -1495,8 +1498,8 @@ func TestPodStore(t *testing.T) {
 	}
 
 	for i, c := range cases {
-		c.Func = generator.ComposeMetricGenFuncs(podMetricFamilies(nil))
-		c.Headers = generator.ExtractMetricFamilyHeaders(podMetricFamilies(nil))
+		c.Func = generator.ComposeMetricGenFuncs(podMetricFamilies(c.AllowLabelsList))
```
lilic, line 1501: This is great but ideally, we want a separate test case for this. Keep testing what we did before and add this as a new test.

docs/cli-arguments.md
```diff
@@ -32,7 +32,7 @@ Usage of ./kube-state-metrics:
...
       --host string                       Host to expose metrics on. (default "::")
       --kubeconfig string                 Absolute path to the kubeconfig file
-      --labels-metric-allow-list string   Allows to pass a list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'
+      --labels-metric-allow-list string   Allows to pass a list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]').
```
lilic, line 35: Just a thought: Wonder if there are scenarios where this might not work. 🤔 the argument having `*`. Like manifests in yaml or cli.

pkg/options/types.go
```diff
@@ -130,6 +130,9 @@ func (n *NamespaceList) Type() string {
 	return "string"
 }
 
+// LabelWildcard allowlists any label
```
lilic, line 133: Do you mind adding a few test cases for the `TestLabelsAllowListSet` test in types_test.go file.

lilic (commented):
Nice work! Thank you! I am wondering if a single * would cause any problems, otherwise, this looks good apart from adding a few more test cases.

cc @tariq1890 @brancz please have a look as well as this changes our flags.

internal/store/pod_test.go
Mikulas: Agreed, updated

Mikulas:
Thanks for the review :) I rebased the PR and updated tests as recommended. I'd be happy to add more tests if deemed necessary but it seems like `pod_test.go` is the catchall for generic behaviour and the rest of stores only test resource specific behaviour.

Mikulas:
@lilic Please take a look when you get a chance :)

/assign lilic

lilic (approved):
/lgtm

I am sorry for the delay, I already reviewed but forgot to press the button 🤦‍♀️ 

Thank you so much for this!! 🎉

Mikulas:
Awesome, thank you :)

PR #1404, opened 2021-03-03
fix labels-metric-allow-list documentation

Mikulas:
The parser defined in types.go does not expect quotes
https://github.com/kubernetes/kube-state-metrics/blob/release-2.0/pkg/options/types.go#L136-L140
https://github.com/kubernetes/kube-state-metrics/blob/release-2.0/pkg/options/types_test.go#L168

This issue is unfortunate because ksm does parse the quoted syntax fine but the labels never match (eg label is `foo` but ksm matches against `"foo"`). As such this is pretty important patch for the 2.0 release. Further improvements could be made but this documentation fix is the bare minimum.

Targeting release-2.0 (or should this be based on master and backported afterwards?)


brancz:
Thank you! lgtm :+1: 

/lgtm

PR #1405, opened 2021-03-03
.github: Update helm chart workflows

mrueg:
* Uses a more recent version of helm
* Include latest fixes of chart-releaser-action

/assign @tariq1890


tariq1890:
@mrueg Thanks. Can you retarget this to `release-2.0` first? 

If we don't have the helm build workflow in release-2.0, we should take care of that first

mrueg:
Currently there''s no helm charts folder in release-2.0, I think if we stay with dedicated chart releases we should not include it there.

tariq1890 (approved):

tariq1890:
Sorry @mrueg I missed this comment. 

This PR LGTM

/lgtm

Issue #1422, opened 2021-03-22
Review permissions we set out of the box


lilic:
Currently, in our example manifests we set some permissions in the [cluster-role](https://github.com/kubernetes/kube-state-metrics/blob/master/examples/standard/cluster-role.yaml), we seemed to have had an old API in there. It would be good to look through all the permissions we suggest and remove any we do not need. 

The stretch goal for this would be to possibly add some automation for this based on the APIs we use in our resources here -> https://github.com/kubernetes/kube-state-metrics/tree/master/internal/store

mrueg:
For CI, we could look into including https://github.com/open-policy-agent/conftest into CI and use this rule set: https://github.com/swade1987/deprek8ion

fpetkovski:
Instead of reverse-engineering permission checks, I wonder if it would make sense to have something like `ksmtool` which is going to run through all of the stores and generate the necessary permissions.

PR #1423, opened 2021-03-22
Remove extensions APIs that are deprecated 

lilic:
We do not support 1.16 and older versions anymore so this can be safely removed https://github.com/kubernetes/kube-state-metrics#compatibility-matrix

@mrueg @brancz @tariq1890 please take a look, thanks!

### DIFF
--- a/examples/autosharding/cluster-role.yaml
+++ b/examples/autosharding/cluster-role.yaml
@@ -25,15 +25,6 @@ rules:
   - list
   - watch
 - apiGroups:
-  - extensions
-  resources:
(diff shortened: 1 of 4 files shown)


mrueg:
/approve
lgtm!

lilic:
I also opened https://github.com/kubernetes/kube-state-metrics/issues/1422 to review the existing permissions we grant or possibly add automation to generate those.

brancz:
Oops.. we don't even watch these APIs anymore, so  ...

/lgtm
/approve

Issue #1421, opened 2021-03-22
[Discussion] Rename --labels-metric-allow-list


mrueg:
**What would you like to be added**:

**Why is this needed**:
`--labels-metric-allow-list` is inconsistent with `--metric-allowlist` and there might be a better name for it, that is more descriptive on what it's doing.
**Describe the solution you'd like**
*  Use one way to write allowlist/allow-list in flags
* Reorder `--labels-metric` to `--metric-labels`
* Consider using `passthrough` or `filter` as with allowlist folks might wonder if there's a denylist as well.

This is how it could read:
```
      --metric-labels-passthrough string  Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'
```

mrueg:
Resolved in https://github.com/mrueg/kube-state-metrics/commit/2a7a96b65952a10dbcfa1213c9893448f0c84a48

PR #1424, opened 2021-03-22
Rename cli flag from label-metrics-allow-list to metric-labels-allowlist

mrueg:
This might be more clear on what the flag is doing. There were two ways
to spell allow-list / allowlist which it removes as well
Fixes: https://github.com/kubernetes/kube-state-metrics/issues/1421

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -25,35 +25,35 @@ spec:
 ```txt
 $ kube-state-metrics -h
 Usage of ./kube-state-metrics:
-      --add_dir_header                    If true, adds the file directory to the header of the log messages
-      --alsologtostderr                   log to standard error as well as files
-      --apiserver string                  The URL of the apiserver to use as a master
-      --enable-gzip-encoding              Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-  -h, --help                              Print Help text
-      --host string                       Host to expose metrics on. (default "0.0.0.0")
-      --kubeconfig string                 Absolute path to the kubeconfig file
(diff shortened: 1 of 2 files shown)


docs/cli-arguments.md
```diff
@@ -32,14 +32,14 @@ Usage of ./kube-state-metrics:
...
       --log_file string                   If non-empty, use this log file
       --log_file_max_size uint            Defines the maximum size a log file can grow to. Unit is megabytes. If the value is 0, the maximum file size is unlimited. (default 1800)
       --logtostderr                       log to standard error instead of files (default true)
       --metric-allowlist string           Comma-separated list of metrics to be exposed. This list comprises of exact metric names and/or regex patterns. The allowlist and denylist are mutually exclusive.
       --metric-denylist string            Comma-separated list of metrics not to be enabled. This list comprises of exact metric names and/or regex patterns. The allowlist and denylist are mutually exclusive.
+      --metric-labels-passthrough string  Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the metric contains only name and namespace labels. To include additional labels provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'
```
lilic, line 42: Either is fine by me, whatever is less confusing for our users. @brancz @tariq1890 thoughts for or against this change? We should get it into rc.1 if we want it.

brancz: hmm I think I preferred allowlist, but whatever we do, the exact spelling should be consistent with other flags .. I don't feel strongly whether `allow-list` or `allowlist`, but it should be consistent.

lilic: Sounds good to me as well, @mrueg would you be okay with making the change to the allowlist instead?

mrueg: Thanks for the feedback. I changed it to metric-labels-allowlist and regenerated the docs.

lilic (approved):
/lgtm

Thanks :shipit:

PR #1426, opened 2021-03-24
Fix docs README link to a nonexistent section

michaelmdresser:
I assume deprecation information is supposed to be contained in the metric tables in the relevant sub-documents for metric categories.

Cleans up a dead link in the docs/ README.md


lilic (approved):
Ah good catch, we removed those now and I guess forgot to remove this section.

/lgtm

Thank you!

PR #1443, opened 2021-04-08
notify about permissions in verticalpodautoscaler documentation

ThaSami:

adding a missing point about permissions to enable VPA.

Which issue(s) this PR fixes:

Fixed kubernetes#1439


lilic:
@ThaSami thank you, this looks great, you just need to sign the CLA and we are good to go. https://github.com/kubernetes/community/blob/master/CLA.md#the-contributor-license-agreement

ThaSami:
@lilic thank you. I want to ask about the v1beta2 CRDs mentioned in the documentation, this is a bit misleading as the kube-state-metrics also works on the V1 API of VPA. without installing v1beta2
do you think this is still needed to be mentioned in the documentation? as I thought that Kube-state-metrics:2.0 only exposes the v1beta2 VPA API...

fpetkovski:
Hi @ThaSami, I suggest we keep the documentation version agnostic to make sure it applies to several versions of KSM. So overall this change looks good to me too, and once you sign the CLA we can get it merged into master :tada:

PR #1450, opened 2021-04-12
tests/e2e.sh: Fix image names

mrueg:
This should remove the regression introduced in e8feba5a81343b21b6cf252bbf1cabffb39a237e


lilic (approved):
Thank you for the quick fix!

/lgtm

PR #1223, opened 2020-09-13
Use promauto to register metrics

yeya24:
Signed-off-by: Ben Ye <yb532204897@gmail.com>

**What this PR does / why we need it**:

This pr uses promauto to register metrics and does some cleanup.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1455, opened 2021-04-15
Fix builder.Builder WithMetrics signature

L3n41c:

The `WithMetrics` method signature changed from

```
func (b *Builder) WithMetrics(r *prometheus.Registry) {
```

to

```
func (b *Builder) WithMetrics(r prometheus.Registerer) {
```

in #1223 but one occurrence was forgotten.

Fixes the following compilation error when `kube-state-metrics` is used as a GO dependency of an external project:
```pkg/kubestatemetrics/builder/builder.go:48:3: cannot use builder.NewBuilder() (type *builder.Builder) as type "k8s.io/kube-state-metrics/v2/pkg/builder/types".BuilderInterface in field value:
	*builder.Builder does not implement "k8s.io/kube-state-metrics/v2/pkg/builder/types".BuilderInterface (wrong type for WithMetrics method)
		have WithMetrics(*prometheus.Registry)
		want WithMetrics(prometheus.Registerer)
```

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -44,7 +44,7 @@ func NewBuilder() *Builder {
 }
 
 // WithMetrics sets the metrics property of a Builder.
-func (b *Builder) WithMetrics(r *prometheus.Registry) {
+func (b *Builder) WithMetrics(r prometheus.Registerer) {
 	b.internal.WithMetrics(r)
 }
 


mrueg (approved):

tariq1890:
/lgtm

Thanks! @L3n41c

lilic:
Thank you! We should backport this to release-2.0 as I suspect it has the same problem?

L3n41c:
@lilic Thanks for bringing the backport point to my attention.
I’ve just opened #1459 to cherry-pick the commit of this PR to branche `release-2.0`. Is this the right thing to do ?

PR #1457, opened 2021-04-16
e2e: Test against multiple versions of k8s at the same time

### LINKED ISSUE #1454: Bump kubernetes dependencies to v1.21 version
Couple of things to do:

* In master branch, we want to bump client-go & co to `v0.21` -> https://github.com/kubernetes/kube-state-metrics/blob/master/go.mod#L41-L43

* We also need to test against Kubernetes v1.21, for that we should change this to v1.21 https://github.com/kubernetes/kube-state-metrics/blob/7fd83751066b39d6f80fd31ee190d1cd2cd6352a/tests/e2e.sh#L27 maybe even adding the ability to test against both v1.20 and v1.21. cc @mrueg what do you think?

* Document compatibility in the README file.

mrueg:
We support multiple versions of k8s, let's run e2e tests for all as suggested by @lilic 
See #1454

### DIFF
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -24,7 +24,7 @@ case $(uname -m) in
 esac
 
 NODE_IMAGE_NAME="docker.io/kindest/node"
-KUBERNETES_VERSION=v1.20.0
+KUBERNETES_VERSION=${KUBERNETES_VERSION:-"v1.20.0"}
 KUBE_STATE_METRICS_LOG_DIR=./log
 KUBE_STATE_METRICS_CURRENT_IMAGE_NAME="k8s.gcr.io/kube-state-metrics/kube-state-metrics"
 KUBE_STATE_METRICS_IMAGE_NAME="k8s.gcr.io/kube-state-metrics/kube-state-metrics-${ARCH}"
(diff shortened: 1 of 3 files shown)


lilic (approved):
Nice thank you! 🎉 
/lgtm

/hold in case any objections from other folks

dgrisonnet:
Just a side note, if this PR is merged as is, it will close #1454 because `resolves` was specified in the description even though it was meant to be a partial resolution.

tests/e2e.sh
```diff
@@ -24,7 +24,7 @@ case $(uname -m) in
 esac
 
 NODE_IMAGE_NAME="docker.io/kindest/node"
-KUBERNETES_VERSION=v1.20.0
+KUBERNETES_VERSION=${KUBERNETES_VERSION:-v1.20.0}
```
dgrisonnet, line 27: We might want to change the default to `v1.21.0` now that it has been released.

dgrisonnet:
It might also be meaningful to update the compatibility matrix as part of this effort since it verifies that kube-state-metrics is compatible with Kubernetes `v1.21.0`

tests/e2e.sh
mrueg: Done thanks!

tariq1890:
/hold cancel

Thanks @mrueg !

PR #1458, opened 2021-04-19
Replace deprecated io/ioutil

mrueg:
io/ioutil was deprecated in golang 1.16: https://golang.org/doc/go1.16#ioutil

Thinking a bit about it:
Not a super urgent change and likely will break builds with <go-1.16. Let's wait till go 1.17 is out.

### DIFF
--- a/main_test.go
+++ b/main_test.go
@@ -20,7 +20,7 @@ import (
 	"bytes"
 	"context"
 	"fmt"
-	"io/ioutil"
+	"io"
 	"net/http/httptest"
 	"sort"
 	"strconv"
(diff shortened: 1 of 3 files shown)


brancz:
Personally, I'd be ok with moving ahead with this already. But no strong opinion.

/hold
/lgtm

tariq1890:
I think we could make a go support matrix in our CI for changes like this.

We can start with go versions that are supported by the K8s versions we support. That should be a good way to test go-related breaking changes

tariq1890:
But for now, I do agree with @brancz as I have no strong reservations against this change.

Do you have any thoughts @lilic ?

PR #1464, opened 2021-04-23
.github/ci.yml: Fixup the Github Action

mrueg:
Broke CI in https://github.com/kubernetes/kube-state-metrics/pull/1457

.github/workflows/ci.yml#L219
The workflow is not valid. .github/workflows/ci.yml (Line: 219, Col: 29): Unrecognized named-value: 'cluster_version'. Located at position 1 within expression: cluster_version


.github/workflows/ci.yml
```diff
@@ -195,10 +195,8 @@ jobs:
     strategy:
       matrix:
         cluster_version:
-         - 'v1.18.0'
-         - 'v1.19.0'
-         - 'v1.20.0'
-         - 'v1.21.0'
```
tariq1890, line 201: Have we chosen to drop explicit support for 1.18 ?

I'd like to add that back at least

mrueg: The test was failing due to unsupported kubernetes object api versions in 1.18, that's why I removed it for now.

mrueg:
```
unable to recognize "tests/manifests/csr.yaml": no matches for kind "CertificateSigningRequest" in version "certificates.k8s.io/v1"

unable to recognize "tests/manifests/ingress.yaml": no matches for kind "Ingress" in version "networking.k8s.io/v1" 
```

lilic: Yes, this was due to deprecated version that was spamming the logs. https://github.com/kubernetes/kube-state-metrics#compatibility-matrix 

> Note: The v2.0.0-alpha.2+ and master releases of kube-state-metrics work on Kubernetes v1.17 and v1.18 excluding Ingress or CertificateSigningRequest resource metrics. If you require those metrics and are on an older Kubernetes version, use v2.0.0-alpha.1 or v1.9.8 kube-state-metrics release.

I since found out you can tweak client-go not to log this deprecated spam msgs, so we can add back the support if there are many requests for it. Up to you all!

mrueg:
I'll have to revert the complete change unfortunately. As @tariq1890 found out currently test-infra will not succeed on the ci-e2e-tests job since it can't find the version-specific k8s jobs. Will rebase this one with the change in, once we have support in test-infra ready.

mrueg:
Rebased on master, this needs to wait on prow to support regex or we need to hardcode k8s versions in https://github.com/kubernetes/test-infra/blob/master/config/prow/config.yaml#L293

fpetkovski:
@mrueg do we want to add 1.21 and 1.22 as well?

mrueg:
@fpetkovski this requires still some work on prow see: https://github.com/kubernetes/test-infra/issues/22164

fpetkovski:
Alright, let's merge what we have so far.

/lgtm

.github/workflows/ci.yml
```diff
@@ -192,6 +192,11 @@ jobs:
   ci-e2e-tests:
     name: ci-e2e-tests
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        cluster_version:
+         - 'v1.19.7'
+         - 'v1.20.2'
```
fpetkovski, line 199: Should we add more versions here?

PR #1466, opened 2021-04-27
[helm] remove github actions for helm continuous deploys

tariq1890:

Fixes #


mrueg:
Should we remove the chart as well and only leave a README pointing to it's new location?

tariq1890:
> Should we remove the chart as well and only leave a README pointing to it's new location?

I think we can remove the chart in the next release cycle

PR #1471, opened 2021-05-04
Add metric for node selectors on deployment

vaibhavkhurana2018:

Add a metric for adding the node selectors as prometheus labels for the deployments. This will help in plotting all the deployments for a given nodeSelector.

### DIFF
--- a/docs/deployment-metrics.md
+++ b/docs/deployment-metrics.md
@@ -15,3 +15,4 @@
 | kube_deployment_metadata_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_labels | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `label_DEPLOYMENT_LABEL`=&lt;DEPLOYMENT_LABEL&gt; | STABLE |
 | kube_deployment_created | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
+| kube_deployment_spec_nodeselector | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `label_SELECTOR_LABEL`=&lt;NODE_SELECTOR_LABEL&gt;| STABLE |
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -269,6 +269,29 @@ func deploymentMetricFamilies(allowLabelsList []string) []generator.FamilyGenera
 				}
 			}),
 		),
+		*generator.NewFamilyGenerator(
+			"kube_deployment_spec_nodeselector",
+			"The node Selector for the deployment",
+			metric.Gauge,
+			"",
+			wrapDeploymentFunc(func(d *v1.Deployment) *metric.Family {
+				if d.Spec.Template.Spec.NodeSelector == nil {
+					return &metric.Family{}
+				}
+
+				selectorlabelKeys, selectorlabelValues := kubeLabelsToPrometheusLabels(d.Spec.Template.Spec.NodeSelector)
+
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   selectorlabelKeys,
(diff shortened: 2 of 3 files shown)


internal/store/deployment_test.go
```diff
@@ -69,6 +69,8 @@ func TestDeploymentStore(t *testing.T) {
 		# TYPE kube_deployment_spec_strategy_rollingupdate_max_surge gauge
 		# HELP kube_deployment_labels Kubernetes labels converted to Prometheus labels.
 		# TYPE kube_deployment_labels gauge
+		# HELP kube_deployment_spec_nodeselector The node Selector lables for the deployment converted to prometheus labels.
```
mrueg, line 72:
```suggestion
		# HELP kube_deployment_spec_nodeselector Node selector labels for the deployment converted to Prometheus labels.
```

docs/deployment-metrics.md
```diff
@@ -15,3 +15,4 @@
 | kube_deployment_metadata_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_labels | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `label_DEPLOYMENT_LABEL`=&lt;DEPLOYMENT_LABEL&gt; | STABLE |
 | kube_deployment_created | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
+| kube_deployment_spec_nodeselector | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `label_SELECTOR_LABEL`=&lt;NODE_SELECTOR_LABEL&gt;| STABLE |
```
mrueg, line 18:
```suggestion
| kube_deployment_spec_nodeselector | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `label_NODE_SELECTOR_LABEL`=&lt;NODE_SELECTOR_LABEL&gt; | STABLE |
```

internal/store/deployment.go
```diff
@@ -269,6 +269,29 @@ func deploymentMetricFamilies(allowLabelsList []string) []generator.FamilyGenera
 				}
 			}),
 		),
+		*generator.NewFamilyGenerator(
+			"kube_deployment_spec_nodeselector",
+			"The node Selector for the deployment",
```
mrueg, line 274:
```suggestion
			"Node selector of the deployment",
```

internal/store/deployment_test.go
```diff
@@ -104,7 +106,8 @@ func TestDeploymentStore(t *testing.T) {
 				},
 			},
 			Want: metadata + `
-        kube_deployment_created{deployment="depl1",namespace="ns1"} 1.5e+09
+		kube_deployment_created{deployment="depl1",namespace="ns1"} 1.5e+09
```
lilic, line 109: Can you please clean up the formating of this and below as well, thanks!

lilic (commented):
Hey 👋 thanks for the PR, but wondering is there an issue where we discussed this to be added?

I fear the cardinality and potential security aspect of this. Node selectors can have potentially sensitive information about the node, as well as potentially a high cardinality out of the box. We right now disable label metrics by default and allowlist to be added, maybe a similar approach could be done here as well?

cc @brancz @tariq1890 for input here as well


/hold

vaibhavkhurana2018:
Hey 👋 thanks for reviewing the PR. No, we have not discussed about the issue earlier, this is one thing we were requiring for a use case, just saw this and thought of adding a PR for it. 

The cardinality of the metrics should not be high as the nature of nodeSelectors in a deployment is AND condition and it will be highly unlikely that anyone will be adding more than a couple of NodeSelectors for a given deployment. Since this is used as a filter to reduce the number of nodes from the overall nodePool, ideally the cardinality should be fine. Pardon if my understanding is wrong here. 

If we still feel on the allowlist approach, can go ahead and add that as well.

internal/store/deployment_test.go
vaibhavkhurana2018: Done. 👍

tariq1890:
Agreed with @lilic on the cardinality concerns. Having this go through the Metrics allow list sounds good to me

PR #1474, opened 2021-05-07
go.mod: Update deps and golang

mrueg:

Use golang 1.16.4 https://groups.google.com/g/golang-announce/c/cu9SP4eSXMc
Bump dependency for prometheus/common to 0.23.0
Update Prometheus CLI to 2.26.0


lilic (commented):
/lgtm

PR #1482, opened 2021-05-17
ISSUE-1465: Added kube_pod_status_ready_time and kube_pod_status_containers_ready_time

### LINKED ISSUE #1465: Support kube_pod_ready_time metric

Hello kube-state-metrics team!
I am a happy user of your metrics software. I would like the kube-state-metrics to report also the time, when pod became ready (passing readiness probe). According to [docs](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md), there are already a couple of gauges in seconds: `kube_pod_start_time`, `kube_pod_container_state_started`, etc. 

**Why is this needed**:

I would like to be able to measure the time needed for the container to become fully operational and healthy. I already have the created and start timestamp, so a simple delta query in prometheus would do the trick if a metric reporting the ready time would be implemented. 

Query the Kubernetes API to get the ready timestamp.

**Additional context**

sgrzemski:

Adds `kube_pod_status_ready_time` and `kube_pod_status_containers_ready_time` metrics to get the information provided by [Pod Lifecycle's](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) [PodConditions](https://pkg.go.dev/k8s.io/api/core/v1#PodCondition) array.

Fixes #1465


docs/pod-metrics.md
```diff
@@ -9,6 +9,8 @@
 | kube_pod_labels | Gauge | Kubernetes labels converted to Prometheus labels | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `label_POD_LABEL`=&lt;POD_LABEL&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
 | kube_pod_status_phase | Gauge | The pods current phase | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `phase`=&lt;Pending\|Running\|Succeeded\|Failed\|Unknown&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
 | kube_pod_status_ready | Gauge | Describes whether the pod is ready to serve requests | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `condition`=&lt;true\|false\|unknown&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
+| kube_pod_status_ready_time | Gauge | Time when pod passed readiness probes. | seconds | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
+| kube_pod_status_containers_ready_time | Gauge | Time when pod containers entered Ready state. | seconds | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
```
mrueg, line 13:
```suggestion
| kube_pod_status_container_ready_time | Gauge | Time when pod the container entered Ready state. | seconds | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE |
```

Would suggest to use singular here, since we already have `kube_pod_container_info` and `kube_pod_container_status_waiting`

sgrzemski: Hello @mrueg,
The k8s documentation specifies this particular condition as 
>ContainersReady: all containers in the Pod are ready. 

, hence the plural. Due to the fact there can be more than 1 containers in a pod and this condition turns into true only when **all** pods are ready, I think we should stick with plural. **What do you think?** If you still want to stick with singular that's fine for me as well, but I found it a little bit confusing.

mrueg: Thanks, good point! I'll defer to @lilic @brancz and @tariq1890 if we want to stick closer to k8s or not

mrueg:
@sgrzemski can you also sign the linuxfoundation/cla?

internal/store/pod.go
```diff
@@ -1362,6 +1364,32 @@ func createPodStartTimeFamilyGenerator() generator.FamilyGenerator {
...
+        "kube_pod_status_containers_ready_time",
+        "Readiness achieved time in unix timestamp for a pod containers.",
+        metric.Gauge,
+        "",
+        wrapPodFunc(func(p *v1.Pod) *metric.Family {
+            ms := []*metric.Metric{}
+
+			for _, c := range p.Status.Conditions {
+				if c.Type == v1.ContainersReady {
+                    ms = append(ms, &metric.Metric{
```
lilic, line 1378: This seems a bit off formating, did you run gofmt here?

internal/store/pod.go
```diff
@@ -1362,6 +1364,32 @@ func createPodStartTimeFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+func createPodStatusContainersReadyTimeFamilyGenerator() generator.FamilyGenerator {
```
lilic, line 1367: Could you add a test for both of these, thank you!

internal/store/pod.go
```diff
@@ -1433,6 +1461,32 @@ func createPodStatusReadyFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+func createPodStatusReadyTimeFamilyGenerator() generator.FamilyGenerator {
+    return *generator.NewFamilyGenerator(
+        "kube_pod_status_ready_time",
```
lilic, line 1466: But we already have kube_pod_status_ready, which is a timeseries entry in Prometheus. What would be your need for timestamp here curious?

docs/pod-metrics.md
lilic: I would prefer keeping to how we have it in kube-state-metrics which is `kube_pod_container_info`, but up to @brancz and @mrueg .

lilic (commented):
Couple of comments. Thanks for the PR!

What is the cardinality of these new metrics?

internal/store/pod.go
sgrzemski: According to [docs](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md) kube_pod_status_ready is like a boolean, 0 or 1. `kube_pod_status_ready_time` would inform you when was the last time pod entered `Ready` state.

internal/store/pod.go
sgrzemski: Added, thank you!

internal/store/pod.go
sgrzemski: Forgot it! Mea maxima culpa. Fixed my mistake.

sgrzemski:
> @sgrzemski can you also sign the linuxfoundation/cla?

Done. Thank you for pointing that out, my initial try was unsuccessful. I've retried today and succeeded.

ankitpatel96:
Hi @sgrzemski - this would be really useful for my organization.  Is there any chance you can get this across the line?  If not I'd be happy to take over the PR and get it merged.

BartVB:
It seems like all that's needed is an opinion from @brancz about:

https://github.com/kubernetes/kube-state-metrics/pull/1482#discussion_r633331970

GopiChandra25:
@k8s-ci-robot / @mrueg  and Team,
Can someone please help me we have to evaluate Pod startuptime (total time that pod tooks to up and running )using kube statemetrics, is there any metrics ?

PR #1487, opened 2021-05-24
Ignore metrics if node is removed

### LINKED ISSUE #1380: Metrics about removed node still exported
**What happened**:
Metrics still appear after a node has been removed from the cluster.
The metric is 
`kube_node_status_condition{condition="Ready",...,node="kubeX", status="true"} 0`
Kubectl has no result for this node, and it's no even alive. 

Those metrics shouldn't appear.

I haven't tried it in an isolated set up. But I think the use case is clear.

**Environment**:
- kube-state-metrics version: v1.9.7
```
Client Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:56:40Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"18", GitVersion:"v1.18.2", GitCommit:"52c56ce7a8272c798dbc29846288d7cd9fbae032", GitTreeState:"clean", BuildDate:"2020-04-16T11:48:36Z", GoVersion:"go1.13.9", Compiler:"gc", Platform:"linux/amd64"}
```
- Cloud provider or hardware configuration: on-premise
- Other info: -

kei6u:
Fixes #1380

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -19,6 +19,7 @@ package store
 import (
 	"context"
 	"strings"
+	"time"
 
 	"k8s.io/kube-state-metrics/v2/pkg/constant"
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
@@ -422,6 +423,13 @@ func wrapNodeFunc(f func(*v1.Node) *metric.Family) func(interface{}) *metric.Fam
 		for _, m := range metricFamily.Metrics {
 			m.LabelKeys = append(descNodeLabelsDefaultLabels, m.LabelKeys...)
 			m.LabelValues = append([]string{node.Name}, m.LabelValues...)
+		}
+
+		// Removed node doesn't need metrics.
+		t := node.GetDeletionTimestamp()
+		if t != nil && time.Now().After(t.Time) {
+			metricFamily.Metrics = nil
+			return metricFamily
 		}
 
 		return metricFamily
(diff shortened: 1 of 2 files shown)


fpetkovski:
I am not sure if this is the best way to fix the problem. This PR is treating the nodes differently from other kubernetes resources, but the behavior should be the same for pods, services and any other resource in the API. 

In addition, I just tried reproducing the issue by deleting a node from a cluster and KSM metrics removed all the series for that node. This leads me to believe that the issue reported in #1380 might actually be the same as https://github.com/kubernetes/kube-state-metrics/issues/694.

dgrisonnet:
I agree with @fpetkovski here, the issue you are experiencing with stale node metrics might be caused by an extended apiserver outage,

kskumgk63:
@fpetkovski @dgrisonnet 
Thanks for your review. 
Do you have any ideas to solve this issue?

fpetkovski:
I think the staleness problem is still under investigation since no one has been able to reproduce it reliably in a development environment yet. How often does this problem happen for you? Is it something that you see once in a while, or is it a regular occurence?

aclowkey:
Hi,
I didn't test it thoroughly. It was a reported *once* in my organization.
It has been a while and it wasn't reported since.

I'm not sure how much context I can give to this issue since so much time
has passed.

fpetkovski:
In that case, I would say that this is an indication of the issue being related to the staleness problem reported in https://github.com/kubernetes/kube-state-metrics/issues/694. My suggestion would be to follow that issue instead.

PR #1491, opened 2021-05-27
Use v1 PodDisruptionBudget and CronJob resources

bison:

Both the PodDisruptionBudget and CronJob resources have been promoted
to v1 as of Kubernetes v1.21 release.  The previous v1beta1 versions
are now deprecated, and will be removed in v1.25 and above.  This
updates all references to the new v1 versions.

The logs for kube-state-metrics eventually have lots of warnings on latest Kubernetes:
```
W0409 02:15:37.386026       1 warnings.go:70] batch/v1beta1 CronJob is deprecated in v1.21+, unavailable in v1.25+; use batch/v1 CronJob
W0409 02:22:41.412498       1 warnings.go:70] policy/v1beta1 PodDisruptionBudget is deprecated in v1.21+, unavailable in v1.25+; use policy/v1 
```

Sorry, not actually sure what the policy is for updating these. Does this need to wait, or should logic be added to make it dependent on underlying Kubernetes version or something?

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -29,12 +29,11 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	autoscaling "k8s.io/api/autoscaling/v2beta2"
 	batchv1 "k8s.io/api/batch/v1"
-	batchv1beta1 "k8s.io/api/batch/v1beta1"
 	certv1 "k8s.io/api/certificates/v1"
 	coordinationv1 "k8s.io/api/coordination/v1"
 	v1 "k8s.io/api/core/v1"
 	networkingv1 "k8s.io/api/networking/v1"
-	policy "k8s.io/api/policy/v1beta1"
+	policyv1 "k8s.io/api/policy/v1"
 	storagev1 "k8s.io/api/storage/v1"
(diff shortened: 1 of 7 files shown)


mrueg:
Does this break compatibility with <k8s-1.21.0?
In general I would suggest to support the supported k8s versions (which are 1.19, 1.20 and 1.21)

bison:
> Does this break compatibility with <k8s-1.21.0?
> In general I would suggest to support the supported k8s versions (which are 1.19, 1.20 and 1.21)

Yeah, it does for these resources. Is it worth it to add logic that requests the non-deprecated API version based on Kubernetes version? Or should this just be closed and revisited later?

lilic (commented):
I would say we add support for both, we had folks complain in the past when we removed support for beta versions.

Sebor:
Hello! Any news?

mrueg:
@bison do you have still plans to update this PR?

bison:
@mrueg: Sorry, I forgot about this. I'm happy to update it, but these versions are only available as of v1.21 and later. Is that acceptable for the master branch at this point?

fpetkovski:
@bison Now that we've updated our deps to 1.23, would you mind rebasing this PR so that we can merge it and resolve the deprecation warnings?

bison:
@fpetkovski, done.

PR #1468, opened 2021-04-29
Add kube_*_annotations metrics

sylr:
**What this PR does / why we need it**:

Adds kube_*_annotations for every existing  kube_*_labels counter parts.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #1495, opened 2021-06-04
Add annotation metrics

### LINKED ISSUE #941: annotation metrics, what is the future direction?

sepich:
This is continuation of https://github.com/kubernetes/kube-state-metrics/pull/1041:
Provides a metric which exposes annotations for that particular API kind. By default it is disabled, but user could specify annotation keys to expose per-each API kind.
- Taken into account https://github.com/kubernetes/kube-state-metrics/issues/941#issuecomment-729560572
- If no `--metric-labels-allowlist/--metric-annotations-allowlist` provided for family - then label/annotation metrics would not be even exposed. Because it seems there is no sense in such metric alone:
```
kube_namespace_labels{namespace="name"} 1
```
Sorry for large PR size, but this feature affects almost all families.

Fixes #941


lilic:
Hey, did you see there is https://github.com/kubernetes/kube-state-metrics/pull/1468. Can you explain the diff between this one and that one, thanks!

sepich:
Sorry, i haven't seen #1468 as it not linked to the issue.
I've checked the implementation and the difference is:
- consider starting ksm with:
`--metric-labels-allowlist=pods=[team] --metric-annotations-allowlist=namespaces=[channel]`
In 1468 for each API type (family) metrics like this would be created 
```
kube_<kind>_labels{namespace="ns",<kind>="name"} 1
kube_<kind>_annotations{namespace="ns",<kind>="name"} 1
```
But only `kube_pod_labels` and `kube_namespace_annotations` would have useful additional metric labels. 
In my case only these metrics would be exposed:
```
kube_pod_labels{namespace="...",pod="...",label_team="..."} 1
kube_namespace_annotations{namespace="ns",annotation_channel="..."} 1
```
(So, no `kube_deployment_labels` for example)
Which is more in line with `--metric-allowlist` usage.
(In case starting ksm without `--metric-labels-allowlist/--metric-annotations-allowlist` cli arguments set, this PR would reduce unneeded cardinality. Idea is from https://github.com/kubernetes/kube-state-metrics/issues/941#issuecomment-729560572
> By default this would be all disabled and users could just enable annotations they are interested in.

- this PR has test cases for both Labels and Annotations, that's why its size is bigger

dgrisonnet:
Since @sylr started looking into this feature first, https://github.com/kubernetes/kube-state-metrics/pull/1468 supersedes this PR. 

That said, both of you could collaborate to make this feature better. Since you have invested some time to look into this feature @sepich, your opinion can be very valuable to improve #1468.

PR #1499, opened 2021-06-07
Replace multiListerWatcher with independent listWatchers per namespace

### LINKED ISSUE #1413: ksm doesn't report any metrics at all if it lacks rights for just 1 subject namespace
**What happened**: ksm doesn't report any metrics at all if just 1 namespace from the list of specified ones is not available.

**What you expected to happen**: ksm should return metrics for the k8s objects from other namespaces (where it has access to).

**How to reproduce it (as minimally and precisely as possible)**: have arg '--namespaces=project1,project2' but give access only to ksm's ServiceAccount 'view' rights only in project1.
ksm will produce **NO** metrics, because its ServiceAccount lacks 'view' rights in project2, although it could return metrics from project1.

**Anything else we need to know?**: no

**Environment**:
- kube-state-metrics version: 1.9.0
- Kubernetes version (use `kubectl version`): 1.18.3+002a51f
- Cloud provider or hardware configuration: OpenShift (on premises)
- Other info: none.

fpetkovski:
The `multiListerWatcher` is a composite object encapsulating multiple
`ListerWatchers` and implements the `ListerWatcher` interface.
With the current implementation, when an individual lister fails, the
entire List operation fails. This causes no metrics to be shown when KSM
has no permissions to a single namespace.

In addition to this, the `multiListerWatcher` takes advantage of internal
implementation details if the client-go library by modifiying and
relying on the `ResourceVersion` metadata field. This introduces a bug
where reconnecting to the API server will break the `multiListerWatcher`
completely.

This commit replaces the `multiListerWatcher` with individual
`ListerWatchers` per each configured namespace, resolving both issues. 
It introduces a `MetricsWriter` interface to writes out metrics from stores, 
and implements it in the `MetricsStore` struct. It also introduces a `MultiNamespaceMetricsWriter` 
struct implementing `MetricsWriter` to allow writing out metrics from multiple 
related stores (stores holding data for objects of the same kind but different namespace).

When KSM has no permissions for resources in at least one namespace, it will not export any metrics on the `/metrics` endpoint.

Fixes #1413 

```release-note
Fixes a bug where KSM did not export any metrics when it had no permissions for resources in at least one namespace
```


pkg/listwatch/listwatch.go
```diff
@@ -135,24 +137,27 @@ type multiListerWatcher []cache.ListerWatcher
...
 		if err != nil {
-			return nil, err
+			klog.Errorf("Failed to list accessor: %v", err)
+			continue
 		}
 		for _, item := range items {
 			l.Items = append(l.Items, runtime.RawExtension{Object: item.DeepCopyObject()})
 		}
-		resourceVersions = append(resourceVersions, metaObj.GetResourceVersion())
+		resourceVersions[i] = metaObj.GetResourceVersion()
```
fpetkovski, line 160: Since we continue in case of errors, we need to switch from appending to to using an indexed based accessor. Otherwise, the resource versions will get out of sync with the listers used to get the resource.

dgrisonnet: I wonder if this change has any impact on the watchers. Originally, the resource versions provided to the watch functions were always valid, but now when the `List` operation fails an empty resource version is provided to the `Watch` function.

As far as I can tell, the bug originally reported will still exist with the watchers since whenever a watch operation fails, the whole multi lister watcher fails [1]. Now that the watch functions are passed incorrect resource versions (`""` in case the `List` failed), failures in the `Watch` operation should occur more often than what it used to, so we will end up with the same issue but somewhere else. 

[1] https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/listwatch/listwatch.go#L210-L214

fpetkovski: This is a good catch! Do you see any problems with modifying the multiWatch to not fail when an individual watcher fails?

dgrisonnet: I don't see any problem with that. However, I am wondering if the failed Listers should really be propagated to the watchers. Theoretically, if the `list` operation failed, I wouldn't expect us to try to setup a watcher for the failed listwatcher, but with the way, the code is currently, that's what we would be doing.

fpetkovski: I am not sure how we would do that. The `List` and `Watch` functions operate on a `multiListerWatcher` which is an alias for `[]cache.ListerWatcher`. There are different reasons why a `List` operation can fail other than permissions issues, for example there could be a temporary 5xx error from the kube-apiserver. If the `List` function modifies the slice it operates on, a relist operation will never take into account the failed list watchers.

fpetkovski: Another thing I noticed was that the ResourceVersion which the `List` function modifies actually persists between relists. When watching two namespaces, this leads to errors in the form of: 

```
Failed to watch *v1.Pod: failed to list *v1.Pod: resourceVersion: Invalid value: "1829687/1829687": strconv.ParseUint: parsing "1829687/1829687": invalid syntax
```
This is because the resource version for the list option is set [here](https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/listwatch/listwatch.go#L159) and is never reset.

I believe this might be causing https://github.com/kubernetes/kube-state-metrics/issues/694.

internal/store/builder.go
```diff
@@ -218,150 +223,169 @@ func availableResources() []string {
...
-		familyHeaders,
-		composedMetricGenFuncs,
-	)
-	b.reflectorPerNamespace(expectedType, store, listWatchFunc)
+	if isAllNamespaces(b.namespaces) {
+		store := metricsstore.NewMetricsStore(
+			familyHeaders,
+			composedMetricGenFuncs,
+		)
+		listWatcher := listWatchFunc(b.kubeClient, "")
```
dgrisonnet, line 356: I would prefer `v1.NamespaceAll` over `""`.

internal/store/builder.go
```diff
@@ -218,150 +223,169 @@ func availableResources() []string {
...
+func (b *Builder) buildVPAStore() []*metricsstore.MetricsStore {
 	return b.buildStoreFunc(vpaMetricFamilies(b.allowLabelsList["verticalpodautoscalers"]), &vpaautoscaling.VerticalPodAutoscaler{}, createVPAListWatchFunc(b.vpaClient))
 }
 
-func (b *Builder) buildLeases() cache.Store {
+func (b *Builder) buildLeases() []*metricsstore.MetricsStore {
 	return b.buildStoreFunc(leaseMetricFamilies, &coordinationv1.Lease{}, createLeaseListWatch)
 }
 
 func (b *Builder) buildStore(
```
dgrisonnet, line 342: We might want to rename this function to buildStores

dgrisonnet (commented):
In the description of your PR, would you mind detailing the changes made to the store?

pkg/listwatch/listwatch.go
fpetkovski: Will resolve this since the file was removed.

pkg/listwatch/namespace_denylist.go
```diff
@@ -17,162 +17,10 @@ limitations under the License.
...
-				next.Stop()
-				return
-			}
-		}
-	}()
-
-	return proxy
-}
-
 // getNamespace returns the namespace of the given object.
```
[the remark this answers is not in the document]
fpetkovski: `getNamespace` seems to be unused. Remove the entire file.

internal/store/builder.go
fpetkovski: I renamed it to `buildStores` in a different commit since the change is a bit bigger.

pkg/metrics_store/metrics_writer_test.go
```diff
@@ -0,0 +1,224 @@
...
+
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
+	metricsstore "k8s.io/kube-state-metrics/v2/pkg/metrics_store"
+)
+
+func TestMultiNamespaceMetricsWriter_SingleStores_WriteAll(t *testing.T) {
```
dgrisonnet, line 31: Using underscores in function names is against the Golang best practices. Also, since it is a test for the metrics_writer package, it is redundant to mention it in the test name. Usually, you want to name your unit tests `Test<FunctionName>`, so in your case, you could have one main test function called `TestWriteAll` which would execute two sub-tests, one testing single store and another for multiple stores.

fpetkovski: Good point. I think tests can have different naming conventions depending on the project, but looking at the ones we have, they are all named in PascalCase. 

I renamed the tests in this commit them to match the project's convention, but I wouldn't try to merge them into a single test since the assertions are a bit different and it might just complicate the test. This is also in line with existing tests [1][2].

[1] https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/endpoint_test.go#L29
[2] https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/endpoint_test.go#L104

dgrisonnet (approved):
lgtm

/cc @kubernetes/kube-state-metrics-maintainers

pkg/builder/types/interfaces.go
```diff
@@ -38,17 +40,17 @@ type BuilderInterface interface {
...
-	DefaultGenerateStoreFunc() BuildStoreFunc
-	Build() []cache.Store
+	WithGenerateStoresFunc(f BuildStoresFunc)
+	DefaultGenerateStoresFunc() BuildStoresFunc
+	Build() []metricsstore.MetricsWriter
 }
 
-// BuildStoreFunc function signature that is use to returns a cache.Store
-type BuildStoreFunc func(metricFamilies []generator.FamilyGenerator,
+// BuildStoresFunc function signature that is use to returns a list of metricsstore.MetricsStore
```
mrueg, line 49:
```suggestion
// BuildStoresFunc function signature that is used to return a list of metricsstore.MetricsStore
```

mrueg (approved):

mrueg:
/assign @tariq1890 

we might want to backport this to release-2.1 as well.

brancz:
Nice work!

/lgtm
/approve

brancz:
I have a suspicion that this might also help with https://github.com/kubernetes/kube-state-metrics/issues/694

PR #1506, opened 2021-06-21
Allow quotes in LabelsAllowList

fpetkovski:
Allows using quotes when setting the `--metric-labels-allowlist` option

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1313

### DIFF
--- a/pkg/options/types_test.go
+++ b/pkg/options/types_test.go
@@ -202,13 +202,48 @@ func TestLabelsAllowListSet(t *testing.T) {
 					"bar",
 					"*"}}),
 		},
+		{
+			Desc:  "with quoted tokens",
+			Value: `"nodes=["kubernetes.io/arch","gpu.infra/intel","network.infra/fast"]"`,
+			Wanted: LabelsAllowList(map[string][]string{
+				"nodes": {
+					"kubernetes.io/arch",
+					"gpu.infra/intel",
+					"network.infra/fast",
+				}}),
+		},
+		{
+			Desc:  "with quoted labels",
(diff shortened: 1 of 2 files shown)


dgrisonnet (approved):
LGTM

fpetkovski:
@lilic could you please take a look at this one?

pkg/options/types_test.go
```diff
@@ -202,13 +202,48 @@ func TestLabelsAllowListSet(t *testing.T) {
 					"bar",
 					"*"}}),
 		},
+		{
+			Desc:  "with quoted tokens",
```
mrueg, line 206: Can you add one negative test with a non-matching quotation?

something like `"node=["kubernetes.io/arch]"`

fpetkovski: Good point. Since we do not validate the expression syntactically, such a test case would actually pass. It seems that the quotes are simply skipped during processing. 

Is this something we want to tackle as well?

mrueg: I don't have a strong opinion here, maybe we can also use an upstream parser for this? https://github.com/kubernetes/apimachinery/tree/master/pkg/labels

fpetkovski: I looked into the labels parser and it also does not accept quotation marks. On the other hand, it will explicitly fail with:
```
Invalid value: "\"nodes": name part must consist of alphanumeric characters
```
We could also do the same as an alternative to disregarding quotes.

mrueg: My preference would be to match the behavior of what's in Kubernetes

PR #1519, opened 2021-07-01
capture start time for containers in terminated state

### LINKED ISSUE #1467: kube_pod_container_state_started is not returned for some pods
**What happened**:
We have multiple pods (jobs) that run for a short period of time (the time from container start to termination varies from 2-3 seconds to a couple of minutes for each pod). We use a query `kube_pod_container_state_started{pod="needePodName"}` in Prometheus and it appears that it works for some pods and doesn't show the result for others (while other queries, including kube_pod_start_time show the result for all of the pods). 

`kube_pod_container_state_started` shows result for every pod, and not just for some of them.

Run multiple pods with short-time living containers, check that a query `kube_pod_container_state_started` to prometheus doesn't show the results for some of the pods.

If we check the pods with kubectl describe pod [pod_name] - the values we need (container start time) are present. When we make a query with `kube_pod_start_time` for every pod - we get the expected result (that is the timestamp when the pod started). But now we want to get the timestamps when the Containers were started.

**Environment**:
- kube-state-metrics version: v.2.0.0
- Kubernetes version (use `kubectl version`): v.1.19.6

harjas27:
In reference to the metric:
```
kube_pod_container_state_started
Start time in unix timestamp for a pod container
```
The metric is collected only for the containers which are in "Running" state and not when the state of the container is "Terminated". The metric can be collected for the "Terminated" containers as well, this PR adds the logic for the same. 
This will fix the issue where this metric is missing for pods which run for a short time and the containers are terminated after completion of the job/task

Fixes #1467

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -292,6 +292,12 @@ func createPodContainerStateStartedFamilyGenerator() generator.FamilyGenerator {
 						LabelValues: []string{cs.Name},
 						Value:       float64((cs.State.Running.StartedAt).Unix()),
 					})
+				} else if cs.State.Terminated != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"container"},
+						LabelValues: []string{cs.Name},
+						Value:       float64((cs.State.Terminated.StartedAt).Unix()),
+					})
 				}
 			}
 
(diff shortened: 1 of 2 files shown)


dgrisonnet (commented):
Would you mind also adding the start time for the `Waiting` status so that we have it for all the status?

harjas27:
@dgrisonnet there is no field for `StartedAt` in the `ContainerStateWaiting` struct:
[https://github.com/kubernetes/api/blob/499b6f90564cff48dc1fba56d974de2e5ec98bb4/core/v1/types.go#L2441-L2448](url)

dgrisonnet:
Ah, indeed I thought it was common to all status.

/lgtm

PR #1525, opened 2021-07-08
Exposing selector metrics for service

### LINKED ISSUE #1473: Add selectors metric to service
Generally, all Kubernetes services have a Selector field. 
The controller for the Service selector continuously scans for Pods that match its selector.

But selectors metric is missing in kube-state-metrics service.
It would be good to have kube_service_selectors metric so that selectors information can be exposed

Example : 

╰─$ kubectl describe svc my-nginx
```
Name:              my-nginx
Namespace:         default
Labels:            app=my-nginx
                   run=my-nginx
Annotations:       <none>
Selector:          run=my-nginx
Type:              ClusterIP
IP:                10.108.77.232
Port:              <unset>  80/TCP
TargetPort:        80/TCP
Endpoints:         10.1.2.63:80
Session Affinity:  None
Events:            <none>
```

harekal7:

Currently, kube-state-metrics does not expose the selectors for services
This PR is adding the kube_service_spec_selectors metric

Fixes #1473


mrueg:
Since selectors can similarly to labels and annotations increase cardinality a lot, we might want to have an allow-list feature for this as well.

lilic:
Agreed with Manuel here, what is your use case for this?

harekal7:
The use case is - in one of our applications, we need to establish a relationship between service and pods. Labels are already there for pods, we just need selectors from the service for this to work.
I can add an allow-list option for selectors similar to labels - if that will help to mitigate the concerns regarding cardinality

PR #1528, opened 2021-07-13
docs: Fix typo

mrueg:
The object is called Certificate Signing Request and not Certificate*s* Signing Request.
 
No change


lilic (commented):
/lgtm

Thanks

PR #1537, opened 2021-07-22
Use BuilderInterface instead of internal/store.Builder in metricshandler

FillZpp:
Signed-off-by: FillZpp <FillZpp.pub@gmail.com>

Use `BuilderInterface` instead of `internal/store.Builder` in metricshandler package.

The MetricsHandler should only care about the interface, not the implementation behind. So that other 3rd party projects can use kube-state-metrics as the basic dependency of their state-metrics subprojects.

### DIFF
--- a/pkg/metricshandler/metrics_handler.go
+++ b/pkg/metricshandler/metrics_handler.go
@@ -33,7 +33,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
 
-	"k8s.io/kube-state-metrics/v2/internal/store"
+	ksmtypes "k8s.io/kube-state-metrics/v2/pkg/builder/types"
 	metricsstore "k8s.io/kube-state-metrics/v2/pkg/metrics_store"
 	"k8s.io/kube-state-metrics/v2/pkg/options"
 )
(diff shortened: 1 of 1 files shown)


fpetkovski:
Nice one, this LGTM

FillZpp:
> Nice one, this LGTM

Thanks!
/cc @tariq1890 @lilic

FillZpp:
PTAL @brancz @tariq1890 @lilic . I'm waiting for it in another 3rd party state-metrics project. Thanks.

pkg/metricshandler/metrics_handler.go
```diff
@@ -56,7 +56,7 @@ type MetricsHandler struct {
 }
 
 // New creates and returns a new MetricsHandler with the given options.
```
lilic, line 58: Thanks, makes sense, we even have a comment about it :D https://github.com/kubernetes/kube-state-metrics/blob/1d61fc146160a94ad6b932b9fc5124bc43a5e00c/internal/store/builder.go#L54

lilic (approved):
We have a /lgtm

Thanks! Curious what project you are using this for? :)

FillZpp:
> Thanks! Curious what project you are using this for? :) 

I'm creating [openkruise/kruise-state-metrics](https://github.com/openkruise/kruise-state-metrics), a component that generate and expose metrics for the [custom resources in OpenKruise](https://github.com/openkruise/kruise#introduction).

I try to bring kube-state-metrics as it's basic dependency. So that it only needs to implement metrics of the CRs in `internal/store`, and then registers the custom builder to the metrics handler of kube-state-metrics.

PR #1514, opened 2021-06-25
Add `WithAllowLabels` to public `BuilderInterface`

ahmed-mez:
**What this PR does / why we need it**:

fixes: https://github.com/kubernetes/kube-state-metrics/issues/1515

Add the method `WithAllowLabels` to public `BuilderInterface`.

It provides an option to configure `--metric-labels-allowlist` when `kube-state-metrics` is used as a Go dependency of another project.

PR #1540, opened 2021-07-28
Cherry pick for v2.1.1 / Cut Release 

mrueg:
Cut a new release and backport some fixes.

    *: Cut v2.1.1 release
    
    * [CHANGE] Replace multiListWatcher with independent listWatchers per namespace #1499
    * [BUGFIX] Use BuilderInterface instead of internal/store.Builder in metricshandler #1537
    * [BUGFIX] Add WithAllowLabels to public BuilderInterface #1514


CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
+## v2.1.1 / 2021-07-28
+
+* [CHANGE] Replace multiListWatcher with independent listWatchers per namespace #1499
```
lilic, line 3: These are changes not bug fixes from what I can tell, do we still want to include them in v2.1.1? Also the bug fixes seem not related to the 2.1 release, so maybe we need to instead cut 2.2? What do you think?

mrueg: I believe they are internal changes (that fix certain issues), so suitable for a patch release.
I would release v2.2 with kubernetes 1.22 support.

lilic: That sounds good to me, thanks!

lilic (approved):
/lgtm

Thank you!

CHANGELOG.md
dgrisonnet, line 3: I think this should be a bugfix since it fixes the permission issue from #1413. Also, we could reuse the release note that @fpetkovski added to his PR:
> Fixes a bug where KSM did not export any metrics when it had no permissions for resources in at least one namespace

CHANGELOG.md
```diff
@@ -1,3 +1,10 @@
...
+* [CHANGE] go.mod: Update and minimize dependencies #1529
+* [BUGFIX] Use BuilderInterface instead of internal/store.Builder in metricshandler #1537
+* [BUGFIX] Add WithAllowLabels to public BuilderInterface #1514
```
dgrisonnet, line 6: I would consider both of these PRs changes instead of bug fixes since they are actual improvements made to the builder interface to make it easier for users to extend ksm functionalities.

CHANGELOG.md
mrueg: I think for 1.9.x we can do that upon request and if someone volunteers doing the backports (so we are sure that there's a need). 
I believe for 2.0 we don't need to create a new release, 2.1 should work for those users as well.

PR #1541, opened 2021-07-28
README.md: Replace go get with go install

mrueg:
go recommends to use `go install` instead of `go get`
None


mrueg:
Invalid, I misread the README. Apologies for the noise.

PR #1544, opened 2021-07-29
cloudbuild.yaml: Update cloudbuild image

mrueg:
I don't have a great way to test this, but I believe in spirit of best practices we should not rely on an almost 2y old image here.

Best option to test it locally was something like: `docker run -it --rm -w /tmp/ksm -v $(pwd):/tmp/ksm -v /var/run:/var/run "gcr.io/k8s-testimages/gcb-docker-gcloud:v20210722-085d930"` and then call make push inside it.

None


fpetkovski:
Looks good
/approve

PR #1548, opened 2021-08-13
Add arg use-api-server-cache to set resourceVersion=0 for ListWatch

### LINKED ISSUE #1518: Support returning cached results from api server (resourceVersion=0)

The ability to specify whether or not to use the k8s api server by setting the `resourceVersion` parameter ([docs](https://kubernetes.io/docs/reference/using-api/api-concepts/#the-resourceversion-parameter)). 

**Why is this needed**:

This would be immensely helpful for us as we run this application at quite a large scale (about 8mil total timeseries, 37k pods / cluster).

We constantly get these errors, which we presume is the cause for many of our metrics to not reflect reality (e.g. terminated pods that still exist as `Running` in metrics):

```
W0630 18:02:48.644785       1 reflector.go:302] k8s.io/kube-state-metrics/internal/store/builder.go:314: watch of *v1.ConfigMap ended with: too old resource version: 2066904261 (2099124935)
```

(not just with `ConfigMaps` but every resource type)

We don't need to be so strongly consistent, thus it would be nice to allow expose this option.

Allow for a CLI arg `--use-api-server-cache` that will explicitly input `resourceVersion=0` to the api server calls.

**Additional context**

Example of a client that does this: https://github.com/fabric8io/fluent-plugin-kubernetes_metadata_filter/commit/3e11f395e5539156bec990d9b71683f0b4fe7c75

liangyuanpeng:

ref #1518

The ability to specify whether or not to use the k8s api server by setting the resourceVersion parameter (docs).

This would be immensely helpful for us as we run this application at quite a large scale (about 8mil total timeseries, 37k pods / cluster).

We constantly get these errors, which we presume is the cause for many of our metrics to not reflect reality (e.g. terminated pods that still exist as Running in metrics):
```
W0630 18:02:48.644785       1 reflector.go:302] k8s.io/kube-state-metrics/internal/store/builder.go:314: watch of *v1.ConfigMap ended with: too old resource version: 2066904261 (2099124935)
```
(not just with ConfigMaps but every resource type)

We don't need to be so strongly consistent, thus it would be nice to allow expose this option.

Fixes #1518

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -54,6 +54,7 @@ Usage of ./kube-state-metrics:
       --telemetry-port int               Port to expose kube-state-metrics self metrics on. (default 8081)
       --tls-config string                Path to the TLS configuration file
       --total-shards int                 The total number of shards. Sharding is disabled when total shards is set to 1. (default 1)
+      --use-api-server-cache             Set resourceVersion=0 for ListWatch
   -v, --v Level                          number for the log level verbosity
       --version                          kube-state-metrics build version information
       --vmodule moduleSpec               comma-separated list of pattern=N settings for file-filtered logging
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -54,7 +54,7 @@ Usage of ./kube-state-metrics:
       --telemetry-port int               Port to expose kube-state-metrics self metrics on. (default 8081)
       --tls-config string                Path to the TLS configuration file
       --total-shards int                 The total number of shards. Sharding is disabled when total shards is set to 1. (default 1)
-      --use-api-server-cache             Set resourceVersion=0 for ListWatch
+      --use-apiserver-cache              Sets resourceVersion=0 for ListWatch requests, using cached resources from the apiserver instead of an etcd quorum read.
   -v, --v Level                          number for the log level verbosity
       --version                          kube-state-metrics build version information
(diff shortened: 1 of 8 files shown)


liangyuanpeng:
Had sign the CLA.
Checked git commit for email.
Checked github for email.

docs/cli-arguments.md
```diff
@@ -54,6 +54,7 @@ Usage of ./kube-state-metrics:
       --telemetry-port int               Port to expose kube-state-metrics self metrics on. (default 8081)
       --tls-config string                Path to the TLS configuration file
       --total-shards int                 The total number of shards. Sharding is disabled when total shards is set to 1. (default 1)
+      --use-api-server-cache             Set resourceVersion=0 for ListWatch
```
mrueg, line 57:
```suggestion
      --use-apiserver-cache             Sets resourceVersion=0 for ListWatch requests, using cached resources from the apiserver instead of an etcd quorum read.
```
rename the flag to be consistent with how we use apiserver in flags

mrueg:
/lgtm
/hold
for another day to give @kubernetes/kube-state-metrics-maintainers the possibility to review as well.

mrueg:
Thanks for your contribution @liangyuanpeng !

jpdstan:
Thanks so much for working on this @liangyuanpeng !

PR #1549, opened 2021-08-13
Endpoint ip addresses is not provided by kube-state-metrics

### LINKED ISSUE #1472: Endpoint ip address is not provided by kube-state-metrics
We are not getting ip address of endpoints from kube-state-metrics. The variable "kube_endpoint_address_available" in https://github.com/kubernetes/kube-state-metrics/blob/master/docs/endpoint-metrics.md doc page only gives count of endpoint ip addresses but not provide all value of ip addresses.
We need these ip addresses for monitoring our k8s cluster using kube-state-metrics.

avinashk442:
We are not getting ip addresses of endpoints from kube-state-metrics. The variable "kube_endpoint_address_available" in https://github.com/kubernetes/kube-state-metrics/blob/master/docs/endpoint-metrics.md doc page only gives count of endpoint ip addresses but not provide all value of ip addresses.
We need these ip addresses for monitoring our k8s cluster using kube-state-metrics.

Fixes #1472


avinashk442:
dgrisonnet , @fpetkovski  Please review the PR 🙂

avinashk442:
@dgrisonnet  @fpetkovski Can you please review the PR.
Thanks!

dgrisonnet:
@avinashk442 would be mind explaining a bit more what your use case is? From my perspective, having the name and namespace of the endpoint should already give you enough information to know which endpoint has unavailable addresses.

avinashk442:
@dgrisonnet  yeah we can get infos about which endpoint has notreadyaddress and available addresses and also we can get count of these addresses too as currently provided by kube-state-metrics but we can't get the ip-addresses values of notreadyaddress and available addresses. So, I am adding to get these addreses value for monitoring purposes using kube-state-metrics.

avinashk442:
@dgrisonnet  can you please review the PR again?
Thanks!

dgrisonnet:
> So, I am adding to get these addreses value for monitoring purposes using kube-state-metrics because we need it a lot of time.

I don't think, there is a valid monitoring use case that would require you to have IP addresses upfront. The `kube_endpoint_address_not_ready` metric is a good fit for alerting and as an administrator the most valuable information when you will be paged is the endpoint that is at fault and its namespace. Once you have that, you can easily go to your cluster and investigate further.

A comment regarding your approach with `kube_endpoint_address_not_ready`, it is not idiomatic to have labels represent a list of values since it is very likely to end up causing cardinality issues. In your case, you should merge IP addresses in a list separated by `|`.

Some other thing to consider here is that kube-state-metrics as a 1:1 relationship with Kubernetes resources, so we shouldn't add new metrics here, but rather updating the existing one. Also, adding IP addresses to the metrics, will increase their cardinality a lot since instead of having one timeserie per endpoint, we would now have one timeserie per IP address.

avinashk442:
@dgrisonnet I agree `kube_endpoint_address_not_ready_count` is good fit for alerting but we also want information about IP addresses from kube-state-metrics so that we don't have to talk to API Server explicitly for only getting IP addresses since kube-state-metrics is already talking and getting informations. Otherwise we have to take some data from kube-state-metrics and some from API Server directly. I don't think this should be ideal. 
If you check I have done the same that is merged IP addresses in a list separated by `|` so avoid cardinality issue. It is one timeseries per endpoint, it is not one timeserie per IP address.

avinashk442:
@dgrisonnet can you check my comment and review the PR again?

avinashk442:
Updated the screenshots result accordingly.

avinashk442:
@dgrisonnet  @fpetkovski  please review the PR

docs/endpoint-metrics.md
```diff
@@ -3,8 +3,10 @@
 | Metric name| Metric type | Labels/tags | Status |
 | ---------- | ----------- | ----------- | ----------- |
 | kube_endpoint_annotations | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; <br> `annotation_ENDPOINT_ANNOTATION`=&lt;ENDPOINT_ANNOTATION&gt;  | EXPERIMENTAL |
-| kube_endpoint_address_not_ready | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; | STABLE |
-| kube_endpoint_address_available | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; | STABLE |
+| kube_endpoint_address_not_ready_total | Counter | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; | EXPERIMENTAL |
```
fpetkovski, line 6: We cannot remove stable metrics because there might be people relying on them. The metric type should also remain a gauge because the number of IPs can go up and down depending on the number of pods behind the endpoint.

The same applies for `kube_endpoint_address_available_total`

internal/store/endpoint.go
```diff
@@ -150,6 +151,90 @@ func endpointMetricFamilies(allowAnnotationsList, allowLabelsList []string) []ge
...
+			"kube_endpoint_address_available",
+			"Addresses available in endpoint",
+			metric.Gauge,
+			"",
+			wrapEndpointFunc(func(e *v1.Endpoints) *metric.Family {
+				var addresses string
+				var ports string
+				ms := []*metric.Metric{}
+				for _, s := range e.Subsets {
+					for _, address := range s.Addresses {
```
fpetkovski, line 164: To simplify the logic here, I would suggest building a new string slice with the `ip:port` for each endpoint and using `strings.join` at the end.

docs/endpoint-metrics.md
```diff
@@ -3,8 +3,10 @@
...
+| kube_endpoint_address_available_total | Counter | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; | EXPERIMENTAL |
+| kube_endpoint_address_not_ready | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; <br> `addresses`=&lt;endpoint-addresses&gt; <br> `ports`=&lt;endpoint-ports&gt; | STABLE |
+| kube_endpoint_address_available | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; <br> `addresses`=&lt;endpoint-addresses&gt; <br> `ports`=&lt;endpoint-ports&gt; | STABLE |
```
fpetkovski, line 9: These two metrics should have a new name and be marked as experimental.

fpetkovski (commented):
> If you check I have done the same that is merged IP addresses in a list separated by | to avoid cardinality issue. It is one timeseries per endpoint, it is not one timeseries per IP address. I have attached screenshot above. Please check.

The IPs behind the endpoint can change for many reasons. E.g. pods can get evicted which would cause replacements with new IPs to come up. Each time there is a new deployment, pods get replaced which means the IPs for an endpoint will also get replaced. In case of rolling updates, each new pod will generate a new series. 

If you still want to add the new metrics, I would consider adding them behind some sort of a flag, just like we do for pod labels. Otherwise, other users might experience an unexpected cardinality increase once they upgrade to a new KSM version.

lilic (commented):
Thanks for the PR!

We sadly cannot just break stable metrics, so we can just be additive here by introducing new labels or metrics.

internal/store/endpoint_test.go
```diff
@@ -167,9 +179,13 @@ func TestEndpointStoreWithLabels(t *testing.T) {
...
 			Want: metadata + `
-				kube_endpoint_address_available{endpoint="test-endpoint",namespace="default"} 6
-				kube_endpoint_address_not_ready{endpoint="test-endpoint",namespace="default"} 6
+				kube_endpoint_address_available_total{endpoint="test-endpoint",namespace="default"} 6
+				kube_endpoint_address_not_ready_total{endpoint="test-endpoint",namespace="default"} 6
 				kube_endpoint_annotations{endpoint="test-endpoint",annotation_app="foobar",namespace="default"} 1
+				kube_endpoint_address_available{endpoint="test-endpoint",namespace="default",addresses="127.0.0.1|10.0.0.1",ports="8080|8081"} 1 
+				kube_endpoint_address_available{endpoint="test-endpoint",namespace="default",addresses="172.22.23.202",ports="8443|9090"} 1
+				kube_endpoint_address_not_ready{endpoint="test-endpoint",namespace="default",addresses="192.168.1.1",ports="1234|5678"} 1
+				kube_endpoint_address_not_ready{endpoint="test-endpoint",namespace="default",addresses="192.168.1.3|192.168.2.2",ports="1234|5678"} 1
```
lilic, line 188: This makes it hard to do queries in promql if they are all part of one list, do you really need the address as well? Having endpoint name should be enough for alerting or dashboard purposes. Afterwards you can always get the endpoint to further debug, no?

PR #1559, opened 2021-08-27
Improve auto-sharding documentation

### LINKED ISSUE #1546: Documentation on Sharding Features / Status of experimental AutoSharding Feature
As @SuperQ made me aware of, the documentation around Sharding is lacking some information on which to use when.
https://github.com/kubernetes/kube-state-metrics/blob/master/README.md#horizontal-scaling-sharding
**Why is this needed**:
Improve User Experience
Document:
* Auto Sharding is marked as experimental since 2019 - is it still?
* Explain when to use AutoSharding vs. fixed argument driven sharding
* Make it clear what is the preferred / supported method.
 
**Additional context**
I believe autosharding is useful when your workloads change a lot in scale (so you can scale the statefulset instead of a redeploy), but comes with the complexity of a statefulset compared to a regular Kubernetes deployment.

fpetkovski:
This PR improves the documentation on auto-sharding. It explains the pros and cons of using auto-sharding vs manual sharding.

Does not affect it

Fixes #1546

### DIFF
--- a/README.md
+++ b/README.md
@@ -3,7 +3,7 @@
 [![Build Status](https://github.com/kubernetes/kube-state-metrics/workflows/continuous-integration/badge.svg)](https://github.com/kubernetes/kube-state-metrics/actions)
 [![Go Report Card](https://goreportcard.com/badge/github.com/kubernetes/kube-state-metrics)](https://goreportcard.com/report/github.com/kubernetes/kube-state-metrics) [![GoDoc](https://godoc.org/github.com/kubernetes/kube-state-metrics?status.svg)](https://godoc.org/github.com/kubernetes/kube-state-metrics)
 
-kube-state-metrics is a simple service that listens to the Kubernetes API
+kube-state-metrics (KSM) is a simple service that listens to the Kubernetes API
 server and generates metrics about the state of the objects. (See examples in
 the Metrics section below.) It is not focused on the health of the individual
 Kubernetes components, but rather on the health of the various objects inside,
@@ -43,7 +43,7 @@ are deleted they are no longer visible on the `/metrics` endpoint.
 - [kube-state-metrics vs. metrics-server](#kube-state-metrics-vs-metrics-server)
 - [Scaling kube-state-metrics](#scaling-kube-state-metrics)
   - [Resource recommendation](#resource-recommendation)
-  - [Horizontal scaling (sharding)](#horizontal-scaling-sharding)
+  - [Horizontal sharding](#horizontal-sharding)
     - [Automated sharding](#automated-sharding)
 - [Setup](#setup)
   - [Building the Docker container](#building-the-docker-container)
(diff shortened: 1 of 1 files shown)


mrueg:
@superq as you asked about more information, is this doc change helpful for you?

README.md
```diff
@@ -215,24 +215,26 @@ metrics-server it too is not responsible for exporting its metrics anywhere.
...
 
-Sharding should be used carefully, and additional monitoring should be set up in order to ensure that sharding is set up and functioning as expected (eg. instances for each shard out of the total shards are configured).
+Sharding should be used carefully and additional monitoring should be set up in order to ensure that sharding is set up and functioning as expected (eg. instances for each shard out of the total shards are configured).
 
-##### Automated sharding
+#### Automated sharding
 
-There is also an experimental feature, that allows kube-state-metrics to auto discover its nominal position if it is deployed in a StatefulSet, in order to automatically configure sharding. This is an experimental feature and may be broken or removed without notice.
+KSM supports a feature which allows each shard to discover its nominal position when deployed in a StatefulSet. which is useful for automatically configuring sharding. This is an experimental feature and may be broken or removed without notice.
```
mrueg, line 231: Do we want to introduce KSM as an abbreviation for the project or stick with "kube-state-metrics"?

README.md
```diff
@@ -215,24 +215,26 @@ metrics-server it too is not responsible for exporting its metrics anywhere.
...
+Sharding is done by taking an md5 sum of the Kubernetes Object's UID and performing a modulo operation on it with the total number of shards. Each shard decides whether the object is handled by the respective instance of kube-state-metrics or not. Note that this means all instances of kube-state-metrics, even if sharded, will have the network traffic and the resource consumption for unmarshaling objects for all objects, not just the ones they are responsible for. To optimize this further, the Kubernetes API would need to support sharded list/watch capabilities. In the optimal case, memory consumption for each shard will be 1/n compared to an unsharded setup. Typically, kube-state-metrics needs to be memory and latency optimized in order for it to return its metrics rather quickly to Prometheus.
```
mrueg, line 225: Should we mention `--use-apiserver-cache` as a way to reduce load on etcd here?

README.md
fpetkovski: Good point, I used it as second nature :smile:. I added the abbreviation next to the first reference of `kube-state-metrics` at the beginning of the readme. I think that it is used commonly used in the community so IMO it makes sense to have it in the docs. Wdyt?

README.md
fpetkovski: Good idea, added the suggestion

README.md
SuperQ: I you change the sentence structure a bit, you could write it like this.

> Automatic sharding allows each shard each shard to discover its nominal position

This avoids needing to be overly self-referential about the fact that this is a kube-state-metrics feature.

fpetkovski: Thanks for the suggestion, applied

SuperQ (approved):
LGTM

README.md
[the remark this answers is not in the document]
mrueg: @kubernetes/kube-state-metrics-maintainers should we start marking this as non-experimental with the next release to get more users test it? Or do we have plans to remove it at some point?

PR #1561, opened 2021-08-28
Add resource metrics for pod containers

### LINKED ISSUE #1488: Pod ephemeral storage resources wanted

kube_pod_container_resource_requests_ephemeral_storage_bytes
kube_pod_container_resource_limits_ephemeral_storage_bytes

**Why is this needed**:

I have metric kube_node_status_allocatable and i want to deduct sum of all running containers requests from this metric to get "how much free ephemeral storage each node has" to see if node soon comes to unschedulable state.

I don't know how this could be done, i just see there is 

kube_pod_init_container_resource_requests_ephemeral_storage_bytes
kube_pod_init_container_resource_limits_ephemeral_storage_bytes

and it is strange to me that there is such metric for init container, but no metric for main container in pod.

**Additional context**

None.

pkbhowmick:
Signed-off-by: Pulak Kanti Bhowmick <pkbhowmick007@gmail.com>

**What this PR does / why we need it**: Add individual resource metrics for pod containers.

Fixes #1488


pkbhowmick:
PR is ready for review.
@dgrisonnet @mrueg

dgrisonnet:
Hello @pkbhowmick, these metrics were removed in https://github.com/kubernetes/kube-state-metrics/pull/1278 in favor of this [Kubernetes Enhancement Proposal](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics.) which cover the topic better than what we can in kube-state-metrics. As such, I don't think we should add these metrics back, and rather we should rely on the more precise metrics from Kubernetes.

pkbhowmick:
Okay, then closing the PR. 
Thanks!

PR #1564, opened 2021-08-31
Add pod preparation period metrics

superbigsea:
add pod preparation period metrics
kube_pod_status_preparation_period{namespace="default",pod="nginx-5dbf784b68-ljz9k",uid="070793df-db03-48ae-bb72-dfcbc5059c3a"} 9

How does this change affect the cardinality of KSM: (increases, decreases or does not change cardinality)
none
Which issue(s) this PR fixes (optional, in fixes #(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -90,6 +90,7 @@ func podMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 		createPodStatusScheduledFamilyGenerator(),
 		createPodStatusScheduledTimeFamilyGenerator(),
 		createPodStatusUnschedulableFamilyGenerator(),
+                createPodPreparationPeriodFamilyGenerator(), 
 	}
 }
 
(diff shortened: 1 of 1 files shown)


fpetkovski:
Hi @superbigsea, thanks for the contribution. 

Instead of exporting a calculated metric, it might be better to export `kube_pod_status_ready_time` instead and do the rest of the calculation in PromQL. This way the `kube_pod_status_ready_time` metric will not be limited to this use case and can be used in different expressions as well.

In addition to this, the documentation for pod metrics needs to be updated as well: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

superbigsea:
ok thank you for your advice
my demand is to monitor if the pod has long preparation time
I try it to use  PromQL.
Thank you

superbigsea:
> Hi @superbigsea, thanks for the contribution.
> 
> Instead of exporting a calculated metric, it might be better to export `kube_pod_status_ready_time` instead and do the rest of the calculation in PromQL. This way the `kube_pod_status_ready_time` metric will not be limited to this use case and can be used in different expressions as well.
> 
> In addition to this, the documentation for pod metrics needs to be updated: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md
ok thank you for your advice
my demand is to monitor if the pod has long preparation time
I try it to use PromQL.
Thank you

PR #1575, opened 2021-09-14
Add NodeAffinity and Shutdown to podStatusReasons

### LINKED ISSUE #1573: Some pod status reasons missing in podStatusReasons var and kube_pod_status_reason metric
**What happened**:
We are trying to monitor pods in `Failed` phase with reasons `Shutdown` or `NodeAffinity`. However, the `kube_pod_status_reason` metric only exposes `NodeLost`, `Evicted` and `UnexpectedAdmissionError` reasons (see [pod-metrics.md](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md) and [pod.go](https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/pod.go#L37))

Here's an example of the status part of a pod we saw (used `kubectl get po <pod name> -o yaml`)
```
status:
  message: Pod Predicate NodeAffinity failed
  phase: Failed
  reason: NodeAffinity
  startTime: "2021-09-09T07:23:22Z"
```
And [here](https://stackoverflow.com/questions/68727863/pods-are-getting-stuck-in-shutdown-status) you can find an example of Failed pods with `Shutdown` reason

pod status reasons `Shutdown`, `NodeAffinity` and possibly others should be added to `kube_pod_status_reason` metric

N/A

See also https://github.com/kubernetes/kube-state-metrics/issues/1389#issuecomment-867945016
We checked the [v1.22 API docs](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.22/#podstatus-v1-core), but couldn't find a predefined list of possible values for `reason`.
`Shutdown` reason was introduced in 1.20 (https://kubernetes.io/docs/concepts/architecture/nodes/#graceful-node-shutdown)

**Environment**:
- kube-state-metrics version: `v2.2.0`
- Kubernetes version (use `kubectl version`): `v1.18.20` for `NodeAffinity` and `v1.20.8` for `Shutdown`
- Cloud provider or hardware configuration: GKE
- Other info:

idsvandermolen:
This pull request implements #1573 to support `NodeAffinity` and `Shutdown` as reasons in `kube_pod_status_reason` metric. This is especially useful when running workloads on preemptible nodes in GKE clusters.

This change increases cardinality of `kube_pod_status_reason` a little bit by adding two values for `reason` label.

Fixes #1573


fpetkovski:
Thanks for the contribution! 

Could you also update the documentation for the `kube_pod_status_reason`[1] by adding the two new values? https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md?

idsvandermolen:
> Thanks for the contribution!
> 
> Could you also update the documentation for the `kube_pod_status_reason`[1] by adding the two new values? https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md?

ah yes, of course... Forgot about that

mrueg:
/lgtm 

Thanks for your contribution!

PR #1581, opened 2021-09-21
Expose configmap labels and annotations

### LINKED ISSUE #1503: Expose annotations and labels on configmaps

Configmap metrics only export info, created, and resource_version ([docs](https://github.com/kubernetes/kube-state-metrics/blob/master/docs/configmap-metrics.md)).  It would be useful to see labels and annotations as well.

**Why is this needed**:

In my usecase, it's possible to use the labels/annotations applied to pods to track the rollout of versions of an application across different clusters (each with its own prometheus instance, aggregated in a single grafana dashboard -- the labels are applied by our deployment automation).  Some of the "applications" have static (or rarely changing) code, and changes in behavior are instead driven by config map updates.  It would be good to have a similar ability to track those rollouts via metrics, which requires exposing the labels and annotations to prometheus.

(In the above writeup I've referred to "labels and annotations".  But perhaps it makes sense to prioritize one or the other -- that would probably work as well.  Our tooling at the moment uses annotations but could probably be adjusted to use labels instead.)

Exposure of `kube_configmap_annotations` and `kube_configmap_labels` -- exactly parallel to `kube_namespace_annotations` `kube_pod_labels` etc.

**Additional context**

mindw:
Adds labels and annotations metrics to configmaps. Allowlist and tests are included.

Not by default, Two new stable metrics added - with the same cardinally impact as other such metrics,

Fixes #1503

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -238,119 +238,264 @@ func availableResources() []string {
 }
 
 func (b *Builder) buildConfigMapStores() []*metricsstore.MetricsStore {
-	return b.buildStoresFunc(configMapMetricFamilies, &v1.ConfigMap{}, createConfigMapListWatch, b.useAPIServerCache)
+	return b.buildStoresFunc(
+		configMapMetricFamilies(b.allowAnnotationsList["configmaps"], b.allowLabelsList["configmaps"]),
+		&v1.ConfigMap{},
+		createConfigMapListWatch,
+		b.useAPIServerCache,
+	)
 }
 
 func (b *Builder) buildCronJobStores() []*metricsstore.MetricsStore {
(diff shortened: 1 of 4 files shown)


internal/store/builder.go
```diff
@@ -238,119 +238,264 @@ func availableResources() []string {
 }
 
 func (b *Builder) buildConfigMapStores() []*metricsstore.MetricsStore {
-	return b.buildStoresFunc(configMapMetricFamilies, &v1.ConfigMap{}, createConfigMapListWatch, b.useAPIServerCache)
+	return b.buildStoresFunc(
```
dgrisonnet, line 241: nit: we should keep the same format as before

dgrisonnet (commented):
Just a small nit from my side otherwise looks good to me.

internal/store/builder.go
mindw: My apologies. I had trouble fitting the line into the (24" 1080p) monitor - I use bigger fonts due to poor eyesight, Thought it may be an acceptable change.
I'll revert,

mrueg:
/lgtm

Thanks for the contribution!

PR #1584, opened 2021-09-22
fix: avoid panic because of VPA objects without target ref

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

VPA is a custom resource. CRDs for this resource can be found [there](https://github.com/kubernetes/autoscaler/tree/vertical-pod-autoscaler/v0.9.2/vertical-pod-autoscaler/deploy). 

The problem is that the `targetRef` field was not required by the spec, so there is a possibility to deploy VPA to the cluster without it, which makes kube-state-metrics panic.
```
I0921 13:39:41.487672       1 main.go:191] Communication with server successful
I0921 13:39:41.487934       1 main.go:225] Starting metrics server: 127.0.0.1:8081
I0921 13:39:41.488119       1 main.go:200] Starting kube-state-metrics self metrics server: 127.0.0.1:8082
I0921 13:39:41.488125       1 metrics_handler.go:96] Autosharding disabled
I0921 13:39:41.489736       1 builder.go:144] Active collectors: certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,ingresses,jobs,limitranges,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,verticalpodautoscalers
E0921 13:39:41.547405       1 runtime.go:73] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 91 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic(0x12c5a60, 0x1d35e80)
	/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190612205821-1799e75a0719/pkg/util/runtime/runtime.go:69 +0x7b
k8s.io/apimachinery/pkg/util/runtime.HandleCrash(0x0, 0x0, 0x0)
	/go/pkg/mod/k8s.io/apimachinery@v0.0.0-20190612205821-1799e75a0719/pkg/util/runtime/runtime.go:51 +0x89
panic(0x12c5a60, 0x1d35e80)
	/usr/local/go/src/runtime/panic.go:969 +0x1b9
k8s.io/kube-state-metrics/internal/store.wrapVPAFunc.func1(0x1403040, 0xc000164738, 0x8)
	/src/kube-state-metrics/internal/store/verticalpodautoscaler.go:250 +0x1f0
k8s.io/kube-state-metrics/pkg/metric.(*FamilyGenerator).Generate(...)
	/src/kube-state-metrics/pkg/metric/generator.go:39
```
Without target ref, VPA objects receive no recommendation update. It is pointless trying to expose metrics for them.

It does not change cardinality.

Did not manage to find one.

### DIFF
--- a/internal/store/verticalpodautoscaler.go
+++ b/internal/store/verticalpodautoscaler.go
@@ -19,7 +19,7 @@ package store
 import (
 	"context"
 
-	k8sautoscaling "k8s.io/api/autoscaling/v1"
+	autoscalingv1 "k8s.io/api/autoscaling/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
@@ -284,7 +284,7 @@ func wrapVPAFunc(f func(*autoscaling.VerticalPodAutoscaler) *metric.Family) func
 		// * to alert about VPA objects without target refs
 		// * to count the right amount of VPA objects in a cluster
 		if targetRef == nil {
-			targetRef = &k8sautoscaling.CrossVersionObjectReference{}
+			targetRef = &autoscalingv1.CrossVersionObjectReference{}
 		}
 
 		for _, m := range metricFamily.Metrics {
--- a/internal/store/verticalpodautoscaler.go
+++ b/internal/store/verticalpodautoscaler.go
@@ -274,9 +274,15 @@ func vpaResourcesToMetrics(containerName string, resources v1.ResourceList) []*m
 func wrapVPAFunc(f func(*autoscaling.VerticalPodAutoscaler) *metric.Family) func(interface{}) *metric.Family {
 	return func(obj interface{}) *metric.Family {
 		vpa := obj.(*autoscaling.VerticalPodAutoscaler)
+		targetRef := vpa.Spec.TargetRef
+
+		// targetRef was not a mandatory field, which can lead to a nil pointer exception here.
+		// Since it is pointless to have a VPA object without target ref, skip exporting metrics.
+		if targetRef == nil {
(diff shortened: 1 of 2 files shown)


fpetkovski:
Awesome, thanks for catching this!

I just wonder if we want to still expose a metric with an empty targetRef apiversion, kind and name in these cases. The rule of thumb is to always expose raw data and use PromQL to filter it later on. A metric with empty values could be beneficial for alerting on VPAs without a target ref. Also, not emitting a metric could break clients that have queries which count the number of VPAs per cluster, namespace, team etc..

nabokihms:
Make sense. My first thoughts were about exporting resources for targets because it is our main use-case. I will do as you've said.

fpetkovski:
Great stuff.

/lgtm

internal/store/verticalpodautoscaler.go
```diff
@@ -19,6 +19,7 @@ package store
 import (
 	"context"
 
+	k8sautoscaling "k8s.io/api/autoscaling/v1"
```
mrueg, line 22:
```suggestion
	autoscalingv1 "k8s.io/api/autoscaling/v1"
```

Can we refactor this to use a similar approach like metav1 below?

nabokihms: I copied it from the ksm code somewhere, but yes, no problem. I will make the change.

ahmed-mez:
@mrueg  @fpetkovski is this fix going to be backported to 2.2 and 2.1? Thanks!

PR #1592, opened 2021-09-29
Add app.kubernetes.io/component label

slashpai:

Adding the [recommended label](https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/#labels) `app.kubernetes.io/component`

NA

NA

### DIFF
--- a/examples/autosharding/statefulset.yaml
+++ b/examples/autosharding/statefulset.yaml
@@ -2,6 +2,7 @@ apiVersion: apps/v1
 kind: StatefulSet
 metadata:
   labels:
+    app.kubernetes.io/component: exporter
     app.kubernetes.io/name: kube-state-metrics
     app.kubernetes.io/version: 2.2.1
   name: kube-state-metrics
@@ -15,6 +16,7 @@ spec:
   template:
     metadata:
       labels:
+        app.kubernetes.io/component: exporter
         app.kubernetes.io/name: kube-state-metrics
         app.kubernetes.io/version: 2.2.1
     spec:
--- a/examples/autosharding/cluster-role-binding.yaml
+++ b/examples/autosharding/cluster-role-binding.yaml
@@ -2,6 +2,7 @@ apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRoleBinding
 metadata:
   labels:
+    app.kubernetes.io/component: exporter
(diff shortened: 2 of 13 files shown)


dgrisonnet:
Nice!

/lgtm

examples/autosharding/statefulset.yaml
```diff
@@ -10,11 +11,13 @@ spec:
   replicas: 2
   selector:
     matchLabels:
+      app.kubernetes.io/component: exporter
```
fpetkovski, line 14: I believe that `matchLabels` in a Deployment object is immutable. Could you please verify whether we can seamlessly upgrade from the latest KSM release to this version?

fpetkovski:
Let's also verify the upgrade path.

/hold

examples/autosharding/statefulset.yaml
slashpai: You are right. `The StatefulSet "kube-state-metrics" is invalid: spec: Forbidden: updates to statefulset spec for fields other than 'replicas', 'template', 'updateStrategy' and 'minReadySeconds' are forbidden`

For such scenarios we delete and recreate object when updates are impossible

Would that mean we cannot new labels in future to avoid this?

fpetkovski: Only `matchLabels` is immutable but we can add new labels to the underlying pods. So we can just remove the label from here but keep it everywhere else. As long as `matchLabels` are a subset of the pod labels, everything should work as expected.

slashpai: Looks like I need to update https://github.com/kubernetes/kube-state-metrics/blob/master/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet#L292 so matchLabels will only take existing labels

```jsonnet
        selector: { matchLabels: ksm.podLabels },
```

to

```jsonnet
        selector: { matchLabels: ksm. commonLabels },
```

fpetkovski: Seems like it. We need to use different label sets for pod labels and selector

slashpai:
@fpetkovski @dgrisonnet I updated with changes please take a look at this again

dgrisonnet (approved):
/lgtm

Issue #1593, opened 2021-09-30
Add namespace denylist option


jacekn:
**What would you like to be added**:

I think it would be useful to have  ability to exclude certain namespaces. Currently it's only possible to get metrics for all namespaces or for explicitly defined namespace list using the `--namespaces` flag.

**Why is this needed**:

There are two use cases that I can think of.

First one is where someone wants to have different scrape interval for a selected list of namespaces. To achieve this we would need to run 2 copies of the kube-state metrics and scrape them at different intervals. One copy would be set up to expose metrics for a fixed list of namespaces and the other one would have to monitor all remaining namespaces except a fixed list. Explicitly listing all namespaces minus a few is not practical because every time we add a namespace we'd also need to update kube-state-metrics config.

Second use case is a situation where there is a namespace that causes potential problems. For example one used for experimentation with very high pod churn. It could be desirable to exclude it from kube-state-metrics completely to avoid scraping large amounts of irrelevant metrics and to save resources.

**Describe the solution you'd like**

I think a new `--namespaces-denylist` flag would be good. It would also be consistent with `--metric-denylist` flag.

**Additional context**

jacekn:
@fpetkovski apologies for pinging your directly but I can see that you normally cut new releases of the kube-state-metrics.
This feature was merged 29 days ago - is there a process or schedule for it to be released? I checked latest version and the feature is still not there.

fpetkovski:
Hi @jacekn, no worries. We are actually planning to cut a new release in the following days. We don't have a regular schedule for now and we make releases either on demand or when we get a significant amount of contributions. Nevertheless, thank you for that feedback, we'll see if it makes sense to have a more regular release schedule that people can rely on.

fpetkovski:
@jacekn since Kubernetes 1.23 is just around the corner, it would be best to wait until it is released and cut KSM 2.3. I hope that's not too long of a wait for you.

PR #1595, opened 2021-09-30
Add namespace denied list

mallow111:
Details: https://github.com/kubernetes/kube-state-metrics/issues/1593

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1593


mallow111:
@dgrisonnet  any idea how to fix the ci-validate-docs check?

fpetkovski:
Thanks for the contribution @mallow111! To fix the docs, you can run `make generate`.

Besides that, the PR seems to need a bit more work. Currently, the `namespace-denylist` option only excludes namespaces from the `namespaces` list. If the `namespaces` list is empty, KSM will export metrics from all namespaces, regardless of what is in the namespace denylist. 

One way to get namespace exclusion to work when watching all namespaces is to propagate the namespace denylist all the way down to the listwatchers. Here is an example of how the `createConfigMapListWatch` could end up looking like:
```
func createConfigMapListWatch(kubeClient clientset.Interface, ns string, nsDenylist []string) cache.ListerWatcher {
	fieldSelector := excludeNamespacesFieldSelector(nsDenylist)
	return &cache.ListWatch{
			opts.FieldSelector = fieldSelector
			return kubeClient.CoreV1().ConfigMaps(ns).List(context.TODO(), opts)
		},
			opts.FieldSelector = fieldSelector
			return kubeClient.CoreV1().ConfigMaps(ns).Watch(context.TODO(), opts)
		},
	}
}

func excludeNamespacesFieldSelector(nsDenylist []string) string {
	if len(nsDenylist) == 0 {
		return ""
	}

	namespaceExcludeSelectors := make([]fields.Selector, len(nsDenylist))
	for i, ns := range nsDenylist {
		selector := fields.OneTermNotEqualSelector("metadata.namespace", ns)
		namespaceExcludeSelectors[i] = selector
	}
	return fields.AndSelectors(namespaceExcludeSelectors...).String()
}
```

mallow111:
@fpetkovski  Please take a look at the PR for the listwatcher change, once it is done, I can go back to this pr and address the rest, trying to break big PR into smaller ones.
https://github.com/kubernetes/kube-state-metrics/pull/1596

mallow111:
It is merged, thank you!

PR #1603, opened 2021-10-09
feat: kustomization for all examples

karancode:
- adds `kustomization.yaml` for all the examples and updates readme.

(with current kustomization.yaml, cannot use the autosharding example with remote resource reference).


karancode:
@fpetkovski we kustomization file present for each type of deployment (standard, autosharding, etc.) it is easier to refer to release specific configs directly from kube-state-metrics repo.  Otherwise everyone have to manage these configs at their end by copy pasting. 
If we are using kustomize, it is very easy to manage these configs and we can easily patch with different overlays. Something like :
```yaml
---
apiVersion: kustomize.config.k8s.io/v1beta1
kind: Kustomization

namespace: kube-system

resources:
- https://github.com/kubernetes/kube-state-metrics.git/examples/standard?ref=v2.2.1
replicas: 3
```

Currently there is kustomization.yaml present at the base of repo which I think is added for enabling this use case. But as we have different type of deployment as well(autosharding) now, we cannot use above method for deploying that.
So I thought it will be better to enable it for each example deployment.

Issue #1605, opened 2021-10-11
Breaking changes in the `BuilderInterface` released in a bug fix version (v2.1.1)


ahmed-mez:
**What happened**:

https://github.com/kubernetes/kube-state-metrics/pull/1499 introduces major breaking changes in the `BuilderInterface` interface.

- `BuildStoresFunc` now returns `[]*metricsstore.MetricsStore` (struct) instead of a `cache.Store` (interface). This made injecting custom stores impossible when kube state metrics is used as a library.
- `Build()` not exposing the store anymore. The returned interface `[]metricsstore.MetricsWriter` is very specific to the regular use-case and isn't generic enough. The builder should at least have a method to expose the store as is.

**What you expected to happen**:

- Public APIs and interfaces should stay backward compatible especially in bug fix releases.
- Stick to using interfaces: `metricsstore.MetricsStore` implements `cache.Store` already. Any reason why `[]*metricsstore.MetricsStore` can't be replaced by `[]cache.Store` ?

PR #1610, opened 2021-10-11
Fix BuilderInterface and BuildStoresFunc to allow using KSM as a library

ahmed-mez:

This PR makes ksm usable again as a library after the breaking changes introduced by https://github.com/kubernetes/kube-state-metrics/pull/1499

- It makes `BuildStoresFunc` returns a generic `[]cache.Store` to allow using custom stores
- It introduces a new method `builder.BuildStores()` which returns the stores as an alternative to `builder.Build()`

Does not change cardinality

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1605

### DIFF
--- a/pkg/builder/builder_test.go
+++ b/pkg/builder/builder_test.go
@@ -1,5 +1,5 @@
 /*
-Copyright 2019 The Kubernetes Authors All rights reserved.
+Copyright 2021 The Kubernetes Authors All rights reserved.
 
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


ahmed-mez:
Thanks @fpetkovski for approving! Happy to open another PR to backport the fix to v2.1 once this gets merged.

ahmed-mez:
@mrueg thanks for the review!

> Are there any test cases etc. we could include to ensure that we don't break compatibility for external consumers?

I just added a test case in a separate [commit](https://github.com/kubernetes/kube-state-metrics/pull/1610/commits/da9f39e6cb99ebcc31d90b437578611d649fb293).

dgrisonnet:
Unholding since we have 3 approvals.

/unhold

PR #1611, opened 2021-10-11
Merge updates from 2.2.3

fpetkovski:
This commit merges user-facing documentation for the 2.2.3 release.


mrueg:
Do we want to include the full changeset? (as in `git merge release-2.2 master`)?

dgrisonnet:
Yes, I think we should also include the changes from 2.2.2 since the changelog is now diverging from what users can get in https://github.com/kubernetes/kube-state-metrics/releases.

I don't think there will be any conflicts since the changes are only manifests and documentation related.

PR #1616, opened 2021-10-18
[e2e] check all default collector metrics are available

olivierlemasle:
This PR supersedes #950, and fixes part of #875.

It is simply a **rebase** of #950 on `master`, as discussed in https://github.com/kubernetes/kube-state-metrics/pull/950#issuecomment-703579091.


dgrisonnet:
Awesome, thanks @olivierlemasle.

/lgtm

PR #1625, opened 2021-11-08
build as non root and keep working dir clean

bavarianbidi:

* build the kube-state-metrics binary into the new output directory
* build the kube-state-metrics binary as current logged in user
* build the kube-state-metrics binary via Dockerfile (docker build .) as
65543 (nobody)
* download and extract the promtool binary into the output directory
* make clean remove all files under output/*

Fixes #

Signed-off-by: Constanti, Mario <mario.constanti@daimler.com>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,17 +1,19 @@
 ARG GOVERSION=1.17
 ARG GOARCH
 FROM golang:${GOVERSION} as builder
+USER 65534
 ARG GOARCH
 ENV GOARCH=${GOARCH}
+ENV GOCACHE=/tmp/.cache
 WORKDIR /go/src/k8s.io/kube-state-metrics/
-COPY . /go/src/k8s.io/kube-state-metrics/
+COPY --chown=65534:65534 . /go/src/k8s.io/kube-state-metrics/
 
 RUN make build-local
 
 FROM gcr.io/distroless/static:latest-${GOARCH}
-COPY --from=builder /go/src/k8s.io/kube-state-metrics/kube-state-metrics /
+USER 65534
 
-USER nobody
+COPY --from=builder --chown=65534:65534 /go/src/k8s.io/kube-state-metrics/output/kube-state-metrics /
 
 ENTRYPOINT ["/kube-state-metrics", "--port=8080", "--telemetry-port=8081"]
 
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,18 +2,18 @@ ARG GOVERSION=1.17
 ARG GOARCH
 FROM golang:${GOVERSION} as builder
 USER 65534
-ARG GOARCH
 ENV GOARCH=${GOARCH}
 ENV GOCACHE=/tmp/.cache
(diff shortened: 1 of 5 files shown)


bavarianbidi:
Build will fail as `.github/workflows/ci.yaml` need a change for https://github.com/Daimler/kube-state-metrics/blob/ad888d10764cf423fa26bfe74b723481ab287dea/.github/workflows/ci.yml#L147

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
+COPY --chown=65534:65534 . /go/src/k8s.io/kube-state-metrics/
 
 RUN make build-local
 
 FROM gcr.io/distroless/static:latest
-COPY --from=builder /go/src/k8s.io/kube-state-metrics/kube-state-metrics /
+USER 65534
 
-USER nobody
+COPY --from=builder --chown=65543:65534 /go/src/k8s.io/kube-state-metrics/output/kube-state-metrics /
```
mrueg, line 15: why does the UID differ from the GID? Shouldn't it be the same?

bavarianbidi: Yes, you're right. Was a typo on my side. Fixed it few seconds ago - sry ;-)

Dockerfile
```diff
@@ -1,17 +1,19 @@
 ARG GOVERSION=1.17
 ARG GOARCH
 FROM golang:${GOVERSION} as builder
+USER 65534
```
mrueg, line 4: This is the first time I see this pattern during build time, do you have more information or other examples that do the same? I always assumed you would be able to control the buildtime environment from outside and drop privileges there as well. Similar to unprivileged builds in kaniko.

Dockerfile
```diff
@@ -1,17 +1,19 @@
...
+COPY --chown=65534:65534 . /go/src/k8s.io/kube-state-metrics/
 
 RUN make build-local
 
 FROM gcr.io/distroless/static:latest-${GOARCH}
-COPY --from=builder /go/src/k8s.io/kube-state-metrics/kube-state-metrics /
+USER 65534
 
-USER nobody
+COPY --from=builder --chown=65534:65534 /go/src/k8s.io/kube-state-metrics/output/kube-state-metrics /
```
mrueg, line 16: Why does the binary need to be owned by the user executing it?

bavarianbidi: there is not really a need for it, that's right.
will revert that change.

Dockerfile
bavarianbidi: i tried to setup the build-stage in the dockerfile the same way as we currently do via the `make build` target.

Within this PR i add the `-u $(USERID):$(USERGROUP)` parameters and i thought it might be a good idea to keep the build-stage configuration "in sync"

bavarianbidi:
As i will leave Daimler within the next couple of weeks i will close this PR as i will lose access to the mercedes-benz github organisation. If we still think, this is something we should have, i will re-open this PR in a few weeks from another forked repo.

WDYT @mrueg ?

PR #1643, opened 2021-11-28
Add the "--opt-in-metrics" argument to the CLI to allow for opt-in metrics

### LINKED ISSUE #1574: Implement a mechanism for opt-in metrics
A way to enable potentially risky metrics through an opt-in command line argument.

**Why is this needed**:
New metrics are regularly added to kube-state-metrics and KSM versions automatically expose these new metrics on the `/metrics` endpoint. Because of this, we have to be very careful when deciding which metrics KSM should include in order to not break users when they upgrade to a new version.

One possible solution would be to add an additional command line argument through which additional metrics can be enabled. Ideally, users should be able to opt-into each of those metrics individually by specifying them as values to the arg. Examples of metrics which would require an explicit opt-in are those that can potentially have high cardinality. This will allow the community to contribute metrics which enable deeper visibility into their cluster and users can enable those metrics explicitly for themselves. Users which run KSM with the default settings will not be affected by such changes.

**Additional context**
Issues that were closed in the past due to a lack of such mechanism:
* https://github.com/kubernetes/kube-state-metrics/issues/348
* https://github.com/kubernetes/kube-state-metrics/issues/167
* https://github.com/kubernetes/kube-state-metrics/issues/1394

Serializator:
It introduces a way of defining metrics as "opt-in", meaning the end-user needs to specifically specify that metrics which are marked as "opt-in" are to be exposed by the store.

`--opt-in-metrics` is used to specify which opt-in metrics to expose from the store. It follows the same notation as the `--metric-allowlist` and `--metric-denylist` arguments. 

It does not directly change or affect the cardinality of existing metrics.

Fixes #1574


Serializator:
I noticed the CI checks and will resolve them this evening!

pkg/options/options.go
```diff
@@ -97,6 +99,7 @@ func (o *Options) AddFlags() {
 	o.flags.Var(&o.NamespacesDenylist, "namespaces-denylist", "Comma-separated list of namespaces not to be enabled. If namespaces and namespaces-denylist are both set, only namespaces that are excluded in namespaces-denylist will be used.")
 	o.flags.Var(&o.MetricAllowlist, "metric-allowlist", "Comma-separated list of metrics to be exposed. This list comprises of exact metric names and/or regex patterns. The allowlist and denylist are mutually exclusive.")
 	o.flags.Var(&o.MetricDenylist, "metric-denylist", "Comma-separated list of metrics not to be enabled. This list comprises of exact metric names and/or regex patterns. The allowlist and denylist are mutually exclusive.")
+	o.flags.Var(&o.OptInMetrics, "opt-in-metrics", "Comma-separated list of metrics which are opt-in and not enabled by default. This is in addition to the metric allow- and denylists")
```
mrueg, line 102: Should it be called metric-opt-in-list or similar then?

mrueg: Maybe we could also just call it --metric-enable-experimental (to enable metrics that are experimental) and have users filter it on the denylist then?

Serializator: "experimental" does not mean / imply the same as "opt-in" does, as metrics which are opt-in are metrics that can be stable but might not be applicable for everyone or only to a small subset of use-cases and thus shouldn't be enabled by default.

I thought about the "--metric-opt-in-list" as well" but I feel like "--opt-in-metrics" reads and pronounces a lot more natural than "--metric-opt-in-list" does, as "allowlist" is more of a verb and "opt-in" more of a noun which indicates the action of "opting-in for a metric", though I understand what you mean (to keep some kind of similarity in the naming of CLI options).

pkg/optin/optin.go
```diff
@@ -0,0 +1,72 @@
...
+)
+
+// MetricFamilyFilter filters metric families which are defined as opt-in by their generator.FamilyGenerator
+type MetricFamilyFilter struct {
+	metrics []*regexp.Regexp
+}
+
+// Test tests if a given generator is an opt-in metric family and was passed as an opt-in metric family at startup
+func (filter MetricFamilyFilter) Test(generator generator.FamilyGenerator) bool {
+	if generator.OptIn {
```
fpetkovski, line 34: Can we reduce the nesting with
```
  if !generator.OptIn {
    return true
  }
```

pkg/optin/optin.go
```diff
@@ -0,0 +1,72 @@
...
+	return true
+}
+
+// Status returns the metrics contained within the filter as a comma-separated string
+func (filter MetricFamilyFilter) Status() string {
+	asStrings := make([]string, 0)
+	for _, metric := range filter.metrics {
+		asStrings = append(asStrings, metric.String())
+	}
+	// sort the strings for the sake of testing so that the resulting status is consistent
```
fpetkovski, line 51: Should the sort be done in the test?

Serializator: In the test, `(optin.MetricFamilyFilter).metrics` is passed in the order in which I expect the resulting status to be (and check on), but the resulting status does not always (sporadically) result in the same order as I passed in the metrics. Thus I added the sort in the method itself for the sake of testing but also to keep a consistent result in the logs.

If you really feel like this is not in place here I could try the following.
```go
asStrings := make([]string, len(filter.metrics))
for i, metric := range filter.metrics {
    asStrings[i] = metric.String()
}
return strings.Join(asStrings, ", ")
```

One downside of this (in my opinion) is that rather than being explicit we use a really specific way of creating the `asStrings` slice without explaining why or making it explicit using e.g. `sort.Strings`, resulting in the possibility that this is changed in the future and resulting in a test "randomly" failing.

pkg/optin/optin.go
Serializator: Will apply this change, agree that the nesting is a bit too much but didn't think about an early-return like you are suggesting. Thanks! 🎉

pkg/optin/optin.go
fpetkovski: I see, if it was only the test that depended on the sorted order, then the test should be doing the sorting. But in this case it makes sense to keep the sort here for consistency in the logs. I would just adjust the comment to reflect that we are sorting to provide a consistent ux instead of making a test pass.

Serializator: I will change the comment to reflect this. Thanks for the feedback! 👍🏼

pkg/options/options.go
Serializator: @fpetkovski, what do you think about this? I wouldn't say to rename the option to `--metric-enable-experimental` (because of the aforementioned reason), but I am in between about @mrueg's suggestion for `--metric-opt-in-list` (or something similar) versus `--opt-in-metrics`.

fpetkovski: Using the same naming convention would also indicate that the flags are related and control the same functionality.

Serializator: Will change it to `--opt-in-metrics` then. I'm here to contribute value to the project, not to nitpick on a name because I like it, hahah. In favor of consistency as well, as long as it does not affect the clarity of its purpose.

Thanks @mrueg and @fpetkovski for the thoroughness! 😉

fpetkovski:
Nice work! 
/lgtm

/hold for @mrueg to take another look

pkg/options/options.go
```diff
@@ -62,6 +63,7 @@ func NewOptions() *Options {
 		Resources:            ResourceSet{},
 		MetricAllowlist:      MetricSet{},
 		MetricDenylist:       MetricSet{},
+		OptInMetrics:         MetricSet{},
```
mrueg, line 66: rename to
```suggestion
		MetricOptInList:         MetricSet{},
```
for consistency

Serializator: Renamed it 👍🏼

mrueg:
/lgtm
/hold cancel 
Thanks for your contribution!

PR #1644, opened 2021-11-29
Extend kube-state-metrics to support Custom Resource metrics

### LINKED ISSUE #1640: Extend kube-state-metrics to support Custom Resource metrics
Since the CRD has been widely used nowadays, is there any chance for `kube-state-metrics` to support Custom Resource metrics extension, a little bit like `kube-scheduler`. The repo [scheduler-plugin](https://github.com/kubernetes-sigs/scheduler-plugins) is based on the out-of-tree scheduler plugins framework. 

I suppose we could make small changes to make users define their own out-of-tree metrics stores so they could easily depend this project to add the Custom Resource metrics.

**Why is this needed**:
Now users should copy this repo and modify the source codes to support their own Custom Resource metrics because the `availableStores` is hard-code [here](https://github.com/kubernetes/kube-state-metrics/blob/0c9580d09c6ac8b35630813afbcc0fe217ce1874/internal/store/builder.go#L222-L253). I hope this could be more flexible so that users just only have to depend this repo in their own project, add their plugin registry and finally build the project to get a new customized `kube-state-metrics`.

A simple solution is add a new func in `builder.go` and handle some details in `main.go`. Here is the new function.
```go
func RegisterOutOfTreeStores(
	name string,
	expectedType interface{},
	listWatchFunc func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher,
) {
	availableStores[name] = func(b *Builder) []cache.Store {
		return b.buildStoresFunc(metricFamilies, expectedType, listWatchFunc, b.useAPIServerCache)
	}
}
```
The out-of-tree metrics store could be added into the `availableStores` to prevent hard-code. 

**Additional context**

Garrybest:
Now kube-state-metrics does not support Kubernetes Custom Resource metrics. According to #1640, I try to make kube-state-metrics to be more flexible to support Custom Resource. Users only need to:

1. Depend this repo in their own project.
2. Implement the new interface `customresource.RegistryFactory` of their Custom Resources as a plugin factory to tell `kube-state-metrics` how to display custom resource metrics.
3. Register the factories as a parameter by calling `RunKubeStateMetrics`.

The inspiration is from `kube-scheduler` out-of-tree plugin registry. The well-know project [scheduler-plugin](https://github.com/kubernetes-sigs/scheduler-plugins) is based on this cool function. It adds new scheduler plugins without modifying any source codes in `kube-scheduler`.

I have taken Custom Resource `samplev1alpha1.Foo` in [samplecontroller](https://github.com/kubernetes/sample-controller) as an example in the interface comments and unit test.

It does not directly change or affect the cardinality of existing metrics.

Fixes #1640

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -43,6 +43,7 @@ import (
 	"k8s.io/klog/v2"
 
 	ksmtypes "k8s.io/kube-state-metrics/v2/pkg/builder/types"
+	"k8s.io/kube-state-metrics/v2/pkg/customresource"
 	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
 	metricsstore "k8s.io/kube-state-metrics/v2/pkg/metrics_store"
 	"k8s.io/kube-state-metrics/v2/pkg/options"
@@ -57,21 +58,23 @@ var _ ksmtypes.BuilderInterface = &Builder{}
 // Builder helps to build store. It follows the builder pattern
 // (https://en.wikipedia.org/wiki/Builder_pattern).
 type Builder struct {
-	kubeClient            clientset.Interface
-	vpaClient             vpaclientset.Interface
-	namespaces            options.NamespaceList
-	namespaceFilter       string
-	ctx                   context.Context
-	enabledResources      []string
-	familyGeneratorFilter generator.FamilyGeneratorFilter
-	listWatchMetrics      *watch.ListWatchMetrics
-	shardingMetrics       *sharding.Metrics
-	shard                 int32
-	totalShards           int
-	buildStoresFunc       ksmtypes.BuildStoresFunc
-	allowAnnotationsList  map[string][]string
-	allowLabelsList       map[string][]string
-	useAPIServerCache     bool
+	kubeClient                    clientset.Interface
+	customResourceClients         map[string]interface{}
+	vpaClient                     vpaclientset.Interface
+	namespaces                    options.NamespaceList
+	namespaceFilter               string
+	ctx                           context.Context
+	enabledResources              []string
+	familyGeneratorFilter         generator.FamilyGeneratorFilter
+	listWatchMetrics              *watch.ListWatchMetrics
+	shardingMetrics               *sharding.Metrics
+	shard                         int32
+	totalShards                   int
+	buildStoresFunc               ksmtypes.BuildStoresFunc
+	buildCustomResourceStoresFunc ksmtypes.BuildCustomResourceStoresFunc
+	allowAnnotationsList          map[string][]string
+	allowLabelsList               map[string][]string
+	useAPIServerCache             bool
 }
 
 // NewBuilder returns a new builder.
@@ -134,6 +137,16 @@ func (b *Builder) WithVPAClient(c vpaclientset.Interface) {
 	b.vpaClient = c
 }
 
+// WithCustomResourceClients sets the customResourceClients property of a Builder.
+func (b *Builder) WithCustomResourceClients(cs map[string]interface{}) {
+	b.customResourceClients = cs
+}
+
+// WithUsingAPIServerCache configures whether using APIServer cache or not.
+func (b *Builder) WithUsingAPIServerCache(u bool) {
+	b.useAPIServerCache = u
+}
+
 // WithFamilyGeneratorFilter configures the family generator filter which decides which
 // metrics are to be exposed by the store build by the Builder.
 func (b *Builder) WithFamilyGeneratorFilter(l generator.FamilyGeneratorFilter) {
@@ -141,14 +154,42 @@ func (b *Builder) WithFamilyGeneratorFilter(l generator.FamilyGeneratorFilter) {
 }
 
 // WithGenerateStoresFunc configures a custom generate store function
-func (b *Builder) WithGenerateStoresFunc(f ksmtypes.BuildStoresFunc, u bool) {
+func (b *Builder) WithGenerateStoresFunc(f ksmtypes.BuildStoresFunc) {
 	b.buildStoresFunc = f
-	b.useAPIServerCache = u
+}
+
+// WithGenerateCustomResourceStoresFunc configures a custom generate custom resource store function
+func (b *Builder) WithGenerateCustomResourceStoresFunc(f ksmtypes.BuildCustomResourceStoresFunc) {
+	b.buildCustomResourceStoresFunc = f
 }
 
 // DefaultGenerateStoresFunc returns default buildStores function
 func (b *Builder) DefaultGenerateStoresFunc() ksmtypes.BuildStoresFunc {
 	return b.buildStores
+}
+
+// DefaultGenerateCustomResourceStoresFunc returns default buildCustomResourceStores function
+func (b *Builder) DefaultGenerateCustomResourceStoresFunc() ksmtypes.BuildCustomResourceStoresFunc {
+	return b.buildCustomResourceStores
+}
+
+// WithCustomResourceStoreFactories returns configures a custom resource stores factory
+func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.RegistryFactory) {
+	for i := range fs {
+		f := fs[i]
+		if _, ok := availableStores[f.Name()]; ok {
+			klog.Warningf("The internal resource store named %s already exists and is overridden by a custom resource store with the same name, please make sure it meets your expectation", f.Name())
+		}
+		availableStores[f.Name()] = func(b *Builder) []cache.Store {
+			return b.buildCustomResourceStoresFunc(
+				f.Name(),
+				f.MetricFamilyGenerators(b.allowAnnotationsList[f.Name()], b.allowLabelsList[f.Name()]),
+				f.ExpectedType(),
+				f.ListWatch,
+				b.useAPIServerCache,
+			)
+		}
+	}
 }
 
 // WithAllowAnnotations configures which annotations can be returned for metrics
@@ -407,6 +448,47 @@ func (b *Builder) buildStores(
 			composedMetricGenFuncs,
 		)
 		listWatcher := listWatchFunc(b.kubeClient, ns, b.namespaceFilter)
+		b.startReflector(expectedType, store, listWatcher, useAPIServerCache)
+		stores = append(stores, store)
+	}
+
+	return stores
+}
+
+// TODO(Garrybest): Merge `buildStores` and `buildCustomResourceStores`
+func (b *Builder) buildCustomResourceStores(resourceName string,
+	metricFamilies []generator.FamilyGenerator,
+	expectedType interface{},
+	listWatchFunc func(customResourceClient interface{}, ns string, fieldSelector string) cache.ListerWatcher,
+	useAPIServerCache bool,
+) []cache.Store {
+	metricFamilies = generator.FilterFamilyGenerators(b.familyGeneratorFilter, metricFamilies)
+	composedMetricGenFuncs := generator.ComposeMetricGenFuncs(metricFamilies)
+	familyHeaders := generator.ExtractMetricFamilyHeaders(metricFamilies)
+
+	customResourceClient, ok := b.customResourceClients[resourceName]
+	if !ok {
+		klog.Warningf("Custom resource client %s does not exist", resourceName)
+		return []cache.Store{}
+	}
+
+	if b.namespaces.IsAllNamespaces() {
+		store := metricsstore.NewMetricsStore(
+			familyHeaders,
+			composedMetricGenFuncs,
+		)
+		listWatcher := listWatchFunc(customResourceClient, v1.NamespaceAll, b.namespaceFilter)
+		b.startReflector(expectedType, store, listWatcher, useAPIServerCache)
+		return []cache.Store{store}
+	}
+
+	stores := make([]cache.Store, 0, len(b.namespaces))
+	for _, ns := range b.namespaces {
+		store := metricsstore.NewMetricsStore(
+			familyHeaders,
+			composedMetricGenFuncs,
+		)
+		listWatcher := listWatchFunc(customResourceClient, ns, b.namespaceFilter)
 		b.startReflector(expectedType, store, listWatcher, useAPIServerCache)
 		stores = append(stores, store)
 	}
--- a/main.go
+++ b/main.go
@@ -19,65 +19,21 @@ package main
 import (
 	"context"
 	"fmt"
-	"net"
-	"net/http"
-	"net/http/pprof"
 	"os"
-	"strconv"
-	"time"
 
-	"github.com/oklog/run"
-	"github.com/pkg/errors"
-	"github.com/prometheus/client_golang/prometheus"
-	"github.com/prometheus/client_golang/prometheus/collectors"
-	"github.com/prometheus/client_golang/prometheus/promauto"
-	"github.com/prometheus/client_golang/prometheus/promhttp"
 	"github.com/prometheus/common/version"
-	"github.com/prometheus/exporter-toolkit/web"
-	vpaclientset "k8s.io/autoscaler/vertical-pod-autoscaler/pkg/client/clientset/versioned"
-	clientset "k8s.io/client-go/kubernetes"
-	_ "k8s.io/client-go/plugin/pkg/client/auth"
-	"k8s.io/client-go/tools/clientcmd"
 	"k8s.io/klog/v2"
 
-	"k8s.io/kube-state-metrics/v2/internal/store"
-	"k8s.io/kube-state-metrics/v2/pkg/allowdenylist"
-	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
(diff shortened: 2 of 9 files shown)


Garrybest:
The PR has been divided to 4 parts.

1. Design a Custom Resource RegistryFactory for out-of-tree implements. Anyone who wants to extend kube-state-metrics should implement this interface to tell us how to get and display their metrics.
2. Reform Builder to adapt custom resource registry. Now the Builder could register custom resource metric stores.
3. Reform main function to adapt custom resource registry. Users could not depend `main` pkg, so I move the codes into a new pkg `app` and call the custom resource register functions for Builder.
4. Add a custom resource unit test based on foo object in samplecontroller. [samplev1alpha1.Foo](https://github.com/kubernetes/sample-controller/blob/master/pkg/apis/samplecontroller/v1alpha1/types.go) could be a nice example so I add `k8s.io/sample-controller` in `go.mod` and make a best practice in `server_test.go`.

internal/store/builder.go
```diff
@@ -146,11 +154,37 @@ func (b *Builder) WithGenerateStoresFunc(f ksmtypes.BuildStoresFunc, u bool) {
 	b.useAPIServerCache = u
 }
 
+// WithGenerateCustomResourceStoresFunc configures a custom generate custom resource store function
+func (b *Builder) WithGenerateCustomResourceStoresFunc(f ksmtypes.BuildCustomResourceStoresFunc, u bool) {
+	b.buildCustomResourceStoresFunc = f
+	b.useAPIServerCache = u
```
Serializator, line 160: `b.useAPIServerCache` is used by `(store.*Builder).WithGenerateStoresFunc` as well, what would happen in the following scenario?

```go
store := ...
store.WithGenerateStoresFunc(..., false)
store.WithGenerateCustomResourceStoresFunc(..., true)
```

`(store.*Builder).useAPIServerCache` is implicitly overwritten by the invocation of `(store.*Builder).useAPIServerCache` though the expectation could be that they both keep a separate property / variable for this (e.g. `(store.*Builder).useAPIServerCacheForCustomResources`).

It could also be that `(store.*Builder).useAPIServerCache` should not be too tightly coupled with the function and thus needs to be configured separately by adding an extra method to do so.

Or this might be an indication the custom resources like this deserve their own builder, this way you wouldn't need to add respective methos for custom resources to the existing builder but create two separate builders and combine the stores later on.

Garrybest: Thanks @Serializator for reminding! I have not even noticed that. I prefer this solution.
```
It could also be that (store.*Builder).useAPIServerCache should not be too tightly coupled with the function and thus needs to be configured separately by adding an extra method to do so.
```

Serializator: I agree when looking specifically at `(store.*Builder).useAPIServerCache` but I'd like to "widen" the "scope" of my initial feedback and look at the bigger picture in terms of extensibility of `store.Builder` itself.

I haven't looked through the issue or pull request completely so please correct me if I'm wrong! CR metrics are basically metrics like the built-in metrics by KSM but registrered a different way.

Why did you make the choice to extend the existing `store.Builder` rather than create a separate instance of it specifically for registering and working with CR metrics, and later on make some kind of "composite" store which combines the built-in and the CR metrics?

I'm asking in terms of future extensability. If there is ever a need to add an additional way of registering metrics then methods will again be added to the `store.Builder` to allow for it ("extending" the `store.Builder`) rather than "building upon it" and keepign the implementation of `store.Builder` light and clean.

I'd love to hear your point of view and opinion on this! I might be completely wrong 😉

Garrybest: Hi @Serializator, it's really a good question.

I choose to extend `store.Builder` because IMO actually the Custom Resource is very similar to Kubernetes built-in resource. The only thing different is that we don't know their type, what their metrics look like, how to list and watch them(the client code is generated by `code-generator`). Once people tell us the necessary information, it behaves like a common built-in resource. It could be built and stored similarly, shared the same logic with `store.Builder`.

```
Rather than create a separate instance of it specifically for registering and working with CR metrics, and later on make some kind of "composite" store which combines the built-in and the CR metrics
```

I think this solution is more complicated. As I mentioned, the CR and built-in resource could share the same building logic once we have the information in `registry_factory.go`. Actually, what I'm thinking is all metrics could be registered by this interface including built-in resource.  From the aspect of extensability, the inner resource could be added by this method. Actually, I suppose removing the built-in resource registry from `store.Builder` and keep the CustomResourceStores may be more light and clean? 

Just some personal idea, lol 😄

Serializator: In that sense I understand and agree (not that you need my approval 😉) in terms of the extending the `store.Builder` and the why behind it.

`(store.Builder).buildCustomResourceStores` though is really similar to `(store.Builder).buildStores`, with the only difference being the "client" which is passed to the `listWatchFunc`. The crutch in this case is in the fact that `(store.Builder).buildStores` is aware of the client (`k8s.io/client-go/kubernetes.Interface` because `(store.Builder).buildStores` is responsible for calling the `listWatchFunc` function. Disappointing (😞 😆), maybe refactor in the future to abstract away the client (as this can differ now with the CR metrics).

pkg/customresource/registry_factory.go
```diff
@@ -0,0 +1,116 @@
...
+	//						{
+	//							Value: float64(f.Status.AvailableReplicas),
+	//						},
+	//					},
+	//				}
+	//			}),
+	//		),
+	//	}
+	// }
+	MetricFamilyGenerators(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator
```
Serializator, line 87: Should the implementation of `(customresource.RegistryFactory).MetricFamilyGenerators` be responsible for filtering labels and annotations based on which are allowed? What are the risks of doing this?

- What does this do to the expectation that `--metric-labels-allowlist` and `--metric-annotations-allowlist` work the same way for all metrics?

- Does this affect the flexibility in terms of adding onto these labels- and annotation allow lists with other types of lists in the future? With backward compatibility for the implementation of `customresource.RegistryFactory` in mind.

internal/store/builder.go
```diff
@@ -134,23 +137,57 @@ func (b *Builder) WithVPAClient(c vpaclientset.Interface) {
...
+func (b *Builder) WithGenerateCustomResourceStoresFunc(f ksmtypes.BuildCustomResourceStoresFunc) {
+	b.buildCustomResourceStoresFunc = f
 }
 
 // DefaultGenerateStoresFunc returns default buildStores function
 func (b *Builder) DefaultGenerateStoresFunc() ksmtypes.BuildStoresFunc {
 	return b.buildStores
 }
 
+// DefaultGenerateCustomResourceStoresFunc returns default buildStores function
```
Serializator, line 171: **nitpick;** should the comment not be:
``` go
// DefaultGenerateCustomResourceStoresFunc returns default **buildCustomResourceStores** function
```

internal/store/builder.go
```diff
@@ -134,23 +137,57 @@ func (b *Builder) WithVPAClient(c vpaclientset.Interface) {
...
+func (b *Builder) DefaultGenerateCustomResourceStoresFunc() ksmtypes.BuildCustomResourceStoresFunc {
+	return b.buildCustomResourceStores
+}
+
+// WithCustomResourceStoreFactories returns configures a custom resource stores factory
+func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.RegistryFactory) {
```
Serializator, line 177: What about checking if `(customresource.RegistryFactory).Name()` overrides any of the already existing
stores inside of the pre-defined built-in list of stores (`store.availableStores`)?

In that light, what would happen if two `customresource.RegistryFactory` with the same name would be passed
in? Would this result in a silent, maybe difficult to debug situation?

I would plead for adding a check which returns an error if a store with the same name as the `customresource.ResourceFactory` already exists. With possibly a boolean somewhere (e.g. inside of `customresource.ResourceFactory`) to forcefully overwrite an existing store.

internal/store/builder.go
Garrybest: Hi @Serializator. You are right! We have the same opinions😃. That also distresses me! 

These two functions are so similar but unfortunately we have no ideas about what type the `customResourceClient` is. If we merge them together, all types of `listWatchFunc` in `internal/store` directory must be revised. This revision may be a little bit huge and does not focus on custom resource extension. So I decided to make a optimization in the future. That would be better for reviewing.

internal/store/builder.go
Garrybest: Good idea. What about warning the overrides when the store already has an existed component with the same name? I'm afraid some guys may be not satisfied with built-in metrics and just want to make an override. We could just give a warning.

pkg/customresource/registry_factory.go
Garrybest: > What does this do to the expectation that --metric-labels-allowlist and --metric-annotations-allowlist work the same way for all metrics?
I don't figure out any good ways to keep the custom resource registry working the same way. Some metrics do not have these labels and some do focusing. It is really hard so I just let the interface implementers do this. Do you have any ideas?

> Does this affect the flexibility in terms of adding onto these labels- and annotation allow lists with other types of lists in the future? With backward compatibility for the implementation of customresource.RegistryFactory in mind.
How about using a `...[]string`? Like `MetricFamilyGenerators(allowanceLists ...[]string) []generator.FamilyGenerator`. It's a lit bit weird😂.

internal/store/builder.go
Serializator: I agree with the fact that some people might want to override built-in metrics. A simple warning to give the end-user an indication of what is happening seems like a good solution, much better than making the `customresource.ResourceFactory` more complex by adding an additional property.

pkg/customresource/registry_factory.go
Serializator: I personally think that to do this properly and prevent future headaches that a refactor of the way labels and annotations are filtered is appropriate here. If it is possible to refactor these filters outside of the `(generator.FamilyGenerator).GenerateFunc` that would make the generate function by definition less complex and (in the most optimal case) there is only one place where labels and annotations are filtered.

Initially I went with the idea of doing this **after** the metrics were generated but then I realized that `--metric-labels-allowlist` and `--metric-annotations-allowlist` are depended on whether the given label on the metric **was** either a label or annotation on the original resource object.

I then went looking and found the function below which eventually calls the `(metricsstore.MetricsStore).generateMetricsFuncs` with the original resource object.

https://github.com/kubernetes/kube-state-metrics/blob/840d30ce57f75d0ec466a2d612879602c433d5e6/pkg/metrics_store/metrics_store.go#L63-L82

I might be completely insane obviously so I'd like to know what the others think about this.

**^ all of the above is said with the intent of making this refactor a separate PR by the way, who might be you, someone else or myself** 😉

Garrybest: Wow, I think it's a pretty good idea! Users don't need to handle the associated label and annotation filter in their own codes. So how about fuction like `MetricFamilyGenerators() []generator.FamilyGenerator`?

internal/store/builder.go
Garrybest: I have added a `TODO` comment. We could optimize it later😉.

pkg/customresource/registry_factory.go
Serializator: I'm sorry, I think I've completely mistaken what `--metric-labels-allowlist` and `--metric-annotations-allowlist` are for and thus "recommended" something that is completely unnecessary.

`--metric-labels-allowlist` and `--metric-annotations-allowlist` are for `_annotations` and `_labels` metrics (e.g. `kubernetes_node_labels` and `kubernetes_node_annotations`).

There is no point in refactoring this whatsoever and the implementation is indeed responsible for using the label and annotation filters appropriately. If a CR metric is added and a similar `_labels` or `_annotations` metric for this CR is added the implementation is obviously responsible for taking the label and annotation allowlist into account.

Please ignore my clumsiness 😆

Garrybest: Ops, that's OK. I didn't realize too 😂

pkg/app/server.go
```diff
@@ -0,0 +1,332 @@
...
+	ctx := context.Background()
+
+	err := opts.Parse()
+	if err != nil {
+		klog.Fatalf("Error: %s", err)
+	}
+
+	if opts.Version {
+		fmt.Printf("%s\n", version.Print("kube-state-metrics"))
+		os.Exit(0)
```
fpetkovski, line 89: I agree that we need to a separate package for reusability, but interactions with the OS, such as parsing CLI args and providing exit codes should remain in main.go. Could you pull back some of these changes there so that server.go only runs KSM (and potentially its own metrics server) based on the options provided as function arguments? In case of errors, `RunKubeStateMetrics` should return an `error` which we can then log in main.go and exit with the proper status code.

fpetkovski (commented):
This is a great start, thanks for taking the initiative :+1:

pkg/app/server.go
Garrybest: Thanks! That's a good idea.

main.go
```diff
@@ -89,222 +45,9 @@ func main() {
...
-	// Informers don't seem to do a good job logging error messages when it
-	// can't reach the server, making debugging hard. This makes it easier to
-	// figure out if apiserver is configured incorrectly.
-	klog.Infof("Testing communication with server")
-	v, err := kubeClient.Discovery().ServerVersion()
-	if err != nil {
-		return nil, nil, errors.Wrap(err, "error while trying to communicate with apiserver")
+	ctx := context.Background()
+	if err := app.RunKubeStateMetrics(ctx, opts); err != nil {
+		klog.Fatalf("Fail to run kube-state-metrics: %v", err)
```
fpetkovski, line 51:
```suggestion
		klog.Fatalf("Failed to run kube-state-metrics: %v", err)
```

fpetkovski (commented):
Good stuff, just one small nit from my side

main.go
Garrybest: OK.

fpetkovski:
Great work, /lgtm

/hold for others to take a look as well

chrischdi:
First of all: thank you for working on this :+1: With this interface it is also possible to use the controller-runtime client :-)

I did take a look at it and have some feedback, but I think this could get fixed in future Pull Requests:

The flag `--resources` does not list the resource names of the CRs. Filtering for the CRs does work so its just about the help text.
Also in regards to that: maybe it would be handy if it is possible to disable the standard kube-state-metrics metric families by default (e.g. the main.go of kube-state-metrics could pass the core metrics to `RunKubeStateMetrics` the same way we can do for Custom Resources).

Garrybest:
Hi @chrischdi, I really agree with you. 

Now the flag `--resources` only list all built-in resources. I don't make a big revision here because I try to focus on CR registery and interface definition in this PR. So I just add all CRs into `enabledResources` but do not print them by default.

It is a good idea for us to fix it in the future PR. Thanks @chrischdi for your kind reminds. 😄

chrischdi:
> Hi @chrischdi, I really agree with you.
> 
> Now the flag `--resources` only list all built-in resources. I don't make a big revision here because I try to focus on CR registery and interface definition in this PR. So I just add all CRs into `enabledResources` but do not print them by default.
> 
> It is a good idea for us to fix it in the future PR. Thanks @chrischdi for your kind reminds. smile

:+1: yes I think that's totally fine.

I already found a workaround which I think is more than good enough: 
https://github.com/chrischdi/cluster-api-state-metrics/blob/chrischdi/ksm-pr-adopt/main.go#L42

Garrybest:
> I already found a workaround which I think is more than good enough:
https://github.com/chrischdi/cluster-api-state-metrics/blob/chrischdi/ksm-pr-adopt/main.go#L42

Great👍, that's what I'm thinking before! See https://github.com/kubernetes/kube-state-metrics/pull/1644#discussion_r758346687. The built-in resource could also implement the registry interface and  be passed to `RunKubeStateMetrics` as a parameter like what you did. And the problem we mentioned could also be solved.

Garrybest:
Hi @brancz @dgrisonnet, would you please take a look? My appreciation. 😄

internal/store/builder.go
```diff
@@ -134,23 +137,61 @@ func (b *Builder) WithVPAClient(c vpaclientset.Interface) {
...
 
+// DefaultGenerateCustomResourceStoresFunc returns default buildCustomResourceStores function
+func (b *Builder) DefaultGenerateCustomResourceStoresFunc() ksmtypes.BuildCustomResourceStoresFunc {
+	return b.buildCustomResourceStores
+}
+
+// WithCustomResourceStoreFactories returns configures a custom resource stores factory
+func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.RegistryFactory) {
+	for i := range fs {
+		f := fs[i]
```
Garrybest, line 179: Hi @chrischdi, I made a mistake here in the previous version. The `f` must be copied because the value of `availableStores` is a func. Otherwise, it will be registed the last ones when custom resources are more than one. I have fixed here, please pay attention. My apology.

Garrybest:
Can we move forward?

PR #1648, opened 2021-12-07
Add "kube_horizontalpodautoscaler_info" metric

### LINKED ISSUE #1635: Add a new metric for HorizontalPodAutoscalar metrics with scaleTargetRef values

This repeats the feature request from the automatically closed Issue #1395 

**Why is this needed**:

I would like to find HPA, which have a non existing target.

See #1395

**Additional context**

Serializator:
Adds a metric which exposes the `scaleTargetRef.apiVersion`, `scaleTargetRef.kind` and `scaleTargetRef.name` of the `HorizontalPodAutoscaler` resource objects in the cluster.

>Adding a new metric for the HPA target will have a slight impact on cardinality, but most clusters have at most a double or triple digit amount of HPAs. So we are talking about at most 1000 newly added time series, which is really not that significant compared to the number of time series we export for pods and nodes.

Fixes #1635

### DIFF
--- a/internal/store/horizontalpodautoscaler.go
+++ b/internal/store/horizontalpodautoscaler.go
@@ -57,6 +57,29 @@ var (
 func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator {
 	return []generator.FamilyGenerator{
 		*generator.NewFamilyGenerator(
+			"kube_horizontalpodautoscaler_info",
+			"Information about this autoscaler.",
+			metric.Gauge,
+			"",
+			wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				labelKeys := []string{"scaletargetref_kind", "scaletargetref_name"}
+				labelValues := []string{a.Spec.ScaleTargetRef.Kind, a.Spec.ScaleTargetRef.Name}
+				if a.Spec.ScaleTargetRef.APIVersion != "" {
+					labelKeys = append([]string{"scaletargetref_api_version"}, labelKeys...)
+					labelValues = append([]string{a.Spec.ScaleTargetRef.APIVersion}, labelValues...)
+				}
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   labelKeys,
+							LabelValues: labelValues,
+							Value:       1,
+						},
+					},
+				}
+			}),
+		),
+		*generator.NewFamilyGenerator(
 			"kube_horizontalpodautoscaler_metadata_generation",
 			"The generation observed by the HorizontalPodAutoscaler controller.",
 			metric.Gauge,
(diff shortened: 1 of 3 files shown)


internal/store/horizontalpodautoscaler.go
```diff
@@ -56,6 +56,23 @@ var (
...
+			"kube_horizontalpodautoscaler_info",
+			"Information about this autoscaler.",
+			metric.Gauge,
+			"",
+			wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{"scaletargetref_api_version", "scaletargetref_kind", "scaletargetref_name"},
+							LabelValues: []string{a.Spec.ScaleTargetRef.APIVersion, a.Spec.ScaleTargetRef.Kind, a.Spec.ScaleTargetRef.Name},
```
fpetkovski, line 69: Can we add a test to see what happens when a.Spec.ScaleTargetRef is not set?

internal/store/horizontalpodautoscaler.go
```diff
@@ -56,6 +56,23 @@ var (
...
+		*generator.NewFamilyGenerator(
+			"kube_horizontalpodautoscaler_info",
+			"Information about this autoscaler.",
+			metric.Gauge,
+			"",
+			wrapHPAFunc(func(a *autoscaling.HorizontalPodAutoscaler) *metric.Family {
+				return &metric.Family{
+					Metrics: []*metric.Metric{
+						{
+							LabelKeys:   []string{"scaletargetref_api_version", "scaletargetref_kind", "scaletargetref_name"},
```
fpetkovski, line 68: We should probably include the HPA's name and namespace in the labels, otherwise we can't associate a target with an HPA resource :)

Serializator: These labels are added by default in `store.wrapHPAFunc` 😄 
https://github.com/kubernetes/kube-state-metrics/blob/f5b42be4455d5ab37063c848d924f28c4e5b0ea1/internal/store/horizontalpodautoscaler.go#L282-L285

fpetkovski: Ah, ok forgot about that completely :+1:

internal/store/horizontalpodautoscaler.go
Serializator: The OpenAPI specification tells us that `a.Spec.ScaleTargetRef` is required.
https://github.com/kubernetes/kubernetes/blob/0d3f2ca371d03cc58c36d3ee3523b719e987fc35/api/openapi-spec/v3/apis__autoscaling__v2beta2_openapi.json#L323-L326

Though the `CrossVersionObjectReference` does not require the `apiVersion` property and thus could be nil.
https://github.com/kubernetes/kubernetes/blob/0d3f2ca371d03cc58c36d3ee3523b719e987fc35/api/openapi-spec/v3/apis__autoscaling__v2beta2_openapi.json#L74-L77

Agree to add validation for `a.Spec.ScaleTargetRef.APIVersion` only?

fpetkovski: Thanks for double checking, I remember that we had a similar issue for VPA. In this case, let's just add a test for the APIVersion field

PR #1652, opened 2021-12-16
Add ingressclass to kube_ingress_info metric

adammw:
Add ingressclass to kube_ingress_info metric

Should be same cardinality since it adds a tag to an existing metric that is already one per resource.

### DIFF
--- a/docs/ingress-metrics.md
+++ b/docs/ingress-metrics.md
@@ -3,7 +3,7 @@
 | Metric name| Metric type | Labels/tags | Status |
 | ---------- | ----------- | ----------- | ----------- |
 | kube_ingress_annotations | Gauge | `ingress`=&lt;ingress-name&gt; <br> `namespace`=&lt;ingress-namespace&gt; <br> `annotation_INGRESS_ANNOTATION`=&lt;ANNOTATION_LABEL&gt; | EXPERIMENTAL |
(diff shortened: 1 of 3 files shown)


fpetkovski:
@dgrisonnet @mrueg do we have any conventions about adding labels to stable metrics, it might break existing rules and queries that use aggregations.

mrueg:
No official guidelines exist I believe. Since this is an info metric, I'd assume folks extract labels from it and I think the risk of breaking anyones workflow there by adding a new label is low. Also it would be part of a new minor/feature release.

fpetkovski:
Thanks Manuel, this makes sense to me.

@adammw I noticed that the ingress class label is determined from an annotation value. Is there any reason why you cannot use the `--metric-annotations-allowlist` flag to expose the ingress class annotation? Also, Kubernetes has already made ingress class a first order citizen [1] and that specific annotation has been declared deprecated [2].

[1] https://kubernetes.io/docs/concepts/services-networking/ingress/#ingress-class
[2] https://kubernetes.io/docs/concepts/services-networking/ingress/#deprecated-annotation

adammw:
Yep, the idea is to fall back to the annotation if the first-class ingressClassName is not used..

fpetkovski:
Ah I missed that part, then this makes sense to me :+1: 

/lgtm

PR #1653, opened 2021-12-17
pdb: use plural form of resource name in allowlist key

arajkumar:

All other allowlist selectors are using plural form of resource name
except PDB. Actually it is a typo introduced by PR #1623.

Signed-off-by: Arunprasad Rajkumar <arajkuma@redhat.com>

N/A

Fixes #

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -325,7 +325,7 @@ func (b *Builder) buildPersistentVolumeStores() []cache.Store {
 }
 
 func (b *Builder) buildPodDisruptionBudgetStores() []cache.Store {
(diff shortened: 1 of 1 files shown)


arajkumar:
As per the https://github.com/kubernetes/kube-state-metrics/blob/master/RELEASE.md, I'm targeting this bug fix to latest release branch rather than master.

arajkumar:
The workaround is to use `--metric-labels-allowlist=,poddisruptionbudget=[*]` flag instead of `--metric-labels-allowlist=,poddisruptionbudgets=[*]`.

Refer https://github.com/openshift/cluster-monitoring-operator/pull/1516

mrueg:
good catch! Thanks for the fix

/approve

PR #1656, opened 2021-12-20
Add new syntax / notation for selecting namespaces and their labels

### LINKED ISSUE #1631: Get metrics from selected namespaces using labels
A way of gathering metrics from multiple namespaces that matches a label/label selector.

**Why is this needed**:
I monitor multiple clusters where I'm interested in platform specific metrics while our developers uses other solutions to gather metrics.
There is no reason why I should gather there metrics as well.
The `--namepsaces` flag is a good start but it's not easy to gather metrics from many namespaces when I'm providing platforms from different cloud providers. For example in AWS and Azure have a few different namespaces for us but they all have a specific platform label.

Something like:
--namespaces-labels=["platfrom=true"]

external-dns have implemented this: https://github.com/kubernetes-sigs/external-dns/blob/3756f60c4763928d851c2ca22f4f6ed6113c5b53/pkg/apis/externaldns/types.go#L368

**Additional context**
Similar as https://github.com/kubernetes/kube-state-metrics/issues/152

Serializator:
**Currently this PR is a draft and meant for receiving feedback on the parsing. Once the parsing is reviewed I will continue refactoring the usages of the "old" namespace list option to the new one.**

This PR extends the existing syntax / notation (BC) to allow for selecting namespaces based on their labels. It does not yet implement the filtering upon those aforementioned labels, this will be a separate PR.

It does not affect cardinality.

**Which issue(s) this PR fixes**: n/a

### DIFF
--- a/pkg/options/namespace_list/parser.go
+++ b/pkg/options/namespace_list/parser.go
@@ -1,0 +1,163 @@
+/*
+Copyright 2021 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package namespacelist
+
+import (
+	"bufio"
+	"fmt"
+	"io"
+	"strings"
+
+	"github.com/pkg/errors"
+	"k8s.io/apimachinery/pkg/labels"
+)
+
+// Parser parses the syntax for namespace lists
+type Parser struct {
+	source string
+	pos    int
+}
+
+type UnexpectedSyntaxError struct {
+	char string
+	pos  int
+}
+
+// UnexpectedSyntaxError is returned when an invalid syntax is being parsed
+func (err UnexpectedSyntaxError) Error() string {
+	return fmt.Sprintf("syntax error, unexpected %v at col %v", err.char, err.pos)
+}
+
+func (parser *Parser) Parse() (map[string]labels.Selector, error) {
+	reader := bufio.NewReader(strings.NewReader(parser.source))
+	selectors := make(map[string]labels.Selector)
+
+	namespace := strings.Builder{}
+
+	err := readRunes(reader, func(r rune, eof bool, close func()) error {
+		parser.pos++
+
+		if eof {
+			if namespace.Len() > 0 {
+				selectors[namespace.String()] = labels.Everything()
+			}
+			close()
+			return nil
+		}
+
+		if r == '=' {
+			// foo-bar=[...], ...
+			r, _, err := reader.ReadRune()
+			if errors.Is(err, io.EOF) || r != '[' {
+				// foo-bar=...
+				return UnexpectedSyntaxError{char: "EOF", pos: parser.pos}
+			} else if err != nil {
+				return err
+			}
+
+			selector, err := parseSelector(parser, reader)
+			if err != nil {
+				return err
+			}
+
+			selectors[namespace.String()] = selector
+			namespace.Reset()
+			return nil
+		}
+
+		if r == ',' {
+			// foo-bar, ...
+			selectors[namespace.String()] = labels.Everything()
+			namespace.Reset()
+			return nil
+		}
+
+		namespace.WriteRune(r)
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	return selectors, nil
+}
+
+func parseSelector(parser *Parser, reader *bufio.Reader) (labels.Selector, error) {
+	str := strings.Builder{}
+
+	err := readRunes(reader, func(r rune, eof bool, close func()) error {
+		parser.pos++
+
+		if eof {
+			// a namespace selector should always be closed with a "]"
+			return UnexpectedSyntaxError{char: "EOF", pos: parser.pos}
+		}
+
+		if r == ']' {
+			// indicates the end of a namespace selector
+			close()
+			return nil
+		}
+
+		str.WriteRune(r)
+		return nil
+	})
+	if err != nil {
+		return nil, err
+	}
+
+	selector, err := labels.Parse(str.String())
+	if err != nil {
+		return nil, err
+	}
+	return selector, nil
+}
+
+func readRunes(r *bufio.Reader, reader func(r rune, eof bool, close func()) error) error {
+	reading := true
+
+	closer := func() {
+		reading = false
+	}
+
+	for reading {
+		r, _, err := r.ReadRune()
+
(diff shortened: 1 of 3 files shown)


Serializator:
I accidentally created this PR as a normal PR. Could someone with write access to the repository change it to a draft please?

pkg/options/namespace_list/parser.go
```diff
@@ -0,0 +1,174 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
```
Serializator, line 2: **nit for myself:** 2021

pkg/options/namespace_list/types.go
```diff
@@ -0,0 +1,58 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
```
Serializator, line 2: **nit for myself:** 2021

pkg/options/namespace_list/types_test.go
```diff
@@ -0,0 +1,128 @@
+/*
+Copyright 2018 The Kubernetes Authors All rights reserved.
```
Serializator, line 2: **nit for myself:** 2021

pkg/options/namespace_list/parser.go
```diff
@@ -0,0 +1,178 @@
...
+	char string
+	pos  int
+}
+
+// Error returns the error as a human-readable string
+func (err UnexpectedSyntaxError) Error() string {
+	return fmt.Sprintf("syntax error, unexpected %v at col %v", err.char, err.pos)
+}
+
+// Parse parses the source into a map
```
fpetkovski, line 45: Can we document what the resulting map contains or how it looks like?

Alternatively, we can return a struct instead of a map. That should also make it easier to understand the contents by just reading the code.

pkg/options/namespace_list/types_test.go
```diff
@@ -0,0 +1,128 @@
...
+package namespacelist
+
+import (
+	"reflect"
+	"testing"
+
+	"github.com/pkg/errors"
+)
+
+func TestString(t *testing.T) {
```
fpetkovski, line 26: Maybe this test should be repurposed for the Parse method instead? It took me a while to realize that Parse was indirectly tested through NamespaceList. Not a blocker though, just a suggestion.

fpetkovski (commented):
Left a few nits, otherwise looks really good.

pkg/options/namespace_list/parser.go
Serializator: Would the struct be something like below? I much rather use the struct approach than document it. Code should explain itself as much as possible.

```go
type NamespaceList []struct {
    Namespace string
    Labels []string
}
```

fpetkovski: Or maybe something like 

```go
type NamespaceMatcher struct {
    Name string
    Labels []string
}
```

Serializator: I believe using "Matcher" in this case insinuates that the struct is meant to do more than be a data structure.

mrueg:
I was always hoping we could profit from upstream logic (e.g. https://github.com/kubernetes/apimachinery/tree/v0.23.1/pkg/labels ) and not invent that parser ourselves. Maybe we could add another flag --namespaces-selectors (or similar) that allows you to use specify conditions on how namespaces get selected?

Serializator:
> I was always hoping we could profit from upstream logic (e.g. https://github.com/kubernetes/apimachinery/tree/v0.23.1/pkg/labels ) and not invent that parser ourselves. Maybe we could add another flag --namespaces-selectors (or similar) that allows you to use specify conditions on how namespaces get selected?

I agree completely in terms of not inventing the parser ourselves. Of course the parser you mentioned is much more extensive and reliable that the one I wrote so I would prefer using that one as well.

I don't think that adding another option (`--namespaces-selectors` in this case) is the way to go. That would mean there would be 3 options acting upon and deciding what namespaces are either in- or excluded.

- `--namespaces`
- `--namespaces-denylist`
- `--namespaces-selectors`

I would like to take this opportunity to merge `--namespaces` and `--namespaces-denylist` together so that there is one `--namespaces` option with the purpose of either in- or excluding namespaces, rather than 2 / 3 separate options.

Though I understand that this approach poses some BC questions.

pkg/options/namespace_list/parser.go
fpetkovski: Now that you brought this up, I don't think that would be a bad idea. We can filter namespaces by labels by passing the labels to the api server, but we can't really filter by prefix when using wildcards. So I assume that logic will have to be in KSM and the NamespaceMatcher can have a `Matches(string) bool` or `Matches(corev1.Namespace) bool` function to facilitate that. It's done similarly here, albeit through an interface: https://pkg.go.dev/k8s.io/apimachinery/pkg/labels#Selector

Do you have some idea of how you intend to do the filtering?

Serializator:
//cc @fpetkovski what do you think of the above?

fpetkovski:
If we can use the parser from api-machinery, that would surely reduce the maintenance burden on KSM. I wouldn't worry too much about the compatibility disclaimer though. Changing the label parser would likely be a disruptive change to the entire ecosystem and we can also add tests to make sure we catch breaking changes if they happen.

I agree that introducing `--namespaces-selector` might be redundant if we can extend `--namespaces` in a backwards compatible manner. However I wouldn't try to merge `--namespaces` and `--namespaces-denylist` yet. We already have this pattern established for different cli args and I would prefer keeping the options separate in order to maintaining consistency.

Serializator:
I can see what you mean in terms of keeping `--namespaces` and `--namespaces-denylist` separate.

I looked through the syntax for parsing labels in API machinery and I do believe we can reuse it but we will still need to do a bit of parsing ourselves as the parser in API machinery is for labels into a selector. In our case the syntax is:

`namespace-a[<selector>], namespace-b`

This means the parser in API machinery can be used for the `<selector>` part but we still need to do the parsing of `namespace[...]` itself. Which is not too bad. `[` and `]` are not characters allowed within the API machinery so we can safely assume that they either indicate the start or end of a selector (`<selector>`) for a given namespace.

fpetkovski:
This looks good to me, we just need to fix the CI before merging. Would you like to continue working on the rest of the implementation in a separate PR?

Serializator:
Sorry for the late response, I've been quite busy the last couple of weeks / days.

@fpetkovski, I've been thinking about ways to implement this but one bottleneck I came across is the fact that the Kubernetes SDK / API does not allow for getting resources from multiple namespaces at once I believe. To implement this we'll need to get all namespaces and get metrics for each one of those. At the same time these namespaces need to be watched in case a namespace is created or deleted to keep the in-memory namespaces in KSM in-sync.

Do you have an idea of ways to implement this in a more performant (and clean) way?

fpetkovski:
@Serializator sorry I completely missed your comment.

Yes you're right that filtering by namespace might not be as simple, especially if we want to filter by namespace labels and also update exported metrics when those labels change. The only solution I see is doing this filtering in memory as you pointed out. 

I think it's fine to do this as long as we don't worsen the performance of the filtering functionality (only by namespace names).

Serializator:
I will try and get some work done on this soon.

/remove-lifecycle stale

PR #1668, opened 2022-01-21
jsonnet: explicitly forbid privilege escalation

ArthurSens:
We just [introduced Kubernetes manifests scanning](https://github.com/prometheus-operator/kube-prometheus/pull/1584) with [Kubescape](https://github.com/armosec/kubescape) in kube-prometheus. Kubescape is a security scanning tool used to help us maintain a good risk analysis and be aligned with security compliances. 

One of the failures found with Kubescape is that we [don't explicitly forbid container privilege escalation](https://github.com/prometheus-operator/kube-prometheus/issues/1588) for several components, including kube-state-metrics. 

Looking at [kubernetes docs](https://kubernetes.io/docs/tasks/configure-pod-container/security-context/): 
```
AllowPrivilegeEscalation: Controls whether a process can gain more privileges than its parent process. This bool directly controls whether the no_new_privs flag gets set on the container process. AllowPrivilegeEscalation is true always when the container is: 1) run as Privileged OR 2) has CAP_SYS_ADMIN.
```

Which means that kube-state-metrics's container is already deployed with `AllowPrivilegeEscalation` set to false because there's no need to run it as a privileged container, but kubescape asks to explicitly set this to false.

We can override this ourselves downstream, but we're wondering if it would be ok to include this in kube-state-metrics repository.

Cheers 👋 

Does not change

None

### DIFF
--- a/examples/autosharding/statefulset.yaml
+++ b/examples/autosharding/statefulset.yaml
@@ -54,6 +54,7 @@ spec:
           initialDelaySeconds: 5
           timeoutSeconds: 5
         securityContext:
+          allowPrivilegeEscalation: false
           runAsUser: 65534
       nodeSelector:
         kubernetes.io/os: linux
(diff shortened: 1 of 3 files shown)


fpetkovski:
Thanks for the contribution @ArthurSens. Could you also regenerate the manifests?

PR #1685, opened 2022-02-16
Change the processing type from int to float in kube_horizontalpodautoscaler_spec_target_metric

### LINKED ISSUE #1594: HPA using a custom metric of type Object is not visible in `kube_horizontalpodautoscaler_spec_target_metric`
**What happened**:

I have a custom metric (via Prometheus Adapter) to drive the HPA of a certain deployment. It's an Object metric, in this case a Deployment. The target type is "Value"

```
apiVersion: autoscaling/v1
kind: HorizontalPodAutoscaler
metadata:
  name: myHPA
spec:
  minReplicas: 10
  maxReplicas: 100
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: myDeployment
  metrics:
  - type: Object
    object:
      metric:
        name: my_custom_metric_from_prometheus_adapter
      describedObject:
        kind: Deployment
        apiVersion: apps/v1
        name: myDeployment
      target:
        type: Value
        value: 0.5

```
The HPA is behaving as expected scaling the deployment. BUT when I query `kube_horizontalpodautoscaler_spec_target_metric` my HPA is not there. Other HPA based on custom metrics but applied to Pods (a more common case?) are visible.

My HPA should be there along with others. To note, in other metrics (e.g. kube_horizontalpodautoscaler_spec_min_replicas) I do see the name of my HPA.

**Environment**:
- kube-state-metrics version: 2.0.0
- Kubernetes version (use `kubectl version`): v1.16.9
- Cloud provider or hardware configuration: ad-hoc k8s cluster
- Other info:

whitebear009:
Since the kube_horizontalpodautoscaler_spec_target_metric metric currently uses the AsInt64() function when processing data, if there is a float type metric between 0 and 1, it will be filtered and will not be displayed.

Fixes #1594

### DIFF
--- a/internal/store/horizontalpodautoscaler.go
+++ b/internal/store/horizontalpodautoscaler.go
@@ -135,39 +135,40 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 					var metricName string
 
 					var v [metricTargetTypeCount]float64
+					var ok [metricTargetTypeCount]bool
 
 					switch m.Type {
 					case autoscaling.ObjectMetricSourceType:
 						metricName = m.Object.Metric.Name
 
 						if m.Object.Target.Value != nil {
-							v[value] = float64(m.Object.Target.Value.MilliValue()) / 1000
+							v[value], ok[value] = float64(m.Object.Target.Value.MilliValue())/1000, true
 						}
 						if m.Object.Target.AverageValue != nil {
-							v[average] = float64(m.Object.Target.AverageValue.MilliValue()) / 1000
+							v[average], ok[average] = float64(m.Object.Target.AverageValue.MilliValue())/1000, true
 						}
 					case autoscaling.PodsMetricSourceType:
 						metricName = m.Pods.Metric.Name
 
-						v[average] = float64(m.Pods.Target.AverageValue.MilliValue()) / 1000
+						v[average], ok[average] = float64(m.Pods.Target.AverageValue.MilliValue())/1000, true
 					case autoscaling.ResourceMetricSourceType:
 						metricName = string(m.Resource.Name)
 
 						if m.Resource.Target.AverageUtilization != nil {
-							v[utilization] = float64(*m.Resource.Target.AverageUtilization)
+							v[utilization], ok[utilization] = float64(*m.Resource.Target.AverageUtilization), true
 						}
 
 						if m.Resource.Target.AverageValue != nil {
-							v[average] = float64(m.Resource.Target.AverageValue.MilliValue()) / 1000
+							v[average], ok[average] = float64(m.Resource.Target.AverageValue.MilliValue())/1000, true
 						}
 					case autoscaling.ExternalMetricSourceType:
 						metricName = m.External.Metric.Name
 
 						if m.External.Target.Value != nil {
-							v[value] = float64(m.External.Target.Value.MilliValue()) / 1000
+							v[value], ok[value] = float64(m.External.Target.Value.MilliValue())/1000, true
 						}
 						if m.External.Target.AverageValue != nil {
-							v[average] = float64(m.External.Target.AverageValue.MilliValue()) / 1000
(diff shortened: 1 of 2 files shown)


whitebear009:
/assign @fpetkovski
Please review. Thanks.

internal/store/horizontalpodautoscaler.go
```diff
@@ -134,51 +134,53 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
...
-							v[average], ok[average] = m.External.Target.AverageValue.AsInt64()
+							v[average], ok[average] = float64(m.External.Target.AverageValue.MilliValue())/1000, true
 						}
 					default:
 						// Skip unsupported metric type
 						continue
 					}
 
-					for i := range ok {
+					for i := range v {
```
fpetkovski, line 178: I assume modifying the slice does not change the behavior because the length of both `v` and `ok` is the same. Is that correct?

internal/store/horizontalpodautoscaler_test.go
```diff
@@ -80,7 +80,7 @@ func TestHPAStore(t *testing.T) {
 								},
 								Target: autoscaling.MetricTarget{
 									Value:        resourcePtr(resource.MustParse("10")),
-									AverageValue: resourcePtr(resource.MustParse("12")),
+									AverageValue: resourcePtr(resource.MustParse("0.5")),
```
fpetkovski, line 83: Could we also add a test case when `Value` is a float?

internal/store/horizontalpodautoscaler.go
whitebear009: Yes. Because I removed `ok` during the coding process and forgot to change it back. 

I think it's better to use `ok` here. It has been changed back now.

internal/store/horizontalpodautoscaler_test.go
whitebear009: Test cases have been updated.

whitebear009:
A new commit has been submitted. Please check again, thanks. @fpetkovski

internal/store/horizontalpodautoscaler.go
```diff
@@ -134,39 +134,41 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
...
 					case autoscaling.ObjectMetricSourceType:
 						metricName = m.Object.Metric.Name
 
-						v[value], ok[value] = m.Object.Target.Value.AsInt64()
+						if m.Object.Target.Value != nil {
+							v[value], ok[value] = float64(m.Object.Target.Value.MilliValue())/1000, true
+						}
 						if m.Object.Target.AverageValue != nil {
-							v[average], ok[average] = m.Object.Target.AverageValue.AsInt64()
+							v[average], ok[average] = float64(m.Object.Target.AverageValue.MilliValue())/1000, true
```
fpetkovski, line 148: Since we are now always storing `true` in `ok`, do we still need the `ok` variable? Can we try removing it to see if the tests break?

whitebear009: Good idea! The `ok` variable seems useless. But removing it directly would break the unit test, since we set the length of the `v` to a constant, it would show the extra zeros (the default value).

Although I can judge whether the value of v[i] is zero (HPA target metric should be required to be greater than 0) to decide whether to display the corresponding metric, but I think this way is not very good.
So Could I change the type of `v` to the map type, and finally displayed the metric that exists in the map?

whitebear009: I have push a new commit. How do you think?

fpetkovski: I think a map could work, but it might also create a flaky test. Iterating over a map does not guarantee that the order of the keys will always be the same. Can this lead to metrics being serialized in a different order each time a test is run? Could you run the tests with -count=100?

fpetkovski: Perfect!

/lgtm

Issue #1679, opened 2022-02-07
 duplicated metrics using kube-state-metrics sharding with --use-apiserver-cache


m-messiah:
**What happened**:
Running kube-state-metrics in shards with enabled `--use-apiserver-cache` resulted in having constantly growing metrics which are not related to actual data.

**What you expected to happen**:
Correct sharding with reduced apiserver latency

**How to reproduce it (as minimally and precisely as possible)**:
1. Start more than 1 shard of kube-state-metrics.
  deployment0: `--shard=0 - --total-shards=2 --use-apiserver-cache`
  deployment1: `--shard=1 - --total-shards=2 --use-apiserver-cache`
2. Scrape through servicemonitor from examples.
3. Continuously start and stop some deployments.
4. Check `count by (pod) (kube_pod_info)` in prometheus

**Anything else we need to know?**:
When I removed the flag `--use-apiserver-cache` the counters dropped to good values.
<img width="2325" alt="image" src="https://user-images.githubusercontent.com/2103030/152810546-a8ac1c7c-20a5-48bb-af6b-38b1c5101ebe.png">

Probably related to https://github.com/kubernetes/kube-state-metrics/issues/1166

**Environment**:
- kube-state-metrics version: 2.3.0
- Kubernetes version (use `kubectl version`): v1.19.13

dgrisonnet:
This might be because the mechanism we rely on to serve cached data from the apiserver is vulnerable to stale reads: https://github.com/kubernetes/kubernetes/issues/59848. I think it is not emphasized enough in the help text that this flag should be used with care because of that.

That said, do you know if this is replicable with only one instance of kube-state-metrics? Because it looks like there are way more stale data then what I would have expected.

jpdstan:
We're seeing the same thing https://github.com/kubernetes/kube-state-metrics/issues/1569

dgrisonnet:
@jpdstan you were seeing the same problem without sharding right?

jpdstan:
@dgrisonnet this is with sharding as well. 10 pods, approx 500k unique metrics each

fpetkovski:
I wonder if we have a bug in https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/sharding/listwatch.go. 

Does the issue happen without sharding?

JohnRusk:
@dgrisonnet 
[...]

Issue #1569, opened 2021-09-02
deleted pods still reporting metrics


jpdstan:
**What happened**:

it seems that _sometimes_ metrics don't get deleted alongside the pod. It isn't until we churn all the kube-state-metrics pods that it fixes it.

What's even stranger is that it won't be _all_ metrics for that pod that will incorrectly exist; for example, for a particular pod that was deleted, we noticed that it was still reporting `kube_pod_container_status_waiting_reason`, but not `kube_pod_container_resource_requests`.

**What you expected to happen**:

When a pod gets deleted, all metrics associated with that pod should also be deleted.

**How to reproduce it (as minimally and precisely as possible)**:

It's unclear as to how this happens - whenever we try to reproduce by manually deleting a pod and querying for all its metrics (`{pod="my_pod"}`), it seems to work just fine, i.e. the metrics all disappear.

**Anything else we need to know?**:

**Environment**:
- kube-state-metrics version: 2.2.0 (though we were experiencing this on 1.5.0 as well)
- Kubernetes version (use `kubectl version`): 
```
Client Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.1", GitCommit:"4485c6f18cee9a5d3c3b4e523bd27972b1b53892", GitTreeState:"clean", BuildDate:"2019-07-18T09:18:22Z", GoVersion:"go1.12.5", Compiler:"gc", Platform:"darwin/amd64"}
Server Version: version.Info{Major:"1", Minor:"15", GitVersion:"v1.15.12", GitCommit:"e2a822d9f3c2fdb5c9bfbe64313cf9f657f0a725", GitTreeState:"clean", BuildDate:"2020-05-06T05:09:48Z", GoVersion:"go1.12.17", Compiler:"gc", Platform:"linux/amd64"}
```
- Cloud provider or hardware configuration: self-hosted k8s on aws
- Other info:
```
      /kube-state-metrics
      --port=9102
      --telemetry-port=8081
      --resources=configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,jobs,limitranges,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets
      --use-apiserver-cache
      --metric-labels-allowlist=daemonsets=[*],deployments=[*],jobs=[*],nodes=[*],pods=[*],secrets=[*]
[...]

PR #1700, opened 2022-03-02
Prevent watch from resetting ResourceVersion.

e-ngo:
Prevent ListWatcher's Watch from setting ResourceVersion=0.  
This seems to be undesired behavior especially when using the `ListWatch` interface, in which the Reflector performs an initial `List` with `RV=0` to get the cached RV, then tries to start `Watch` from the cached RV (https://github.com/kubernetes/client-go/blob/v0.22.0/tools/cache/reflector.go#L401-L415). Resetting the RV to 0 after this `List` introduces potential inconsistencies. Additionally, when a Watch client retries on a different API server, the watch caches may be behind what the client has seen, and potentially introduces stale metrics that way as well. There is also a Warning against using RV=0 for Watch semantics which suggests additional inconsistencies/staleness from using Watch RV=0 semantics:

> Caution: Watches initialized this way may return arbitrarily stale data. Please review this semantic before using it, and favor the other semantics where possible.

(https://kubernetes.io/docs/reference/using-api/api-concepts/#semantics-for-watch)
Does not change cardinality
Fixes #
Potentially related to: https://github.com/kubernetes/kube-state-metrics/issues/1679
https://github.com/kubernetes/kube-state-metrics/issues/1569


dgrisonnet:
This is indeed wrong, the initial LIST with RV = 0 should initialize the list watcher to get data from the apiserver watch cache. Currently, the objects returned by the initial LIST and the one by the first "Added" WATCH event may have different RVs which can definitely lead to some inconsistencies.

/lgtm
/approve

PR #1709, opened 2022-03-17
pkg/options: Remove flag import

mrueg:
Small change so we use pflag everywhere in the code.

None
Fixes #

### DIFF
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -17,7 +17,6 @@ limitations under the License.
 package options
 
 import (
-	"flag"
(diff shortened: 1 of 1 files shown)


fpetkovski:
Looks like it broke the CI 🤔

fpetkovski:
Looks like it broke the CI 🤔

mrueg:
> Looks like it broke the CI thinking

I think I made a mistake here and the change is not useful at all. Thanks for taking a look though! :) Closing the PR.

PR #1710, opened 2022-03-23
feat: enable customresource metrics by configuration

iamnoah:

Building off of the new `customresource.RegistryFactory` (thanks @Garrybest !), I found myself making a list of metrics I wanted on my custom resources and thought it made more sense to drive collection of those metrics from a config file than having to write pretty similar code for each one.

I've put together some examples in the .md file, and the tests cover most of the ins and outs.

This is a pretty early rough draft that suits my needs, but I thought I would put it out there for feedback and see if it's useful. 

Users may add unlimited custom metrics, but the default cardinality remains unchanged.


Garrybest:
Thanks @iamnoah, it is a really good practice!

main.go
```diff
@@ -20,9 +20,13 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/prometheus/common/version"
-	"k8s.io/klog/v2"
+	"gopkg.in/yaml.v3"
+	"k8s.io/klog"
```
mrueg, line 27: Is the switch to klog/v1 here intentional?

docs/customresourcestate-metrics.md
```diff
@@ -0,0 +1,220 @@
+# Custom Resource State Metrics
+
+This section describes how to add metrics based on the state of a custom resource without writing a custom resource 
+registry and running your own build of KSM.
+
+## Configuration
+
+The `KSM_CUSTOM_RESOURCE_METRICS` environment variable can be used to configure the metrics for custom resources:
```
mrueg, line 8: As environment variables are limited in size to 32k chars, have you considered using a file-based config instead?

mrueg (commented):
Overall looks really interesting. Thanks for the contribution!

main.go
iamnoah: nope, fixed

docs/customresourcestate-metrics.md
iamnoah: Oh sure easy as, just prototyping with something simple.

iamnoah:
Is that linter error related to my changes in some way?

fpetkovski:
I think it's an issue in golangci-lint: https://github.com/golangci/golangci-lint/issues/2374. Seems to be happening with go 1.18. I can take a look today to see if we can get it quickly resolved.

fpetkovski:
This PR should fix the CI issue: https://github.com/kubernetes/kube-state-metrics/pull/1713

fpetkovski:
The CI should be fixed with https://github.com/kubernetes/kube-state-metrics/pull/1713. Can you rebase your branch against main?

go.mod
```diff
@@ -15,13 +15,16 @@ require (
...
+	github.com/stretchr/testify v1.7.0
 	golang.org/x/perf v0.0.0-20220317191433-96728ec1d780
+	gopkg.in/yaml.v3 v3.0.0-20210107192922-496545a6307b
 	k8s.io/api v0.23.5
 	k8s.io/apimachinery v0.23.5
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.10.0
 	k8s.io/client-go v0.23.5
 	k8s.io/klog/v2 v2.60.0
 	k8s.io/sample-controller v0.23.4
+	sigs.k8s.io/controller-runtime v0.11.1
```
fpetkovski, line 27: Can we use client-go instead of controller-runtime? I think having both libraries in the same project can become challenging to maintain over time. It can also be confusing for contributors about which one to use.

pkg/customresourcestate/registry_factory.go
```diff
@@ -0,0 +1,517 @@
...
+	case int64:
+		v = float64(vv)
+	case uint:
+		v = float64(vv)
+	case uint32:
+		v = float64(vv)
+	case uint64:
+		v = float64(vv)
+	case float32:
+		v = float64(vv)
```
fpetkovski, line 487: Can these cases be merged into one like this?

```suggestion
	case byte, int, int32, int64, uint, uint32, uint64, float32:
		v = float64(vv)
```

pkg/customresourcestate/registry_factory.go
```diff
@@ -0,0 +1,517 @@
...
+		if i > 0 {
+			b.WriteRune(',')
+		}
+		b.WriteString(op.part)
+	}
+	b.WriteRune(']')
+	return b.String()
+}
+
+func compilePath(path []string) (out valuePath, _ error) {
```
fpetkovski, line 338: Can we use something like https://github.com/kubernetes/client-go/blob/master/util/jsonpath/jsonpath.go to reduce or replace this function?

fpetkovski (commented):
I think this would be a great addition to KSM 👍

pkg/customresourcestate/registry_factory.go
iamnoah: I attempted to drop it in, but there are a few problems:

* Bit of an impedance mismatch. jsonpath is expecting to parse a template string, and while I can get at the raw reflect values, a lot of "valid" expressions are going to result in output that doesn't make sense (multiple result values, etc.)
* Numeric handling is too strict. json.Unmarshal(..., map[string]any) parses all numerics as float64. So an expression like `$.Items[?(@.id==3)]` will never match anything because the types won't match. 

I don't love inventing a path expression syntax, but it's simple and flexible and gets the job done. JSONPath seems like overkill.

pkg/customresourcestate/registry_factory.go
iamnoah: This did not work as a multi-type case results in an interface variable that I can't cast to flaot64.

fpetkovski: Ok, thanks for giving it a shot!

muffl0n: Can't this be shortened with `default:`? Or has this the same drawback as the multi-type case?

fpetkovski (commented):
Other than one last nit, I'm okay with merging this as it is. I think we should just announce it as an experimental feature in the changelog and keep an eye on incoming feedback.

@mrueg or @dgrisonnet any thoughts?

main.go
```diff
@@ -20,9 +20,13 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 
 	"github.com/prometheus/common/version"
+	"gopkg.in/yaml.v3"
 	"k8s.io/klog/v2"
```
mrueg, line 27:
```
main.go:27: File is not `goimports`-ed with -local k8s.io/kube-state-metrics (goimports)
	"k8s.io/klog/v2"
```

Can you sort this to make the linter pass?

mrueg:
@iamnoah we're planning for a next release in the upcoming days, are you interested in getting your contribution into that one?

docs/customresourcestate-metrics.md
```diff
@@ -0,0 +1,212 @@
...
```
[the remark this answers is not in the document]
iamnoah: I use the variable as I don't like having to create a ConfigMap, add it as a volume, and mount it just to set a few options, and CLI flags for all the variables would be impossible to read.

I added a flag for the config file location as a middle ground.

I agree something like viper would be an improvement.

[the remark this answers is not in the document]
dgrisonnet: I also think inlining the config would be a better option here. It is always better to have all the configurations in the same place. If we start spreading it, it will just be harder for users to figure out how their apps are configured.

pkg/options/options.go
```diff
@@ -114,7 +115,8 @@ func (o *Options) AddFlags() {
 	o.flags.BoolVarP(&o.Version, "version", "", false, "kube-state-metrics build version information")
 	o.flags.BoolVar(&o.EnableGZIPEncoding, "enable-gzip-encoding", false, "Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.")
 
-	o.flags.StringVar(&o.CustomResourceConfigFile, "custom-resource-state-config", "", "Path to a Custom Resource State Metrics config file")
+	o.flags.StringVar(&o.CustomResourceConfig, "custom-resource-state.config", "", "Inline Custom Resource State Metrics config YAML")
```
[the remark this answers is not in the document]
iamnoah: Sure, I was following the convention from the example @fpetkovski gave.

pkg/options/options.go
```diff
@@ -111,6 +114,9 @@ func (o *Options) AddFlags() {
 	o.flags.StringVar(&o.Namespace, "pod-namespace", "", "Name of the namespace of the pod specified by --pod. "+autoshardingNotice)
 	o.flags.BoolVarP(&o.Version, "version", "", false, "kube-state-metrics build version information")
 	o.flags.BoolVar(&o.EnableGZIPEncoding, "enable-gzip-encoding", false, "Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.")
+
+	o.flags.StringVar(&o.CustomResourceConfig, "custom-resource-state-config", "", "Inline Custom Resource State Metrics config YAML")
```
mrueg, line 118:
```suggestion
	o.flags.StringVar(&o.CustomResourceConfig, "custom-resource-state-config", "", "Inline Custom Resource State Metrics config YAML (experimental)")
```

mrueg (approved):

mrueg:
Thanks for your contribution @iamnoah !

PR #1731, opened 2022-05-09
internal/store: fix metrics slice length

jan--f:
Problem: In https://github.com/kubernetes/kube-state-metrics/pull/1723 a
potential panic in the pod metrics gathering was fixed by working around
a disconnect of `Spec.Containers` and `Status.ContainerStatuses`. The
slice storing the resulting metrics however was still defined based on
the length of the `Status.ContainerStatuses` list.

Solution: Make the slice dynamic and append metrics to it.

Signed-off-by: Jan Fajerski <jfajersk@redhat.com>

Fixes #


fpetkovski:
Thanks. Instead of trying to guess the slice length, I think this is the more sane approach.

/lgtm

honpey:
\cc @jan--f

PR #1732, opened 2022-05-16
Add missing LastSuccessfulTime metric

splitice:

A metric for CronJob LastSuccessfulTime is missing.

This is needed to evaluate cronjob rules (such as backup job must be run every $x hours)

This increases the metrics output per CronJob by one.


fpetkovski (commented):
Thanks, this looks good to me.

/approve

splitice:
@fpetkovski Able to weigh in on that failing unit test? I can't seem to figure it out (must be something simple).

fpetkovski:
I believe the metric type and helptext should be present even when the metric is not exposed (e.g. due to a nil value). So you should be able to fix the test by adding those to the expected output.

splitice:
Ok @fpetkovski, I've done just that.

mrueg:
/lgtm

thanks!

PR #1733, opened 2022-05-16
Fix Unknown and Terminating query examples

honnix:
This fixes the example queries for getting list of `Unknown` or `Terminating` pods.

It seems `pod` has been changed to `podname`, so the examples don't work.

Not applicable.

Not applicable.

### DIFF
--- a/docs/pod-metrics.md
+++ b/docs/pod-metrics.md
@@ -57,9 +57,9 @@ So to mimic the [logic](https://github.com/kubernetes/kubernetes/blob/v1.17.3/pk
 
 For example:
 
-* To get the list of pods that are in the `Unknown` state, you can run the following PromQL query: `sum(kube_pod_status_phase{phase="Unknown"}) by (namespace, pod) or (count(kube_pod_deletion_timestamp) by (namespace, pod) * sum(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod))`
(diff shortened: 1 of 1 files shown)


docs/pod-metrics.md
```diff
@@ -57,9 +57,9 @@ So to mimic the [logic](https://github.com/kubernetes/kubernetes/blob/v1.17.3/pk
 
 For example:
 
-* To get the list of pods that are in the `Unknown` state, you can run the following PromQL query: `sum(kube_pod_status_phase{phase="Unknown"}) by (namespace, pod) or (count(kube_pod_deletion_timestamp) by (namespace, pod) * sum(kube_pod_status_reason{reason="NodeLost"}) by(namespace, pod))`
+* To get the list of pods that are in the `Unknown` state, you can run the following PromQL query: `kube_pod_status_phase{phase="Unknown"} or count(kube_pod_deletion_timestamp) by (namespace, podname) * sum(kube_pod_status_reason{reason="NodeLost"}) by(namespace, podname)`
```
honnix, line 60: Remove the `sum` part of `sum(kube_pod_status_phase{phase="Unknown"})` because the example says to get a list.

fpetkovski: I don't think we have a podname label. Do you have some relabeling configs which might be changing the label in your setup?

https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

honnix: Ah, right. I missed that. I found the replacement in our setup. Thank you.

PR #1740, opened 2022-05-23
Add kube_pod_ips gauge to pod metrics

### LINKED ISSUE #1737: Expose additional pod IPs
**What would you like to be added**: Currently, `kube_pod_info` returns `.status.podIP`, which is usually the pod's IPv4 address. With the stabilization of dual-stack networking, pods will also have an IPv6 address. This is exposed on `.status.podIPs` but doesn't appear to be exposed anywhere in kube-state-metrics.

**Why is this needed**: To be able to record all IPs assigned to a pod for investigation and info.

**Describe the solution you'd like**: It seems to make the most sense to expose a new metric series, call it `kube_pod_ips`:

| Metric name| Metric type | Description | Unit (where applicable) | Labels/tags | Status | Opt-in |
| ---------- | ----------- | ----------- | ----------------------- | ----------- | ------ | ------ |
| kube_pod_ips | Gauge | Pod IP addresses | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `ip`=&lt;pod ip address&gt; <br> `ip_family`=&lt;v4 OR v6=&gt; <br> `uid`=&lt;pod-uid&gt; | EXPERIMENTAL | -

**Additional context**

I'm open to writing this patch, but wanted to put it out here for discussion/comments -- or if I'm missing functionality that doesn't exist.

bjschafer:

Add experimental `kube_pod_ips` metric to show all IPs on a pod. This is useful for dual-stack clusters where pods will have at least one IPv4 and one IPv6 address. Current `kube_pod_info` metric only shows the single IP exposed by `.status.podIP`.

**How does this change affect the cardinality of KSM**: *Increases* (in most cases). Clusters with dual-stack IPv4/IPv6 will mean pods will have two IP addresses instead of one.

Fixes #1737


bjschafer:
Looks like the `ci-go-lint` CI stage is broken in CI. It passes when run locally.

fpetkovski:
This was fixed in https://github.com/kubernetes/kube-state-metrics/pull/1742.

Could you rebase your branch on main to see if it will pass now?

docs/pod-metrics.md
```diff
@@ -4,6 +4,7 @@
 | ---------- | ----------- | ----------- | ----------------------- | ----------- | ------ | ------ |
 | kube_pod_annotations | Gauge | Kubernetes annotations converted to Prometheus labels | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `annotation_POD_ANNOTATION`=&lt;POD_ANNOTATION&gt; <br> `uid`=&lt;pod-uid&gt; | EXPERIMENTAL | -
 | kube_pod_info | Gauge | Information about pod | | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `host_ip`=&lt;host-ip&gt; <br> `pod_ip`=&lt;pod-ip&gt; <br> `node`=&lt;node-name&gt;<br> `created_by_kind`=&lt;created_by_kind&gt;<br> `created_by_name`=&lt;created_by_name&gt;<br> `uid`=&lt;pod-uid&gt;<br> `priority_class`=&lt;priority_class&gt;<br> `host_network`=&lt;host_network&gt;| STABLE | - |
```
mrueg, line 6: For a kube-state-metrics 3.x we should consider removing the pod_ip field here.

bjschafer: I agree with this. Would you consider it in scope for this PR?

internal/store/pod.go
```diff
@@ -580,6 +583,37 @@ func createPodInfoFamilyGenerator() generator.FamilyGenerator {
...
+		"kube_pod_ips",
+		"Pod IP addresses",
+		metric.Gauge,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			ms := make([]*metric.Metric, len(p.Status.PodIPs))
+			labelKeys := []string{"ip", "ip_family"}
+
+			for i, ip := range p.Status.PodIPs {
+				netIP := net.ParseIPSloppy(ip.IP)
```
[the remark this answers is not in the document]
bjschafer: > why did you prefer `ParseIPSloppy` over `ParseIP` here?

I was trying to avoid having to munge import names -- `k8s.io/utils/net` contains `ParseIPSloppy` as well as the `IsIPv4` function. I don't have a strong opinion here though, so if you'd like to see `ParseIP`, I'm happy to make that change.

> Should we handle any errors (e.g. if ip.IP is not an actual IP it will return nil)?

How reliable do we consider the IPs from pod status? I can check for nil and not include it in the returned metric family.

docs/pod-metrics.md
mrueg: No it shouldn't be in scope for this one. More a remark for other maintainers. :)

internal/store/pod.go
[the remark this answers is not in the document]
mrueg: Thanks for the clarification. Let's stick with ParseIPSloppy then and not include in the metric output if ParseIPSloppy returns nil.

fpetkovski:
/lgtm

Thanks for the contribution 👍

Issue #1724, opened 2022-04-25
PV/PVC creation timestamp metrics


aidan-og:
**What would you like to be added**:
Addition of 'created' metric on PV/PVC metrics, using `creationTimestamp` field, similar to existing metrics for other components.

**Why is this needed**:
We have Prometheus alerts which alert if a PV is predicted to fill up in X days. However, disk usage is variable during creation which causes quite a few false positives each time a PV is provisioned for a new resource. 

We want to suppress these alerts until the PVC is >N seconds old. In order to do this, we want to know the creation time of the resource.

**Describe the solution you'd like**
Add `*_created` metric to PVC metrics, similar to other resources.

**Additional context**

dgrisonnet:
@aidan-og this sounds like a good idea, would you be interested in contributing these new metrics?

aidan-og:
@dgrisonnet sure, I'll tackle it!

bavarianbidi:
As #1741 got merged, this issue could be closed i guess?

PR #1741, opened 2022-05-23
Adding Unix creationTimestamp to PV/PVC metrics

aidan-ogorman-dev:
Adding PV/PVC creationTimestamp to metrics

How does this change affect the cardinality of KSM: (increases, decreases or does not change cardinality)
does not change cardinality

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes #

https://github.com/kubernetes/kube-state-metrics/issues/1724


internal/store/persistentvolume.go
```diff
@@ -263,6 +263,27 @@ func persistentVolumeMetricFamilies(allowAnnotationsList, allowLabelsList []stri
...
 		),
+		*generator.NewFamilyGenerator(
+			"kube_persistentvolume_created",
+			"Unix creation timestamp",
+			metric.Gauge,
+			"",
+			wrapPersistentVolumeFunc(func(p *v1.PersistentVolume) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if !p.CreationTimestamp.IsZero() {
```
fpetkovski, line 274: Can `CreationTimestamp` be nil?

docs/persistentvolume-metrics.md
```diff
@@ -1,11 +1,12 @@
 # PersistentVolume Metrics
 
-| Metric name| Metric type | Labels/tags | Status |
```
fpetkovski, line 3: Is this diff large because we're adding two new columns?

internal/store/persistentvolume.go
aidan-ogorman-dev: I think you're right - looks like it is generated before the object is persisted (https://kubernetes.io/docs/reference/using-api/api-concepts/#generated-values). 

This check exists in multiple places - should I update it everywhere in a separate PR?

docs/persistentvolume-metrics.md
aidan-ogorman-dev: Yes

docs/persistentvolume-metrics.md
```diff
@@ -1,11 +1,12 @@
...
+| ------------------------------------ | ----------- | ----------------------- | ----------------------- | ------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------- | ------------ |
```
mrueg, line 4: I believe you can use `-----------` per column and markdown will still render it nicely.

docs/persistentvolume-metrics.md
```diff
@@ -1,11 +1,12 @@
...
+| kube_persistentvolume_annotations    | Gauge       |                         |                         | `persistentvolume`=&lt;persistentvolume-name&gt; <br> `annotation_PERSISTENTVOLUME_ANNOTATION`=&lt;PERSISTENTVOLUME_ANNOTATION&gt;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                        | EXPERIMENTAL |
```
mrueg, line 5: You can omit the additional spacing here, as markdown will render it.

internal/store/persistentvolume.go
fpetkovski: According to those docs, it should actually be be safe to ignore nils. KSM gets objects only after they are persisted, so the creationTimestamp would be set. It's probably why usage of this field in other metrics works without issues

fpetkovski (commented):
Other than @mrueg's comments, this looks good to me.

aidan-ogorman-dev:
Thanks @fpetkovski, I've fixed up the formatting now.

mrueg:
/hold cancel

Thanks for your contribution!

PR #1750, opened 2022-06-07
Switch registry from k8s.gcr.io to registry.k8s.io

mrueg:
Use new domain for k8s image registry, as per https://groups.google.com/a/kubernetes.io/g/dev/c/DYZYNQ_A6_c/m/FpHqeVR2BAAJ

More context: https://github.com/kubernetes/kubernetes/pull/109938

None

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ release.
 #### Container Image
 
 The latest container image can be found at:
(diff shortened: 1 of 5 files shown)


mrueg:
@fpetkovski can you LGTM again? thanks!

PR #1751, opened 2022-06-08
fix: Change the label of kube_pod_nodeselector from nodeselector to n…

yosshi825:
…odeselectors

The docs says nodeselector`s`_NODE_SELECTOR as follows:
> nodeselectors_NODE_SELECTOR=<NODE_SELECTOR>

ref : https://github.com/kubernetes/kube-state-metrics/blob/master/docs/pod-metrics.md

Check the metric in kms v 2.5.0 and the label name is `nodeselector_ NODE _ SELECTOR`

```
# HELP kube_pod_nodeselectors Describes the Pod nodeSelectors.
# TYPE kube_pod_nodeselectors gauge
kube_pod_nodeselectors{namespace="kube-system",pod="node-local-dns-mpm2x",uid="0f954293-170b-4f0c-91f3-bf4b4e334ce2",nodeselector_role="worker"} 1
```

```diff
 # actual  metrics
- nodeselector_NODE_SELECTOR
 # docs description
+ nodeselectors_NODE_SELECTOR
```

unchanging

Fixes #

Fix the parts that were not changed in the following PR
- https://github.com/kubernetes/kube-state-metrics/commit/9a6fae5fc390dd801a74894feacfc42d1b9081e8

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1410,7 +1410,7 @@ func createPodNodeSelectorsFamilyGenerator() generator.FamilyGenerator {
 		metric.Gauge,
 		"",
 		wrapPodFunc(func(p *v1.Pod) *metric.Family {
-			labelKeys, labelValues := kubeMapToPrometheusLabels("nodeselector", p.Spec.NodeSelector)
+			labelKeys, labelValues := kubeMapToPrometheusLabels("nodeselectors", p.Spec.NodeSelector)
 			m := metric.Metric{
 				LabelKeys:   labelKeys,
 				LabelValues: labelValues,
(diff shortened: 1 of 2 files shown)


mrueg:
Thanks for the PR! I guess we should fix the doc instead. I believe the dimensions of a metric usually would be singular.

mrueg:
Thanks for the PR! I guess we should fix the doc instead. I believe the dimensions of a metric usually would be singular.

yosshi825:
@mrueg 
Thank you for your comment
I created the following to fix the documentation https://github.com/kubernetes/kube-state-metrics/pull/1756

yosshi825:
The notation of kube_pod_nodeselector is different in the following part of the test code.
The fix was included in this PR, but should I make another one?

https://github.com/kubernetes/kube-state-metrics/blob/e3bc8c22c21ac091969fa3e4e730311420974bdc/internal/store/pod_test.go#L1879-L1880

https://github.com/kubernetes/kube-state-metrics/blob/e3bc8c22c21ac091969fa3e4e730311420974bdc/internal/store/pod_test.go#L1903-L1904

Issue #1481, opened 2021-05-14
add exitcode to labels of kube_pod_init_container_status_terminated_reason


dafenko:
**What would you like to be added**:
Add exitcode to labels of kube_pod_init_container_status_terminated_reason

**Why is this needed**:
Our application has defined exit codes. We would like to be alerted when app ends up with a specified error code.

**Describe the solution you'd like**
We would like to define alert that uses kube_pod_init_container_status_terminated_reason metric and be alerted when container exists with certain exit code. 

**Additional context**
We are using node.js. Node.js has limit for memory allocation. (see NODE_OPTIONS --max-old-space-size=) 
We intentionally keep limit for node.js a few percent less then K8S resource limit.
In that way node.js has possibility to handle OOM (e.g. create dump file).
POD gets then signal SIGABRT 6 (exit code 128+6 = 134).
We would like to be alerted when such situation happens.

slashpai:
I would like to work on this. I will review and update back

/assign

dannylee-:
Would it be better to create a new metric with the exit code?  Or else we would introduce a label with a large cardinality (256 possible values)

prashantv:
/reopen

Having exit code would be useful for monitoring different types of container exits.

fgiannetti:
> I would like to work on this. I will review and update back
> 
> /assign

Hey @slashpai 
Have some news about this? It would be really useful to have this feature ;)
Or maybe this message will reopen this issue...
Thank you!!

PR #1752, opened 2022-06-08
add exit code

ssabo:

Adds a new metric with the exit code from LastTerminantedStatus. 

A new metric will be added with the same cardinality as `kube_pod_container_status_last_terminated_reason` but with a value of the exit code

Fixes #
https://github.com/kubernetes/kube-state-metrics/issues/1481


fpetkovski:
Thanks for the contribution. As discussed in one of the previous issues, I think this should be a new metric where the exit code is a value. Otherwise the cardinality impact will be very high.

If you need to filter by exit codes, you could always add `== <exit_code>` at the end of a PromQL expression.

ssabo:
Ah I hadn't realized that PromQL supported that grammar. That makes perfect sense to me, I'll make those updates.

internal/store/pod.go
```diff
@@ -334,6 +335,31 @@ func createPodContainerStatusLastTerminatedReasonFamilyGenerator() generator.Fam
...
+		"kube_pod_container_status_last_terminated_exitcode",
+		"Describes the exit code for the last container in terminated state.",
+		metric.Gauge,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			ms := make([]*metric.Metric, 0, len(p.Status.ContainerStatuses))
+			for _, cs := range p.Status.ContainerStatuses {
+				if cs.LastTerminationState.Terminated != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"container", "reason"},
```
ssabo, line 349: Should I keep reason attached to the exit code as a tag? or remove it entirely?

fpetkovski: I think we should remove it because it's already present in a different metric

fpetkovski:
This looks good, but the CI needs to be fixed before we can merge it.

ssabo:
@fpetkovski i'll probably get some some time to look at fixing the CI tests next week. In the meantime, what's the best way to run those locally? I don't think I have the capability to do `/retest` in this project yet and I was running into missing CLI tools when I had tried to invoke them manually. Any way to run them containerized? or just have to install those dependencies?

fpetkovski:
You can check directly what the CI commands are. It seems that [`make test-benchmark-compare`](https://github.com/kubernetes/kube-state-metrics/runs/6873557351?check_suite_focus=true#step:5:2) and [`make doccheck`](https://github.com/kubernetes/kube-state-metrics/runs/6873557547?check_suite_focus=true#step:5:2) are failing.

ssabo:
@fpetkovski sorry for the long delay. Life got busy unexpectedly. Tests should be all fixed up now

fpetkovski (commented):
Thanks

/lgtm

Issue #1624, opened 2021-11-08
kube_endpoint_address_available and kube_endpoint_address_not_ready get computed by kube-state-metrics


bavarianbidi:
**What would you like to be added**:

`kube_endpoint_address_available` and `kube_endpoint_address_not_ready` should be marked as deprecated and replaced in the future by an other metric which only export these information and calculation could be done afterwards.

**Why is this needed**:

> Yes, kube-state-metrics shouldn't be responsible for the computation. As an exporter, its only goal is to expose kubernetes objects data as metrics. Any computation such as getting the total number of ports should be done in promQL.

xref: https://github.com/kubernetes/kube-state-metrics/pull/1571#issuecomment-936327316

**Describe the solution you'd like**

* [ ] `kube_endpoint_address_available` and `kube_endpoint_address_not_ready` should be marked as deprecated
* [ ] new metrics is implemented
* [ ] `kube_endpoint_address_available` and `kube_endpoint_address_not_ready` will be removed

**Additional context**

Similar implementation got rejected in https://github.com/kubernetes/kube-state-metrics/pull/1571 and was reimplemented to do the computation via promQL later on.

bavarianbidi:
#1549 is also doing some changes on this topic

bavarianbidi:
> @dgrisonnet if #1761 got merged, what's the process to mark `kube_endpoint_address_available` and `kube_endpoint_address_not_ready` as deprecated?

+ @fpetkovski :point_up:

mangkoran:
Apologize for the necrobump. The new `kube_endpoint_address` would have cardinality of all IPs in the endpoint subset. This seems to be a major cost compromise if we have service with a lot of endpoints and only want to count the available addresses and readiness of these addresses. Whereas `kube_endpoint_address_available` and `kube_endpoint_address_not_ready` should always have one series per endpoint.

receperdogan:
> Apologize for the necrobump. The new `kube_endpoint_address` would have cardinality of all IPs in the endpoint subset. This seems to be a major cost compromise if we have service with a lot of endpoints and only want to have the number of available addresses and readiness of these addresses. Whereas `kube_endpoint_address_available` and `kube_endpoint_address_not_ready` should always have one series per endpoint. In our case we would like to check for service readiness and `kube_endpoint_address_available` and `kube_endpoint_address_not_ready`seems the better option to keep our cost down.

I agree with @mangkoran. for this reason, we downgraded the version.

PR #1761, opened 2022-06-21
add new kube_endpoint_address metric

bavarianbidi:

As KSM shouldn't be responsible for computation of metrics, a new metric
is needed to expose available and unavailable addresses of an endpoint.

This new metric is generated as a future replacement for
kube_endpoint_address_available and kube_endpoint_address_not_ready.

Increases the cardinality of KSM: sum of all IPs in an endpoint subset (`endpoints.subsets.addresses.ip` and `endpoints.subsets.notReadyAddresses.ip`)

Partially solves #1624

### DIFF
--- a/internal/store/endpoint.go
+++ b/internal/store/endpoint.go
@@ -161,15 +161,15 @@ func endpointMetricFamilies(allowAnnotationsList, allowLabelsList []string) []ge
 				for _, s := range e.Subsets {
 					for _, available := range s.Addresses {
 						ms = append(ms, &metric.Metric{
-							LabelValues: []string{available.IP},
-							LabelKeys:   []string{"available_ip"},
+							LabelValues: []string{available.IP, "true"},
+							LabelKeys:   []string{"ip", "ready"},
 							Value:       1,
 						})
 					}
 					for _, notReadyAddresses := range s.NotReadyAddresses {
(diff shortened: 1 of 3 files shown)


internal/store/endpoint.go
```diff
@@ -151,6 +151,34 @@ func endpointMetricFamilies(allowAnnotationsList, allowLabelsList []string) []ge
...
+			"Information about Endpoint available and non available addresses.",
+			metric.Gauge,
+			"",
+			wrapEndpointFunc(func(e *v1.Endpoints) *metric.Family {
+				ms := []*metric.Metric{}
+				for _, s := range e.Subsets {
+					for _, available := range s.Addresses {
+						ms = append(ms, &metric.Metric{
+							LabelValues: []string{available.IP},
+							LabelKeys:   []string{"available_ip"},
```
fpetkovski, line 165: Should we have a single `ready` label which has `true` or `false` values?

bavarianbidi: > Should we have a single ready label which has true or false values?

yep, make sense.
should we than change `available_ip` and `unavailable_ip` to `ip`?

fpetkovski: Yes, I think making that change would be consistent with the `ready` label.

PR #1799, opened 2022-08-06
Add support for rolebindings

### LINKED ISSUE #1717: Add metrics for role, rolebinding and serviceaccount objects
Add metrics for role, rolebinding and serviceaccount objects 

**Why is this needed**:
In one of my usecases - I need to access label, annotations from role, rolebinding and serviceaccount objects 

Add metrics for role, rolebinding and serviceaccount objects much liek we have for configmap, secret etc and add support for exposing label/annotations. However, please let me know if there's a solution that exists currently as I couldnt find done.

**Additional context**

kaitoii11:
Adds support for generating metrics for `clusterrolebindings` and `rolebindings`

Fixes #1717


tests/manifests/clusterolebinding.yaml
```diff
@@ -0,0 +1,12 @@
+apiVersion: rbac.authorization.k8s.io/v1
```
mrueg, line 1: Can you rename this file, it should be cluster*r*olebinding.yaml ?

pkg/options/resource.go
```diff
@@ -50,6 +51,7 @@ var (
 		"replicationcontrollers":          struct{}{},
 		"resourcequotas":                  struct{}{},
 		"roles":                           struct{}{},
+		"rolebindings":                    struct{}{},
```
mrueg, line 54: Same here, please alphabetically sorted.

examples/standard/cluster-role.yaml
```diff
@@ -112,7 +112,9 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterroles
+  - clusterrolebindings
   - roles
+  - rolesbindings
```
mrueg, line 117:
```suggestion
  - rolebindings
```

examples/autosharding/cluster-role.yaml
```diff
@@ -112,7 +112,9 @@ rules:
   - rbac.authorization.k8s.io
   resources:
   - clusterroles
+  - clusterrolebindings
   - roles
+  - rolesbindings
```
mrueg, line 117:
```suggestion
  - rolebindings
```

internal/store/clusterrolebinding.go
```diff
@@ -0,0 +1,156 @@
...
+				}
+			}),
+		),
+		*generator.NewFamilyGenerator(
+			"kube_clusterrolebinding_info",
+			"Information about clusterrolebinding.",
+			metric.Gauge,
+			"",
+			wrapClusterRoleBindingFunc(func(r *rbacv1.ClusterRoleBinding) *metric.Family {
+				labelKeys := []string{"roleref_kind", "roleref_name"}
```
mrueg, line 85: I assume e2e is failing due to this?

kaitoii11: I had a typo in the document. It should be passing now.

mrueg:
/lgtm

/hold
for @dgrisonnet and @fpetkovski to review.

pkg/options/resource.go
```diff
@@ -27,6 +27,7 @@ var (
 	// DefaultResources represents the default set of resources in kube-state-metrics.
 	DefaultResources = ResourceSet{
 		"certificatesigningrequests":      struct{}{},
+		"clusterrolebindings":             struct{}{},
```
dgrisonnet, line 30: I wonder if we should really have these exposed by default. The use cases for this kind of metric seem really niche to me and I wouldn't want to add even more timeseries to the default setup if it is not really useful to the majority of ksm users. Same comment can be applied to the recently added `role` and `clusterrole` since they are not part of a release yet, we can still afford to remove them from the default if deemed meaningful.

Any thoughts on that @fpetkovski @mrueg?

fpetkovski: I think not exposing them by default makes sense. I agree that the amount of metrics has crept up over time and we can do a better job at providing sane defaults.

mrueg: sounds good to have standard set, should we resolve that in a follow-up PR since it will affect other resources as well?

kaitoii11: So for this PR, should I remove this line?

mrueg: No all good, we will continue in a follow up PR!

mrueg:

PR #1810, opened 2022-08-19
Prevent definition of same gvk in custom resource configuration

chrischdi:
I hope it is okay that I directly created a PR for this. I can also create an issue for that first if this is preferred 👍 

This PR adjusts the options for custom resource configuration to allow defining them multiple times.
This allows to split potentially large configurations to multiple files.

It also adds validation to detect if the configuration includes multiple definitions for the same resource name.
This prevents the following unexpected behaviour:

Configuration:
```yaml
spec:
  resources:
  - groupVersionKind:
      group: cluster.x-k8s.io
      kind: Cluster
      version: v1beta1
    metrics:
    - each:
        info: {}
        type: Info
      name: foo
  - groupVersionKind:
      group: cluster.x-k8s.io
      kind: Cluster
      version: v1beta1
    metrics:
    - each:
        info: {}
        type: Info
      name: bar
```

Expected result:
* either return an error and fail to start (like proposed in the PR)
* or the following metrics:
    ```
    # HELP kube_cluster_x_k8s_io_v1beta1_Cluster_bar
    # TYPE kube_cluster_x_k8s_io_v1beta1_Cluster_bar gauge
    kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
    kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
    # HELP kube_cluster_x_k8s_io_v1beta1_Cluster_foo
    # TYPE kube_cluster_x_k8s_io_v1beta1_Cluster_foo gauge
    kube_cluster_x_k8s_io_v1beta1_Cluster_foo 1
    kube_cluster_x_k8s_io_v1beta1_Cluster_foo 1
    ```

Actual result metrics:

```
# HELP kube_cluster_x_k8s_io_v1beta1_Cluster_bar
# TYPE kube_cluster_x_k8s_io_v1beta1_Cluster_bar gauge
kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
# HELP kube_cluster_x_k8s_io_v1beta1_Cluster_bar
# TYPE kube_cluster_x_k8s_io_v1beta1_Cluster_bar gauge
kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
kube_cluster_x_k8s_io_v1beta1_Cluster_bar 1
```

No changes to cardinality

Fixes #

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -9,12 +9,14 @@ A YAML configuration file described below is required to define your custom reso
 
 Two flags can be used:
 
- * `--custom-resource-state-config "inline yaml (see example)"` or
- * `--custom-resource-state-config-file /path/to/config.yaml`
+* `--custom-resource-state-config "inline yaml (see example)"` or
+* `--custom-resource-state-config-file /path/to/config.yaml`
 
 If both flags are provided, the inline configuration will take precedence.
+When multiple entries for the same resource exist, kube-state-metrics will exit with an error.
+This includes configuration which refers to a different API version.
 
-In addition to specifying one of `--custom-resource-state-config*` flags, you should also add the custom resource *Kind*s in plural form to the list of exposed resources in the `--resources` flag. If you don't specify `--resources`, then all known custom resources configured in `--custom-resource-state-config-*` and all available default kubernetes objects will be taken into account by kube-state-metrics.
+In addition to specifying one of `--custom-resource-state-config*` flags, you should also add the custom resource *Kind*s in plural form to the list of exposed resources in the `--resources` flag. If you don't specify `--resources`, then all known custom resources configured in `--custom-resource-state-config*` and all available default kubernetes objects will be taken into account by kube-state-metrics.
 
 ```yaml
 apiVersion: apps/v1
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -162,8 +162,10 @@ type ConfigDecoder interface {
 }
 
 // FromConfig decodes a configuration source into a slice of customresource.RegistryFactory that are ready to use.
-func FromConfig(decoder ConfigDecoder) (factories []customresource.RegistryFactory, err error) {
+func FromConfig(decoder ConfigDecoder) ([]customresource.RegistryFactory, error) {
 	var crconfig Metrics
+	var factories []customresource.RegistryFactory
+	factoriesIndex := map[string]bool{}
 	if err := decoder.Decode(&crconfig); err != nil {
 		return nil, fmt.Errorf("failed to parse Custom Resource State metrics: %w", err)
 	}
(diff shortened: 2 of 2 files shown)


dgrisonnet:
I have a feeling that this will lead to some unexpected behavior. It is always safer to override the data completely to avoid having data leaking from past configurations.

Generally speaking, I am not a fan of complexifying how configuration is consumed. Reading a configuration file should be really straightforward for a software. I'd rather have the configuration file generated from a data templating language than adding additional complexity to the code.

chrischdi:
> I have a feeling that this will lead to some unexpected behavior. It is always safer to override the data completely to avoid having data leaking from past configurations.
> 
> Generally speaking, I am not a fan of complexifying how configuration is consumed. Reading a configuration file should be really straightforward for a software. I'd rather have the configuration file generated from a data templating language than adding additional complexity to the code.

Seems to be a fair point.

Where I was coming from was: using kustomize to provision kube-state-metrics by using the official helm-chart of kube-state-metrics. When using kustomize I would then have to use some other pre-processing to create the single CRD configuration (e.g. maybe via helm?). My first thought was about having a single config file for each CR I want to produce metrics for and use a single configmap which has multiple files for providing them to the pod. 

However: I'm ok to drop that cli parts 👍

At least the check in https://github.com/kubernetes/kube-state-metrics/pull/1810/files#diff-bf91d16f16c99a97ad09eb6587f27d8fc075a68825d396e035c8100f864a7582R174-R176 should still get added to the `FromConfig` to prevent misconfigurations (duplicates) when using a single file.

mrueg:
Would having CustomResourceStateMetrics CRDs and a way for kube-state-metrics to discover them, solve this problem as well? This might make things way more complex but would allow other deployments to ship their own CRSM config.

chrischdi:
> Would having CustomResourceStateMetrics CRDs and a way for kube-state-metrics to discover them, solve this problem as well? This might make things way more complex but would allow other deployments to ship their own CRSM config.
> 
> > At least the check in https://github.com/kubernetes/kube-state-metrics/pull/1810/files#diff-bf91d16f16c99a97ad09eb6587f27d8fc075a68825d396e035c8100f864a7582R174-R176 should still get added to the `FromConfig` to prevent misconfigurations (duplicates) when using a single file.
> 
> 👍

Yes, that even seems to be the best method to provide the configuration 👍

chrischdi:
Updated to only prevent the configuration issue.

docs/customresourcestate-metrics.md
```diff
@@ -9,12 +9,14 @@ A YAML configuration file described below is required to define your custom reso
...
 Two flags can be used:
 
- * `--custom-resource-state-config "inline yaml (see example)"` or
- * `--custom-resource-state-config-file /path/to/config.yaml`
+* `--custom-resource-state-config "inline yaml (see example)"` or
+* `--custom-resource-state-config-file /path/to/config.yaml`
 
-If both flags are provided, the inline configuration will take precedence.
+Both flags can be provided multiple times to load multiple configurations.
+When multiple entries for the same resource exist, kube-state-metrics will refuse to start.
```
fpetkovski, line 16:
```suggestion
When multiple entries for the same resource exist, kube-state-metrics will exit with an error.
```

fpetkovski (commented):
Thanks, just a small nit

sbueringer:
> Would having CustomResourceStateMetrics CRDs and a way for kube-state-metrics to discover them, solve this problem as well? This might make things way more complex but would allow other deployments to ship their own CRSM config.

@mrueg @chrischdi Would it make sense to track this in a new issue? If I saw correctly, it's currently not tracked in an issue.

PR #1812, opened 2022-08-23
graduate new endpoint metrics to STABLE

### LINKED ISSUE #1624: kube_endpoint_address_available and kube_endpoint_address_not_ready get computed by kube-state-metrics

bavarianbidi:
* graduate kube_endpoint_ports and `kube_endpoint_address` to STABLE
  * `kube_endpoint_address` was introduced in KSM `v2.3.0`
  * `kube_endpoint_ports` was introduced in KSM `v2.5.0` (but IMO it's safe to graduate to STABLE)
* graduate kube_endpoint_address_not_ready and
  kube_endpoint_address_available to DEPRECATED as the information is
precomputed during metrics-scraping

Fixes #1624

### DIFF
--- a/docs/endpoint-metrics.md
+++ b/docs/endpoint-metrics.md
@@ -3,10 +3,10 @@
 | Metric name| Metric type | Labels/tags | Status |
 | ---------- | ----------- | ----------- | ----------- |
 | kube_endpoint_annotations | Gauge | `endpoint`=&lt;endpoint-name&gt; <br> `namespace`=&lt;endpoint-namespace&gt; <br> `annotation_ENDPOINT_ANNOTATION`=&lt;ENDPOINT_ANNOTATION&gt;  | EXPERIMENTAL |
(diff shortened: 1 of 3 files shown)


mrueg:
/approve

Thanks!

PR #1813, opened 2022-08-23
introduce custom-resources-only flag ...

### LINKED ISSUE #1778: Allow monitoring of CRs only

As i'm only interested in monitoring CRs i would have an additional flag e.g. `--resources-crds-only` to not list every CR in the `resources` argument

**Why is this needed**:

I'm currently setting up a second `KSM` instance which is only responsible for CRDs.

Any given GVK in my `helm template` will extend the `ClusterRole` with the required permissions to monitor these objects.
In the current implementation i have to extend the GVK list with a `plural` field to generate and concat the values for the `--resources` flag.

A boolean called `resourcesCrdsOnly` which is `false` per default/if not set and can be set to `true` to only monitor CRs.

logic in `pgk/app/server.go` could look like:

```diff
	var resources []string
-	if len(opts.Resources) == 0 {
+	if len(opts.Resources) == 0 && resourcesCrdsOnly == false {
		klog.Info("Using default resources")
		resources = options.DefaultResources.AsSlice()
		// enable custom resource
		for _, factory := range factories {
			resources = append(resources, factory.Name())
		}
+	} else if resourcesCrdsOnly { 
+		// enable custom resource
+		for _, factory := range factories {
+			resources = append(resources, factory.Name())
+		}
	} else {
		klog.Infof("Using resources %s", opts.Resources.String())
		resources = opts.Resources.AsSlice()
	}
```

**Additional context**

bavarianbidi:
... to only monitor all known custom-resource configurations instead of
listing each of them explicitly

Fixes #1778

### DIFF
--- a/docs/cli-arguments.md
+++ b/docs/cli-arguments.md
@@ -30,6 +30,7 @@ Usage of ./kube-state-metrics:
       --apiserver string                           The URL of the apiserver to use as a master
       --custom-resource-state-config string        Inline Custom Resource State Metrics config YAML (experimental)
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
+      --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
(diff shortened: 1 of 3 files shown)


dgrisonnet:
Agreed, having this flag would be a great addition to ksm.

In v3, we can definitely improve the existing flags to make it easier since the current list kind of grew on us over time, but I wouldn't block that effort on that. It would be great to be able to split ksm and the crd version of ksm already in v2.

mrueg:
@bavarianbidi do you have plans to get back to this change? We could try getting it into the next release.

bavarianbidi:
> @bavarianbidi do you have plans to get back to this change? We could try getting it into the next release.

:wave: sorry for the late reply :see_no_evil: 
@mrueg i've rebased the PR and force pushed again.

mrueg (approved):
/lgtm
/hold

dgrisonnet:
Looking forward to start using this feature, thank you @bavarianbidi :)

/lgtm
/approve
/unhold

PR #1814, opened 2022-08-23
feat: Add local storage labels to kube_persistentvolume_info

nabokihms:
Signed-off-by: m.nabokikh <maksim.nabokikh@flant.com>

Local volumes in Kubernetes differ from other volumes, e.g., they do not have the max cap as other persistent volumes. Instead, local PVs can consume all the available space on the disk. Thus, there should be a way to distinguish local PVs (and other rules to monitor them). 

Yes, it make local storage volumes metrics looks like `kube_persistentvolume_info{local_path="/mnt/data", local_fs="ext4"}`


docs/persistentvolume-metrics.md
```diff
@@ -1,12 +1,12 @@
...
-| kube_persistentvolume_created | Gauge | Unix Creation Timestamp | seconds | `persistentvolume`=&lt;persistentvolume-name&gt; <br> | EXPERIMENTAL |
+| Metric name | Metric type | Description | Unit (where applicable) | Labels/tags                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                             | Status |
```
dgrisonnet, line 3: would you mind reverting the changes made to this file and just add the two labels that you added to the `kube_persistentvolume_info` metric?

nabokihms: Oh, sorry. Yeah, it is some kind of an autoformatting in my IDE.

nabokihms:
The idea to add specific labels to a dedicated metric seems more than ok to me.

Something like:
```
kube_persistentvolume_info{type="csi", persistentvolume="test-1"}
kube_persistentvolume_csi_info{driver="ABC", volume_handle="csi-volume-handle"}
kube_persistentvolume_info{type="local", persistentvolume="test-1"}
kube_persistentvolume_local_info{path="/mnt", fs="ext4"}
```

@dgrisonnet  Does it seem ok to you? If so, I can dig to implement it more generic way.

nabokihms:
I reverted table changes and add host path labels.

nabokihms:
Related https://github.com/dexidp/dex/pull/2562

dgrisonnet:
LGTM, thank you @nabokihms for the contribution :)

/lgtm
/approve

dgrisonnet:
Yeah your suggestion is pretty similar to what I had in mind, I'll start a github discussion around that topic.

nabokihms:
@dgrisonnet I'd like to participate, if it is possible.
😄

PR #1816, opened 2022-08-23
Remove RBAC resources from default exposed metric set

mrueg:
These resources might create a lot of metrics, so we won't enable them by default
for now.

Reduces it.
Fixes #


docs/README.md
```diff
@@ -27,7 +27,9 @@ As of v2.3.0, kube-state-metrics supports additional opt-in metrics via the CLI
 
 ## Exposed Metrics
 
-Per group of metrics there is one file for each metrics. See each file for specific documentation about the exposed metrics:
+Per group of metrics there is one file for each metrics. 
+Some resources (ClusterRole, ClusterRoleBinding, Role, RoleBinding and ServiceAccount) are not exposed by default.
```
dgrisonnet, line 31: These aren't the only metrics not exposed by default, there are also VPA metrics and maybe some others. Maybe we should make two lists, one for metrics exposed by default and another one with the additional ones.

mrueg: Good point, I split it up in two lists.

dgrisonnet:
You also need to update the list inside the `TestDefaultCollectorMetricsAvailable` test to skip the optional collectors:
https://github.com/kubernetes/kube-state-metrics/blob/9063fa6b179219aaea9fc421330067b89620e777/tests/e2e/main_test.go#L270-L273

mrueg:
> You also need to update the list inside the `TestDefaultCollectorMetricsAvailable` test to skip the optional collectors:
> 
> https://github.com/kubernetes/kube-state-metrics/blob/9063fa6b179219aaea9fc421330067b89620e777/tests/e2e/main_test.go#L270-L273

Thanks good catch! I cleaned it up a bit.

PR #1819, opened 2022-08-26
Build with go 1.19 / update to k8s 1.25

mrueg:
Build with go 1.19 and bump the k8s libraries to 1.25.x

None
Fixes #


fpetkovski (commented):
/lgtm

README.md
```diff
@@ -68,13 +68,13 @@ All additional compatibility is only best effort, or happens to still/already be
...
-| **v2.6.0**         |         -/✓          |          -/✓          |          ✓          |          ✓          |          ✓          |
-| **master**         |         -/✓          |          -/✓          |          ✓          |          ✓          |          ✓          |
+| kube-state-metrics |  **Kubernetes 1.21** | **Kubernetes 1.22** | **Kubernetes 1.23** | **Kubernetes 1.24** | **Kubernetes 1.25** |
+|--------------------|:--------------------:|:-------------------:|:-------------------:|:-------------------:|:-------------------:|
+| **v2.3.0**         |          ✓           |          ✓          |          ✓          |          -          |          -          |
+| **v2.4.2**         |         -/✓          |          ✓          |          ✓          |          -          |          -          |
+| **v2.5.0**         |         -/✓          |          ✓          |          ✓          |          -          |          -          |
+| **v2.6.0**         |         -/✓          |          ✓          |          ✓          |          ✓          |          -          |
```
dgrisonnet, line 76: I don't think it is a good idea to add Kubernetes 1.25 to the matrix if we don't have any official versions supporting it. Users shouldn't have to rely on the "unstable" master branch if they upgrade their cluster.

I haven't tested it but I am pretty sure `v2.6.0` is compatible with v1.25 since it is using the v1.24 client, so IMHO we should at least add support for v2.6.0.

PR #1823, opened 2022-09-01
Support filtering label allowlist by "*"

### LINKED ISSUE #1815: Allow all labels/annotations for resources
**Why is this needed**:
As of today, users need to specify allowed resources in two places (if they want to also have label metrics), e.g.:
```
--resources=cronjobs,daemonsets,ingresses
--metric-labels-allowlist=cronjobs=[*],daemonsets=[*],ingresses=[*]
```

I'd like to use asterisk (or another common name, e.g., `apis/all` for the [kube-apiserver](https://kubernetes.io/docs/reference/command-line-tools-reference/kube-apiserver/) --runtime-config flag):
```
--resources=cronjobs,daemonsets,ingresses
--metric-labels-allowlist="*=[*]"
```
or
```
--resources=cronjobs,daemonsets,ingresses
--metric-labels-allowlist="all=[*]"
```

**Additional context**

rexagod:
**What this PR does / why we need it**: Support filtering label allowlist by "*", which will expand to the enabled resources, while inferring their values based on its value(s).

**Which issue(s) this PR fixes**: Fixes #1815.


rexagod:
cc @dgrisonnet @mrueg

.gitignore
```diff
@@ -37,6 +37,7 @@ _testmain.go
 
 *.iml
 .idea/
+.run/
```
dgrisonnet, line 40: since this is not generated by ksm, we should remove it from the gitignore

dgrisonnet (commented):
I think this will be a pretty useful feature, thank you for taking it up @rexagod.

There are a couple of things that would need to be done though before merging this effort:
- Update the description of the command to explain the new behavior: https://github.com/kubernetes/kube-state-metrics/blob/a3a509acdff1f525c3ca2ff76344d0cedafd8a60/pkg/options/options.go#L110 
- Add a new test for the `all` wildcard in https://github.com/kubernetes/kube-state-metrics/blob/a3a509acdff1f525c3ca2ff76344d0cedafd8a60/pkg/options/types_test.go#L197
- Test that the `all` wildcard really returns label metrics for all the resources. This could be an e2e test like the one we have to test the `--resources` option in or a unit test in for the builder: https://github.com/kubernetes/kube-state-metrics/blob/a3a509acdff1f525c3ca2ff76344d0cedafd8a60/tests/e2e/main_test.go#L247

internal/store/builder.go
```diff
@@ -202,7 +202,16 @@ func (b *Builder) WithAllowAnnotations(annotations map[string][]string) {
 // WithAllowLabels configures which labels can be returned for metrics
 func (b *Builder) WithAllowLabels(labels map[string][]string) {
 	if len(labels) > 0 {
-		b.allowLabelsList = labels
+		// "all" takes precedence over other specifications
+		if allowedLabels, ok := labels["all"]; ok {
```
mrueg, line 206: Can we use `*` instead of `all`?

rexagod:
@dgrisonnet @mrueg PTAL.

rexagod:
Ready for review.

internal/store/builder.go
```diff
@@ -203,11 +203,21 @@ func (b *Builder) WithAllowAnnotations(annotations map[string][]string) {
 func (b *Builder) WithAllowLabels(labels map[string][]string) error {
 	if len(labels) > 0 {
 		for label := range labels {
-			if !resourceExists(label) {
+			if !resourceExists(label) && label != "*" {
 				return fmt.Errorf("resource %s does not exist. Available resources: %s", label, strings.Join(availableResources(), ","))
 			}
 		}
 		b.allowLabelsList = labels
+		// "all" takes precedence over other specifications
```
mrueg, line 211:
```suggestion
		// "*" takes precedence over other specifications
```

internal/store/builder.go
```diff
@@ -203,11 +203,21 @@ func (b *Builder) WithAllowAnnotations(annotations map[string][]string) {
...
 		}
 		b.allowLabelsList = labels
+		// "*" takes precedence over other specifications
+		if allowedLabels, ok := labels["*"]; ok {
+			m := make(map[string][]string)
+			for _, resource := range b.enabledResources {
+				m[resource] = allowedLabels
+			}
+			b.allowLabelsList = m
+		} else {
```
mrueg, line 218: Is this `else` path needed? It's already set before in line 210

rexagod: Must've got in while resolving conflicts during the rebase, fixed.

mrueg (approved):
/lgtm 
/hold
Thanks!

dgrisonnet:
Awesome, great stuff @rexagod!

/lgtm
/approve

/unhold

PR #1825, opened 2022-09-02
Validate metrics

### LINKED ISSUE #1793: Streamline Metrics
Define standards and look through all metrics and check if the dimensions match certain standards (e.g. all plural / singular)

rexagod:
**What this PR does / why we need it**: Validate metrics that are served over the metrics endpoint using promtool in the e2e test flow.

**Which issue(s) this PR fixes**: Fixes #1793

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -214,4 +214,4 @@ jobs:
 
     - name: End-to-end tests
       run: |
-        make e2e
+        CHECK_METRICS=true make e2e
--- a/.gitignore
+++ b/.gitignore
@@ -38,7 +38,8 @@ _testmain.go
 *.iml
 .idea/
 
-# Generated CLI help file
+# Generated artefacts
+promtool
 help.txt
 
 # jsonnet dependency management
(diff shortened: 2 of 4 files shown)


rexagod:
cc @dgrisonnet @mrueg

dgrisonnet:
Since this is a step of e2e testing and we want to get rid of shell scripts as much as possible in kube-state-metrics, I would be in favor of doing the check in a Golang test with the promlint library rather than using the tool directly: https://godocs.io/github.com/prometheus/client_golang/prometheus/testutil/promlint.

This is essentially what `promlint check metrics` is doing internally: https://github.com/prometheus/prometheus/blob/4927e1353786c89e847e39c41a05a6f8ee464465/cmd/promtool/main.go#L713-L748

rexagod:
I just looked at the e2e suite, and it seems we are already doing this: https://github.com/kubernetes/kube-state-metrics/blob/master/tests/e2e/main_test.go#L78-L95?

dgrisonnet:
Then I guess it was already done and we somehow forgot :sweat_smile:, sorry about that @rexagod.

PR #1826, opened 2022-09-02
Handle singular labels in allowlist

### LINKED ISSUE #1786: Improve --metric-labels-allowlist handling of non-plural resource names
`--metric-labels-allowlist` to support singular form `pod` vs `pods` for resources or warn when user uses singular form of common, [predefined resources](https://github.com/kubernetes/kube-state-metrics/blob/335e9e1235150fb62ee965b3ba15ce0f7dc769d8/internal/store/builder.go#L263-L296).

**Why is this needed**:
Currently, it is trivial to make a mistake when configuring KSM's labels allowlist and provide a singular form of resource name `pod`, `namespace` etc. This would not result in any error or warning, but also will not make KSM capture the metrics. This behaviour can confuse the end user.

Option 1: Support both singular and plural forms.
Option 2: Warn when singular form is discovered.

**Additional context**
Happy to code the solution once we agree on the approach to take :)

rexagod:
**What this PR does / why we need it**: Handle singular labels in allowlist failing when such a label is supplied, in order to keep the behaviour in sync with --resources.

**Which issue(s) this PR fixes**: Fixes #1786

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -200,10 +200,16 @@ func (b *Builder) WithAllowAnnotations(annotations map[string][]string) {
 }
 
 // WithAllowLabels configures which labels can be returned for metrics
-func (b *Builder) WithAllowLabels(labels map[string][]string) {
+func (b *Builder) WithAllowLabels(labels map[string][]string) error {
 	if len(labels) > 0 {
+		for label := range labels {
+			if !resourceExists(label) {
+				return fmt.Errorf("resource %s does not exist. Available resources: %s", label, strings.Join(availableResources(), ","))
+			}
+		}
 		b.allowLabelsList = labels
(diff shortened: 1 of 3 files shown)


rexagod:
cc @dgrisonnet @mrueg

.gitignore
```diff
@@ -37,6 +37,7 @@ _testmain.go
 
 *.iml
 .idea/
+.run
```
dgrisonnet, line 40: This is not generated by the kube-state-metrics repository so I wouldn't want to add it to the repository. If this is annoying in your environment I would recommend using a global gitignore.

dgrisonnet (commented):
@rexagod could you also add a test verifying that this works as intended?

Issue #1791, opened 2022-07-19
Replace CLI toolkit


mrueg:
Currently we're using pflag to parse arguments. This comes with a certain simplicity, a more sophisticated toolkit like viper/cobra or others would allow users to use other methods to configure KSM as well.
In particular an idea, that came up for the CustomResource Collector is to use EnvironmentVariables or a Configuration File based approach.

rexagod:
If I understand the ask here correctly,  `--custom-resource-state-config-file`allows you to do just that (file-based input).

mrueg:
The request here is to look into https://github.com/spf13/viper or similar libraries that have built-in support for a likely yaml-based config.

rexagod:
This issue is divided into the following sub-issues (after discussing with @dgrisonnet).
* Incorporate `cobra`: https://github.com/kubernetes/kube-state-metrics/pull/1834
* Incorporate `viper`: https://github.com/kubernetes/kube-state-metrics/pull/1827
* Future plans (after v3.0): https://github.com/kubernetes/kube-state-metrics/issues/1738#issuecomment-1241802049

mrueg:
I think this issue is resolved. Thanks for your work @rexagod

PR #1827, opened 2022-09-04
Incorporate Viper

### LINKED ISSUE #1738: Ability to configure --metric-allowlist and --metric-denylist after startup
Ability to configure --metric-allowlist and --metric-denylist after startup. Today, we can only configure these as cli arguments at startup.

**Why is this needed**:
We need this functionality because we are installing kube-state-metrics as a dependent chart and we will need to redeploy the container if these settings have to be changed.

One of the ways we could configure this after startup is by having a configmap which could contain this setting and which is mounted as a volume to the container. If there is any change to the configmap a liveness probe can be configured to restart the container to pick up the new settings.

rexagod:
**What this PR does / why we need it**: Add `--config` flag to support supplying options from a config file, and hot-reload on any changes to it.

**Which issue(s) this PR fixes**: Fixes #1738


mrueg:
/hold
I would suggest to resolve this as part of https://github.com/kubernetes/kube-state-metrics/issues/1791

dgrisonnet:
I agree but I would still prefer to have the efforts split into different PRs since it will be easier to review.

1. Update CLI toolkit
2. Add config-file option
3. Implement config reloading

rexagod:
ACK @dgrisonnet, I'll break this PR up.

rexagod:
@dgrisonnet Split the commits, PTAL. cc @mrueg @fpetkovski

pkg/app/server.go
```diff
@@ -66,10 +70,20 @@ func (pl promLogger) Log(v ...interface{}) error {
 	return nil
 }
 
-// RunKubeStateMetrics will build and run the kube-state-metrics.
+// RunKubeStateMetricsWrapper runs KSM with context cancellation.
+func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
+	err := runKubeStateMetrics(ctx, opts, factories...)
+	if ctx.Err() == context.Canceled {
+		klog.Infoln("RESTARTING kube-state-metrics")
```
rexagod, line 77: Decided to make this uppercase, to make it sort of "pop out" in the sea of logs, since I thought it made sense for the reader to be able to know where the older logs ended and where the newer ones started from.

rexagod:
Ready for review.

rexagod:
Will rebase once https://github.com/kubernetes/kube-state-metrics/pull/1834 is in.

docs/cli-arguments.md
```diff
@@ -60,6 +60,7 @@ Flags:
       --namespaces-denylist string                 Comma-separated list of namespaces not to be enabled. If namespaces and namespaces-denylist are both set, only namespaces that are excluded in namespaces-denylist will be used.
       --node string                                Name of the node that contains the kube-state-metrics pod. Most likely it should be passed via the downward API. This is used for daemonset sharding. Only available for resources (pod metrics) that support spec.nodeName fieldSelector. This is experimental.
       --one_output                                 If true, only write logs to their native severity level (vs also writing to each lower severity level; no effect when -logtostderr=true)
+      --options-config-file string                 Path to the kube-state-metrics options config file
```
mrueg, line 63: Can we rename it to `--config` ? makes it easier to discover as a user.

mrueg: (I know there are different config files for custom resource metrics, but this is probably the main config)

mrueg:
Nice work!

A couple of questions and suggestions:

* When changing the config file, will metrics be reset as the ksm routine restarts? If so, maybe this would be a good additional log message.
* The reloading features uses fsnotify, is this polling or using inotify? Do we need to set any poll rate here? 
* What happens if I change it to a bad config? Will it continue running with the old version? If this is expected behavior, maybe we can implement additional metrics as below and send some warning log lines about the bad config?


Alertmanager uses these, so maybe it would be helpful to have similar metrics in ksm:
```
# HELP alertmanager_config_hash Hash of the currently loaded alertmanager configuration.
# TYPE alertmanager_config_hash gauge
alertmanager_config_hash 2.58844882410869e+14
# HELP alertmanager_config_last_reload_success_timestamp_seconds Timestamp of the last successful configuration reload.
# TYPE alertmanager_config_last_reload_success_timestamp_seconds gauge
alertmanager_config_last_reload_success_timestamp_seconds 1.592489682e+09
# HELP alertmanager_config_last_reload_successful Whether the last configuration reload attempt was successful.
# TYPE alertmanager_config_last_reload_successful gauge
alertmanager_config_last_reload_successful 1
```



Another item (likely outside of this PR and once this is implemented):
We should add the reloading feature for custom-state-config file as well.

rexagod:
> When changing the config file, will metrics be reset as the ksm routine restarts? If so, maybe this would be a good additional log message.

ACK. Also, there's a suggestion (https://github.com/kubernetes/kube-state-metrics/issues/1791#issuecomment-1247326376) to selectively update/invalidate the cache on-the-fly, instead of restarting KSM down-the-line. I'll open an issue to track this effort.

> The reloading features uses fsnotify, is this polling or using inotify? Do we need to set any poll rate here?

As it is right now, `fsnotify` [continuously](https://github.com/fsnotify/fsnotify/blob/0d1c8342fcec1c0e2ff5ae2e460fdaf908fd2188/backend_fen.go#L321) watches for events, as there is [no polling support](https://github.com/fsnotify/fsnotify#readme) at the moment: https://github.com/fsnotify/fsnotify/issues/9.

> What happens if I change it to a bad config? Will it continue running with the old version?

If there's an issue with unmarshalling the config, KSM will report the error but won't end it's process, otherwise if it's a option-specific issue, it'll error out, same way if we specified a bad value to an option through an argument to the binary.

rexagod:
Regarding the alert manager metrics above, I'm absolutely +1, now that we have hot-reloading. I'll open an issue for this as well. :)

pkg/app/server.go
```diff
@@ -66,10 +70,20 @@ func (pl promLogger) Log(v ...interface{}) error {
 	return nil
 }
 
-// RunKubeStateMetrics will build and run the kube-state-metrics.
+// RunKubeStateMetricsWrapper runs KSM with context cancellation.
+func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
+	err := runKubeStateMetrics(ctx, opts, factories...)
+	if ctx.Err() == context.Canceled {
+		klog.Infoln("RESTARTING kube-state-metrics, metrics will be reset")
```
fpetkovski, line 77:
```suggestion
		klog.Infoln("Restarting: kube-state-metrics, metrics will be reset")
```

pkg/app/server.go
```diff
@@ -66,10 +70,20 @@ func (pl promLogger) Log(v ...interface{}) error {
...
+		return nil
+	}
+	return err
+}
+
+// runKubeStateMetrics will build and run the kube-state-metrics.
 // Any out-of-tree custom resource metrics could be registered by newing a registry factory
 // which implements customresource.RegistryFactory and pass all factories into this function.
-func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
+func runKubeStateMetrics(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
```
fpetkovski, line 86: KSM is often used as a library by external projects. It's better to keep this method exporter as it was.

main.go
```diff
@@ -63,11 +65,37 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
+			klog.ErrorS(err, "Failed to run kube-state-metrics")
+			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
+		}
+	}
+	ctx, cancel := context.WithCancel(context.Background())
+	if file := options.GetOptsConfigFile(*opts); file != "" {
+		viper.SetConfigType("yaml")
+		viper.SetConfigFile(file)
+		if err := viper.ReadInConfig(); err != nil {
+			if _, ok := err.(viper.ConfigFileNotFoundError); ok {
```
fpetkovski, line 79: We can use `errors.Is` for better detection if it ends up being wrapped at one point.

pkg/app/server.go
fpetkovski: Does this mean the container will shut down when the config file is changed?

rexagod: No the container will stay up, only the context will be cancelled, and a new goroutine will be created in place of the older one, which will be destroyed.

fpetkovski (commented):
Nice work, I like the approach. I would suggest adding a test for this behavior since it can easily be broken by future changes.

go.mod
```diff
@@ -4,6 +4,7 @@ require (
 	github.com/brancz/gojsontoyaml v0.1.0
 	github.com/campoy/embedmd v1.0.0
 	github.com/dgryski/go-jump v0.0.0-20211018200510-ba001c3ffce0
+	github.com/fsnotify/fsnotify v1.5.4
```
mrueg, line 7: Can we use a later version here (v1.6.0+)?
```
inotify: replace epoll() with non-blocking inotify (https://github.com/fsnotify/fsnotify/pull/434)

Non-blocking inotify was not generally available at the time this library was written in 2014, but now it is. As a result, the minimum Linux version is bumped from 2.6.27 to 2.6.32. This hugely simplifies the code and is faster.```

See: https://github.com/fsnotify/fsnotify/releases/tag/v1.6.0

File: go.mod [hunk 7]
```diff
@@ -15,6 +16,7 @@ require (
 	github.com/prometheus/exporter-toolkit v0.8.1
 	github.com/robfig/cron/v3 v3.0.1
 	github.com/spf13/cobra v1.6.1
+	github.com/spf13/viper v1.13.0
```
mrueg, line 19: This can be 1.14.0

docs/node-metrics.md
```diff
@@ -12,4 +12,4 @@
 | kube_node_status_allocatable | Gauge | The allocatable for different resources of a node that are available for scheduling | `cpu`=&lt;core&gt; <br> `ephemeral_storage`=&lt;byte&gt; <br> `pods`=&lt;integer&gt; <br> `attachable_volumes_*`=&lt;byte&gt; <br> `hugepages_*`=&lt;byte&gt; <br> `memory`=&lt;byte&gt; |`node`=&lt;node-address&gt; <br> `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt;| STABLE |
 | kube_node_status_condition | Gauge | The condition of a cluster node | |`node`=&lt;node-address&gt; <br> `condition`=&lt;node-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; | STABLE |
 | kube_node_created | Gauge | Unix creation timestamp | seconds |`node`=&lt;node-address&gt;| STABLE |
-| kube_node_deletion_timestamp | Gauge | Unix creation timestamp | seconds |`node`=&lt;node-address&gt;| STABLE |
+| kube_node_deletion_timestamp | Gauge | Unix creation timestamp | seconds |`node`=&lt;node-address&gt;| EXPERIMENTAL |
```
rexagod, line 15: I missed this: https://github.com/kubernetes/kube-state-metrics/pull/1890/files#diff-380eca5a922c0ddbf67f04daefc6823e7ef0e197434d3a826d39c7063cdfa6d6R15.

rexagod:
Test added here: [`60a7373` (#1827)](https://github.com/kubernetes/kube-state-metrics/pull/1827/commits/60a7373c95c6886d61fabc037e62b97e3f43f6e0) (3rd last fixup).

rexagod:
Tidying up the commits.

mrueg:
/hold cancel
/lgtm

Thanks for this contribution @rexagod !

PR #1831, opened 2022-09-13
add ContainerResourceSourceType for hpa metrics

whitebear009:
kubernetes hpa support ContainerResourceSourceType(https://kubernetes.io/docs/tasks/run-application/horizontal-pod-autoscale/#container-resource-metrics) since v1.20, but there is no metrics about it in the kube-state-metrics.
The PR add ContainerResourceSourceType for hpa metrics kube_horizontalpodautoscaler_spec_target_metric and kube_horizontalpodautoscaler_status_target_metric.
Then decrease cyclomatic complexity by reusing metricTarget objects, so as to pass golint-ci.

Increases cardinality.  Add metrics for ContainerResourceSourceType hpa.

Fixes #

### DIFF
--- a/internal/store/horizontalpodautoscaler.go
+++ b/internal/store/horizontalpodautoscaler.go
@@ -131,46 +131,39 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 				ms := make([]*metric.Metric, 0, len(a.Spec.Metrics))
 				for _, m := range a.Spec.Metrics {
 					var metricName string
-
+					var metricTarget autoscaling.MetricTarget
 					// The variable maps the type of metric to the corresponding value
 					metricMap := make(map[metricTargetType]float64)
 
 					switch m.Type {
 					case autoscaling.ObjectMetricSourceType:
 						metricName = m.Object.Metric.Name
-
-						if m.Object.Target.Value != nil {
-							metricMap[value] = float64(m.Object.Target.Value.MilliValue()) / 1000
-						}
-						if m.Object.Target.AverageValue != nil {
-							metricMap[average] = float64(m.Object.Target.AverageValue.MilliValue()) / 1000
-						}
+						metricTarget = m.Object.Target
 					case autoscaling.PodsMetricSourceType:
 						metricName = m.Pods.Metric.Name
-
-						metricMap[average] = float64(m.Pods.Target.AverageValue.MilliValue()) / 1000
+						metricTarget = m.Pods.Target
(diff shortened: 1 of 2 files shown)


whitebear009:
/assign @mrueg
Please take a look. Thanks :)

internal/store/horizontalpodautoscaler.go
```diff
@@ -193,48 +186,41 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
...
-						if m.Resource.Current.AverageUtilization != nil {
-							metricMap[utilization] = float64(*m.Resource.Current.AverageUtilization)
-						}
-
-						if m.Resource.Current.AverageValue != nil {
-							metricMap[average] = float64(m.Resource.Current.AverageValue.MilliValue()) / 1000
-						}
+						currentMetric = m.Resource.Current
+					case autoscaling.ContainerResourceMetricSourceType:
+						metricName = string(m.ContainerResource.Name)
```
dgrisonnet, line 204: this is already a string

internal/store/horizontalpodautoscaler.go
```diff
@@ -131,48 +131,41 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
...
-						if m.Resource.Target.AverageUtilization != nil {
-							metricMap[utilization] = float64(*m.Resource.Target.AverageUtilization)
-						}
-
-						if m.Resource.Target.AverageValue != nil {
-							metricMap[average] = float64(m.Resource.Target.AverageValue.MilliValue()) / 1000
-						}
+						metricTarget = m.Resource.Target
+					case autoscaling.ContainerResourceMetricSourceType:
+						metricName = string(m.ContainerResource.Name)
```
dgrisonnet, line 149: this is already a string

dgrisonnet:
@whitebear009 my previous comments are still unresolved, could you have a look at them?

internal/store/horizontalpodautoscaler.go
whitebear009: I am afraid of not. It's type corev1.ResourceName, like `case autoscaling.ResourceMetricSourceType`. Refer to https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/autoscaling/types.go#L296

internal/store/horizontalpodautoscaler.go
whitebear009: the same as above.

whitebear009:
@dgrisonnet My reply was not sent before, sorry.  please check again

Issue #1718, opened 2022-04-15
Support verticalpodautoscaler v1


iamcan13:
**What would you like to be added**:
Support verticalpodautoscalers/v1

**Why is this needed**:
Since vertical pod autoscalers CRD has been updated from v1beta2 to v1, kube-state-metrics should support new version of API. ([release note](https://github.com/kubernetes/autoscaler/releases/tag/vertical-pod-autoscaler-0.10.0))

**Describe the solution you'd like**
Generate client codes from version `v1` and modify the dependencies.
The legacy client dependency is `v1beta2`, below is one of them I found.
https://github.com/kubernetes/kube-state-metrics/blob/41eea36f69efd9824836089aae67d308111f4e01/internal/store/builder.go#L38

**Additional context**

Serializator:
@fpetkovski, is this something to be looked into or would the quote below from the README negate this issue?
>Resources in Kubernetes can evolve, i.e., the group version for a resource may change from alpha to beta and finally GA in different Kubernetes versions. For now, kube-state-metrics will only use the oldest API available in the latest release.

I'm not too familiar with the compatibility and versioning of KSM. The quote says "will only use the oldest API available in the latest release", what does this mean exactly?

Would this mean if `autoscaling.k8s.io/v1beta2` is still available in Kubernetes 1.24 then KSM will keep using `v1beta2` until this API version is removed in one of the next Kubernetes releases and then start using `autoscaling.k8s.io/v1`?

fpetkovski:
KSM aims to have backwards compatibility with the latest 3 Kubernetes versions. However, VPA is an external CRD so I'm not sure if we should apply the same rules since the CRD is not linked to Kubernetes releases. I think it should be safe to switch to `v1` and users that use the old API can upgrade VPA independently of Kubernetes.

@mrueg @dgrisonnet any thoughts from your side?

dgrisonnet:
Adding VPA to kube-state-metrics was a mistake since it is out of its scope to only expose metrics about the core resources of Kubernetes. This leads to the problem we have here where we can't really move forward with an update since the API isn't tied to Kubernetes directly.

In my opinion, we should slowly deprecate it in favor of adding it as a configuration-based CRD. In the meantime, I think we shouldn't break the existing compatibility and stick to v1beta2.

As for the deprecation of VPA, I would recommend doing it over 2 minor releases of ksm: 

v2.6.0:
[...]

PR #1835, opened 2022-09-15
Deprecate VPA

### LINKED ISSUE #1790: Remove Verticalpodautoscaler Collector (VPA)
As we have now an implementation to collect metrics from Custom Resources,  we want to remove the VPA Collector in KSM v3.0

rexagod:
- [x] Deprecate VPA metrics.
- [x] Inform users about the deprecation (print a warning to `stdout`).
- [x] Document an alternative approach using the CRD capabilities we have now.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

**How does this change affect the cardinality of KSM**: No change (VPA metrics will be removed in v2.9.0).

**Which issue(s) this PR fixes**: Partially targets #1718.

### DIFF
--- a/docs/verticalpodautoscaler-metrics.md
+++ b/docs/verticalpodautoscaler-metrics.md
@@ -1,16 +1,53 @@
 # Vertical Pod Autoscaler Metrics
 
-| Metric name                                                                | Metric type | Labels/tags                                                                                                                                                                                                                                                | Status                                                                                                                                                      |
-| --------------------------------                                           | ----------- | -------------------------------------------------------------                                                                                                                                                                                              | ------                                                                                                                                                      |
-| kube_verticalpodautoscaler_annotations                                          | Gauge       | `annotation_app`=&lt;foo&gt; <br> `namespace`=&lt;namespace&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt;   | EXPERIMENTAL                                                                                                                                                |
-| kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_minallowed                   | Gauge       | `container`=&lt;container name&gt; <br> `namespace`=&lt;namespace&gt; <br> `resource`=&lt;cpu memory&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `unit`=&lt;core byte&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt;                | EXPERIMENTAL |
-| kube_verticalpodautoscaler_spec_resourcepolicy_container_policies_maxallowed                   | Gauge       | `container`=&lt;container name&gt; <br> `namespace`=&lt;namespace&gt; <br> `resource`=&lt;cpu memory&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `unit`=&lt;core byte&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt;                | EXPERIMENTAL |
-| kube_verticalpodautoscaler_status_recommendation_containerrecommendations_lowerbound     | Gauge       | `container`=&lt;container name&gt; <br> `namespace`=&lt;namespace&gt; <br> `resource`=&lt;cpu memory&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `unit`=&lt;core byte&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt;                | EXPERIMENTAL |
-| kube_verticalpodautoscaler_status_recommendation_containerrecommendations_target          | Gauge       | `container`=&lt;container name&gt; <br> `namespace`=&lt;namespace&gt; <br> `resource`=&lt;cpu memory&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `unit`=&lt;core byte&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt;                | EXPERIMENTAL |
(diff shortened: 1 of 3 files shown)


fpetkovski (approved):
Thanks.

/lgtm
/hold

internal/store/builder.go
```diff
@@ -95,6 +96,18 @@ func (b *Builder) WithEnabledResources(r []string) error {
 		if !resourceExists(col) {
 			return fmt.Errorf("resource %s does not exist. Available resources: %s", col, strings.Join(availableResources(), ","))
 		}
+
+		deprecationNoticeHeader := "DEPRECATION NOTICE"
```
dgrisonnet, line 100: Instead of adding the deprecation notice here, we should reuse the existing mechanism in the family generators: https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/metric_generator/generator.go#L30-L37

docs/verticalpodautoscaler-metrics.md
```diff
@@ -1,16 +1,37 @@
...
+        group: autoscaling.k8s.io
+        kind: "VerticalPodAutoscaler"
+        version: "v1"
+      metrics:
+        - name: "generation"
+          help: "VPA Generation"
+          each:
+            type: Gauge
+            gauge:
+              path: [metadata, generation]
```
dgrisonnet, line 20: Could we provide them with a manifest that generates exactly the same metrics as we had before? I know that we won't be able to get the exact same names so it would get great to also point the name changes.

docs/verticalpodautoscaler-metrics.md
```diff
@@ -1,16 +1,37 @@
...
-| kube_verticalpodautoscaler_spec_updatepolicy_updatemode                                     | Gauge       | `namespace`=&lt;namespace&gt; <br> `target_api_version`=&lt;api version&gt; <br> `target_kind`=&lt;target kind&gt; <br> `target_name`=&lt;target name&gt; <br> `update_mode`=&lt;foo&gt; <br> `verticalpodautoscaler`=&lt;vertical pod autoscaler name&gt; | EXPERIMENTAL                                                                                                                                                |
+## DEPRECATION NOTICE
+
+From v2.8.0 onwards, `vericalpodautoscalers` will be removed from the list of default resources. This means that specifying that in the `--resource` flag will **not** generate metrics for the same. In order to generate `verticalpodautoscalers` metrics, you will have to explicitly specify it in `--custom-resource-state-config*` (either the inline yaml, or the configuration file), like so:
```
dgrisonnet, line 5: Since v2.6.0 has already been released, we should only start removing VPA in v2.9.0 since the deprecation period should be of at least 2 releases.

mrueg: IMHO, we should bump version to v3 if we remove anything.

dgrisonnet:
The unit tests are failing, you need to the update the test metadata to include the deprecation notice: https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/verticalpodautoscaler_test.go#L33-L40

docs/verticalpodautoscaler-metrics.md
dgrisonnet: Breaking changes used to be reserved for major releases when following semantic versioning, but what we've seen in recent years is that projects with only minor releases are healthy and can still handle breaking changes properly as long as they have clear deprecation policies, which is the case for Kubernetes for example. IMHO major releases should only be reserved for major rewrites, which don't include VPA metrics removal. If we are not confident that the deprecation period is enough, I am fine with increasing it, but I don't think waiting for a major release is the right call here.

What do you think @fpetkovski?

mrueg: I can understand that major releases can become quite painful with go versioning. Personally, I prefer going with the least surprise for users. Users won't be happy when their VPA metrics are gone and they did not see it coming.

rexagod:
@dgrisonnet @mrueg Moved to v3, PTAL.

rexagod:
Moved to v3, ready for review.

rexagod:
cc @dgrisonnet @mrueg @logicalhan

docs/verticalpodautoscaler-metrics.md
```diff
@@ -1,16 +1,53 @@
...
+        target_name: [spec, targetRef, name]
+      metrics:
+        - name: "annotations"
+          help: "(Deprecated since v3.0.0) Kubernetes annotations converted to Prometheus labels."
+          each:
+            type: Gauge
+            gauge:
+              path: [metadata, annotations]
+# This will output the following metric:
+# HELP kube_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations (Deprecated since v3.0.0) Kubernetes annotations converted to Prometheus labels.
```
logicalhan, line 33: is this output still correct given your change with the CRD metrics in your other PR? Shouldn't this be `kube_cred_annotations`?

docs/verticalpodautoscaler-metrics.md
dgrisonnet: The problem is that there is no reason to cut a new major today besides removing the VPA metrics, which is not a big reason to have a new major version just for. Leaving a long enough deprecation period should already be good enough to not surprise the users. At the very least if we are not confident enough that our users are going through our changelogs, we can add a log line when the vpa store is used saying that it is deprecated and will be removed in version x.y.z.

dgrisonnet:
I am still very much against cutting a new major version of ksm just for deprecating VPA metrics. This is not an overhaul of the codebase, we are just deprecating some metrics and replacing them with new ones. See it as a renaming of metrics rather than a complete removal.

Also, it is worth noting that all the VPA metrics are experimental so although they have been in the codebase for a while, users shouldn't expect them to never change.

Holding until we've reached a consensus.
/hold

logicalhan:
> I am still very much against cutting a new major version of ksm just for deprecating VPA metrics. This is not an overhaul of the codebase, we are just deprecating some metrics and replacing them with new ones. See it as a renaming of metrics rather than a complete removal.
> 
> Also, it is worth noting that all the VPA metrics are experimental so although they have been in the codebase for a while, users shouldn't expect them to never change.
> 
> Holding until we've reached a consensus. /hold

Yeah, cutting a release for deleting experimental metrics seems a bit much to me too.

docs/verticalpodautoscaler-metrics.md
logicalhan: Experimental features should not have stability guarantees.

mrueg: Apologies here, I wasn't aware that VPA never made it stable. I'm fine with deprecating and removing without bumping major here. Thanks for explaining!

rexagod:
Reverted to `v2.9.0`.

docs/verticalpodautoscaler-metrics.md
```diff
@@ -1,16 +1,53 @@
...
+        version: "v1"
+      labelsFromPath:
+        verticalpodautoscaler: [metadata, name]
+        namespace: [metadata, namespace]
+        target_api_version: [apiVersion]
+        target_kind: [spec, targetRef, kind]
+        target_name: [spec, targetRef, name]
+      metrics:
+        - name: "annotations"
+          help: "(Deprecated since v2.9.0) Kubernetes annotations converted to Prometheus labels."
```
mrueg, line 27: Should this should this say "Deprecated since v2.7.0, will be removed in v2.9.0" instead?

rexagod:
@mrueg Can you please `/lgtm` (if all looks good here)?

mrueg:
/lgtm
/hold cancel
Proceeding here, thanks for your contribution @rexagod!

dgrisonnet:
There is a doc explaining how to get them back in https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#verticalpodautoscaler

Issue #1833, opened 2022-09-14
Presubmit checks for stable metrics


CatherineF-dev:
**What would you like to be added**:
A presubmit to validate stable metrics don't change metric name, add labels or delete labels. 

Similar to [k8s stable metrics presubmit](https://github.com/kubernetes/kubernetes/tree/master/test/instrumentation).

**Why is this needed**:
Guarantee stable metrics are stable.

Due to lack of presubmit check, some stable metrics are changed. For example, 
- stable metric `kube_persistentvolume_info` is changed in https://github.com/kubernetes/kube-state-metrics/pull/1814
- stable metric `kube_service_status_load_balancer_ingress` is changed in https://github.com/kubernetes/kube-state-metrics/pull/1687

**Describe the solution you'd like**
1. Import k8s metrics stability framework https://github.com/kubernetes/kube-state-metrics/pull/1844 and https://github.com/kubernetes/kube-state-metrics/pull/1976
2. Presubmit check https://github.com/kubernetes/kube-state-metrics/pull/1836

**Additional context**

PR #1836, opened 2022-09-17
Check stable metrics aren't changed

CatherineF-dev:
Check stable metrics are stable

**What this PR does / why we need it**: Guarantee metric name, labels aren't changed for stable metrics. 

does not change cardinality

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1833


CatherineF-dev:
cc @dgrisonnet

dgrisonnet (commented):
I generally like the idea behind that PR but this is quite challenging to bring to kube-state-metrics since the metrics are generated at runtime and so you might not get all the labels by default.

One solution could be to pass a runtime object to the generator that satisfies all the conditions to generate metrics with all the labels, but that means that we should make sure that the object is updated every time a new condition is added so it isn't very safe.

Another solution that would be safer is to statically visit the generators to build a metrics object. From the generator itself, we could get all the metadata of the metrics and we could get the full list of labels by walking through the generator function and finding assignation to `LabelKeys`. That would be a beefy work to do, but it could be reused to both generate and test the stable metrics, so that would be my suggestion. (maybe also to generate some doc in the future)

-- 

With regards to the actual content of the PR, would you mind first opening a PR to introduce the stability framework and the help text annotation? That would make it easier to review. As I commented in the PR, I am generally more in favor of using the Kubernetes framework compare to the very basic one we have in KSM, so feel free to go with the component-base one.

CatherineF-dev:
Sure! Thanks!

Agree on visiting the generators to build a metrics object.

CatherineF-dev:
It's not easy to do static analysis for KSM metrics. But e2e tests expose all stable metrics, as shown in https://github.com/kubernetes/kube-state-metrics/pull/1853.

So plan to use e2e tests to validate stable metrics (suggested by han). Also will support SKIP_METRICS_LIST, given future stable metrics might be only exposed when certain conditions happen.   For these metrics which aren't exposed in e2e tests, we can use unit-test.

CatherineF-dev:
cc @rexagod @dgrisonnet @logicalhan it's ready to review this PR which adds presubmit check for stable metrics.

tests/stable/check_test.go
```diff
@@ -0,0 +1,194 @@
...
+	fatal(err)
+	collectedStableMetrics := extractStableMetrics(mf)
+	printMetric(collectedStableMetrics)
+
+	expectedStableMetrics, err := readYaml(stableYaml)
+	if err != nil {
+		t.Fatalf("Can't read stable metrics from file. err = %v", err)
+	}
+
+	err = compare(collectedStableMetrics, *expectedStableMetrics, skipStableMetrics)
```
logicalhan, line 68: Honestly, i would just sort the metrics, output a yaml string and do a simple string comparison. 

This is not only simpler, but when you generate the yaml string, you'll end up with deterministic ordering which will make the diff more pleasant.

rexagod (changes requested):

tests/e2e.sh
```diff
@@ -163,6 +164,10 @@ mkdir -p ${KUBE_STATE_METRICS_LOG_DIR}
 echo "access kube-state-metrics metrics endpoint"
 curl -s "http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state-metrics:http-metrics/proxy/metrics" >${KUBE_STATE_METRICS_LOG_DIR}/metrics
 
+cat ${KUBE_STATE_METRICS_LOG_DIR}/metrics
```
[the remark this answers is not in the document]
CatherineF-dev: Done.

tests/stable/check_test.go
```diff
@@ -0,0 +1,207 @@
...
+func TestMain(m *testing.M) {
+	flag.StringVar(&promText, "collectedMetricsFile", "", "input prometheus metrics text file, text format")
+	flag.StringVar(&stableYaml, "stableMetricsFile", "", "expected stable metrics yaml file, yaml format")
+	flag.Parse()
+	m.Run()
+}
+
+func TestStableMetrics(t *testing.T) {
+	mf, err := parsePromText(promText)
+	fatal(err)
```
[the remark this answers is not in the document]
CatherineF-dev: Done.

tests/stable/check_test.go
```diff
@@ -0,0 +1,207 @@
...
+	}
+	c := &[]Metric{}
+	err = yaml.Unmarshal(buf, c)
+	if err != nil {
+		return nil, fmt.Errorf("error %q: %w", filename, err)
+	}
+	return c, err
+}
+
+func convert2Map(metrics []Metric, skipMap map[string]int) map[string]Metric {
```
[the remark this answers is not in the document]
CatherineF-dev: Done.

CatherineF-dev:
Ping~

tests/stable/testdata/README.md
```diff
@@ -0,0 +1,23 @@
+These stable metrics aren't covered in stablemetrics.yaml, because they aren't
```
logicalhan, line 1: Can you add more documentation on how to run this stuff?

dgrisonnet, line 1: This is a bit problematic since ideally we would want a system that none of the stable metrics are changed.

Another problem we might encounter is that some labels will not verified because they are not present in the e2e tests. I don't have any example of stable metrics like that in mind, but we have for experimental ones: https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/endpointslice.go#L77-L148.
If we were to verify that metric today in the e2e tests, we would have to make sure that the endpointslice object is complete which is very prone to errors.

I recall we discussed implementing a static analysis solution to make sure that we were going over all the metrics and labels, was that not possible in the end?

CatherineF-dev: > a static analysis solution

It's hard to use static analysis for current kube-state-metrics repo, because labels are too dynamic.
For example,  it uses`labelKeys = append(labelKeys, "ready")` in [kube_endpointslice_endpoints](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/endpointslice.go#L77-L148).

dgrisonnet: 1. There are two things that are difficult to handle in that example:
1.1 Having an intermediate labelKeys variable mean that we can't simply have the tool look for the labelKeys field assignation but might need to also build a constant version of the variable assigned to the field. An idea could be to have the tool first look for the labelKeys field assignation and in case the RValue is a variable and not a constant, iterate over the code again to build a const version of that variable.
1.2 `append` should be the only callExp that we will ever see so having a case in the tool just for it would make sense to me. The idea would to replace append(labelKeys, "ready") but building a temporary slice and putting the "ready" string inside of it to convert the variable array to a constant one for our static analysis. 
2. We won't have to handle this scenario because it only affects `_labels` and `_annotations` metrics which have variable labels by design and as such shouldn't have any stability guarantees.

> One possible way to make static analysis easier is to require defining all possible labels first.

This could make sense but I am afraid that this might make label values assignation error-prone since we would have to use the indexes instead of just appending the new labels.

The only option I see today in order to catch all the labels of all the metrics is via static analysis and if we don't want to make that intrusive to the developers to avoid errors, we will need to make the tool intelligent.

For the static analysis I am thinking that the following could be viable:
1. Create an AST of each store and iterate over every generator code
2. Find the labelKeys field assignation
3. If the RValue isn't constant, reiterate over the generators to build a constant version of the variable. That should be possible because even though the slices are dynamic, there values are all constant.

Writing this I noticed that there is another scenario that we would need to handle which is the one of default labels: https://github.com/kubernetes/kube-state-metrics/blob/559bb28d6f35ab797e31cc361605732121a1603e/internal/store/endpointslice.go#L38

These should always be define at the top-level of each store as a constant slice so we should be able to easily add them to each metrics.

This is becoming quite complex so if you have any other suggestions I would gladly hear them but I don't think that we can avoid static analysis if we want to build consistent guarantees.

CatherineF-dev: > I don't think that we can avoid static analysis if we want to build consistent guarantees.

Agree

> This could make sense but I am afraid that this might make label values assignation error-prone since we would have to use the indexes instead of just appending the new labels.

Could you provide an example for this case?

dgrisonnet: > Could you provide an example for this case?

I think my reasoning was based on examples such as https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/endpointslice.go#L77-L148. In these cases order is super important since we build the list of label values one label at a time. I thought that if we were to define label keys upfront, future addition of labels might mess with existing ordering.

But thinking about it again we could transform the code into something like:

```go
const labelKeys []string = {"ready", "serving", ...} 

ready := ""
if ep.Conditions.Ready != nil {
  ready = ep.Conditions.Ready
}

serving := ""
if ep.Conditions.Serving != nil {
  serving = ep.Conditions.Serving
}

m = append(m, &metric.Metric,
  LabelKeys:   labelKeys,
  LabelValues: []string{ready, serving},
  Value:       1,
})
```

This original reasoning for how it is coded in the example I shared above was to reduce the network footprint by removing empty labels. But it could be argued that it doesn't have much impact + we could do processing later on when we build the HTTP response.

So I think your idea of defining const label keys everywhere makes sense and would ease very much static analysis.

rexagod:
Could we perhaps lint docs so any doc changes for stable metrics are disallowed entirely?

No PR can be merged without updating docs, which won't be possible if the lint step fails first due to a stable metric manipulation, so one hand would essentially feed the other.

PR #1839, opened 2022-09-19
[FEATURE] Add kube_*_labels for custom resources if enabled

### LINKED ISSUE #1832: Add *_labels metrics for custom resource configurations

Allow to expose `*_labels` for Custom Resource configuration which behaves the same as `kube_*_labels` metrics.
This means it would also use the configured value for the flag `--metric-labels-allowlist` for filtering the exposed labels.

**Why is this needed**:

To have the same kind of features for custom resource metrics.

One option would be to add a boolean for enabling the labels for a custom resource configuration, e.g. at the `Resource` struct level https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/customresourcestate/config.go#L42

**Additional context**

As of today it would require some custom configuration to expose something like that for a custom resource.

```yaml
spec:
  resources:
    - groupVersionKind:
        group: myteam.io
        kind: "Foo"
        version: "v1"
      metrics:
        - name: "labels"
          each:
            type: Info
            info:
              labelsFromPath:
                label_foo: [metadata, labels, foo]
                label_bar: [metadata, labels, bar]
```

QuentinBisson:

This PR allows the creation of a kube_*_ labels metrics for the custom resources

This change will add a new metric per enabled CRD in KSM

Fixes #1832

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -181,9 +181,13 @@ func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.Registry
 			klog.InfoS("The internal resource store already exists and is overridden by a custom resource store with the same name, please make sure it meets your expectation", "registryName", f.Name())
 		}
 		availableStores[f.Name()] = func(b *Builder) []cache.Store {
+			metricFamilyGenerators := f.MetricFamilyGenerators(b.allowAnnotationsList[f.Name()], b.allowLabelsList[f.Name()])
+			if f.IsLabelsMetricEnabled() {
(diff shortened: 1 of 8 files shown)


QuentinBisson:
I could use some help on the documentation part as I'm not sure how to document this properly

QuentinBisson:
@fpetkovski I included your comments :)

dgrisonnet:
I commented about this in https://github.com/kubernetes/kube-state-metrics/issues/1832#issuecomment-1322518103.

I am not in favor of this change since it is already possible to add labels and annotations metrics via the custom resource configuration. I don't think these metrics are worth to have by default and I would expect anyone needing these metrics to write a configuration for them.

PR #1844, opened 2022-09-27
Import k8s metrics stability framework

### LINKED ISSUE #1833: Presubmit checks for stable metrics

CatherineF-dev:
**What this PR does / why we need it**: #1833

Fixes #1833

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -22,6 +22,7 @@ require (
 	k8s.io/apimachinery v0.25.2
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.12.0
 	k8s.io/client-go v0.25.2
+	k8s.io/component-base v0.25.2
 	k8s.io/klog/v2 v2.80.1
 	k8s.io/sample-controller v0.25.2
 	k8s.io/utils v0.0.0-20220922133306-665eaaec4324
@@ -40,7 +41,7 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
-	github.com/blang/semver v3.5.0+incompatible // indirect
+	github.com/blang/semver/v4 v4.0.0 // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/emicklei/go-restful/v3 v3.8.0 // indirect
--- a/go.mod
+++ b/go.mod
@@ -40,6 +40,7 @@ require (
 	github.com/alecthomas/template v0.0.0-20190718012654-fb15b899a751 // indirect
 	github.com/alecthomas/units v0.0.0-20211218093645-b94a6e3cc137 // indirect
 	github.com/beorn7/perks v1.0.1 // indirect
+	github.com/blang/semver v3.5.0+incompatible // indirect
 	github.com/cespare/xxhash/v2 v2.1.2 // indirect
 	github.com/davecgh/go-spew v1.1.1 // indirect
 	github.com/emicklei/go-restful/v3 v3.8.0 // indirect
--- a/pkg/metric_generator/generator.go
+++ b/pkg/metric_generator/generator.go
@@ -20,8 +20,8 @@ import (
 	"fmt"
 	"strings"
 
-	"k8s.io/kube-state-metrics/v2/pkg/metric"
 	basemetrics "k8s.io/component-base/metrics"
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
 )
 
 // FamilyGenerator provides everything needed to generate a metric family with a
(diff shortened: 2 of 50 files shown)


CatherineF-dev:
cc @dgrisonnet

pkg/metric_generator/generator.go
```diff
@@ -52,6 +56,11 @@ func NewFamilyGenerator(name string, help string, metricType metric.Type, deprec
 	return f
 }
 
+// NewFamilyGenerator creates new FamilyGenerator instances.
+func NewFamilyGenerator(name string, help string, metricType metric.Type, deprecatedVersion string, generateFunc func(obj interface{}) *metric.Family) *FamilyGenerator {
+	return NewFamilyGeneratorWithStability(name, help, metricType, basemetrics.ALPHA, deprecatedVersion, generateFunc)
```
rexagod, line 61: I'm not sure why all metrics are being set to `alpha`?

CatherineF-dev: Not all. 
In replacing NewFamilyGenerator with NewFamilyGeneratorWithStability. 

For example: https://github.com/kubernetes/kube-state-metrics/pull/1844/files#diff-28c04d8ad4d0fd23a2b14d61e94ca54e757c1e5ab971cea37974fe1b2f260085

rexagod: Ah, okay. So for now we are setting all `node` metrics to `alpha`?

go.mod
```diff
@@ -22,6 +22,7 @@ require (
 	k8s.io/apimachinery v0.24.4
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.11.0
 	k8s.io/client-go v0.24.4
+	k8s.io/component-base v0.18.3
```
mrueg, line 25: looks like this is introducing an old version of component-base, while newer releases are available?

pkg/metric_generator/generator.go
```diff
@@ -75,7 +84,13 @@ func (g *FamilyGenerator) generateHeader() string {
 	header.WriteString("# HELP ")
 	header.WriteString(g.Name)
 	header.WriteByte(' ')
-	header.WriteString(g.Help)
+	// Will remove if-else after all metrics are attached with right
```
mrueg, line 87: Can you add "TODO" key to the comment, so it won't be forgotten? (and folks will find it when grepping for TODO)

rexagod: +1.

pkg/metric_generator/generator.go
rexagod: It seems only two of these metrics were experimental before. So IIUC, if the `node` metrics are being set to `alpha`, should it only be for these two?
```
| kube_node_annotations        | Gauge       | Kubernetes annotations converted to Prometheus labels
| kube_node_role               | Gauge       | The role of a cluster node
```

CatherineF-dev: > what state were they in before? Experimental?

Yes, we are mapping into k8s stability level:
Experimental -> Alpha
Stable -> Stable

> So IIUC, if the node metrics are being set to alpha, should it only be for these two?

Working in progress for other Stable metrics. Because there are many test codes need to be changed. 
I planned to create 2+ CLs. The first CL (this CL) to import stability framework, other CLs migrate NewFamilyGenerator 
 to NewFamilyGeneratorWithStability.

CatherineF-dev: Added all STABLE metrics just now.

rexagod: Oh wow! This looks amazing, thank you! 🎉

CatherineF-dev: Oh, only stable metrics are using NewFamilyGeneratorWithStability now.

Will create another CL to migrate non-stable metrics to use NewFamilyGeneratorWithStability. Because this way has less merge conflicts in this CL.

CatherineF-dev:
QQ: do we have other comments? Or could someone approve this CL?

go.mod
```diff
@@ -18,20 +18,21 @@ require (
...
-	k8s.io/api v0.24.4
-	k8s.io/apimachinery v0.24.4
+	k8s.io/api v0.25.2
+	k8s.io/apimachinery v0.25.2
 	k8s.io/autoscaler/vertical-pod-autoscaler v0.11.0
-	k8s.io/client-go v0.24.4
+	k8s.io/client-go v0.25.2
+	k8s.io/component-base v0.25.2
 	k8s.io/klog/v2 v2.70.1
 	k8s.io/sample-controller v0.24.4
```
dgrisonnet, line 27: could you also bump that one to v0.25.2, just to make sure that all the k8s deps are in line

dgrisonnet (commented):
One small comment from my side, but otherwise it looks good to me. Great work @CatherineF-dev :+1:

go.mod
dgrisonnet: actually, I forgot that this one was in flight: https://github.com/kubernetes/kube-state-metrics/pull/1819, if you rebase on master you should be able to remove the dependency bump from this PR alltogether.

CatherineF-dev: Done.

PR #1845, opened 2022-09-27
Allow Lease metrics to be exported across all namespaces

lantingchiang:
**What this PR does / why we need it**: This PR modifies the Lease store's `createLeaseListWatch()` method so that the metrics for the Lease custom resource to be exported across all namespaces, as opposed to only the `kube-node-lease` namespace. This allows us to get metrics on Lease resources used by pods as well. 

**How does this change affect the cardinality of KSM**: *increases cardinality by the number of namespaces*

Fixes #

### DIFF
--- a/internal/store/lease.go
+++ b/internal/store/lease.go
@@ -40,7 +40,7 @@ var (
 			metric.Gauge,
 			"",
 			wrapLeaseFunc(func(l *coordinationv1.Lease) *metric.Family {
-				labelKeys := []string{"owner_kind", "owner_name", "namespace"}
+				labelKeys := []string{"owner_kind", "owner_name", "namespace", "lease_holder"}
 
 				owners := l.GetOwnerReferences()
 				if len(owners) == 0 {
@@ -48,7 +48,7 @@ var (
 						Metrics: []*metric.Metric{
 							{
 								LabelKeys:   labelKeys,
-								LabelValues: []string{"<none>", "<none>", l.Namespace},
+								LabelValues: []string{"<none>", "<none>", l.Namespace, *l.Spec.HolderIdentity},
 								Value:       1,
 							},
 						},
@@ -59,7 +59,7 @@ var (
 				for i, owner := range owners {
 					ms[i] = &metric.Metric{
 						LabelKeys:   labelKeys,
-						LabelValues: []string{owner.Kind, owner.Name, l.Namespace},
+						LabelValues: []string{owner.Kind, owner.Name, l.Namespace, *l.Spec.HolderIdentity},
 						Value:       1,
 					}
 				}
--- a/internal/store/lease.go
+++ b/internal/store/lease.go
@@ -42,13 +42,18 @@ var (
 			wrapLeaseFunc(func(l *coordinationv1.Lease) *metric.Family {
 				labelKeys := []string{"owner_kind", "owner_name", "namespace", "lease_holder"}
 
+				var holder string
+				if l.Spec.HolderIdentity != nil {
+					holder = *l.Spec.HolderIdentity
+				}
+
 				owners := l.GetOwnerReferences()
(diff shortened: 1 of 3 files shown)


fpetkovski:
Thanks for this PR. The change makes sense to me, but I am just curious why KSM was exporting lease metrics from that single namespace only.

/lgtm
/hold

fpetkovski:
Looking at the tests, we don't expose the namespace yet: https://github.com/kubernetes/kube-state-metrics/blob/master/internal/store/lease_test.go#L57-L58

We should update the lease store to export the fields that you added to the docs, and update tests accordingly.

lantingchiang:
> Thanks for this PR. The change makes sense to me, but I am just curious why KSM was exporting lease metrics from that single namespace only.

I think it's because that namespace contains all node leases, so it was sufficient if pod leases weren't of interest.

fpetkovski (commented):
Thanks

/lgtm

lantingchiang:
Hi @dgrisonnet, could I get a review please? Thanks!

lantingchiang:
Exported `Lease.Spec.HolderIdentity` field because some pod leases don't have a `metadata.ownerReferences` field.

internal/store/lease.go
```diff
@@ -40,15 +40,15 @@ var (
...
+				labelKeys := []string{"owner_kind", "owner_name", "namespace", "lease_holder"}
 
 				owners := l.GetOwnerReferences()
 				if len(owners) == 0 {
 					return &metric.Family{
 						Metrics: []*metric.Metric{
 							{
 								LabelKeys:   labelKeys,
-								LabelValues: []string{"<none>", "<none>"},
+								LabelValues: []string{"<none>", "<none>", l.Namespace, *l.Spec.HolderIdentity},
```
fpetkovski, line 51: Can `*l.Spec.HolderIdentity` be nil in some cases?

lantingchiang: Yes HolderIdentity is optional. Thanks for catching this!

internal/store/lease.go
```diff
@@ -42,13 +42,20 @@ var (
 			wrapLeaseFunc(func(l *coordinationv1.Lease) *metric.Family {
 				labelKeys := []string{"owner_kind", "owner_name", "namespace", "lease_holder"}
 
+				var holder string
+				if l.Spec.HolderIdentity == nil {
+					holder = "<none>"
```
fpetkovski, line 47:
```suggestion
					holder = ""
```

internal/store/lease.go
```diff
@@ -42,13 +42,20 @@ var (
...
+				} else {
+					holder = *l.Spec.HolderIdentity
+				}
+
```
fpetkovski, line 51:
```suggestion
				var holder string
				if l.Spec.HolderIdentity != nil {
					holder = *l.Spec.HolderIdentity
				}

```

fpetkovski (commented):
Just one small nit, otherwise looks good

internal/store/lease.go
```diff
@@ -40,15 +40,20 @@ var (
 			metric.Gauge,
 			"",
 			wrapLeaseFunc(func(l *coordinationv1.Lease) *metric.Family {
-				labelKeys := []string{"owner_kind", "owner_name"}
+				labelKeys := []string{"owner_kind", "owner_name", "namespace", "lease_holder"}
+
+				holder = "<none>"
```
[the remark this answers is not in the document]
lantingchiang: Got it! I was trying to match the `owner_name` and `owner_kind` values when they're empty, but thanks for the clarification!

fpetkovski:
Thank you for the contribution!

/lgtm

dgrisonnet (approved):
/lgtm
/approve
/unhold

PR #1849, opened 2022-10-07
Recommend kube-scheduler's metrics

### LINKED ISSUE #1846: Deprecate pod resource request and limit metrics
We should look into whether we should deprecate the container-level `kube_pod_container_resource_limits` and `kube_pod_container_resource_requests` metrics in favor of pod-level resource request and limit metrics exposed by the kube-scheduler. The scheduler metrics were introduced by https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics in 1.21, as a replacement for the metrics exposed by kube-state-metrics. The goal is to improve accuracy when doing capacity planning.

The k8s effort is currently in beta phase, but we will be looking into graduating it to stable for 1.27. As such, to avoid having duplicate metrics in the ecosystem, it would be better to deprecate the ksm metrics in favor of the ones exposed by the scheduler.

One thing we need to consider is that these metrics are exposed on a separate endpoint so we should make sure that if we do the transition, we leave some time for the kube-state-metrics users to start scraping the /metrics/resources endpoint.

rexagod:
**What this PR does / why we need it**: Recommend metrics exposed by the kube-scheduler, in case of the ones below:
* `kube_pod_container_resource_limits`
* `kube_pod_container_resource_requests`

Refer: https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics#expose-new-metrics.

**How does this change affect the cardinality of KSM**: No.

**Which issue(s) this PR fixes**: Fixes #1846.


internal/store/pod.go
```diff
@@ -157,7 +157,7 @@ func createPodContainerInfoFamilyGenerator() generator.FamilyGenerator {
 func createPodContainerResourceLimitsFamilyGenerator() generator.FamilyGenerator {
 	return *generator.NewFamilyGenerator(
 		"kube_pod_container_resource_limits",
-		"The number of requested limit resource by a container.",
+		"The number of requested limit resource by a container. However, it is recommended to use the `kube_pod_resource_limits` metric exposed by kube-scheduler instead, as it is more precise.",
```
dgrisonnet, line 160:
```suggestion
		"The number of requested limit resource by a container. It is recommended to use the `kube_pod_resource_limits` metric exposed by kube-scheduler instead, as it is more precise.",
```

docs/pod-metrics.md
```diff
@@ -1,54 +1,54 @@
...
+| kube_pod_status_phase                                 | Gauge       | The pods current phase                                                                                                                                                                           |                                                | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `phase`=&lt;Pending\                                                                                                                                                                                                                                                                           | Running\      | Succeeded\                               |Failed\|Unknown&gt; <br> `uid`=&lt;pod-uid&gt; | STABLE | - |
```
dgrisonnet, line 13: The doc seems to be broken starting from here

mrueg:
@rexagod do you have plans to fix the test failures and include the feedback here? I will work on the next release soon.

rexagod:
Ah, my bad. Let me force push.

PR #1850, opened 2022-10-07
Represent GVK information as labels

### LINKED ISSUE #1847: CRD metrics are extremely brittle
**What happened**:

CRD metrics are currently extremely brittle due to the way it is implemented. The name of the CRD is interpolated into the name of the metric which makes it impossible to aggregate a CRD across different versions (or at least very very annoying to do so). 

The name, version, group of the CRD should be a value in a metric label.

https://github.com/kubernetes/kube-state-metrics/blob/master/docs/customresourcestate-metrics.md

rexagod:
**What this PR does / why we need it**: Represent GVK information as labels in the metrics, instead of appending them to the metric name itself. This would allow users to aggregate varying GVKs of a CR under the same metric, making operations much more easier.

**How does this change affect the cardinality of KSM**: Increases, or stays the same, depending on the cardinality of the group(s), version(s), or kind(s) of a CR present in the cluster at the given time.

**Which issue(s) this PR fixes**: Fixes #1847

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -48,6 +48,8 @@ spec:
           - --resources=certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,foos,horizontalpodautoscalers,ingresses,jobs,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments,verticalpodautoscalers
 ```
 
+NOTE: The `group`, `version`, and `kind` common labels are reserved, and will be overwritten by the values from the `groupVersionKind` field.
+
 ### Examples
 
 The examples in this section will use the following custom resource:
@@ -114,7 +116,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_myteam_io_v1_Foo_uptime 43.21
+uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
 ```
 
 #### Multiple Metrics/Kitchen Sink
@@ -165,8 +167,8 @@ spec:
 Produces the following metrics:
 
 ```prometheus
-kube_myteam_io_v1_Foo_active_count{active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 1
-kube_myteam_io_v1_Foo_active_count{active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 3
+active_count{group="myteam.io", kind="Foo", version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 1
+active_count{group="myteam.io", kind="Foo", version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 3
 ```
 
 ### Metric types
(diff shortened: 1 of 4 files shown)


pkg/customresourcestate/registry_factory_test.go
```diff
@@ -324,7 +324,7 @@ func Test_fullName(t *testing.T) {
 				resource: r(nil),
 				f:        count,
 			},
-			want: "kube_apps_v1_Deployment_count",
+			want: "count",
```
logicalhan, line 327: should we prefix these with `crd_`?

logicalhan (commented):
This looks awesome by the way. But I have some reservations about the unqualified metric names, because they seem way too generic.

pkg/customresourcestate/registry_factory_test.go
mrueg: I'd prefer kube_crd

docs/customresourcestate-metrics.md
```diff
@@ -287,7 +289,7 @@ spec:
 
 Produces:
 ```prometheus
-myteam_foos_uptime 43.21
+myteam_foos_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
```
logicalhan, line 292:
```suggestion
uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
```

logicalhan: Would you mind changing the prefix to "kube_crd" as the other comment suggested?

logicalhan: https://github.com/kubernetes/kube-state-metrics/pull/1850#discussion_r991572154

pkg/customresourcestate/config.go
```diff
@@ -63,17 +62,11 @@ type Resource struct {
...
 func (r Resource) GetMetricNamePrefix() string {
-	if r.MetricNamePrefix == nil {
-		return strings.NewReplacer(
-			"/", "_",
-			".", "_",
-			"-", "_",
-		).Replace(fmt.Sprintf("kube_%s_%s_%s", r.GroupVersionKind.Group, r.GroupVersionKind.Version, r.GroupVersionKind.Kind))
+	p := r.MetricNamePrefix
+	if p == nil {
+		return "kube"
```
logicalhan, line 67:
```suggestion
		return "kube_crd"
```

rexagod:
@logicalhan Addressed the reviews, PTAL.

pkg/customresourcestate/registry_factory_test.go
```diff
@@ -324,7 +324,7 @@ func Test_fullName(t *testing.T) {
 				resource: r(nil),
 				f:        count,
 			},
-			want: "kube_apps_v1_Deployment_count",
+			want: "kube_count",
```
logicalhan, line 327:
```suggestion
			want: "kube_crd_count",
```

docs/customresourcestate-metrics.md
rexagod: This metric would be unchanged as the configuration specified a `metricNamePrefix: myteam_foos`.

https://github.com/kubernetes/kube-state-metrics/blob/633cbf49d0ecf607c716860b21e03c4afd596b8c/docs/customresourcestate-metrics.md?plain=1#L279-L293

docs/customresourcestate-metrics.md
```diff
@@ -305,6 +305,11 @@ spec:
           ...
 ```
 
+Produces:
+```prometheus
+uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
+```
+
```
rexagod, line 312: @logicalhan JFYI I added a doc snippet here to make things a bit more explicit to the reader.

logicalhan: sgtm

PR #1856, opened 2022-10-17
customresourcestate fix type indentation in example

### LINKED ISSUE #1855: Typo in docs of custom resources after PR #1777
**What happened**:
The new 'type' field for custom metrics, as added in https://github.com/kubernetes/kube-state-metrics/pull/1777, should be under 'each', see:

https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/customresourcestate/registry_factory.go#L65
https://github.com/kubernetes/kube-state-metrics/blob/master/pkg/customresourcestate/registry_factory.go#L136

The problem is in line 46 in https://github.com/kubernetes/kube-state-metrics/edit/master/docs/customresourcestate-metrics.md

**Environment**:
- kube-state-metrics version: 
- Other info:

chrischdi:

Fixes document to match configuration syntax.

Fixes #1855

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -43,8 +43,9 @@ spec:
                     metrics:
                       - name: active_count
                         help: "Count of active Foo"
-                        type: Gauge
-                        ...
+                        each:
+                          type: Gauge
+                          ...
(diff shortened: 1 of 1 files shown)


fpetkovski:
Thanks for the quick fix.

/lgtm

PR #1864, opened 2022-10-20
Sharding metrics per node via fieldSelector

### LINKED ISSUE #1863: Support sharding by node_name for pod metrics

**Why is this needed**: Reduce KSM agent memory/cpu usage with multiple replicas, while not watching duplicated information.

**Describe the solution you'd like**: Add selector `spec.nodeName=`  in ListOptions

**Additional context**

Env $(NODE_NAME) can be passed to each replica

- [x] Add nodename FieldSelector https://github.com/kubernetes/kube-state-metrics/pull/1864
- [ ] Add sharding example under https://github.com/kubernetes/kube-state-metrics/pull/1873

CatherineF-dev:
Increase scalability for big k8s cluster by sharding metrics per node. It only supports for resources which support nodeName fieldSelector. For example, pod metrics.

**Which issue(s) this PR fixes** *:
Fixes #1863


CatherineF-dev:
Ping~

CatherineF-dev:
cc @rexagod, could you help review as well? Thanks!

internal/store/builder.go
```diff
@@ -461,7 +472,13 @@ func (b *Builder) buildStores(
...
 			composedMetricGenFuncs,
 		)
-		listWatcher := listWatchFunc(b.kubeClient, v1.NamespaceAll, b.namespaceFilter)
+		merged, err := b.MergeFieldSelector(b.namespaceFilter, b.nodenameFilter)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to merge fieldSelector %s and %s", b.namespaceFilter, b.nodenameFilter))
+		}
+		klog.Infof("FieldSelector is used ", merged)
+		listWatcher := listWatchFunc(b.kubeClient, v1.NamespaceAll, merged)
+		b.startReflector(expectedType, store, listWatcher, useAPIServerCache)
```
fpetkovski, line 481: Not sure if I am reading the diff correctly, but does this start the same reflector twice?

fpetkovski (commented):
Thanks, overall this looks good. I would suggest using `NodeName` instead of `Nodename` since it's more readable and it's also not a single word.

internal/store/builder.go
CatherineF-dev: Yes, you're right. I need to remove this `b.startReflector(expectedType, store, listWatcher, useAPIServerCache)`.  Thank you!

CatherineF-dev:
Reply @mrueg 
- Pod metrics: daemonset
- Other metrics:  deployment

When deployed as a daemonset, set nodename to current NODE_NAME.
```
apiVersion: apps/v1
kind: DaemonSet
metadata:
  name: kube-state-metrics
spec:
  template:
    spec:
      containers:
      - name: kube-state-metrics
        args:
        - --resources=pods
        - --nodename=$(NODE_NAME)
```

internal/store/builder.go
```diff
@@ -472,7 +489,12 @@ func (b *Builder) buildStores(
 			familyHeaders,
 			composedMetricGenFuncs,
 		)
-		listWatcher := listWatchFunc(b.kubeClient, ns, b.namespaceFilter)
+		merged, err := b.MergeFieldSelector(b.namespaceFilter, b.nodenameFilter)
+		if err != nil {
+			panic(fmt.Sprintf("Failed to merge fieldSelector %s and %s", b.namespaceFilter, b.nodenameFilter))
```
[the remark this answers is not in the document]
CatherineF-dev: In builder.go, it also has some panics at https://github.com/kubernetes/kube-state-metrics/blob/71b4d04b1c6ee14924ea2e9a25dc5ed0e4b730c4/internal/store/builder.go#L225

Maybe it's fine to keep that panic?  Or should we move `merged, err := b.MergeFieldSelector(b.namespaceFilter, b.nodenameFilter)` to here: https://github.com/kubernetes/kube-state-metrics/blob/8118f0271be2bf849f10746d94e829d1bd5db101/pkg/app/server.go#L107-L111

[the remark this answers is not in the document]
CatherineF-dev: Moved `merged, err := b.MergeFieldSelector(b.namespaceFilter, b.nodenameFilter)`  to pkg/app/server.go.

See in this commit https://github.com/kubernetes/kube-state-metrics/pull/1864/commits/06acc57f9678113773f3e95889ad5d8d2f8f9a89

CatherineF-dev:
Reply @rexagod

It's a daemonset. Each kube-state-metrics agent will only collect pod metrics from the same node.

- Pod metrics: daemonset
- Other metrics: deployment/statefulset

pkg/options/options.go
```diff
@@ -100,6 +101,7 @@ func (o *Options) AddFlags() {
 	o.flags.Var(&o.Resources, "resources", fmt.Sprintf("Comma-separated list of Resources to be enabled. Defaults to %q", &DefaultResources))
 	o.flags.Var(&o.Namespaces, "namespaces", fmt.Sprintf("Comma-separated list of namespaces to be enabled. Defaults to %q", &DefaultNamespaces))
 	o.flags.Var(&o.NamespacesDenylist, "namespaces-denylist", "Comma-separated list of namespaces not to be enabled. If namespaces and namespaces-denylist are both set, only namespaces that are excluded in namespaces-denylist will be used.")
+	o.flags.StringVar((*string)(&o.NodeName), "nodename", "", "Set spec.nodeName=nodename when watching resources. Only available for resources which support nodeName filter.")
```
rexagod, line 104: Additionally, could add something explicitly stating that only these resources will be watched.
```suggestion
	o.flags.StringVar((*string)(&o.NodeName), "nodename", "", "Name of the node that contains the kube-state-metrics pod, only available for resources that support nodeName filter. Most likely this should be passed via the downward API.")
```

pkg/options/types.go
```diff
@@ -104,6 +104,41 @@ func (r *ResourceSet) Type() string {
...
+	if err != nil {
+		return EmptyFieldSelector(), err
+	}
+	if selector1.Empty() {
+		return selector2.String(), nil
+	}
+	if selector2.Empty() {
+		return selector1.String(), nil
+	}
+	return fields.AndSelectors(selector1, selector2).String(), nil
```
rexagod, line 139: Is it safe to assume that we'd always be working with two selectors here? What if we add another filter in the future? Since this is not a method on `*NodeNameType` but a general function inside `options`, do we want to generalize this a bit further, so as to be able to merge `n` selectors?

main.go
```diff
@@ -33,6 +33,19 @@ import (
...
+		return nil
+	}
+	for _, x := range opts.Resources.AsSlice() {
+		if x != shardableResource {
+			return fmt.Errorf("Resource %s can't be sharding by field selector nodeName", x)
+		}
+	}
+	return nil
+}
+
```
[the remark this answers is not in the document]
CatherineF-dev: Yes, it will be better!

internal/store/builder.go
```diff
@@ -461,7 +478,10 @@ func (b *Builder) buildStores(
 			familyHeaders,
 			composedMetricGenFuncs,
 		)
-		listWatcher := listWatchFunc(b.kubeClient, v1.NamespaceAll, b.namespaceFilter)
+		if b.fieldSelectorFilter != "" {
+			klog.Infof("FieldSelector is used ", b.fieldSelectorFilter)
+		}
```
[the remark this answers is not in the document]
CatherineF-dev: Yes, users will know enabled field selectors.

CatherineF-dev:
Will add scale model example in https://github.com/kubernetes/kube-state-metrics/pull/1873 after it has a valid tag name.

README.md
```diff
@@ -235,6 +236,31 @@ This way of deploying shards is useful when you want to manage KSM shards throug
...
+  template:
+    spec:
+      containers:
+      - name: kube-state-metrics
+        args:
+        - --resources=pods
+        - --nodename=$(NODE_NAME)
+```
+
+For other metrics, they can sharded via [Horizontal sharding](#horizontal-sharding).
```
mrueg, line 262:
```suggestion
Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
```

README.md
```diff
@@ -235,6 +236,31 @@ This way of deploying shards is useful when you want to manage KSM shards throug
...
+metadata:
+  name: kube-state-metrics
+spec:
+  template:
+    spec:
+      containers:
+      - name: kube-state-metrics
+        args:
+        - --resources=pods
+        - --nodename=$(NODE_NAME)
```
mrueg, line 259: Can you include the downward API in here https://kubernetes.io/docs/concepts/workloads/pods/downward-api/ It's not clear where `$NODE_NAME` is coming from otherwise.

CatherineF-dev: Updated. NODE_NAME is  from:
```
        env:
        - name: NODE_NAME
          valueFrom:
            fieldRef:
              apiVersion: v1
              fieldPath: spec.nodeName

```

mrueg:
This looks good, thanks for your work @CatherineF-dev  
/approve

Waiting for @dgrisonnet or @fpetkovski to take another look

tests/compare_benchmarks.sh
```diff
@@ -5,6 +5,8 @@ set -e
 set -o pipefail
 # error on unset variables
 set -u
+# TODO: delete `set -x` after debugging
```
fpetkovski, line 8: Is this TODO for yourself? :)

tests/e2e/main_test.go
```diff
@@ -177,8 +203,15 @@ func getLabelsDocumentation() (map[string][]string, error) {
 			}
 			metric := params[1]
 			labelsDoc := params[2]
+			if metric == "kube_node_spec_taint" {
+				fmt.Println("## labelsDoc", labelsDoc)
```
fpetkovski, line 207: Might be better to use `t.Log`

tests/e2e/main_test.go
```diff
@@ -312,3 +345,141 @@ func TestDefaultCollectorMetricsAvailable(t *testing.T) {
...
+						},
+					},
+				})
+				break
+			}
+		}
+		_, updateErr := deploymentsClient.Update(context.TODO(), result, metav1.UpdateOptions{})
+		return updateErr
+	})
+	time.Sleep(time.Second * 100)
```
fpetkovski, line 434: What is the purpose of this sleep statement? Can we use something like `wait.Poll`?

fpetkovski (commented):
Overall looks good, I think tests need a bit more work.

CatherineF-dev:
Cleaned up. @fpetkovski

Will add E2E tests with PR https://github.com/kubernetes/kube-state-metrics/pull/1873.

docs/cli-arguments.md
```diff
@@ -46,6 +46,7 @@ Usage of ./kube-state-metrics:
       --metric-opt-in-list string                  Comma-separated list of metrics which are opt-in and not enabled by default. This is in addition to the metric allow- and denylists
       --namespaces string                          Comma-separated list of namespaces to be enabled. Defaults to ""
       --namespaces-denylist string                 Comma-separated list of namespaces not to be enabled. If namespaces and namespaces-denylist are both set, only namespaces that are excluded in namespaces-denylist will be used.
+      --nodename string                            Name of the node that contains the kube-state-metrics pod, only available for resources (pod metrics) that support spec.nodeName fieldSelector. Each kube-state-metrics pod will only exposes metrics related to this node.
```
dgrisonnet, line 49: To make sure that users don't use this option under normal circumstances, could you perhaps mention that this used for sharding? In a similar way as what is done today for the `--pod` option.
https://github.com/kubernetes/kube-state-metrics/blob/master/docs/cli-arguments.md?plain=1#L50

Also, to make it consistent with the existing `--pod` option I would suggest renaming the nodename option to `--node`.

pkg/options/options.go
```diff
@@ -132,3 +134,17 @@ func (o *Options) Parse() error {
...
+
+// Validate validates arguments
+func (o *Options) Validate() error {
+	shardableResource := "pods"
+	if o.NodeName == "" {
+		return nil
+	}
+	for _, x := range o.Resources.AsSlice() {
+		if x != shardableResource {
+			return fmt.Errorf("Resource %s can't be sharding by field selector spec.nodeName", x)
```
dgrisonnet, line 146:
```suggestion
			return fmt.Errorf("Resource %s can't be sharded by field selector spec.nodeName", x)
```

pkg/options/types.go
```diff
@@ -104,6 +104,54 @@ func (r *ResourceSet) Type() string {
...
 
+// NodeNameType represents a nodeName to query from.
+type NodeNameType string
+
+// GetNodeNameFieldSelector returns a nodename field selector.
+func (n *NodeNameType) GetNodeNameFieldSelector() string {
+	if string(*n) != "" {
+		return fields.OneTermEqualSelector("spec.nodeName", string(*n)).String()
+	}
+	return fields.Nothing().String()
```
dgrisonnet, line 115: it would be cleaner to use the shorthand `EmptyFieldSelector()` here

dgrisonnet (commented):
A few comments, but the implementation looks good to me, great job @CatherineF-dev!

README.md
```diff
@@ -235,6 +236,36 @@ This way of deploying shards is useful when you want to manage KSM shards throug
...
 The downside of using an auto-sharded setup comes from the rollout strategy supported by `StatefulSet`s. When managed by a `StatefulSet`, pods are replaced one at a time with each pod first getting terminated and then recreated. Besides such rollouts being slower, they will also lead to short downtime for each shard. If a Prometheus scrape happens during a rollout, it can miss some of the metrics exported by kube-state-metrics.
 
+### Daemonset sharding for pod metrics
+
+For pod metrics, they can be sharded per node with the following flag:
+* `--nodename`
+
+Each kube-state-metrics pod uses FieldSelector (spec.nodeName) to watch/list pod metrics only on the same node.
+
+A daemonset kube-state-metrics example:
```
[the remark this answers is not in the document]
CatherineF-dev: Yes. In adding example in https://github.com/kubernetes/kube-state-metrics/pull/1873, which needs a valid [kube-state-metrics image version]( https://github.com/kubernetes/kube-state-metrics/pull/1873/files#diff-ccfe7f5f6d632e90aa1610bbe66063b81c593bd8fff4e04342653f89c078424eR23).

docs/cli-arguments.md
CatherineF-dev: Done.

dgrisonnet:
Awesome work!

/lgtm
/unhold

PR #1865, opened 2022-10-20
Harden and add gosec linter

mrueg:
Remediate a couple of findings from the gosec linter
G104: Errors unhandled.
G109: Potential Integer overflow made by strconv.Atoi result conversion to int16/32 G112: Potential Slowloris Attack because ReadHeaderTimeout is not configured in the http.Server G304: Potential file inclusion via variable
G601: Implicit memory aliasing in for loop.

I've ignored G104 Error logging on a couple of other items, this remains a TODO.

None
Fixes #

### DIFF
--- a/internal/store/job.go
+++ b/internal/store/job.go
@@ -212,7 +212,8 @@ func jobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 					}
 				}
 
-				for _, condition := range j.Status.Conditions {
+				for _, c := range j.Status.Conditions {
+					condition := c
 					if condition.Type == v1batch.JobFailed {
 						reasonKnown := false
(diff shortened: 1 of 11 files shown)


internal/store/job.go
```diff
@@ -212,7 +212,8 @@ func jobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 					}
 				}
 
-				for _, condition := range j.Status.Conditions {
+				for _, c := range j.Status.Conditions {
```
fpetkovski, line 215: Is this something that was also caught by the new linter?

mrueg: This is G601: Implicit memory aliasing in for loop.

PR #1873, opened 2022-10-31
docs: Daemonset sharding example

CatherineF-dev:
**What this PR does / why we need it**: provide scaling mode example for https://github.com/kubernetes/kube-state-metrics/pull/1864 

Fixes https://github.com/kubernetes/kube-state-metrics/issues/1863

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -134,7 +134,7 @@ examples/prometheus-alerting-rules/alerts.yaml: jsonnet $(shell find jsonnet | g
 	mkdir -p examples/prometheus-alerting-rules
 	jsonnet -J scripts/vendor scripts/mixin.jsonnet | gojsontoyaml > examples/prometheus-alerting-rules/alerts.yaml
 
-examples: examples/standard examples/autosharding mixin
+examples: examples/standard examples/autosharding examples/daemonsetsharding mixin
 
 examples/standard: jsonnet $(shell find jsonnet | grep ".libsonnet") scripts/standard.jsonnet scripts/vendor VERSION
 	mkdir -p examples/standard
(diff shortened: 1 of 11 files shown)


CatherineF-dev:
Also need to add a deployment which watches other resources.

examples/standard/deployment.yaml
```diff
@@ -1,5 +1,5 @@
 apiVersion: apps/v1
-kind: Deployment
+kind: Daemonset
```
mrueg, line 2: I believe this should probably stay a deployment?

fpetkovski:
Do we generate examples from jsonnet, or are they managed separately? I noticed that we use an older image version and keeping the examples up to date will need some work if they're not auto-generated.

examples/standard/deployment.yaml
CatherineF-dev: Yes, I was testing that. I should add a TODO(revert to Deployment after testing) there.

fpetkovski:
Thanks, looks great so far. I just wonder if it's possible to name the folder `daemonset-sharding` instead of `daemonsetSharding` to consistency with file names. Not a blocker though, I understand this can be a limitation of how we generate examples from jsonnet.

mrueg:
If we don't generate it with jsonnet, will container image versions need to be updated manually?

CatherineF-dev:
Yes, it has a TODO (use jsonnet to generate yaml) in [previous comment](https://github.com/kubernetes/kube-state-metrics/pull/1873#issuecomment-1418350442). Will add jsonnet in this PR.

The generated Yamls are similar with/without jsonnet.

CatherineF-dev:
Have fixed presubmits, ready for review~

mrueg:
/lgtm

Thanks @CatherineF-dev

PR #1874, opened 2022-11-02
Bump dependencies

mrueg:
* Bumps go dependencies
* Bumps golang to 1.19.3
* Bumps e2e tests to test against k8s 1.25.0

This is in preparation of the next ksm release. 
I omitted the exporter-toolkit bump since it requires more code changes.


fpetkovski (commented):
/lgtm
/hold

go.mod
```diff
@@ -92,4 +93,4 @@ require (
 	sigs.k8s.io/yaml v1.3.0 // indirect
 )
 
-go 1.17
+go 1.18
```
rexagod, line 96: Shouldn't this be `1.19`?

rexagod: Also the vulns addressed in https://github.com/kubernetes/kube-state-metrics/pull/1878#issue-1434848166 will be fixed when moving to `1.19`.

mrueg: Doesn't need to be go1.19, this defines the minimal version that we support. So we can be a bit more liberal here.

rexagod: ACK, sounds good. However, would it be a good idea to move to `1.19.3` as it would address all 2 of our vulns?

mrueg: it has happened here: https://github.com/kubernetes/kube-state-metrics/pull/1874/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R19 

It looks like actions/setup-go@v3 doesn't consume it yet for CI

rexagod: Ah, makes sense! 👍🏼

PR #1878, opened 2022-11-03
Detect vulns actually used by KSM

rexagod:
Add a cron-job to detect vulns that are actually present in the code that KSM uses (for eg., the functions and procedures it actually calls, etc.), and not just false positives due to considering the entire package in general. cc @dgrisonnet 

Caveats: https://pkg.go.dev/golang.org/x/vuln@v0.0.0-20221103225512-4f561ca73b59/cmd/govulncheck#hdr-Limitations.

***

Visual: https://github.com/kubernetes/kube-state-metrics/pull/1883#issuecomment-1303472700.

PR #1879, opened 2022-11-03
do not merge: testing security checks

rexagod:


mrueg:
This should rather be a scheduled job and not something that is run on a PR base as it will break unrelated benign changes.

My personal opinion: these scanners in general are too noisy output right now for this project (we're effectively only using gcr.io/distroless/static container and golang).
If there's a need to have a scanner, we should look into something like https://pkg.go.dev/golang.org/x/vuln/cmd/govulncheck which only checks the code in path.

mrueg:
As your PR title is do not merge, I'll add a 

/hold

rexagod:
Closing this as this was only for visualization purposes of the changes introduced in #1878, all related changes will be pushed there.

Issue #1871, opened 2022-10-26
labelFromKey is just for Gauge, other types (eg. Info) don't have the attribute 


ningan123:
**What would you like to be added**:
I need a labelFromKey for Info, not just Gauge.

**Why is this needed**:
If a CR (eg. Foo) is like this, 
```
kind: Foo
apiVersion: myteam.io/vl
metadata:
    annotations:
        bar: baz
        qux: quxx
    labels:
        foo: bar
    name: foo
spec:
    version: v1.2.3
    order:
        - id: 1
          value: true
        - id: 3
          value: false
    replicas: 1
status:
    phase: Pending
    active:
        type-a: 1   # focus on this
        type-b: 3   # focus on this
```
we can get the metric label using the following YAML:
```
kind: CustomResourceStateMetrics
spec:
  resources:
    - groupVersionKind:
        group: myteam.io
        kind: "Foo"
        version: "v1"
      metrics:
        - name: "active"
          help: "Foo active"
          each:
            type: Gauge
            gauge:
              path: [status, active]
              labelFromKey: type
```
The output metrics are like this:
```
kube_xxx_active{type=type-a} 1
kube_xxx_active{type=type-b} 3
```

However, if a CR (eg. Foo) is like this:

```
kind: Foo
apiVersion: myteam.io/vl
metadata:
    annotations:
        bar: baz
        qux: quxx
    labels:
        foo: bar
    name: foo
spec:
    version: v1.2.3
    order:
        - id: 1
          value: true
        - id: 3
          value: false
    replicas: 1
status:
    phase: Pending
    active:
        type-a: running  # focus on this
        type-b: failed   # focus on this
```
Because the value of  "type-a" is not a number, we need to use Info, not Gauge.  How can I get a metric with the label "type=type-a"?

**Describe the solution you'd like**
1) I need a labelFromKey for Info to get a metric with the label "type=type-a"
2) I want to get the value of "type-a" to use as a label "state=running"

The expected YAML is like this:
```
kind: CustomResourceStateMetrics
spec:
  resources:
    - groupVersionKind:
        group: myteam.io
        kind: "Foo"
        version: "v1"
      metrics:
        - name: "active"
          help: "Foo active"
          each:
            type: Info
            info:
              path: [status, active]
              labelFromKey: type
              labelsFromPath:
                state: [???]  # I don't know how to write it
```

The expected output metrics are like this:
```
kube_xxx_active{type=type-a, state=running} 1
[...]

Issue #1868, opened 2022-10-25
"LabelFromKey" not avaliable


hdxj-kaoy:
`labelFromKey` does not work when I use the config below

`kube-state-metrics --port=8880 --telemetry-port=8881 --kubeconfig=/config --custom-resource-state-config-file=/crsm.yaml`

crsm.yaml
```
kind: CustomResourceStateMetrics
spec:
  resources:
    - groupVersionKind:
        group: bbb
        kind: "xxx"
        version: "v1alpha1"
      labelsFromPath:
        name: [metadata, name]
      metrics:
        - name: "np"
          help: "xxx"
          each:
            type: Gauge
            gauge:
              path: [status, nodeProgress]
              # does not work
              labelFromKey: type
```

CR looks like:
```
apiVersion: bbb/v1alpha1
kind: xxx
  ...
  ...
  ...
status:
  np:
    key1: 100
    key2: 100
```

I got :
```
kube_bbb_v1alpha1_xxx_np{name="aaa"} 100
kube_bbb_v1alpha1_xxx_np{name="aaa"} 100
```
instead of 
```
kube_bbb_v1alpha1_xxx_np{name="aaa",type="key1"} 100
kube_bbb_v1alpha1_xxx_np{name="aaa",type="key2"} 100
```

After checking the source code and changing 

https://github.com/kubernetes/kube-state-metrics/blob/a3a509acdff1f525c3ca2ff76344d0cedafd8a60/pkg/customresourcestate/registry_factory.go#L156-L160

to

```
 return &compiledGauge{ 
 	compiledCommon: *cc, 
 	ValueFrom:      valueFromPath, 
 	NilIsZero:      m.Gauge.NilIsZero, 
	LabelFromKey:   m.Gauge.LabelFromKey,
 }, nil 
```
 I get the result I want.

I wonder whether this is intentional or there are other solutions that do not need to modify the source code.

rexagod:
/assign
I don't think this was intentional. This makes sense to me. #1871 is more or less similar to this as well.

mrueg:
v2.7.0 has been released, feel free to reopen if this is not resolved.

PR #1880, opened 2022-11-03
Allow labelFromKey field for all applicable types

rexagod:
Allow `labelFromKey` field for the following types:
* Gauge: Done.
* Info: Done.
* StateSet: N/A (redundant use case, see doc changes for more info).

Fixes: #1871 and #1868.

**How does this change affect the cardinality of KSM**: Increases (we are adding labels from keys, which may increase cardinality).

### DIFF
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -268,6 +268,12 @@ func (c *compiledInfo) Values(v interface{}) (result []eachValue, errs []error) 
 			onError(err...)
 			break
 		}
+		for _, ev := range value {
+			if _, ok := ev.Labels[c.labelFromKey]; ok {
+				onError(fmt.Errorf("labelFromKey (%s) generated labels conflict with labelsFromPath, consider renaming it", c.labelFromKey))
+				continue
+			}
+		}
 		// labelFromKey logic
 		for key := range iter {
 			if key != "" && c.labelFromKey != "" {
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -214,6 +214,10 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
 				onError(fmt.Errorf("[%s]: %w", key, err))
 				continue
 			}
+			if _, ok := ev.Labels[c.labelFromKey]; ok {
+				onError(fmt.Errorf("labelFromKey (%s) generated labels conflict with labelsFromPath, consider renaming it", c.labelFromKey))
+				continue
+			}
 			if key != "" && c.labelFromKey != "" {
 				ev.Labels[c.labelFromKey] = key
 			}
@@ -285,9 +289,7 @@ func (c *compiledInfo) Values(v interface{}) (result []eachValue, errs []error) 
 				})
 			}
 		}
-		if len(result) == 0 {
-			result = value
-		}
+		result = append(result, value...)
 	default:
 		result, errs = c.values(v)
 	}
@@ -301,7 +303,9 @@ func (c *compiledInfo) values(v interface{}) (result []eachValue, err []error) {
 	}
 	value := eachValue{Value: 1, Labels: map[string]string{}}
 	addPathLabels(v, c.labelFromPath, value.Labels)
-	result = append(result, value)
+	if len(value.Labels) != 0 {
+		result = append(result, value)
+	}
 	return
 }
 
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -262,33 +262,33 @@ func (c *compiledInfo) Values(v interface{}) (result []eachValue, errs []error) 
 			}
 			result = append(result, ev...)
 		}
-	default:
+	case map[string]interface{}:
 		value, err := c.values(v)
 		if err != nil {
 			onError(err...)
 			break
 		}
 		// labelFromKey logic
-		if vv, ok := v.(map[string]interface{}); ok {
-			for key, val := range vv {
-				if key != "" && c.labelFromKey != "" {
-					n, err := toFloat64(val, false)
-					if err != nil {
-						onError(err)
-						continue
-					}
-					result = append(result, eachValue{
-						Labels: map[string]string{
-							c.labelFromKey: key,
-						},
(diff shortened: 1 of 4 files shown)


rexagod:
cc @dgrisonnet @mrueg @fpetkovski

rexagod:
Ready for review.

CatherineF-dev:
Wondering if it's possible to have an E2E test for CRD metrics.

pkg/customresourcestate/registry_factory.go
```diff
@@ -544,6 +564,7 @@ func famGen(f compiledFamily) generator.FamilyGenerator {
 	errLog := klog.V(f.ErrorLogV)
 	return generator.FamilyGenerator{
 		Name: f.Name,
+		// TODO(@rexagod): This should be dynamic.
```
mrueg, line 567: Should we create an issue for this, so it's not only a TODO sitting in the code?

rexagod: Opened https://github.com/kubernetes/kube-state-metrics/issues/1886.

mrueg: Thanks!

mrueg:
/lgtm
/hold 
for @dgrisonnet @fpetkovski @logicalhan to have another look

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,54 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+							c.labelFromKey: key,
+						},
+						Value: n,
+					})
+				}
+			}
+		}
+		if len(result) == 0 {
+			result = value
+		}
```
dgrisonnet, line 291: Could we perhaps split that into two different cases, one for `map[string]interface{}` and the other for `default`, having it embedded into the default case confused me at first, I would prefer a logic similar to the one with have today with the gauge as it is easier to understand.

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,54 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
 		}
-		return
+	case map[string]interface{}:
+		value, err := c.values(v)
+		if err != nil {
+			onError(err...)
+			break
+		}
+		// labelFromKey logic
+		for key, val := range v.(map[string]interface{}) {
```
logicalhan, line 272:
```suggestion
		for key, val := range iter) {
```

logicalhan (commented):
/lgtm

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,54 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+				n, err := toFloat64(val, false)
+				if err != nil {
+					onError(err)
+					continue
+				}
+				result = append(result, eachValue{
+					Labels: map[string]string{
+						c.labelFromKey: key,
+					},
+					Value: n,
```
dgrisonnet, line 283: value should always be 1 for info metrics

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,54 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+			if key != "" && c.labelFromKey != "" {
+				n, err := toFloat64(val, false)
+				if err != nil {
+					onError(err)
+					continue
+				}
+				result = append(result, eachValue{
+					Labels: map[string]string{
+						c.labelFromKey: key,
+					},
```
dgrisonnet, line 282: this new label needs to be appended to `value.Labels` otherwise we will be missing the labels related to the path

rexagod: Not sure IIUC, but it seems to be done here: https://github.com/rexagod/kube-state-metrics/blob/1871-1868/pkg/customresourcestate/registry_factory.go#L279-L289.

dgrisonnet: it is done correctly if you don't have `labelFromKey` set, but when it is set, you are manually setting label map of the result to just `c.labelFromKey: key` and thus don't take into account the list of labels that was previously created when calling `c.values(v)`.

dgrisonnet: The thing is that you either only add the eachValue from the labelFromKey logic or the one computed by the `values()` method so both are never present together in the final list of eachValue that you return

rexagod: I added logic to check for the `labelFromKey` value in the set of label keys collected beforehand. PTAL the latest fixup commit.

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,55 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+		if err != nil {
+			onError(err...)
+			break
+		}
+		for _, ev := range value {
+			if _, ok := ev.Labels[c.labelFromKey]; ok {
+				onError(fmt.Errorf("labelFromKey (%s) generated labels conflict with labelsFromPath, consider renaming it", c.labelFromKey))
+				continue
+			}
+		}
```
dgrisonnet, line 276: That's a good check, it might be worth to also add it to the gauge type

pkg/customresourcestate/registry_factory.go
```diff
@@ -257,22 +244,55 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+					Labels: map[string]string{
+						c.labelFromKey: key,
+					},
+					Value: 1,
+				})
+			}
+		}
+		if len(result) == 0 {
+			result = value
+		}
```
dgrisonnet, line 290: This is what I was referring to in https://github.com/kubernetes/kube-state-metrics/pull/1880#discussion_r1017494200, we need value to be added to the list everytime no? Otherwise the result will not contain path labels when we already added some stuff to the list.
```suggestion
	        result = append(result, value)
```

rexagod: Makes sense, patched.

mrueg:
/lgtm

/hold cancel

Thanks for your contribution @rexagod !

PR #1890, opened 2022-11-12
Add node deletionTimestamp metric

### LINKED ISSUE #1889: Node Termination/Removal Time
A timestamp which records when a Node was removed from the cluster. 
**Why is this needed**:
So as to track the performance and history of the physical infrastructure supporting the cluster, for alerting and monitoring purposes. Additionally this would be useful for things like auditing. 
Metrics would be exposed similar to the `kube_pod_deletion_timestamp` for pods, just for nodes. 
**Additional context**

I'm not totally sure this is even possible within Kubernetes or if more information would be needed. I just noticed that there is no termination metric inside the `Nodes` category in the docs: https://github.com/kubernetes/kube-state-metrics/blob/master/docs/node-metrics.md

Additionally I have a question open on stack overflow about this request/issue 
https://stackoverflow.com/questions/74351702/prometheus-scrape-node-terminate-time

rexagod:
**What this PR does / why we need it**: Adds deletionTimestamp metric, for nodes.

**How does this change affect the cardinality of KSM**: No change (no variable labels were introduced).

**Which issue(s) this PR fixes**: Fixes #1889.

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -44,9 +44,10 @@ var (
 
 func nodeMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator {
 	return []generator.FamilyGenerator{
-		createNodeCreatedFamilyGenerator(),
-		createNodeInfoFamilyGenerator(),
 		createNodeAnnotationsGenerator(allowAnnotationsList),
+		createNodeCreatedFamilyGenerator(),
+		createNodeDeletionTimestampFamilyGenerator(),
+		createNodeInfoFamilyGenerator(),
 		createNodeLabelsGenerator(allowLabelsList),
 		createNodeRoleFamilyGenerator(),
 		createNodeSpecTaintFamilyGenerator(),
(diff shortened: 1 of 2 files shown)


mrueg:
/approve

Thanks for your contribution!

internal/store/node.go
```diff
@@ -57,6 +58,29 @@ func nodeMetricFamilies(allowAnnotationsList, allowLabelsList []string) []genera
 	}
 }
 
+func createNodeDeletionTimestampFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGeneratorWithStability(
+		"kube_node_deletion_timestamp",
```
CatherineF-dev, line 63: Do we want to set this metric as Alpha/Experimental first, and then promote this into Stable?

Because, labels/metric name can't be changed if it's a stable metric.

mrueg: Good catch! @rexagod can you introduce it as experimental?

rexagod: ACK, fixed.

PR #1894, opened 2022-11-14
update kube_node_status_{capacity/allocatable} doc to clarify difference

### LINKED ISSUE #1859: Improve kube_node_status_{allocatable,capacity} description in the doc
I'd like someone to explain the difference between _kube_node_status_allocatable_ and _kube_node_status_capacity_ metrics.

**Why is this needed**:
The names are confusing. Existing description doesn't give any clue why those return the same values in most cases.

I'd add a few words and links. For example, for allocatable I'd attach this one https://kubernetes.io/docs/tasks/administer-cluster/reserve-compute-resources/
For capacity I'd put a few more words telling those metrics describe how much resources a node has in general.

**Additional context**

exitflynn:
**What this PR does / why we need it**: The current description for `kube_node_status_capacity` and `kube_node_status_allocatable` uses the terms capacity and allocatable. What differentiates them is not clear from the description and is a cause of confusion to those who haven't encountered these terms before and are unaware of their definition in this context.

**How does this change affect the cardinality of KSM**: does not change cardinality

**Which issue(s) this PR fixes** :
Fixes #1859


dgrisonnet:
Thank you for the contribution @exitflynn! :slightly_smiling_face: 

/lgtm
/approve

PR #1909, opened 2022-11-25
endpoint: Rename createEndpointsListWatch() to createEndpointListWatch()

mrueg:
Small change for consistency.

All other ListWatch functions use the singular for the resource they watch in their function name.


mrueg:
Ah nvm, this makes sense because the resource is v1.Endpoints{}

PR #1910, opened 2022-11-25
Add Metrics for EndpointSlices

mrueg:
Implements https://pkg.go.dev/k8s.io/api/discovery/v1#EndpointSlice
Metrics from endpointslices can be used to identify if specific pods are part of an endpoint and thus discoverable through a service.

```
kube_endpointslice_info{endpointslice="my-app-75zcw",addresstype="IPv4"} 1
kube_endpointslice_created{endpointslice="my-app-75zcw"} 1.605287362e+09
kube_endpointslice_endpoints{endpointslice="my-app-75zcw",ready="true",serving="true",terminating="false",targetref_kind="Pod",targetref_name="my-app-73967648-75dgx",targetref_namespace="my-namespace",endpoint_nodename="my-node",endpoint_zone="west",address="192.168.1.1"} 1
kube_endpointslice_ports{endpointslice="my-app-75zcw",port_name="http",port_protocol="TCP",port_number="8080"} 1
```

This resource type is disabled by default as they are very verbose and have a high cardinality.

### DIFF
--- a/internal/store/endpointslice.go
+++ b/internal/store/endpointslice.go
@@ -1,0 +1,234 @@
+/*
+Copyright 2022 The Kubernetes Authors All rights reserved.
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+    http://www.apache.org/licenses/LICENSE-2.0
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package store
+
+import (
+	"context"
+	"strconv"
+
+	basemetrics "k8s.io/component-base/metrics"
+
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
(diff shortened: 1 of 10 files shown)


fpetkovski (approved):
/lgtm
/hold

internal/store/endpointslice.go
```diff
@@ -0,0 +1,234 @@
...
+						Value: float64(s.CreationTimestamp.Unix()),
+					})
+				}
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_endpointslice_endpoints",
```
dgrisonnet, line 78: I think we are plumbing too maybe information inside of this metric. What would you think about moving the conditions, targetref and addresses into their own metrics? There are some precedence to that kind of split with statuses metrics and `kube_endpoint_address`.

mrueg: I decided to combine everything in it that is in https://pkg.go.dev/k8s.io/api/discovery/v1#Endpoint (which is used in 
https://pkg.go.dev/k8s.io/api/discovery/v1#EndpointSlice). This would make it easier for users to query it instead of combining two or more metrics. Do you see any downsides?

logicalhan (commented):
/lgtm

PR #1916, opened 2022-11-29
Add metrics for config file changes

### LINKED ISSUE #1893: Add support for meta-metrics
**What would you like to be added**: Add meta-metrics that hint towards the (hot-)reload status of KSM.

**Why is this needed**: This could be a initial effort towards a more dedicated way to introduce helpful meta-metrics for monitoring KSM itself (https://github.com/kubernetes/kube-state-metrics/pull/1827#issuecomment-1311703079, as Alertmanager does).

mrueg:
Provides metrics for config file updates when viper picks up a new config.

This uses code pieces from prometheus/alertmanager in https://github.com/prometheus/alertmanager/blob/main/config/coordinator.go#LL56C26-L56C26 licensed under Apache-2.0.

kube_state_metrics_config_hash 4.0061079457904e+13 kube_state_metrics_config_last_reload_success_timestamp_seconds 1.6697483049487052e+09 kube_state_metrics_config_last_reload_successful 1

Adds three metric series to the ksm exporter, not the kubernetes exporter. 

Fixes #1893

### DIFF
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -18,6 +18,8 @@ package app
 
 import (
 	"context"
+	"crypto/md5" //nolint:gosec
+	"encoding/binary"
 	"fmt"
 	"net"
 	"net/http"
@@ -99,16 +101,32 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
 			ConstLabels: prometheus.Labels{"handler": "metrics"},
 		}, []string{"method"},
 	)
+	configHash := promauto.With(ksmMetricsRegistry).NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_config_hash",
+			Help: "Hash of the currently loaded configuration.",
+		}, []string{"type", "filename"})
+	configSuccess := promauto.With(ksmMetricsRegistry).NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_last_config_reload_successful",
+			Help: "Whether the last configuration reload attempt was successful.",
+		}, []string{"type", "filename"})
+	configSuccessTime := promauto.With(ksmMetricsRegistry).NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_last_config_reload_success_timestamp_seconds",
+			Help: "Timestamp of the last successful configuration reload.",
+		}, []string{"type", "filename"})
+
 	storeBuilder.WithMetrics(ksmMetricsRegistry)
 
-	got := options.GetOptsConfigFile(*opts)
+	got := options.GetConfigFile(*opts)
 	if got != "" {
-		optsConfigFile, err := os.ReadFile(filepath.Clean(got))
+		configFile, err := os.ReadFile(filepath.Clean(got))
 		if err != nil {
 			return fmt.Errorf("failed to read opts config file: %v", err)
 		}
 		// NOTE: Config value will override default values of intersecting options.
-		err = yaml.Unmarshal(optsConfigFile, opts)
+		err = yaml.Unmarshal(configFile, opts)
 		if err != nil {
 			// DO NOT end the process.
 			// We want to allow the user to still be able to fix the misconfigured config (redeploy or edit the configmaps) and reload KSM automatically once that's done.
@@ -116,7 +134,13 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
 			// Wait for the next reload.
 			klog.Infof("misconfigured config detected, KSM will automatically reload on next write to the config")
 			klog.Infof("waiting for config to be fixed")
+			configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(0)
 			<-ctx.Done()
+		} else {
+			configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(1)
+			configSuccessTime.WithLabelValues("config", filepath.Clean(got)).SetToCurrentTime()
+			hash := md5HashAsMetricValue(configFile)
+			configHash.WithLabelValues("config", filepath.Clean(got)).Set(hash)
 		}
 	}
 	var resources []string
@@ -370,4 +394,15 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
              </html>`))
 	})
 	return mux
+}
+
+// md5HashAsMetricValue creates an md5 hash and returns the most significant bytes that fit into a float64
+// Taken from https://github.com/prometheus/alertmanager/blob/6ef6e6868dbeb7984d2d577dd4bf75c65bf1904f/config/coordinator.go#L149
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data) //nolint:gosec
+	// We only want 48 bits as a float64 only has a 53 bit mantissa.
+	smallSum := sum[0:6]
+	bytes := make([]byte, 8)
+	copy(bytes, smallSum)
+	return float64(binary.LittleEndian.Uint64(bytes))
 }
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -62,8 +62,8 @@ type Options struct {
 	cmd *cobra.Command
 }
 
-// GetOptsConfigFile is the getter for --options-config-file value.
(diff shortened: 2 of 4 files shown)


pkg/options/options.go
```diff
@@ -62,7 +62,7 @@ type Options struct {
 	cmd *cobra.Command
 }
 
-// GetOptsConfigFile is the getter for --options-config-file value.
+// GetOptsConfigFile is the getter for --config value.
```
mrueg, line 65: @rexagod Let me know if you think we should rename the function as well here

rexagod: Yep, makes sense! 👍

pkg/app/server.go
```diff
@@ -371,3 +395,12 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
 	})
 	return mux
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data) //nolint:gosec
```
rexagod, line 400: Just wondering, why is the linter not happy with this?

pkg/app/server.go
```diff
@@ -371,3 +395,12 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
...
+	// We only want 48 bits as a float64 only has a 53 bit mantissa.
+	smallSum := sum[0:6]
```
rexagod, line 402: What about `0:7` to include all 53 bits of the mantissa? We'd waste a bit, sure, but this would make sure the hash is more unique.

pkg/app/server.go
```diff
@@ -371,3 +395,12 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
...
+	bytes := make([]byte, 8)
+	copy(bytes, smallSum)
+	return float64(binary.LittleEndian.Uint64(bytes))
```
rexagod, line 405: +1! Machine endianness would've caused the final result to differ due to the multi-byte intermediary.

rexagod (commented):
These metrics will be very useful, thanks Manuel!

pkg/app/server.go
```diff
@@ -371,3 +395,12 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
 	})
 	return mux
 }
+
+func md5HashAsMetricValue(data []byte) float64 {
+	sum := md5.Sum(data) //nolint:gosec
+	// We only want 48 bits as a float64 only has a 53 bit mantissa.
```
rexagod, line 401: Would _significand_ be a better term to use here (in place of mantissa)? PTAL https://developer.apple.com/documentation/swift/float/significand.

pkg/app/server.go
```diff
@@ -99,9 +101,25 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
...
 		}, []string{"method"},
 	)
+	configHash := promauto.With(ksmMetricsRegistry).NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_config_hash",
+			Help: "Hash of the currently loaded configuration.",
+		})
+	configSuccess := promauto.With(ksmMetricsRegistry).NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_config_last_reload_successful",
```
logicalhan, line 111: I'd prefer `kube_state_metrics_last_config_reload_successful`

pkg/app/server.go
mrueg: md5 is considered a weak cryptographic hash function and gosec has the module on a warn list. Usage here is fine though, as we're checking for file contents and are not hashing passwords or similar.

logicalhan: md5 is cryptographically broken, so the linter doesn't like uses of it. But for checksumming it should be fine.

pkg/app/server.go
logicalhan: I prefer mantissa.

pkg/app/server.go
logicalhan: I'm not sure this will matter much practically.

pkg/app/server.go
mrueg: I tend to agree, in future we might have other config files as well though, that's why I added the infix.
Another config we might want to create metrics for once the hot reload for it is in place as well:
`kube_state_metrics_customresourcestateconfig_last_reload_successful`

Alternatively we could drop it from the name and use:

```
kube_state_metrics_last_reload_successful{type="config"}
kube_state_metrics_last_reload_successful{type="customresourcestateconfig"}
```

fpetkovski: I'd vote for using labels, seems to reduce verbosity and users can also monitor all config updates through a single metric.

pkg/app/server.go
```diff
@@ -99,9 +101,25 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
...
+			Help: "Whether the last configuration reload attempt was successful.",
+		})
+	configSuccessTime := promauto.With(ksmMetricsRegistry).NewGauge(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_config_last_reload_success_timestamp_seconds",
```
CatherineF-dev, line 116: How about `kube_state_metrics_last_config_uptime`?
So that we can easily know whether there are some changes to config recently.

It equals to current_timestamp - kube_state_metrics_config_last_reload_success_timestamp_seconds

pkg/app/server.go
mrueg: I took the function from the alertmanager repository , so I would suggest to keep the delta to a minimum

pkg/app/server.go
mrueg: I prefer to keep it as it, as a timestamp over a counter is usually preferred and has a couple of advantages:
- few static values inside the TSDB for this metric
- Resets can also be detected if the timestamp changes
- You never have the same counter twice.
See also: https://prometheus.io/docs/practices/instrumentation/#timestamps-not-time-since

mrueg:
```
# HELP kube_state_metrics_config_hash Hash of the currently loaded configuration.
# TYPE kube_state_metrics_config_hash gauge
kube_state_metrics_config_hash{filename="config.yml",type="config"} 2.76423981094048e+14
# HELP kube_state_metrics_last_config_reload_success_timestamp_seconds Timestamp of the last successful configuration reload.
# TYPE kube_state_metrics_last_config_reload_success_timestamp_seconds gauge
kube_state_metrics_last_config_reload_success_timestamp_seconds{filename="config.yml",type="config"} 1.669900259628547e+09
# HELP kube_state_metrics_last_config_reload_successful Whether the last configuration reload attempt was successful.
# TYPE kube_state_metrics_last_config_reload_successful gauge
kube_state_metrics_last_config_reload_successful{filename="config.yml",type="config"} 1
```

So this is how metrics currently look like. Any thoughts on the config hash? We could make the value static (=1) and add it as a label instead.

rexagod:
Seems fine to me to keep it as the value.

rexagod:
@mrueg How about allowing folks to switch between YAML and JSON? We can add more decoders as well in the future.

Changing the unmarshalling behavior for specific flags will have to be followed-up by some documentation that specifies the format these should be in, diverging from the format that one would assume by looking at the types, which may lead to issues requesting the current behavior, or interchangeability between older and newer formats.

pkg/app/server.go
```diff
@@ -99,9 +101,25 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
...
+			Name: "config_last_reload_success_timestamp_seconds",
+			Help: "Timestamp of the last successful configuration reload.",
+		}, []string{"type", "filename"})
+
 	storeBuilder.WithMetrics(ksmMetricsRegistry)
 
-	got := options.GetOptsConfigFile(*opts)
+	got := options.GetConfigFile(*opts)
 	if got != "" {
 		optsConfigFile, err := os.ReadFile(filepath.Clean(got))
```
rexagod, line 124: Sorry for the last minute blocker, but any chance we can rename this, and all such occurrences to just `configFile`, in accordance with the same change that was done in the PR earlier: https://github.com/kubernetes/kube-state-metrics/pull/1916#discussion_r1035171754?

mrueg: Sure, done! :)

mrueg:
/hold cancel

Let's proceed here

PR #1921, opened 2022-12-02
Remove duplicated MetricsWriter implementation

dgrisonnet:

Simplify the implementation of the MetricsWriter to avoid code duplication between single and multi stores scenarios.

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -233,12 +233,12 @@ func (b *Builder) WithAllowLabels(labels map[string][]string) error {
 // Build initializes and registers all enabled stores.
 // It returns metrics writers which can be used to write out
 // metrics from the stores.
-func (b *Builder) Build() []metricsstore.MetricsWriter {
+func (b *Builder) Build() metricsstore.MetricsWriterList {
 	if b.familyGeneratorFilter == nil {
 		panic("familyGeneratorFilter should not be nil")
 	}
 
-	var metricsWriters []metricsstore.MetricsWriter
+	var metricsWriters metricsstore.MetricsWriterList
 	var activeStoreNames []string
 
 	for _, c := range b.enabledResources {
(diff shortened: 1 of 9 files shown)


pkg/builder/types/interfaces.go
```diff
@@ -51,7 +51,7 @@ type BuilderInterface interface {
 	DefaultGenerateStoresFunc() BuildStoresFunc
 	DefaultGenerateCustomResourceStoresFunc() BuildCustomResourceStoresFunc
 	WithCustomResourceStoreFactories(fs ...customresource.RegistryFactory)
-	Build() []metricsstore.MetricsWriter
+	Build() []*metricsstore.MetricsWriter
```
dgrisonnet, line 54: Note that since this is changing the signature of one of the function of the `BuilderInterface` it might break projects using kube-state-metrics as a library.

That said, I don't think this should be blocking this change since:
1. Kube-state-metrics isn't a library
2. Consumers of the library should switch to custom-state-metrics

dgrisonnet: Opened https://github.com/kubernetes/kube-state-metrics/discussions/1922 to discuss that particular topic

internal/store/builder.go
```diff
@@ -233,24 +233,20 @@ func (b *Builder) WithAllowLabels(labels map[string][]string) error {
 // Build initializes and registers all enabled stores.
 // It returns metrics writers which can be used to write out
 // metrics from the stores.
-func (b *Builder) Build() []metricsstore.MetricsWriter {
+func (b *Builder) Build() []*metricsstore.MetricsWriter {
```
mrueg, line 236: should we create a type for the return value like MetricsWriterList?

pkg/metrics_store/metrics_writer.go
```diff
@@ -46,7 +40,7 @@ func NewMultiStoreMetricsWriter(stores []*MetricsStore) MetricsWriter {
 //
 // WriteAll writes metrics so that the ones with the same name
 // are grouped together when written out.
-func (m MultiStoreMetricsWriter) WriteAll(w io.Writer) {
+func (m MetricsWriter) WriteAll(w io.Writer) {
```
mrueg, line 43: unrelated to this change, can we improve error handling in this function if the write fails?

dgrisonnet: That a good idea, I added the error handling in a new commit

dgrisonnet:
Forced-pushed to move the new MetricsWriterList type inside of the metricsstore package. I incorrectly added it to the type interface package at first, but that's not where the MetricsWriter type is defined.

logicalhan (approved):
/lgtm

fpetkovski (approved):
Thanks!

/lgtm
/hold cancel

PR #1928, opened 2022-12-08
Reload CustomResourceState Config File on Change

### LINKED ISSUE #1892: Add support for hot-reloading on CRS config changes
**What would you like to be added**: Hot-reloading when CRS config changes are detected.

**Why is this needed**: Allow users to make changes to their config and have KSM pick up on that automatically.

**Describe the solution you'd like**: Extend https://github.com/kubernetes/kube-state-metrics/pull/1827 for CRS config.

mrueg:
This change adds hot reloading support for the customresourcestate config file.

It also resolves a bug in which the customresourcestate config file was
included in the ksm config file, in which it did not get detected.

It also resolves a bug in which customresourcestatemetrics were not
added when resources were non-default resources.

Code for this file change detection was reused from https://github.com/spf13/viper licensed under MIT license.

None
Fixes #1892

### DIFF
--- a/internal/wrapper.go
+++ b/internal/wrapper.go
@@ -21,7 +21,6 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
-	"strings"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
@@ -30,34 +29,25 @@ import (
 	"k8s.io/klog/v2"
 
 	"k8s.io/kube-state-metrics/v2/pkg/app"
-	"k8s.io/kube-state-metrics/v2/pkg/customresource"
-	"k8s.io/kube-state-metrics/v2/pkg/customresourcestate"
 	"k8s.io/kube-state-metrics/v2/pkg/options"
 )
 
 // RunKubeStateMetricsWrapper is a wrapper around KSM, delegated to the root command.
 func RunKubeStateMetricsWrapper(opts *options.Options) {
-	var factories []customresource.RegistryFactory
-	if config, set := resolveCustomResourceConfig(opts); set {
-		crf, err := customresourcestate.FromConfig(config)
-		if err != nil {
-			klog.ErrorS(err, "Parsing from Custom Resource State Metrics file failed")
-			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
-		}
-		factories = append(factories, crf...)
-	}
 
 	KSMRunOrDie := func(ctx context.Context) {
-		if err := app.RunKubeStateMetricsWrapper(ctx, opts, factories...); err != nil {
+		if err := app.RunKubeStateMetricsWrapper(ctx, opts); err != nil {
 			klog.ErrorS(err, "Failed to run kube-state-metrics")
 			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		}
 	}
+
 	ctx, cancel := context.WithCancel(context.Background())
 	if file := options.GetConfigFile(*opts); file != "" {
-		viper.SetConfigType("yaml")
-		viper.SetConfigFile(file)
-		if err := viper.ReadInConfig(); err != nil {
+		cfgViper := viper.New()
+		cfgViper.SetConfigType("yaml")
+		cfgViper.SetConfigFile(file)
+		if err := cfgViper.ReadInConfig(); err != nil {
 			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
 				klog.ErrorS(err, "Options configuration file not found", "file", file)
 			} else {
@@ -65,7 +55,7 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 			}
 			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		}
-		viper.OnConfigChange(func(e fsnotify.Event) {
+		cfgViper.OnConfigChange(func(e fsnotify.Event) {
 			klog.Infof("Changes detected: %s\n", e.Name)
 			cancel()
 			// Wait for the ports to be released.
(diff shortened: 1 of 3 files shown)


pkg/app/server.go
```diff
@@ -157,23 +174,21 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
...
 	case opts.CustomResourcesOnly:
 		// enable custom resource only
-		for _, factory := range factories {
-			resources = append(resources, factory.Name())
-		}
 		klog.InfoS("Used CRD resources only", "resources", resources)
 	default:
-		klog.InfoS("Used resources", "resources", opts.Resources.String())
-		resources = opts.Resources.AsSlice()
+		resources = append(resources, opts.Resources.AsSlice()...)
```
rexagod, line 190: +1!

internal/wrapper.go
```diff
@@ -75,22 +63,76 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
 }
 
-func resolveCustomResourceConfig(opts *options.Options) (customresourcestate.ConfigDecoder, bool) {
-	if s := opts.CustomResourceConfig; s != "" {
-		return yaml.NewDecoder(strings.NewReader(s)), true
-	}
-	if file := opts.CustomResourceConfigFile; file != "" {
-		f, err := os.Open(filepath.Clean(file))
+// WatchConfig is taken and adapted from https://github.com/spf13/viper/blob/b89e554a96abde447ad13a26dcc59fd00375e555/viper.go#L429 MIT-Licensed
+func WatchConfig(filename string, onConfigChange func(fsnotify.Event)) {
```
dgrisonnet, line 83: Couldn't we just create a Viper config for the custom resource config and reuse the original function?

pkg/app/server.go
```diff
@@ -85,11 +87,11 @@ func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options, fact
 // RunKubeStateMetrics will build and run the kube-state-metrics.
 // Any out-of-tree custom resource metrics could be registered by newing a registry factory
 // which implements customresource.RegistryFactory and pass all factories into this function.
-func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
+func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
+	var factories []customresource.RegistryFactory
```
dgrisonnet, line 91: could we move that closer to where it is used in the function?

pkg/app/server.go
```diff
@@ -157,23 +174,21 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories .
 	}
 
 	var resources []string
```
dgrisonnet, line 176: since we know the size of the array, it would be better to initialize to the correct size directly, I would avoid having to resize it multiple times with `append`

mrueg: Good catch, fixed it!

pkg/app/server.go
mrueg: Done!

internal/wrapper.go
mrueg: I'm not sure if Viper is able to handle multiple configs at the same time and independently.

dgrisonnet: We could create new viper instances with `viper.New()` and stop using the global viper config for the general config.

A benefit to reusing viper directly here is that we would be able to use `viper.ReadInConfig()` and `viper.ReadInConfig()` like we do for the general config.

internal/wrapper.go
```diff
@@ -30,67 +29,73 @@ import (
...
-	if file := opts.CustomResourceConfigFile; file != "" {
-		f, err := os.Open(filepath.Clean(file))
+		// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
+		configFile, err := os.ReadFile(filepath.Clean(file))
 		if err != nil {
-			klog.ErrorS(err, "Custom Resource State Metrics file could not be opened")
+			klog.ErrorS(err, "failed to read options configuration file", "file", file)
+		}
+
+		yaml.Unmarshal(configFile, opts)
```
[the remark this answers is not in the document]
mrueg: I believe this would require further logic (as it could also be set via command line), so merging seemed easier (to use the same logic that we have in https://github.com/kubernetes/kube-state-metrics/blob/b7a7070e87ba4299a90587c8c2d5c7647db1ba42/pkg/app/server.go#L129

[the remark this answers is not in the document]
dgrisonnet: sounds good :+1:

dgrisonnet:
Leaving lgtm up to @rexagod 

/approve
/unhold

pkg/app/server.go
```diff
@@ -85,11 +87,10 @@ func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options, fact
 // RunKubeStateMetrics will build and run the kube-state-metrics.
 // Any out-of-tree custom resource metrics could be registered by newing a registry factory
 // which implements customresource.RegistryFactory and pass all factories into this function.
-func RunKubeStateMetrics(ctx context.Context, opts *options.Options, factories ...customresource.RegistryFactory) error {
```
Garrybest, line 88: Hi @mrueg, I found that this parameter `factories ...customresource.RegistryFactory` is deleted from function `RunKubeStateMetrics`.

As #1644 shows, this parameter is for some repository which does not use custom resource config file and wants to add their own metrics collecting logic in their codes. So a user could construct a new repository and vendor KSM as a dependency in `go.mod`. I believe some users have already built their project like this, e.g. https://github.com/kubernetes/kube-state-metrics/pull/1644#issuecomment-990892061.

So do you have any ideas about this scenario?

PR #1929, opened 2022-12-11
Warn on transient-state metrics

### LINKED ISSUE #1828: kube_pod_container_resource.* showing duplicate metrics with and without node label
**What happened**:
for these resource metrics, it is expected that just one metric appears per container, aka one limit and one request metric.

however, we're seeing two of each - one without a `node` label that appears for a brief moment, and then one with a `node` label, which replaces it (see image below)

image

this is problematic because there is a time in point where both metrics co-exist, and that breaks some of our recording rules that expect both of these metrics to be unique. you can see that the without-node metric exists for a brief moment before the with-node metric appears, and subsequently the without-node metric disappears (though you can't really see it dropping off in the screenshot since the with-node one is on top).

i'm curious as to why this happens and why the `node` label is special here.
 

that each of these metrics (`kube_pod_container_resource_requests` and `kube_pod_container_resource_limits`) are unique per container. 

This happens when any pod gets newly created.

**Environment**:
- kube-state-metrics version: 2.2.0
- Kubernetes version (use `kubectl version`): 1.18.20
- Cloud provider or hardware configuration: self-hosted
- Other info:
```
        - /kube-state-metrics
        - --port=9102
        - --telemetry-port=8081
        - --resources=configmaps,cronjobs,daemonsets,deployments,endpoints,horizontalpodautoscalers,jobs,limitranges,namespaces,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets
        - --use-apiserver-cache
        - --metric-labels-allowlist=daemonsets=[*],deployments=[*],jobs=[*],nodes=[*],pods=[*],secrets=[*]
        - -v4
        - --pod=$(POD_NAME)
        - --pod-namespace=$(POD_NAMESPACE)
```

rexagod:
**What this PR does / why we need it**:Warn on showing transient-state metrics for `kube_pod_container_resource_.+`.

**Which issue(s) this PR fixes**: Fixes #1828.

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -18,7 +18,10 @@ package store
 
 import (
 	"context"
+	"reflect"
 	"strconv"
+
+	"k8s.io/klog/v2"
 
 	basemetrics "k8s.io/component-base/metrics"
 	"k8s.io/utils/net"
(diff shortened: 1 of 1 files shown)


rexagod:
Closing in favor of https://github.com/kubernetes/kube-state-metrics/issues/1828#issuecomment-1349533406.

Issue #1792, opened 2022-07-19
Promote experimental metrics


mrueg:
We should take a look at the current metrics and promote experimental metrics to stable.

CatherineF-dev:
Hi team, could we promote kube_pod_container_status_waiting_reason to stable?

This metric is super useful.

cpanato:
what is the process to promote a metric? i would like to help here

rexagod:
@CatherineF-dev Would it be possible to document the promotion process somewhere in the code-base, maybe `docs/developer/guide.md`? This would greatly help newer contributors looking to work on this issue.

suryapandian:
I could also help on this if there is a document on how to

CatherineF-dev:
I think the hard part is how to decide a metric X can be promoted to STABLE. One criteria we want to use is whether metric X is not changed in last X months.

But because we don't have data on the number of users are using metric X. Even though a metric hasn't been changed for 1 year, it might be because no one is using it. If someone begin to use, they might want to make some changes again.

suryapandian:
if there is no change for a metric for 1 year, i would suggest that we could move to stable or plan on deprecating it. no point waiting to see if someone would potentially make some change.

Omswastik-11:
Hi @rexagod and @CatherineF-dev !!
Is this issue is still open ? 
I am a new contributor and want to contribute to KSM as it labeled 'good-first -issue' Can i work on it ?

PR #1941, opened 2023-01-02
promote namespace annotations metrics

### LINKED ISSUE #1940: Promote namespace annotation metrics
Promoting the experimental `kube_namespace_annotations` to STABLE
**Why is this needed**:
Namespaces often store annotations that are essential for our monitoring purposes at AKS, and to truely have end to end observability, we need these in a stable stage

Referencing #1792

Bryce-Soghigian:
Promotes kube_namespace_annotations to stable. For KSM to act as an e2e monitoring solution of kubernetes state, we need this metric to be stable.  Read more about the why in the linked issue.
Does not Change cardinality as metrics are already in the metrics endpoint, we are just promoting them
Fixes #1940

Regarding #1792

### DIFF
--- a/docs/namespace-metrics.md
+++ b/docs/namespace-metrics.md
@@ -2,7 +2,7 @@
 
 | Metric name| Metric type | Labels/tags | Status |
 | ---------- | ----------- | ----------- | ----------- |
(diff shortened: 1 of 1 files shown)


dgrisonnet:
I don't think it make sense to promote annotations/labels metrics because the promotion is meant to make sure that the metric and label names don't change over time. But since these metrics have changing labels, we won't be able to freeze their state.

PR #1942, opened 2023-01-03
prefix GVK labels in CustomResourceMonitoring

bavarianbidi:

This PR will prefix the auto-generated GVK labels for CustomResources with `cr_` to make it a bit more clear that these labels got generated.
It also make it more easier to reflect e.g. a `version` of a `CR` in a metric by just label it `version` instead of some specific `app_version` (or similar).

Fixes #

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -117,7 +117,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_crd_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
+kube_customresource_uptime{customresource_group="myteam.io", kind="Foo", version="v1"} 43.21
 ```
 
 #### Multiple Metrics/Kitchen Sink
@@ -169,8 +169,8 @@ spec:
 Produces the following metrics:
 
 ```prometheus
-kube_crd_ready_count{group="myteam.io", kind="Foo", version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 2
-kube_crd_ready_count{group="myteam.io", kind="Foo", version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 4
+kube_customresource_ready_count{customresource_group="myteam.io", kind="Foo", version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 2
+kube_customresource_ready_count{customresource_group="myteam.io", kind="Foo", version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 4
 ```
 
 ### Metric types
@@ -205,7 +205,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_crd_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
+kube_customresource_uptime{customresource_group="myteam.io", kind="Foo", version="v1"} 43.21
 ```
 
 #### StateSet
@@ -237,9 +237,9 @@ The value will be 1, if the value matches the one in list.
 Produces the metric:
 
 ```prometheus
-kube_crd_status_phase{group="myteam.io", kind="Foo", version="v1", phase="Pending"} 1
-kube_crd_status_phase{group="myteam.io", kind="Foo", version="v1", phase="Bar"} 0
-kube_crd_status_phase{group="myteam.io", kind="Foo", version="v1", phase="Baz"} 0
+kube_customresource_status_phase{customresource_group="myteam.io", kind="Foo", version="v1", phase="Pending"} 1
+kube_customresource_status_phase{customresource_group="myteam.io", kind="Foo", version="v1", phase="Bar"} 0
+kube_customresource_status_phase{customresource_group="myteam.io", kind="Foo", version="v1", phase="Baz"} 0
 ```
 
 #### Info
@@ -269,7 +269,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_crd_version{group="myteam.io", kind="Foo", version="v1", version="v1.2.3"} 1
+kube_customresource_version{customresource_group="myteam.io", kind="Foo", version="v1", version="v1.2.3"} 1
 ```
 
 ### Naming
@@ -291,7 +291,7 @@ spec:
 
 Produces:
 ```prometheus
-myteam_foos_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
+myteam_foos_uptime{customresource_group="myteam.io", kind="Foo", version="v1"} 43.21
 ```
 
 To omit namespace and/or subsystem altogether, set them to the empty string:
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -49,6 +49,40 @@ spec:
           - --resources=certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,foos,horizontalpodautoscalers,ingresses,jobs,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments,verticalpodautoscalers
 ```
 
+It's also possible to configure kube-state-metrics to run in a `custom-resource-mode` only. In addition to specifying one of `--custom-resource-state-config*` flags, you could set `--custom-resource-state-only` to `true`.
+With this configuration only the known custom resources configured in `--custom-resource-state-config*` will be taken into account by kube-state-metrics.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kube-state-metrics
+  namespace: kube-system
+spec:
+  template:
+    spec:
+      containers:
+      - name: kube-state-metrics
+        args:
+          - --custom-resource-state-config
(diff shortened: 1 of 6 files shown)


CatherineF-dev:
I think "group" is fine and it aligns with k8s documentations, cc @rexagod to have another review.

dgrisonnet:
I am a bit on the fence here as I like the current naming because it aligns with the labels we have in the kube-apiserver HTTP metrics. But at the same time I can see how one could want to use a version label for other purposes. Maybe we could prefix them with `crd` similarly to how we prefix the metrics name `kube_crd`?

Do you have an opinion on that @logicalhan?

bavarianbidi:
Don't get me wrong, but it's really nice to get the `group`, `version` and `kind` data as label values in the metric without additional work. 
But the main purpose from a user point of view is to reflect information from CRs. 
As many third-party-components have a `version` field which i'm more interested in than the version of the `CRD` itself (as i already have to define the GVK for a CRD in the configuration) i'm very happy if we find a solution for that.

As an alternative we could also think about an additional configuration value/flag where a user can opt-in/opt-out the auto-generated `GVK` labels.

Beside the origin issue with the `version` field,  i've already stumbled 2 times this morning about the `version` field in a `CR` where i have to decided how to continue.

from `cluster-api-azure`, where an `AzureMachinePool` can reflect the `version` of an image in the `status` field:

```
k explain azuremachinepool.status.image.marketplace.version
KIND:     AzureMachinePool
VERSION:  infrastructure.cluster.x-k8s.io/v1beta1

FIELD:    version <string>

DESCRIPTION:
     Version specifies the version of an image sku. The allowed formats are
     Major.Minor.Build or 'latest'. Major, Minor, and Build are decimal numbers.
     Specify 'latest' to use the latest version of an image available at deploy
     time. Even if you use 'latest', the VM image will not automatically update
     after deploy time even if a new version becomes available.
```

from `flux`, where you can define a `HelmRelease` and point to a specific version:

```
$ k explain helmreleases.spec.chart.spec.version
KIND:     HelmRelease
VERSION:  helm.toolkit.fluxcd.io/v2beta1

FIELD:    version <string>

DESCRIPTION:
     Version semver expression, ignored for charts from v1beta2.GitRepository
     and v1beta2.Bucket sources. Defaults to latest when omitted.

```

dgrisonnet:
> As many third-party-components have a version field which i'm more interested in than the version of the CRD itself (as i already have to define the GVK for a CRD in the configuration) i'm very happy if we find a solution for that.

Yeah this scenario totally makes sense.

> As an alternative we could also think about an additional configuration value/flag where a user can opt-in/opt-out the auto-generated GVK labels.

I wouldn't be in favor of that since I don't think that there are any advantages to removing the GVK labels since they provide key information for barely any overhead.

docs/customresourcestate-metrics.md
```diff
@@ -49,7 +49,7 @@ spec:
           - --resources=certificatesigningrequests,configmaps,cronjobs,daemonsets,deployments,endpoints,foos,horizontalpodautoscalers,ingresses,jobs,limitranges,mutatingwebhookconfigurations,namespaces,networkpolicies,nodes,persistentvolumeclaims,persistentvolumes,poddisruptionbudgets,pods,replicasets,replicationcontrollers,resourcequotas,secrets,services,statefulsets,storageclasses,validatingwebhookconfigurations,volumeattachments,verticalpodautoscalers
 ```
 
-NOTE: The `group`, `version`, and `kind` common labels are reserved, and will be overwritten by the values from the `groupVersionKind` field.
+NOTE: The `cr_group`, `cr_version`, and `cr_kind` common labels are reserved, and will be overwritten by the values from the `groupVersionKind` field.
```
dgrisonnet, line 52: I would prefer if we had some consistency between this prefix and the one in the metric name. Currently the metric name is using the `crd` prefix which might not be well suited. Might be better to go with `kube_cr` or `kube_custom_resources`.

bavarianbidi: do you mean we should use the future prefix in the metric name and in the label itself?
I like `kube_custom_resource` more than `kube_cr`.

CatherineF-dev:
Another way might be kube-state-metrics to support renaming labels for CRs.
So that kube-state-metrics collects HelmRelease.version and renames it into helm_version.

```
$ k explain helmreleases.spec.chart.spec.version
KIND:     HelmRelease
VERSION:  helm.toolkit.fluxcd.io/v2beta1

FIELD:    version <string>

DESCRIPTION:
     Version semver expression, ignored for charts from v1beta2.GitRepository
     and v1beta2.Bucket sources. Defaults to latest when omitted.
```

docs/customresourcestate-metrics.md
dgrisonnet: Yes, I'd like to avoid having `crd` somewhere and `cr` elsewhere, it's better be consistent.

mrueg: I agree, would be nice to have one thing. I believe right now we mix Custom Resource, Custom Resource Definition and Custom Resource State.  `kube_customresource` would align with `kube_horizontalpodautoscaler` (no additional underscores).

Should we rename everything custom resource state into simply custom resource and abbreviate it with "cr" where necessary?

bavarianbidi: i've pushed a new commit (https://github.com/kubernetes/kube-state-metrics/pull/1942/commits/ed85920888a27a984f5f1a65c6e2d7caedee2948) - this should reflect the way we want to go further:

bavarianbidi: FYI: Based on the previous commit i started to improve the e2e tests because they didn't cover the default generated `group`,`version`,`kind` labels atm. will continue on monday morning.

pkg/customresourcestate/custom_resource_metrics_test.go
```diff
@@ -58,18 +130,118 @@ func TestNewCustomResourceMetrics(t *testing.T) {
...
+					Kind:    "Deployment",
+				},
+				ResourceName: "deployments",
+				Families: []compiledFamily{
+					{
+						Name: "apps_deployment_test_metrics",
+						Help: "metrics for testing",
+						Each: &compiledInfo{},
+						Labels: map[string]string{
+							"customresource_group":   "apps",
```
CatherineF-dev, line 212: customresource_group to cr_group?

bavarianbidi: comming from [this comment](https://github.com/kubernetes/kube-state-metrics/pull/1942#discussion_r1061934899)

> I agree, would be nice to have one thing. I believe right now we mix Custom Resource, Custom Resource Definition and Custom Resource State. `kube_customresource` would align with `kube_horizontalpodautoscaler` (no additional underscores).
> 
> Should we rename everything custom resource state into simply custom resource and abbreviate it with "cr" where necessary?

i've started with `customresource` (because i like it a bit more) but as i'm not very opinionated i'm also fine to switch to `cr`

pkg/customresourcestate/config.go
```diff
@@ -64,7 +68,7 @@ type Resource struct {
 func (r Resource) GetMetricNamePrefix() string {
 	p := r.MetricNamePrefix
 	if p == nil {
-		return "kube_crd"
+		return "kube_" + customResourceState
```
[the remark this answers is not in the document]
bavarianbidi: thanks for the hint @mrueg - done

docs/verticalpodautoscaler-metrics.md
```diff
@@ -30,9 +30,9 @@ spec:
             gauge:
               path: [metadata, annotations]
 # This will output the following metric:
-# HELP kube_crd_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations Kubernetes annotations converted to Prometheus labels.
-# TYPE kube_crd_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations gauge
-# kube_crd_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations{namespace="default",target_api_version="autoscaling.k8s.io/v1",target_kind="Deployment",target_name="hamster",verticalpodautoscaler="hamster-vpa"} 123
+# HELP kube_customresource_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations Kubernetes annotations converted to Prometheus labels.
```
mrueg, line 33: It looks like these also use the old naming scheme which we replaced (and this MR updates its fields).

I believe they should look like this: 
```
# HELP kube_customresource_annotations Kubernetes annotations converted to Prometheus labels.
# TYPE kube_customresource_autoscaling_k8s_io_v1_VerticalPodAutoscaler_annotations gauge
# kube_customresource_annotations{customresource_group="autoscaling.k8s.io", customresource_version="v1", customresource_kind="VerticalPodAutoscaler", namespace="default",target_api_version="autoscaling.k8s.io/v1",target_kind="Deployment",target_name="hamster",verticalpodautoscaler="hamster-vpa"} 123
```

docs/customresourcestate-metrics.md
```diff
@@ -205,7 +205,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_crd_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
+kube_customresource_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
```
mrueg, line 208: group, kind, version should be prefixed with customresource_ in this file

pkg/customresourcestate/custom_resource_metrics_test.go
CatherineF-dev: cc @mrueg which label name do you prefer? `cr_group` or `customresource_group`

mrueg: My personal preference is the second one, two letter acronyms can become ambiguous

bavarianbidi:
I've got 2 approvals... Who could give an LGTM to get this merged?

mrueg:
I'll make sure to merge it by the end of this week, want to provide other maintainers the opportunity to give further feedback

dgrisonnet (approved):
Thank you for bearing with us @bavarianbidi :)

/lgtm
/approve

PR #1953, opened 2023-01-14
Support kustomization install

### LINKED ISSUE #1952: Kustomization for standard install
Kustomization for standard install 

**Why is this needed**:
Better installation experience.

PR: https://github.com/kubernetes/kube-state-metrics/pull/1953

kehao95:

Add Kustomization support for the standard install. 
*We have that for [examples/autosharding](https://github.com/kubernetes/kube-state-metrics/tree/main/examples/autosharding) already.*

Also, as instructed in the README the standard installation can be done without having to clone the project locally. 

None

Fixes #1952

**You can test it out before merging**
```
kubectl apply -k github.com/kehao95/kube-state-metrics/examples/standard\?ref=support-kustomization-install --dry-run=client
```


README.md
```diff
@@ -308,7 +308,7 @@ If you want to revise the default configuration for kube-prometheus, for example
 
 #### Kubernetes Deployment
 
-To deploy this project, you can simply run `kubectl apply -f examples/standard` and a Kubernetes service and deployment will be created. (Note: Adjust the apiVersion of some resource if your kubernetes cluster's version is not 1.8+, check the yaml file for more information).
+To deploy this project, you can simply run `kubectl apply -k github.com/kubernetes/kube-state-metrics/examples/standard` and a Kubernetes service and deployment will be created. (Note: Adjust the apiVersion of some resource if your kubernetes cluster's version is not 1.8+, check the yaml file for more information).
```
mrueg, line 311: I have a slight preference to keep the default deployment without kustomize on the doc. Could we keep the `kubectl apply -f` here or will this fail if the kustomization file is there?

kehao95: [Kustomization](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/kustomization/) is considered one standard tool of kubectl already.
I prefer kustomization for the both remote/local install option. You can also install locally with `kubectl apply -k standard`
Otherwise, we imply to the users that they have to clone the project before installation. 


The `kubectl apply -f` will not fall but will generate an error when trying to apply the `kustomization.yaml` to cluster. (other files are applied correctly)
```bash
$ kubectl apply -f standard
clusterrolebinding.rbac.authorization.k8s.io/kube-state-metrics created
clusterrole.rbac.authorization.k8s.io/kube-state-metrics created
deployment.apps/kube-state-metrics created
serviceaccount/kube-state-metrics created
service/kube-state-metrics created
error: resource mapping not found for name: "" namespace: "" from "standard/kustomization.yaml": no matches for kind "Kustomization" in version "kustomize.config.k8s.io/v1beta1"
ensure CRDs are installed first
```

mrueg: This might break other users, see: https://github.com/kubernetes/kube-state-metrics/pull/1603#issuecomment-950156579

I would prefer not to add kustomize to this repository if there's no backwards compatible way.

CatherineF-dev:
/remove-lgtm
/lgtm cancel

`kubectl apply -f` has error messages.

mrueg:
Thinking about it, I can also see that it makes it easier to be consumed for kustomize users this way.
How about  generating a `standard-with-kustomize` folder via `make manifests`, which contains the same files as standard but with an additional kustomize config?

kehao95:
@mrueg How about now?

README.md
```diff
@@ -308,7 +308,7 @@ If you want to revise the default configuration for kube-prometheus, for example
 
 #### Kubernetes Deployment
 
-To deploy this project, you can simply run `kubectl apply -f examples/standard` and a Kubernetes service and deployment will be created. (Note: Adjust the apiVersion of some resource if your kubernetes cluster's version is not 1.8+, check the yaml file for more information).
+To deploy this project, you can simply run `kubectl apply -k github.com/kubernetes/kube-state-metrics/` and a Kubernetes service and deployment will be created. (Note: Adjust the apiVersion of some resource if your kubernetes cluster's version is not 1.8+, check the yaml file for more information).
```
mrueg, line 311: I would still prefer to keep the default instruction without kustomize.

We can have another sentence after that that explains how to install with kustomize.

kehao95: You make the call. Then we can close this PR. You may just commit directly as you like.

PR #1956, opened 2023-01-18
remove broken --version flag, replace by version command

bjorand:

`--version` flag is broken in 2.7.0. It should return the version instead of trying to run kube-state-metrics.

here is the current broken behaviour:
```
$ kube-state-metrics --version
I0118 21:27:02.638112   80415 wrapper.go:98] Starting kube-state-metrics
I0118 21:27:02.640674   80415 server.go:185] "Used default resources"
I0118 21:27:02.641064   80415 types.go:184] "Using all namespaces"
I0118 21:27:02.641174   80415 server.go:218] "Metric allow-denylisting" allowDenyStatus="Excluding the following lists that were on denylist: "
W0118 21:27:02.641423   80415 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
W0118 21:27:02.641503   80415 client_config.go:623] error creating inClusterConfig, falling back to default config: unable to load in-cluster configuration, KUBERNETES_SERVICE_HOST and KUBERNETES_SERVICE_PORT must be defined
E0118 21:27:02.642813   80415 wrapper.go:40] "Failed to run kube-state-metrics" err="failed to create client: invalid configuration: no configuration has been provided, try setting KUBERNETES_MASTER environment variable"
```

with this fix:
```
$ kube-state-metrics --version
kube-state-metrics, version  (branch: , revision: 689a1f5f4fd60cdb8a285d39875b0328923a5cc6)
  build user:
  build date:
  go version:       go1.19.5
  platform:         darwin/amd64
```

Note that I'm also removing useless code.


mrueg:
There are currently two ways to show the version:

via a version command and via a version flag. The `kube-state-metrics version` command still works.
I don't think we need both and rather would remove one of them.
Any thoughts/preferences @kubernetes/kube-state-metrics-maintainers ?

fpetkovski:
Is the help text of the command sufficient for users to find out they need to use `kube-state-metrics version`? I agree that we should not have two ways to show the version.

bjorand:
> Is the help text of the command sufficient for users to find out they need to use `kube-state-metrics version`? I agree that we should not have two ways to show the version.

I agree. I removed the broken --version flag and we now have only the version command (which is documented in the help message)

mrueg:
Can you also remove the version flag from `help.txt`?  You can generate it via ./scripts/generate-help-text.sh

bjorand:
done in 57a196b

mrueg:
I think this change only updated the markdown file and not the help.txt.

bjorand:
> I think this change only updated the markdown file and not the help.txt.

help.txt is in .gitignore

mrueg:
> > I think this change only updated the markdown file and not the help.txt.
> 
> help.txt is in .gitignore

Ah my mistake. I looked at my local checkout.


/approve

Thanks!

PR #1957, opened 2023-01-20
tools: Use own go.mod

mrueg:
This reduces the number of deps that will be fetched for building ksm

None
Fixes #


Makefile
```diff
@@ -150,7 +150,8 @@ scripts/vendor: scripts/jsonnetfile.json scripts/jsonnetfile.lock.json
 
 install-tools:
 	@echo Installing tools from tools.go
-	@cat tools/tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
+	@cd tools && cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go get %
```
dgrisonnet, line 153: why is this needed?

Makefile
```diff
@@ -150,7 +150,8 @@ scripts/vendor: scripts/jsonnetfile.json scripts/jsonnetfile.lock.json
...
+	@cd tools && cat tools.go | grep _ | awk -F'"' '{print $$2}' | xargs -tI % go install %
```
dgrisonnet, line 154: instead of changing directory we could use `go install -mod=readonly -modfile=tools/go.mod`

Makefile
mrueg: The go.mod in the parent directory will record dependencies if we are not changing to a different directory.

dgrisonnet: It shouldn't be the case if we specify "-mod=readonly" as commented below

mrueg: Yep, that way it works. thanks for the suggestion!

dgrisonnet: You're welcome!

dgrisonnet (approved):
/lgtm
/approve

PR #1959, opened 2023-01-23
Add StatefulSet Start Ordinal metrics for KEP-3335

pwschuurman:
Add metrics for new StatefulSet fields from KEP-3335
 * KEP: https://github.com/kubernetes/enhancements/issues/3335

Does not change the cardinality of KSM

### DIFF
--- a/internal/store/statefulset.go
+++ b/internal/store/statefulset.go
@@ -179,6 +179,26 @@ func statefulSetMetricFamilies(allowAnnotationsList, allowLabelsList []string) [
 			}),
 		),
 		*generator.NewFamilyGeneratorWithStability(
+			"kube_statefulset_ordinals_start",
+			"Start ordinal of the StatefulSet.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapStatefulSetFunc(func(s *v1.StatefulSet) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if s.Spec.Ordinals != nil {
+					ms = append(ms, &metric.Metric{
+						Value: float64(s.Spec.Ordinals.Start),
+					})
+				}
+
+				return &metric.Family{
+					Metrics: ms,
+				}
+			}),
+		),
+		*generator.NewFamilyGeneratorWithStability(
 			"kube_statefulset_metadata_generation",
 			"Sequence number representing a specific generation of the desired state for the StatefulSet.",
 			metric.Gauge,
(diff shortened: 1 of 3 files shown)


CatherineF-dev:
QQ: how to set alerts around metric kube_statefulset_ordinals_start?

Does it compare with actual pod deletion/creation order?

pwschuurman:
> QQ: how to set alerts around metric kube_statefulset_ordinals_start?

This metric is used to inform a cluster administrator about usage, rather than correctness of the feature.

> Does it compare with actual pod deletion/creation order?

kube_statefulset_ordinals_start reports the specified .spec.ordinals.start field, if .spec.ordinals is set. This enables a cluster administrator to monitor the usage of this feature. I considered adding a metric that indicated the status of the start ordinal, but in the KEP review, it was decided this wasn't super valuable to measure. See https://github.com/kubernetes/enhancements/pull/3768#discussion_r1097512692 for context.

soltysh (approved):
/lgtm

internal/store/statefulset_test.go
```diff
@@ -65,6 +65,7 @@ func TestStatefulSetStore(t *testing.T) {
 				# HELP kube_statefulset_metadata_generation [STABLE] Sequence number representing a specific generation of the desired state for the StatefulSet.
 				# HELP kube_statefulset_persistentvolumeclaim_retention_policy Count of retention policy for StatefulSet template PVCs
 				# HELP kube_statefulset_replicas [STABLE] Number of desired pods for a StatefulSet.
+                # HELP kube_statefulset_ordinals_start Start ordinal of the StatefulSet.
```
mrueg, line 68: Looks like indentation here is off

internal/store/statefulset.go
```diff
@@ -178,6 +178,26 @@ func statefulSetMetricFamilies(allowAnnotationsList, allowLabelsList []string) [
...
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_statefulset_ordinals_start",
+			"Start ordinal of the StatefulSet.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapStatefulSetFunc(func(s *v1.StatefulSet) *metric.Family {
+				ms := []*metric.Metric{}
+
+				if s.Spec.Ordinals != nil {
```
mrueg, line 190: What if .Ordinals exists but .Ordinals.Start does not?

pwschuurman: .Ordinals is a pointer, while .Ordinals.Start is an integer. So .Ordinals.Start will default to zero. The user doesn't have a way to set .Ordinals, without specifying Start.

internal/store/statefulset_test.go
pwschuurman: Thanks for catching (tabs vs. spaces issue). Fixed here and on line 371.

dgrisonnet (approved):
/lgtm
/approve

PR #1960, opened 2023-01-23
Enrich UserAgent with more information

### LINKED ISSUE #1955: UserAgent should include program name, instead of just version number
**What happened**:

Kube State Metrics user agent appears to be just a version number. E.g. "v2.1.0".  This makes it hard to tell in places like kube-audit, whether the app in question really is Kube State Metrics.

The issue appears to be here https://github.com/kubernetes/kube-state-metrics/blob/63db4839e921479ebf5b8835267beaa0515d7b6e/pkg/app/server.go#L337 

IMHO the line should not be `config.UserAgent = version.Version`.  Instead it should be `config.UserAgent = restclient.DefaultKubernetesUserAgent()`. That's what the rest of the K8s ecosystem normally uses. 

I can identify Kube State Metrics reliably from its user agent string, because it names itself in that string.

Look at any Kube Audit record created by Kube State Metrics traffic.

mrueg:
Before:
"User-Agent: v2.7.0"

After:
"User-Agent: kube-state-metrics/v2.7.0 (linux/amd64) kubernetes/1cda0bf9"

None
Fixes #1955


rexagod:
On `--kubeconfig=... --resources=pods` (this commit) I got the following. Am I missing something?

image

***
On `HEAD`.
image

logicalhan (approved):
/lgtm
/approve

PR #1938, opened 2022-12-29
Add metrics of kube_pod_status_ready_time and kube_pod_status_containers_ready_time redux

ryanrolds:
> This is a copy of #1837 with the merge conflict resolved. All credit should go to @liangyuanpeng. The text below is copied from the other PR.

What this PR does / why we need it:
Base PR https://github.com/kubernetes/kube-state-metrics/pull/1482

Ref: Adds kube_pod_status_ready_time and kube_pod_status_containers_ready_time metrics to get the information provided by [Pod Lifecycle's](https://pkg.go.dev/k8s.io/api/core/v1#PodCondition) [PodConditions](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#pod-conditions) array.

Maybe we should support param to open this feature or not, juse like feature-gates.

How does this change affect the cardinality of KSM: (increases, decreases or does not change cardinality)
increases metrics number that Pod number * 2

Which issue(s) this PR fixes (optional, in fixes #<issue number>(, fixes #<issue_number>, ...) format, will close the issue(s) when PR gets merged):
Fixes https://github.com/kubernetes/kube-state-metrics/issues/1465,https://github.com/kubernetes/kube-state-metrics/issues/1830

Co-authored-by: Szymon Grzemski [sz.grzemski@gmail.com](mailto:sz.grzemski@gmail.com)
Signed-off-by: Lan Liang [gcslyp@gmail.com](mailto:gcslyp@gmail.com)

PR #1961, opened 2023-01-24
Backporting #1938 to 2.3

ryanrolds:

Backports #1938 to the 2.3 release branch. We are in process of updating our K8s cluster, but current OKRs require tracking these metrics sooner rather than later. Happy to backport to all release branches from 2.3 on.

It shouldn't affect it beyond what is already in the main branch.


mrueg:
We consider new metrics a new feature and therefore would like to limit backporting to bug and security vulnerability fixes only.

ryanrolds:
Makes sense. Thanks.

PR #1963, opened 2023-01-26
support "True" and "False" as string in custom-resource-state

### LINKED ISSUE #1962: CustomResourceStateMetrics Gauges do not work for Kubernetes Controller Status Conditions
**What happened**:
We are trying to export a metric based on the standard status conditions CRD for objects of our kubernetes controllers (i.e. cert-manager or custom controllers based on kubebuilder).

We defined `--custom-resource-state-config` in kube-state-metics 2.7.0:

```
      - --custom-resource-state-config
      # in YAML files, | allows a multi-line string to be passed as a flag value
      # see https://yaml-multiline.info
      -  |
          spec:
            resources:
            - groupVersionKind:
                group: xxxxx.xxx
                kind: "xxx"
                version: "v1"
              metrics:
              - name: "xxx_ready"
                help: "xxx ready state"
                each:
                  type: Gauge
                  gauge:
                    path: [status, conditions, "[type=Ready]", status]
```

Unfortunately, by default kubernetes controllers use a string condition:
```
status:
  conditions:
    - lastTransitionTime: "2019-10-22T16:29:31Z"
      status: "True"
      type: Ready
```

And `kube-state-metrics` yields a parse error:

```
registry_factory.go:589] \"kube_crd_xxxx_ready\" err=\"[status,conditions,[type=Ready],status]: []: strconv.ParseFloat: parsing \\\"True\\\": invalid syntax\""}
```

I would expect it to be able to parse "True"/"False" or a way to map them to 1/0. 

1. Install a controller which exposes status conditions in its CRD. For instance cert-manager would work.
2. Add the config line from above
3. Run kube-state-metrics

**Environment**:
- kube-state-metrics version: 2.7.0
- Kubernetes version (use `kubectl version`): v1.22.17
- Cloud provider or hardware configuration: AWS via Kops
- Other info:

jabdoa2:
Kubernetes Operators usually store status conditions in the following format:
```
status:
  conditions:
    - lastTransitionTime: "2019-10-22T16:29:31Z"
      status: "True"
      type: Ready
```

Note that `status` is actually a string. This previously broke KSM custom-resource-state gauges because it could not convert strings to float. This PR maps "True", "true", "Yes" and "yes" to `1.0` and "False", "false", "No", "no" to `0.0`. This allows us to use custom-resource-state for all CRD objects of standard kubernetes operators.

For a single condition you can use (as in unit test `status_conditions`):
```
      - --custom-resource-state-config
      # in YAML files, | allows a multi-line string to be passed as a flag value
      # see https://yaml-multiline.info
      -  |
          spec:
            resources:
            - groupVersionKind:
                group: xxxxx.xxx
                kind: "xxx"
                version: "v1"
              metrics:
              - name: "xxx_ready"
                help: "xxx ready state"
                each:
                  type: Gauge
                  gauge:
                    path: [status, conditions, "[type=Ready]", status]
```

To expose one metric per status `type` (as in unit test `status_conditions_all`):
```
      - --custom-resource-state-config
      # in YAML files, | allows a multi-line string to be passed as a flag value
      # see https://yaml-multiline.info
      -  |
          spec:
            resources:
            - groupVersionKind:
                group: xxxxx.xxx
                kind: "xxx"
                version: "v1"
              metrics:
              - name: "xxx_status"
                help: "xxx status"
                each:
                  type: Gauge
                  gauge:
                    path: [status, conditions]
                    labelsFromPath:
                      type: ["type"]
                    valueFrom: ["status"]
```

Guess this could be something for the documentation as it will be a very common usecase for controller CRDs.

It does not change cardinality. It allows KSM to support `.status.conditions` used by standard kubernetes operators.

Fixes #1962

### DIFF
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -635,10 +635,11 @@ func toFloat64(value interface{}, nilIsZero bool) (float64, error) {
 		}
 		return 0, nil
 	case string:
-		if value.(string) == "True" || value.(string) == "true" || value.(string) == "Yes" || value.(string) == "yes" {
+		normalized := strings.ToLower(value.(string))
+		if normalized == "true" || normalized == "yes" {
 			return 1, nil
 		}
-		if value.(string) == "False" || value.(string) == "false" || value.(string) == "No" || value.(string) == "no" {
+		if normalized == "false" || normalized == "no" {
 			return 0, nil
 		}
 		if t, e := time.Parse(time.RFC3339, value.(string)); e == nil {
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -635,6 +635,12 @@ func toFloat64(value interface{}, nilIsZero bool) (float64, error) {
 		}
 		return 0, nil
 	case string:
+		if value.(string) == "True" || value.(string) == "true" || value.(string) == "Yes" || value.(string) == "yes" {
+			return 1, nil
+		}
+		if value.(string) == "False" || value.(string) == "false" || value.(string) == "No" || value.(string) == "no" {
+			return 0, nil
+		}
 		if t, e := time.Parse(time.RFC3339, value.(string)); e == nil {
 			return float64(t.Unix()), nil
 		}
--- a/pkg/customresourcestate/registry_factory_test.go
+++ b/pkg/customresourcestate/registry_factory_test.go
@@ -64,7 +64,7 @@ func init() {
 				},
 			},
 			"uptime": 43.21,
-			"conditions": Array{
+			"condition_values": Array{
 				Obj{
 					"name":  "a",
 					"value": 45,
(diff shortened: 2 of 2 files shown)


mrueg:
Since this really depends on the implementor of the CR, having some defaults make sense. Anything else really should be defined via a user supplied valuemap (e.g.if the implementation returns "Ok" etc.).

pkg/customresourcestate/registry_factory.go
```diff
@@ -635,6 +635,12 @@ func toFloat64(value interface{}, nilIsZero bool) (float64, error) {
 		}
 		return 0, nil
 	case string:
+		if value.(string) == "True" || value.(string) == "true" || value.(string) == "Yes" || value.(string) == "yes" {
+			return 1, nil
+		}
+		if value.(string) == "False" || value.(string) == "false" || value.(string) == "No" || value.(string) == "no" {
+			return 0, nil
+		}
```
mrueg, line 643: Let's normalize the string first so we catch also different capitalizations.

```suggestion
                normalized := strings.ToLower(value.(string))
		if normalized == "true" || normalized == "yes" {
			return 1, nil
		}
		if normalized == "false" || normalized == "no" {
			return 0, nil
		}
```

jabdoa2: Good one. Thanks!

mrueg:
It would be good to add a sentence or two to https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md about this conversion.

dgrisonnet (approved):
/approve
/lgtm

PR #1971, opened 2023-01-31
Fixing emitting of ready time metrics when condition is false

ryanrolds:

Pod status conditions have a status flag that indicates if the status applies or not. That status not being factored in causing a "ready time" metric to be emitted when the pod/containers are not ready.

It does not change cardinality, but does reduce the number of times the metrics are emitted.

NA

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1331,7 +1331,7 @@ func createPodStatusContainerReadyTimeFamilyGenerator() generator.FamilyGenerato
 			ms := []*metric.Metric{}
 
 			for _, c := range p.Status.Conditions {
-				if c.Type == v1.ContainersReady {
+				if c.Type == v1.ContainersReady && c.Status == v1.ConditionTrue {
 					ms = append(ms, &metric.Metric{
 						LabelKeys:   []string{},
 						LabelValues: []string{},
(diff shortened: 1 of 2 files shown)


internal/store/pod.go
```diff
@@ -1331,7 +1331,7 @@ func createPodStatusContainerReadyTimeFamilyGenerator() generator.FamilyGenerato
 			ms := []*metric.Metric{}
 
 			for _, c := range p.Status.Conditions {
-				if c.Type == v1.ContainersReady {
+				if c.Type == v1.ContainersReady && c.Status == v1.ConditionTrue {
```
mrueg, line 1334: since the value is c.LastTransitionTime, I wonder if it would make sense to have a dimension "ready=True|False" instead? That way you can also track when a transition to not ready happened.

ryanrolds: The pattern I saw in other pod metrics was to have a [separate metric for the false case](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/pod.go#L1543).

mrueg: Good point, since the metric is still experimental we can look at it later as well and work out things.

PR #1972, opened 2023-01-31
Doc updates to for custom resource metrics

### LINKED ISSUE #1970: Custom Resources reference kube_customresource_ and it looks to have changed to kube_crd_
**What happened**:
While reading through the docs https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md I was searching and debugging why the custom resources weren't being displayed. It looks (at least in my instance) that the resources have shown up now as `kube_crd_` instead of `kube_customresource_`. 
 
I was searching for `kube_customresource_` in our prom instance and unable to find the desired custom metrics. 

Basic setup with a custom resource config file as outlined here:  https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md

Example configuration: 
```
        - groupVersionKind:
            group: "runwhen.com"
            kind: "Location"
            version: "v1"
          metrics:
            - name: "heartbeat"
              help: "Latest Heartbeat"
              each:
                type: Gauge
                gauge:
                  path: [status]
                  valuesFrom: [heartbeat]
              labelsFromPath:
                locationType: [spec, locationType ]

```

Example result from the kube-state-metrics /metrics endpoint: 
```
kube_crd_heartbeat{group="runwhen.com",kind="Location",locationType="public",version="v1"} 1.675200259e+09
```

I *think* this is just a doc update to change `customresource` to `crd` so that other readers can quickly find their custom resource metrics while implementing this solution. 

**Environment**:
- kube-state-metrics version: 2.7.0
- Kubernetes version (use `kubectl version`): 1.23.14-gke-401
- Cloud provider or hardware configuration: GKE
- Other info:

stewartshea:
The PR attempts to update the docs to resolve https://github.com/kubernetes/kube-state-metrics/issues/1970. 
In short, the documentation states that the metrics are named `kube_customresource_` and a quick implementation test on my system indicates they now show up as `kube_crd_`. 

If the issue was environment or configuration specific, feel free to close this out. Thanks!

Does not change. Documentation updates only. 

Fixes #1970

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -151,7 +151,7 @@ spec:
 Produces the metric:
 
 ```prometheus
-kube_crd_uptime{crd_group="myteam.io", crd_kind="Foo", crd_version="v1"} 43.21
+kube_crd_uptime{group="myteam.io", kind="Foo", version="v1"} 43.21
 ```
 
 #### Multiple Metrics/Kitchen Sink
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -83,7 +83,7 @@ spec:
           - --custom-resource-state-only=true
 ```
 
-NOTE: The `customresource_group`, `customresource_version`, and `customresource_kind` common labels are reserved, and will be overwritten by the values from the `groupVersionKind` field.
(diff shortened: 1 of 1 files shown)


mrueg:
This change is not needed, as the renaming from kube_crd to kube_customresource happened in the main branch and will be released in v2.8.0
See: https://github.com/kubernetes/kube-state-metrics/issues/1970#issuecomment-1415752763

PR #1989, opened 2023-02-10
feat(CustomResourceState): Support quantities and percentages

mrueg:
In order to be able to deprecate the VPA completely, CRS needs to be aware of quantities. See https://github.com/kubernetes/kube-state-metrics/issues/1790#issuecomment-1228477919

None

See also https://github.com/kubernetes/kube-state-metrics/issues/1790

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -255,6 +255,7 @@ Supported types are:
   * `"true"` and `"yes"` are mapped to `1.0` and `"false"` and `"no"` are mapped to `0.0` (all case insensitive)
   * RFC3339 times are parsed to float timestamp  
   * Quantities like "250m" or "512Gi" are parsed to float using https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go
+  * Percentages ending with a "%" are parsed to float
   * finally the string is parsed to float using https://pkg.go.dev/strconv#ParseFloat which should support all common number formats. If that fails an error is yielded
 
 ##### Example for status conditions on Kubernetes Controllers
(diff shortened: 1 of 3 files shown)


rexagod:
Can we incorporate support for percentages as well? IIRC that's the only other format that's causing issues.

mrueg:
@rexagod good idea, I've added support for it as well.

rexagod:
I see this incorporates supports for `path` pointing to such quantities, but `valueFrom` still needs to inherit this behavior. We can do that in a different PR, or get it addressed in this one itself, whichever sounds better.

```yaml
            # Works.
            type: Gauge
            gauge:
              path: [foo, bar, baz, que]
```
```yaml
            # Needs to be addressed.
            type: Gauge
            gauge:
              path: [foo]
              valueFrom: [bar, baz, que]
```

mrueg:
@rexagod I've added a valueFrom testcase, this seems to work for both. Can you clarify if I am missing anything?

rexagod:
Ah, yes I see it now. The PR is good for `valueFrom`. I forgot the issue with multiple relative-path in `valueFrom` still persists. Basically, I was testing this on this,
```yaml
            # Same as in the example above.
            type: Gauge
            gauge:
              path: [foo]
              valueFrom: [bar, baz, que]
```
when I should've tested it on,
```yaml
            # Works.
            type: Gauge
            gauge:
              path: [foo, bar, baz]
              valueFrom: [que] # Only single-paths are supported as of now.
```
***
We just need to resolve the failing test and this should be good! 🚀

mrueg:
@rexagod I rebased on the latest main branch and fixed the test. :)

rexagod:
/lgtm
Thank you for the patch! :)

djfinnoy:
Any chance of getting this included in a release soon? We need it :)

PR #1841, opened 2022-09-22
Do not expose ingress path metric when service is nil

evir35:
**What this PR does / why we need it**:

Currently, the kube-state-metrics can down suddenly when an ingress without backend service is created.

https://github.com/kubernetes/api/blob/79091dac6a32a1365edbc6de9576ffa74768a75c/networking/v1/types.go#L484-L496

The `Service` of `IngressBackend` is a nullable variable. But, the current code doesn't check the `Service` is nil or not.
So, I created PR to solve this using an additional if statement that checks `Service` is nil or not.
And I checked both `path.Backend.Service.Name` and `path.Backend.Service.Port.Number` variables can't be nil.

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)*

This PR doesn't affect the cardinality of KSM.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #1660

PR #1994, opened 2023-02-15
fix: public Builder compatibility with the BuilderInterface

clamoriniere:

This pull request add the changes introduced by https://github.com/kubernetes/kube-state-metrics/pull/1864/ into the "public" Builder (`pkg/builder/Builder`). 

To avoid having future similar issue, this PR also add a interface type check between the `pkg/build/type/BuilderInterface` Interface and the `pkg/build/Builder` struct. It should allow to catch at compilation time any discrepancy. 

It doesn't affect cardinality of KSM because it doesn't change any metrics generation logic.

Fixes #

### DIFF
--- a/pkg/builder/builder.go
+++ b/pkg/builder/builder.go
@@ -32,6 +32,10 @@ import (
 	"k8s.io/kube-state-metrics/v2/pkg/options"
 )
 
+// Make sure the public Builder implements the public BuilderInterface.
+// New internal Builder methods should be added to the public BuilderInterface.
+var _ ksmtypes.BuilderInterface = &Builder{}
+
 // Builder helps to build store. It follows the builder pattern
 // (https://en.wikipedia.org/wiki/Builder_pattern).
 type Builder struct {
@@ -59,6 +63,11 @@ func (b *Builder) WithEnabledResources(c []string) error {
 // WithNamespaces sets the namespaces property of a Builder.
 func (b *Builder) WithNamespaces(n options.NamespaceList) {
 	b.internal.WithNamespaces(n)
+}
+
+// WithFieldSelectorFilter sets the fieldSelector property of a Builder.
+func (b *Builder) WithFieldSelectorFilter(fieldSelectorFilter string) {
+	b.internal.WithFieldSelectorFilter(fieldSelectorFilter)
 }
 
 // WithSharding sets the shard and totalShards property of a Builder.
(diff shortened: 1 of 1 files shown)


L3n41c (approved):

clamoriniere:
Hi @CatherineF-dev 

thanks for the review. 

Do you think, it can be back-ported after to previous release branch? I think the first release with this change is `v2.7.0`

CatherineF-dev:
QQ: want to know more about the context. What are the impacts of not backporting into `v2.7.0`?  

Some new features can't be used or the binary will be broken.

clamoriniere:
> QQ: want to know more about the context. What are the impacts of not backporting into `v2.7.0`?
> 
> Some new features can't be used or the binary will be broken.

Sure,

For context, we are vendoring part of the "kube-state-metrics" code in the [datadog-agent](github.com/datadog/datadog-agent) (more details about why we are doing it in this [blog post](https://www.datadoghq.com/blog/engineering/our-journey-taking-kubernetes-state-metrics-to-the-next-level/)).
Currently we are vendoring `kube-state-metrics` version `v2.4.2`, but we discover a [bug](https://github.com/kubernetes/kube-state-metrics/pull/1841) with `Ingresses` resource that is fixed in `v2.7.0`. So our initial tentative was to update to the last release version `v2.8.0`, but this version now vendor `k8s.io/api v0.26.1` but our `datadog-agent` project only vendor `v0.25.5`, and we know that if we upgrade to to `v0.26.0` now, we will break some compatibility with some deprecated Kubernetes resources on old Kubernetes versions.

So we tried to vendor the last version that uses `k8s.io/api v0.25.x` and that also has the `Ingresses` [bug fix ](https://github.com/kubernetes/kube-state-metrics/pull/1841)which corresponds to the version `v2.7.0`. And it is how we also discover that the `pkg/builder/Builder` struct is not anymore compliant with the `BuilderInterface` since the same version.
Here is the full story 😄

pkg/builder/builder.go
```diff
@@ -32,6 +32,10 @@ import (
 	"k8s.io/kube-state-metrics/v2/pkg/options"
 )
 
+// Make sure the public Builder implements the public BuilderInterface.
+// New internal Builder methods should be added to the public BuilderInterface.
+var _ ksmtypes.BuilderInterface = &Builder{}
```
mrueg, line 37: Can we move this to builder_test.go ?

clamoriniere: we can.
I added it the `builder.go` to get quick feedback at build time

mrueg: Looks like we have the same in pkg/internal/store/builder.go so I guess it should be fine here to keep things consistent.

PR #1996, opened 2023-02-15
[waiting KSM 3.x] Map KSM STABLE metrics to k8s BETA metrics and declare meaning of KSM STABLE metrics

### LINKED ISSUE #1995: Define what's stable metric in kube-state-metrics repo
**What would you like to be added**:  Define what's stable metric and  add this definition into documentations.
1. metric name can't be changed
2. existing metric labels can't be changed 
3. new fields can be added to existing metrics without having to create a new version

**Why is this needed**: 

**Additional context**

https://github.com/kubernetes/kube-state-metrics/issues/1991#issuecomment-1431628553

CatherineF-dev:
**What this PR does / why we need it**: https://github.com/kubernetes/kube-state-metrics/issues/1995

Fixes #1995

```
	ALPHA Stability = Stability(basemetrics.ALPHA)
	STABLE Stability = Stability(basemetrics.BETA)
```

### DIFF
--- a/docs/developer/guide.md
+++ b/docs/developer/guide.md
@@ -28,6 +28,6 @@ The following steps are needed to introduce a new resource and its respective re
 
 | Metric stability level |                    |
 |------------------------|--------------------|
-| EXPERIMENTAL           | basemetrics.ALPHA  |
-| STABLE                 | basemetrics.STABLE |
+| EXPERIMENTAL           | metricsstability.ALPHA  |
+| STABLE                 | metricsstability.STABLE |
 
--- a/internal/store/certificatesigningrequest.go
+++ b/internal/store/certificatesigningrequest.go
@@ -19,7 +19,7 @@ package store
 import (
 	"context"
 
-	basemetrics "k8s.io/component-base/metrics"
+	metricsstability "k8s.io/kube-state-metrics/v2/pkg/stability"
 
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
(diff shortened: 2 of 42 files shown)


mrueg:
Thanks for coming up with this! 
Maybe it would be better to have a paragraph around stable metrics as I believe we allow everything for experimental ones.

Imho for stable metrics only two things are allowed:
- Additional labels are allowed
- Bug fixes to the implementation (e.g. change specific label values or edge cases on nil value)
- Description change

Things that we expect from stable metrics:
- No renaming, dropping of labels
- Metric name is considered frozen
- Meaning/Type of the value is considered frozen
- Stable metric can be deprecated in minor release: Lifecycle is: STABLE (2.n)-> DEPRECATED (2.n+1)-> REMOVED (2.n+2). A removal is a breaking change and will be announced prominently in the CHANGELOG.

CatherineF-dev:
We discussed in the sig-instrumentation meeting that: allow adding new labels in KSM stable metrics.

ksm stable metrics -> k8s beta metrics (it allows adding new labels)
ksm alpha metrics -> k8s alpha metrics

CatherineF-dev:
Ready for review~

CatherineF-dev:
cc @mrueg 
Thanks! Have added what you mentioned.

logicalhan:
I personally find this confusing, since we're still using stable to mean the API can change.

CatherineF-dev:
It seems that KSM repo's STABLE means BETA in k8s repo.  In this way, users/contributors don't need to learn something new `BETA`.

If we want to have three stability levels in the future, it has to rename all STABLE to BETA now.

I am fine with both decisions: 1. map KSM STABLE level to k8s BETA 2. rename all STABLE to BETA.

cc @dgrisonnet @mrueg How about your opinions?

dgrisonnet:
> It seems that KSM repo's STABLE means BETA in k8s repo. In this way, users/contributors don't need to learn something new BETA.

Yes we are currently treating it in a similar fashion as how k8s treats BETA.

I am totally onward with moving to k8s stability definition, but this can't be done today without a new major version of kube-state-metrics since having a STABLE level will require to rework a couple of metrics to make sure that they will never need to change.

My suggestion for now would be to stick to only have ALPHA and STABLE in ksm with different definitions from k/k and converge to a common ground with the next major.

PR #2000, opened 2023-02-17
Correct link for customising kube prometheus

edwardmlyte:

Updates documentation to correct the customising kube prometheus link. Helps new users get to the correct docs more easily.

Does not change


mrueg:
/lgtm
Thanks!

PR #2007, opened 2023-02-22
Fix label resolution for maps

rexagod:
**What this PR does / why we need it**: The `map[string]interface{}` resolution searched for values within various nested fields, but not the relative path itself.

***
https://github.com/kubernetes/kube-state-metrics/blob/ca1da6b22ed28fe3aa6cac98705cff71c614eac0/pkg/customresourcestate/registry_factory.go#L598-L604


logicalhan (approved):
/lgtm
/approve
/triage accepted

PR #2017, opened 2023-03-06
feat!: Remove VerticalPodAutoscaler

### LINKED ISSUE #1790: Remove Verticalpodautoscaler Collector (VPA)

mrueg:
BREAKING CHANGE: VPA is now supported only via Custom Resource Metrics

Removes a resource
Fixes #1790


internal/store/networkpolicy_test.go
```diff
@@ -30,8 +30,8 @@ func TestNetworkPolicyStore(t *testing.T) {
 	metav1StartTime := metav1.Unix(int64(startTime), 0)
 
 	const metadata = `
-		# HELP kube_verticalpodautoscaler_labels Kubernetes labels converted to Prometheus labels.
-		# TYPE kube_verticalpodautoscaler_labels gauge
+		# HELP kube_networkpolicy_labels Kubernetes labels converted to Prometheus labels.
+		# TYPE kube_networkpolicy_labels gauge
```
dgrisonnet, line 34: is it just me or that is a bit sus?

Ah I see, it is unused. Normally you would have that on top of tests were you have headers in commons and want to simplify the header verificaiton as in: https://github.com/kubernetes/kube-state-metrics/blob/80e67d01a20f1a16147d8acb4d474a46ccc4971b/internal/store/certificatesigningrequest_test.go

let's remove it for now and fix it in a follow up?

mrueg: I found it while I was looking for vpa/verticalpodautoscaler and thought this will improve it at least a bit. I can drop it.

dgrisonnet: it doesn't seem to be used right, so I am with either add the logic to use it in this PR as in https://github.com/kubernetes/kube-state-metrics/blob/80e67d01a20f1a16147d8acb4d474a46ccc4971b/internal/store/certificatesigningrequest_test.go#L231 or drop it and do it in a follow-up

mrueg: Sounds good, I dropped it for now.

docs/customresourcestate-metrics.md
```diff
@@ -207,6 +207,45 @@ kube_customresource_ready_count{customresource_group="myteam.io", customresource
 kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", customresource_version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 4
 ```
 
+#### VerticalPodAutoscaler
+
+In v2.9.0 the `vericalpodautoscalers` resource was removed from the list of default resources. In order to generate metrics for `verticalpodautoscalers`, you can use the following Custom Resource State config:
+
+```yaml
+# Using --resource=verticalpodautoscalers, we get the following output:
+# HELP kube_verticalpodautoscaler_annotations (Deprecated since v2.9.0) Kubernetes annotations converted to Prometheus labels.
```
dgrisonnet, line 216: wasn't this deprecated since v2.7.0?

mrueg: Good catch, this was a copy'n'paste issue.

dgrisonnet (approved):
/lgtm
/approve

PR #2018, opened 2023-03-08
Improve wording for Prometheus rule example in pod-metrics

arturhoo:
**What this PR does / why we need it**: it clarifies the intention of a useful Prometheus rule.

**How does this change affect the cardinality of KSM**: n/a - documentation change only.


mrueg:
Thanks for improving the docs @arturhoo 
/lgtm

/hold 
for others to review.

PR #2019, opened 2023-03-09
Update Dockerfile

lihantao3:

Fixes #


dgrisonnet:
These image are meant to be built from the Makefile, so it is normal that the arg are not specified in the Dockerfile.

https://github.com/kubernetes/kube-state-metrics/blob/main/Makefile#L92-L100

Issue #1973, opened 2023-02-02
CustomResourceMetrics `info` and `stateset` are now broken in main when scraping with prometheus


primeroz:
**What happened**:
before https://github.com/kubernetes/kube-state-metrics/pull/1930 all `metrics type` for CustomResourceMetrics were reported as type `gauge` but now ( i do run a build from main for some testing ) the `type` defined in the CR is respected and a metrics are reported as type `info` and `stateset` 

for example
```
 λ curl localhost:8000/metrics 2>/dev/null | egrep "*_info" | grep TYPE                                                                                                                                            
# TYPE capi_azurecluster_info info 
```

while those are valid `OpenTelemtry` types they are not valid `Prometheus` Types and so Prometheus fail to scrape with `invalid metric type "info"` 

**What you expected to happen**:

Prometheus be able to scrape those metrics

**How to reproduce it (as minimally and precisely as possible)**:

Run a build of KSM from main with `CustomResourceMetrics` with type set to `Info` or `StateSet` ( i do use the cluster-api example https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/observability/kube-state-metrics/crd-config.yaml ) 

**Anything else we need to know?**:

I think it would be great to support both Prometheus and OpenTelemetry , maybe with a field to set the type of metrics to expose and revert to present all metrics as gauges when Prometheus is the intended target

**Environment**:
- kube-state-metrics version: main
- Kubernetes version (use `kubectl version`): 1.24.10
- Cloud provider or hardware configuration:
- Other info:

pinging @bavarianbidi for reference

mrueg:
@primeroz thanks for the report! Can you test if https://github.com/kubernetes/kube-state-metrics/pull/1974 fixes your issue?

primeroz:
I will test on Monday and report back. Thanks for the quick feedback

primeroz:
testing a build out of `[49af6b3](https://github.com/mrueg/kube-state-metrics/commit/49af6b3945a071ec63de8b2cadaf4d5eec1bdd57)` and is looking good. thank you

Efrat19:
Hi! I see this happening again for the new version 2.11.0 (2.10.1 is fine)
A curl to a ksm pod `10.254.30.65` with image `registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.11.0`:
```
curl --header "Accept: application/openmetrics-text" -v http://10.254.30.65:8080/metrics 
[...]

PR #2024, opened 2023-03-13
Only use OpenMetrics and Text in contentType

### LINKED ISSUE #2022: KSM 2.8.0 can't be scraped by Prometheus if Native Histograms is enabled
**What happened**:
Starting from version 2.8.0, if Prometheus is started with `--enable-feature=native-histograms` flag it fails to scrape KSM, instead erroring with the error:
```
proto: wrong wireType = 0 for field Metric
```

KSM should be able to be scraped by Prometheus

I can reproduce this just on minikube with the standard helm prometheus, chart https://github.com/prometheus-community/helm-charts/tree/main/charts/prometheus

```
helm install prometheus prometheus-community/prometheus --set server.extraFlags={"enable-feature=native-histograms"\,"log.level=debug"} --version=19.7.2
```
This deploys kube-state-metrics version: `registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.8.0`
And Prometheus version: `quay.io/prometheus/prometheus:v2.41.0`
With the `--enable-feature=native-histograms` set.

The problem was introduced in this PR https://github.com/kubernetes/kube-state-metrics/pull/1974

I think the issue, is that when you enable Native Histograms, Prometheus is sending headers with a content type of:
```
content type: %v application/vnd.google.protobuf; proto=io.prometheus.client.MetricFamily; encoding=delimited
```
Which we're then mirroring in our headers:
```go
resHeader.Set("Content-Type", string(contentType))
```
But we're not actually changing our response so we're replying with invalid protobuf hence the error from Prometheus.

I did this as a simple workaround and it fixed the issue:
```go
if contentType == expfmt.FmtProtoDelim {
    resHeader.Set("Content-Type", `text/plain; version=`+"0.0.4")
} else {
    resHeader.Set("Content-Type", string(contentType))
}
```
However I'm not sure this is the right approach, is it intended that kube-state-metrics always responds with content headers that match the request? Or was this just a change for Open Metrics? In which case perhaps the header should only get changed if the `contentType == expfmt.FmtOpenMetric`? e.g reverse the logic, only change the headers if we detect open metrics headers:

```go
if contentType == expfmt.FmtOpenMetrics {
   resHeader.Set("Content-Type", string(contentType))
} else {
   resHeader.Set("Content-Type", `text/plain; version=`+"0.0.4")
}
```
That way the code would behave the same as it did prior to 2.8.0 if the open metrics headers are not present.

**Environment**:
- kube-state-metrics version: 2.8.0
- Kubernetes version (use `kubectl version`): 1.26.1
- Cloud provider or hardware configuration: n/a
- Other info:

CatherineF-dev:
**What this PR does / why we need it**: #2022

Fixes #2022 #1973

### DIFF
--- a/pkg/metricshandler/metrics_handler.go
+++ b/pkg/metricshandler/metrics_handler.go
@@ -187,6 +187,9 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	contentType := expfmt.NegotiateIncludingOpenMetrics(r.Header)
 
+	if contentType != expfmt.FmtOpenMetrics {
+		contentType = expfmt.FmtText
+	}
 	resHeader.Set("Content-Type", string(contentType))
 
 	if m.enableGZIPEncoding {


pkg/metricshandler/metrics_handler.go
```diff
@@ -187,6 +187,9 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 
 	contentType := expfmt.NegotiateIncludingOpenMetrics(r.Header)
 
+	if contentType != expfmt.FmtOpenMetrics {
```
mrueg, line 190:
```suggestion
        # We do not support protobuf at the moment. Fall back to FmtText if the negotiated exposition format is not FmtOpenMetrics See: https://github.com/kubernetes/kube-state-metrics/issues/2022
	if contentType != expfmt.FmtOpenMetrics {
```

Add a comment so it's clear why we're doing it.

mrueg:
Thanks for providing this fix @CatherineF-dev 
/lgtm 

/hold for other to review

PR #2032, opened 2023-03-17
feat: include resources with kubernetes.io prefix.

### LINKED ISSUE #1939: Do not collect  resources that start with ’kubernetes.io/‘
**What happened**:

my pod
```
limits:
        kubernetes.io/batch-cpu: 2k
        kubernetes.io/batch-memory: 4Gi
      requests:
         kubernetes.io/batch-cpu: "100"
         kubernetes.io/batch-memory: 819Mi
```
but not to be collected.

reason is  ’kubernetes.io/‘ Is considered to be a native resource:
https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/utils.go#L155

**Environment**:
- kube-state-metrics version: 
- Other info:

fengyehong:
Fixes #1939

### DIFF
--- a/internal/store/utils.go
+++ b/internal/store/utils.go
@@ -165,12 +165,7 @@ func isExtendedResourceName(name v1.ResourceName) bool {
 }
 
 func isNativeResource(name v1.ResourceName) bool {
-	return !strings.Contains(string(name), "/") ||
-		isPrefixedNativeResource(name)
-}
-
-func isPrefixedNativeResource(name v1.ResourceName) bool {
-	return strings.Contains(string(name), v1.ResourceDefaultNamespacePrefix)
+	return !strings.Contains(string(name), "/")
 }
 
 // createPrometheusLabelKeysValues takes in passed kubernetes annotations/labels
--- a/internal/store/utils_test.go
+++ b/internal/store/utils_test.go
@@ -93,7 +93,7 @@ func TestIsExtendedResourceName(t *testing.T) {
 		},
 		{
 			resourceName: "kubernetes.io/resource-foo",
-			expectVal:    false,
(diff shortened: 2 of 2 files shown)


internal/store/utils.go
```diff
@@ -169,7 +169,7 @@ func isNativeResource(name v1.ResourceName) bool {
 }
 
 func isPrefixedNativeResource(name v1.ResourceName) bool {
-	return strings.Contains(string(name), v1.ResourceDefaultNamespacePrefix)
+	return false
 }
```
logicalhan, line 173: If we're just going to return false, then we should just get rid of this method.

logicalhan: Also there should be a test for this, we shouldn't just be able to change this function without some test breaking.

dgrisonnet: I think we should remove the code that check whether the resource is native or not as a whole.

The idea would be to make `Integer` the default unit if nothing is matched after https://github.com/kubernetes/kube-state-metrics/blob/d3afe69fd784b7ea31a42ce2eaceb7f7b1ead7a0/internal/store/pod.go#L193-L204 and remove the condition for extended resources: https://github.com/kubernetes/kube-state-metrics/blob/d3afe69fd784b7ea31a42ce2eaceb7f7b1ead7a0/internal/store/pod.go#L205-L211

eahydra:
ping @fengyehong

internal/store/utils.go
fengyehong: updated, i noticed there is already a test case for "kubernetes.io/resource-foo" so just updated expectVal

eahydra:
@dgrisonnet @logicalhan PTAL

PR #2046, opened 2023-04-13
Fix casing while scraping failure reason for `kube_job_status_failed`

### LINKED ISSUE #2045: Scraping the failure reason for a job does a case-sensitive search that fails
**What happened**:
When scraping the failure reason for a job, KSM does a [case-sensitive search](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L432) for the [keyword `DeadLineExceeded`](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L42). Since Kubernetes is generating the string `DeadlineExceeded`, the search fails to scrape the reason.

A case-insensitive search should be done, so changes in reason's casing would not impact scraping of the metric.

Create a job that fails to meet the `.spec.activeDeadlineSeconds`, forcing Kubernetes to kill the job with reason `DeadlineExceeded`.

I'm proposing a solution in PR #2046

**Environment**:
- kube-state-metrics version: 2.8.2
- Kubernetes version (use `kubectl version`): 1.26.3

juanjjaramillo:
When scraping the failure reason for a job, KSM does a [case-sensitive search](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L432) for the [keyword `DeadLineExceeded`](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L42). Since Kubernetes is generating the string `DeadlineExceeded`, the search fails to scrape the reason. This PR proposes to do a case-insensitive search instead, so changes in casing do not affect the scraping of the reason.

Does not change cardinality

Fixes #2045

### DIFF
--- a/internal/store/job.go
+++ b/internal/store/job.go
@@ -19,6 +19,7 @@ package store
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	basemetrics "k8s.io/component-base/metrics"
 
@@ -39,7 +40,7 @@ var (
 	descJobLabelsName          = "kube_job_labels"
 	descJobLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descJobLabelsDefaultLabels = []string{"namespace", "job_name"}
-	jobFailureReasons          = []string{"BackoffLimitExceeded", "DeadLineExceeded", "Evicted"}
+	jobFailureReasons          = []string{"BackoffLimitExceeded", "DeadlineExceeded", "Evicted"}
 )
 
 func jobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator {
@@ -429,5 +430,5 @@ func failureReason(jc *v1batch.JobCondition, reason string) bool {
 	if jc == nil {
 		return false
 	}
-	return jc.Reason == reason
+	return strings.EqualFold(jc.Reason, reason)
 }
--- a/internal/store/job.go
+++ b/internal/store/job.go
@@ -19,7 +19,6 @@ package store
 import (
 	"context"
 	"strconv"
-	"strings"
 
 	basemetrics "k8s.io/component-base/metrics"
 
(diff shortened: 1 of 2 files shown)


CatherineF-dev:
This metric is a stable metric. https://github.com/kubernetes/kube-state-metrics/blob/main/docs/job-metrics.md

Changing this might break alerts using this reason. 

cc @dgrisonnet do we allow this change for stable metrics?

mrueg:
Was it ever DeadLineExceeded in upstream kubernetes? If not this seems like a bug in kubestatemetrics

dgrisonnet:
Yes this is a bug, I believe it was always `DeadlineExceeded`, but we might have made a typo when introducing the change.

@juanjjaramillo could you perhaps also update the value in https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L42?

dgrisonnet:
@CatherineF-dev yes we do. It is very rare that label value changes are breaking changes and even if they are I think we should always go forward with them, but if there is any doubt, we should always call the change out in the changelog/release note.

Moreover in this particular case it fixes an actual bug that prevent the actual reason to be exposed, so we should definitely fix it.

internal/store/job.go
```diff
@@ -429,5 +430,5 @@ func failureReason(jc *v1batch.JobCondition, reason string) bool {
 	if jc == nil {
 		return false
 	}
-	return jc.Reason == reason
+	return strings.EqualFold(jc.Reason, reason)
```
dgrisonnet, line 433: We want to map the value exactly to make sure that we have a bounded list of potential values, so I don't think we should make that case insensitive. We should rather update the list of reasons and make sure we are more careful in the future to avoid these kind of typos.

juanjjaramillo: Makes sense @dgrisonnet 

I have already undo this change. Casing in https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/job.go#L42 was already part of the previous commit, so I only needed to update unit testing to make tests pass. Please let me know if something else is needed.

Thank you for the review!

dgrisonnet (approved):
Thank you @juanjjaramillo! :)

/lgtm
/approve

Issue #1899, opened 2022-11-17
Automate generation of Custom Resource configuration


chrischdi:
**What would you like to be added**:

As a developer, I would like the kube-state-metrics Custom Resource configuration to be automatically generated from the API code.

Thanks to @sbueringer and @fabriziopandini  for helping in driving this efforts 👍 

**Why is this needed**:

kube-state-metrics requires a configuration file for Custom Resources which could [grow to a very large file](https://github.com/kubernetes-sigs/cluster-api/blob/main/hack/observability/kube-state-metrics/crd-config.yaml) and may be hard to maintain over long-term together with changing Custom Resource definitions.

Instead of writing the configuration file manually it would be awesome if the configuration could get generated from the code where the custom resource gets defined instead.

**Describe the solution you'd like**

Prior art:

Kubebuilder makes use of a tool called [controller-gen] to generate the yaml's for e.g. Custom Resource definitions from code.
To do that it makes heavy use of markers inside the go code, which are comments in the form of `// +foo:bar:key1=value1,key2=value2`.

[controller-gen]: https://book.kubebuilder.io/reference/controller-gen.html#controller-gen-cli

Similar to `controller-gen`, kube-state-metrics could provide an additional tool to generate the custom resource metrics configuration file from pre-defined markers at the go code level.

**Additional context**

We already thought about adding a tool like this at the Cluster API project, but we think it would be a better fit for the kube-state-metrics project, as it could help authors of any custom resource and is not specific to Cluster API.

We also have an idea how the markers / design could look like, which builds on top of the currently manually written configuration in Cluster API: https://github.com/kubernetes-sigs/cluster-api/issues/7158#issuecomment-1317701277 .
I could also migrate the first implementation design idea over to this issue or however it would be decided to proceed.

If the maintainers of kube-state-metrics also think that the described idea would be worth to implement: I'd be of course happy to volunteer to help or contribute for this efforts.

Small example:

```go
// +Metrics:namePrefix=capi_cluster
// +Metrics:labelFromPath:name=name,JSONPath=".metadata.name"
// +Metrics:labelFromPath:name=namespace,JSONPath=".metadata.namespace"
// +Metrics:labelFromPath:name=uid,JSONPath=".metadata.uid"
type Cluster struct {
	metav1.ObjectMeta `json:"metadata,omitempty"`
	...
	Spec ClusterSpec `json:"spec"` 
}

type ClusterSpec struct {
[...]

PR #2049, opened 2023-04-17
feat: add CustomResourceMonitor CRD (WIP)

### LINKED ISSUE #1948: Read custom resource metrics configuration from a Custom Resource Definition

As of today, kube-state-metrics allows to define a configuration file which is read on startup to define metrics for custom resources via configuration.

Instead of using a configuration file to define custom metrics, kube-state-metrics could define its own Custom Resource Definition, which defines the metric configuration for a single Custom Resource instead.

**Why is this needed**:

Custom Resource Definitions can be deployed dynamically. The same could apply to the configuration of metrics for the same.

There are also had been several comments and discussions which now results in this issue (see additional context below).

* Define a custom resource definition for kube-state-metrics-customresourcemetrics (maybe a replacement/successor for the customresource configuration), it should allow the same featureset as the already existing configuration.
* Adjust kube-state-metrics to dynamically read/add/remove customresourcemetrics by watching the kube-state-metrics-customresourcemetrics custom resources

UX:

* Deploy kube-state-metrics (including a CustomResourceDefinition for customresourcemetrics)
* Deploy my Custom Resource Definition `foo`
* Deploy my "kube-state-metrics customresourcemetrics" CR which defines the metrics for my Custom Resource Definition `foo`

**Additional context**

* > We would like to use kube-state-metrics with crossplane where we create dozens of CRD's
  >
  > It would be great if kube-state-metrics could provide CustomResourceStateMetrics CRD. Whenever object of that kind will be created it will generate new crd-cofig.yaml and reload metrics.
  >
  > With such CRD we can define for our every crossplane component CustomResourceStateMetrics configuration saying how to produce metrics
  [[source]](https://github.com/kubernetes/kube-state-metrics/issues/1899#issuecomment-1375318020)

* Different issue which mentions this idea: https://github.com/kubernetes/kube-state-metrics/pull/1810#issuecomment-1363737471
* There also had been some slack discussions which mentioned this idea.

The pattern would be similar to what the `ServiceMonitor` CRD is for the Prometheus operator.

### LINKED ISSUE #2050: Decouple custom resource config with kube-state-metrics deployment

**Why is this needed**:
Add a custom resource metric requires changing KSM deployment.
In a company with platform monitoring team and application team, it means application team needs to change codes managed by platform monitoring team.

#2049

**Additional context**

CatherineF-dev:
What: A more user-friendly UX to collect custom resource metrics.
Why: 
1.  Using CustomResourceState monitoring feature is not user-friendly. Current ways on configuring CustomResourceState monitoring are:
* `--custom-resource-state-config "inline yaml (see example)"` or
* `--custom-resource-state-config-file /path/to/config.yaml`

2.  Current CustomResourceState monitoring feature doesn't support multiple configuration files.
For example, for a company with 10 teams, each team wants to collect CustomResourceState metrics for their owned Custom Resources. 

See more in design doc. 

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)* no

Fixes #2050 #1948

Design doc is ready to review while codes are proof-of-concepts.
Tested: 1. apply examples/cr.yaml 2. apply examples/cr2.yaml 3. delete two yamls.


CatherineF-dev:
fyi: @chrischdi

rexagod:
Hey, thank you for working on this. Just a question right off the top of my mind, would this be better suited to be in a KSM _operator_? I remember a discussion started my @mrueg in the room a while back that made us hold off on adding this feature for the sole reason of setting up an operator first, IIRC.

chrischdi:
> Hey, thank you for working on this. Just a question right off the top of my mind, would this be better suited to be in a KSM _operator_? I remember a discussion started my @mrueg in the room highlighting the need for this feature a while back but we eventually held off on adding it for the sole reason of setting up an operator first, IIRC.

Let's move that discussion over to the issue itself https://github.com/kubernetes/kube-state-metrics/issues/1948, happy to collect the facts and opinions there to properly document a decision :-)

Thanks @CatherineF-dev , I'll try to figure out time next week after kubecon to take a look at it 👍 🥳

CatherineF-dev:
> Hey, thank you for working on this. Just a question right off the top of my mind, would this be better suited to be in a KSM _operator_? I remember a discussion started my @mrueg in the room highlighting the need for this feature a while back but we eventually held off on adding it for the sole reason of setting up an operator first, IIRC.

1. The reason why for this PR is that coupled custom resources metrics targets and kube-state-metrics deployment causes managing custom resource configuration not easy.  And then I found a CRD can handle this issue and also solve #1948.

Currently, one of these flags is added into kube-state-metrics deployment.
* `--custom-resource-state-config "inline yaml (see example)"` or
* `--custom-resource-state-config-file /path/to/config.yaml`

For example, for a company with two teams (monitoring platform team and application team), application team needs to change kube-state-metrics deployment managed by monitoring platform team.

2. This isn't make KSM as KSM operator fully. It won't add CRD for non custom resource metrics. I agree that CRD makes control KSM memory harder, since new metrics might be added overtime. A flag `--custom_resources_ksm_cr_watched_labels` is added to handle this case.

3. KSM already watches pods/endpoints and reconcile around autosharding. This PR only adds one additional watch on CustomResourceMonitor CR and update custom metrics store dynamically.

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,149 @@
...
+}
+```
+
+Kube-state-metrics listens on add, update and delete events of CustomResourceMonitor via Kubernetes
+client-go reflectors. On these events kube-state-metrics lists all CustomResourceMonitor CRs and concatenate CRs into one config `customresourcestate.Metrics` which has the same format with configs of `--custom-resource-state-config`. This generated custom resource config updates CustomResourceStore by adding monitored custom resource stores and deleting unmonitored custom resource stores.
+
+
+```yaml
+# example cr
+apiVersion: customresource.ksm.io/v1alpha1
```
CatherineF-dev, line 74: I find k8s.io needs additional procedures. Planned to use customresource.ksm.k8s.io before.

CatherineF-dev:
Hi @dgrisonnet, could you help review this design doc when you're available? Thanks!

Related issues https://github.com/kubernetes/kube-state-metrics/issues/2050 https://github.com/kubernetes/kube-state-metrics/issues/1948

dgrisonnet:
Sure, I'll try to catch up on the various discussion asap

docs/design/custom-resource-metrics-crd.md
rexagod: I'd prefer `*.k8s.io` group for the CR given we exist under the same organization here.

logicalhan (commented):
/assign

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,207 @@
...
+
+### New flags
+Apart from existing two flags (`--custom-resource-state-config ` and `--custom-resource-state-config-file`), these three flags will be added: 
+* `--custom_resource_monitor_enabled`: whether watch CustomResourceMonitor CRs or not.
+* `--custom_resource_monitor_labels`: only watch CustomResourceMonitor with label=x. It's used to avoid double custom metrics collection when multiple kube-state-metrics are installed.
+* `--custom_resource_monitor_namespaces`: only watch CustomResourceMonitor under namespaces=x.
+
+If `--custom_resources_monitor_enabled` is set, `--custom-resource-state-config` and `--custom-resource-state-config-file` will be ignored.
+
+### CustomResourceMonitor Definition
```
bwplotka, line 64: Interesting. Wonder if it wouldn't be simpler and easier for both maintainers and users to have some additional field/annotation in each CRD we want to create metrics for? 🤔 

If not, can we add discuss this in some "alternatives" section?

bwplotka: @dgrisonnet mentioned something related: https://github.com/kubernetes/kube-state-metrics/issues/1899

Still, wonder if it's worth to capture the alternative.

dgrisonnet: Yeah I really like that idea aswell. The UX is pretty clean with that approach and it would be a great alternative to the CRD + CEL combo.

> If not, can we add discuss this in some "alternatives" section?

:100: we should definitely add it to the document

CatherineF-dev: Thx, added into alternative.

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+
+## Proposal
+
+Add a custom resource definition (CustomResourceMonitor) which contains customresourcestate.Metrics.
+
+kube-state-metrics watched on CustomResourceMonitor CRs and concatenate these CRs into one config `customresourcestate.Metrics` which has the same content using `--custom-resource-state-config`. 
+
+### New flags
+Apart from existing two flags (`--custom-resource-state-config ` and `--custom-resource-state-config-file`), these three flags will be added: 
+* `--custom_resource_monitor_enabled`: whether watch CustomResourceMonitor CRs or not.
```
mrueg, line 58:
```suggestion
* `--custom-resource-monitor`: whether watch CustomResourceMonitor CRs or not.
```

I would prefer not to have underscores in flag names, if it's enabling it, let's just set it to true/false

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+* `--custom_resource_monitor_labels`: only watch CustomResourceMonitor with label=x. It's used to avoid double custom metrics collection when multiple kube-state-metrics are installed.
+* `--custom_resource_monitor_namespaces`: only watch CustomResourceMonitor under namespaces=x.
+
+If `--custom_resources_monitor_enabled` is set, `--custom-resource-state-config` and `--custom-resource-state-config-file` will be ignored.
```
mrueg, line 62: Is there a need to make it mutually exclusive?

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+
+### CustomResourceMonitor Definition
+
+* GroupName: kubestatemetrics.io
+   * Alternative kubestatemetrics.k8s.io: 1. *.k8s.io needs approval 2. ksm isn't inside k/k repo (need to double confirm)
```
mrueg, line 67: Do you know if this really needs approval and by whom? 

autoscaler also uses *.k8s.io: https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/examples/hamster.yaml#L8 

I would prefer using ksm.k8s.io or kubestatemetris.k8s.io

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+- kind: CustomResourceStateMetrics
++ kind: CustomResourceMonitor
++ metadata:
++  name: crm_nodepool
++  labels:
++    monitoring.backend.io: true
+spec: # copy content from --custom-resource-state-config-file
+```
+
+## CUJ
```
mrueg, line 208: What is a CUJ?

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+## Proposal
+
+Add a custom resource definition (CustomResourceMonitor) which contains customresourcestate.Metrics.
+
+kube-state-metrics watched on CustomResourceMonitor CRs and concatenate these CRs into one config `customresourcestate.Metrics` which has the same content using `--custom-resource-state-config`. 
+
+### New flags
+Apart from existing two flags (`--custom-resource-state-config ` and `--custom-resource-state-config-file`), these three flags will be added: 
+* `--custom_resource_monitor_enabled`: whether watch CustomResourceMonitor CRs or not.
+* `--custom_resource_monitor_labels`: only watch CustomResourceMonitor with label=x. It's used to avoid double custom metrics collection when multiple kube-state-metrics are installed.
```
mrueg, line 59: Is this supposed to be a labelSelector similar to Kubernetes implements them? Can I add multiple labels?

docs/design/custom-resource-metrics-crd.md
```diff
@@ -0,0 +1,211 @@
...
+* `--custom_resource_monitor_namespaces`: only watch CustomResourceMonitor under namespaces=x.
```
mrueg, line 60: Should we consider a denylist for namespaces as well? 
In case someone has a "platform KSM" (only selected namespaces) and a "product KSM" (everything else)?

docs/design/custom-resource-metrics-crd.md
CatherineF-dev: To simplify UX. 

The migration from `--custom-resource-state-config/--custom-resource-state-config-file` to `--custom_resources_monitor` is simple.

docs/design/custom-resource-metrics-crd.md
CatherineF-dev: Updated: CUJ = "Critical User Journey"

docs/design/custom-resource-metrics-crd.md
CatherineF-dev: Yes, it is [labelSelector](https://kubernetes.io/docs/concepts/overview/working-with-objects/labels/#list-and-watch-filtering). Multiple labels will be joined by AND.

guettli:
@CatherineF-dev I think it would be createa to have a CustomResourceMonitor CRD. Are you still working on this PR?

rexagod:
@CatherineF-dev Are you working on this? I see how this could be a large-enough issue entailing multiple cases to warrant more hands, and if so, feel free to bring this up in the next SIG call and we can have more folks contributing to your fork, if that sounds good!

CatherineF-dev:
Yes, from previous discussion, I remember we plan to split KSM and KSM Custom Metrics into 2 binaries first. Then add CRD feature.

(This prototype was working in my k8s cluster)

rexagod:
I believe we can get this in regardless, and include this change whenever the split happens. Getting this one in would resolve a lot of issues, in addition to deserving its own minor release.

I'm +1 for making this happen pre-split, since we don't plan on doing that for a good while.

tyxeron:
@CatherineF-dev 
Whats the status on this issue?

This PR would be very helpful

mrueg:
Closing this as we'll focus on resource-state-metrics for custom-defined metrics.

PR #2052, opened 2023-04-20
feat: Allow prefixing wildcard `labelsFromPath`

rexagod:
**What this PR does / why we need it**: Allow prefixing * paths, for instance, a "foo_" prefix like so, foo_*: [foo, bar, baz], will result in foo_baz being the key for all object resolutions under baz.

**Which issue(s) this PR fixes**: Self-driven.

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -193,6 +193,8 @@ spec:
             # whole objects may be copied into labels by prefixing with "*"
             # *anything will be copied into labels, with the highest sorted * strings first
             "*": [metadata, labels]
+            # a prefix before the asterisk will be used as a label prefix
+            "lorem_*": [metadata, annotations]
             "**": [metadata, annotations]
             
             # or specific fields may be copied. these fields will always override values from *s
@@ -203,8 +205,12 @@ spec:
 Produces the following metrics:
 
 ```prometheus
-kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", customresource_version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 2
-kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", customresource_version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 4
+kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", 
+customresource_version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a",
+lorem_bar="baz",lorem_qux="quxx",} 2
+kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", 
+customresource_version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b",
+lorem_bar="baz",lorem_qux="quxx",} 4
 ```
 
 #### VerticalPodAutoscaler
--- a/pkg/customresourcestate/registry_factory_test.go
+++ b/pkg/customresourcestate/registry_factory_test.go
@@ -134,12 +134,15 @@ func Test_addPathLabels(t *testing.T) {
 		{name: "*", args: args{
 			obj: cr,
 			labels: map[string]valuePath{
-				"*1":  mustCompilePath(t, "metadata", "annotations"),
-				"bar": mustCompilePath(t, "metadata", "labels", "foo"),
+				"*1":             mustCompilePath(t, "metadata", "annotations"),
+				"bar":            mustCompilePath(t, "metadata", "labels", "foo"),
+				"label_object_*": mustCompilePath(t, "metadata", "annotations"),
 			},
 			want: map[string]string{
-				"qux": "quxx",
-				"bar": "bar",
+				"qux":              "quxx",
+				"bar":              "bar",
+				"label_object_qux": "quxx",
+				"label_object_bar": "baz",
 			},
 		}},
 	}
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -507,21 +507,24 @@ func addPathLabels(obj interface{}, labels map[string]valuePath, result map[stri
 	// always do that first so other labels can override
 	var stars []string
 	for k := range labels {
-		if strings.HasPrefix(k, "*") {
+		if strings.HasPrefix(k, "*") || strings.HasSuffix(k, "*") {
 			stars = append(stars, k)
 		}
 	}
 	sort.Strings(stars)
-	for _, k := range stars {
-		m := labels[k].Get(obj)
+	for _, star := range stars {
+		m := labels[star].Get(obj)
 		if kv, ok := m.(map[string]interface{}); ok {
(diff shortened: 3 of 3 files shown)


rexagod:
_TODO: Add a doc entry._

mrueg:
After looking a bit closer at what this does, do the semantics of this syntax look intentional to users?
 
If I see `*` I wouldn't think that this is adding a prefix but more a filter to only include keys with this specific prefix. 
Can we find a better syntax?

rexagod:
Closing in favor of the [newer model](https://github.com/kubernetes/kube-state-metrics/pull/2068#issuecomment-1591437858).

rexagod:
Reopening in favor of https://github.com/kubernetes/kube-state-metrics/pull/2068#issuecomment-1598669259.

docs/customresourcestate-metrics.md
```diff
@@ -193,6 +193,8 @@ spec:
             # whole objects may be copied into labels by prefixing with "*"
             # *anything will be copied into labels, with the highest sorted * strings first
             "*": [metadata, labels]
+            # a prefix before the asterisk will be used as a label prefix
```
dgrisonnet, line 196: That sounds like a fair feature request to me. I'll have a look at the code later.

We also need to consider if this would add some complexity when converting to CEL

rexagod:
Up for reviews.

dgrisonnet (commented):
Should we just support globbing in general for labels from path and use https://github.com/gobwas/glob instead of doing the parsing manually?

rexagod:
For the CRS snippet below, the output will contain labels with the prefix `foo_` for every key present in the `fooObj` and it's values marshalled into the respective label values. There isn't actually any [`Match`](https://github.com/gobwas/glob/blob/v0.2.3/glob.go#L10) happening, and the `*` is but a placeholder to denote where the label keys will be inserted at.
```yaml
labelsFromPath:
  "foo_*": ["bar", "fooObj"]
```
An actual regex match over a sample set of `fooObjA`, `fooObjB` for `foo_*` will return empty (which is why I doubt I might be failing to see here how we'd benefit from this library, and where). This syntax, in this context, addresses the need to manipulate generated labels, as opposed to validating them against an expression.

rexagod:
This basically builds on top of the originally present `*` `labelFromPath` key, which would marshal every key-value pair in an object as is, into label-values, whereas this allows prepending text to these "as-is" keys.

rexagod:
Do you mean supporting expressions such as `foo_b?r.*` for `labelFromPaths` using the library? As in, all `fooObj` keys that match the `b?r.*` pattern and include only those as labels? This may prove ambiguous for estimating where the prefix ends and the regexp begins (do we want to move prefixes to a different field)?

pkg/customresourcestate/registry_factory_test.go
```diff
@@ -134,12 +134,15 @@ func Test_addPathLabels(t *testing.T) {
 		{name: "*", args: args{
 			obj: cr,
 			labels: map[string]valuePath{
-				"*1":  mustCompilePath(t, "metadata", "annotations"),
-				"bar": mustCompilePath(t, "metadata", "labels", "foo"),
+				"*1":             mustCompilePath(t, "metadata", "annotations"),
```
dgrisonnet, line 137: Unrelated to the actual PR, but why does `*1` yield `"qux": "quxx"`. Is this the way we tell the code to take the first annotation?

dgrisonnet:
Disregard https://github.com/kubernetes/kube-state-metrics/pull/2052#pullrequestreview-1602937314, I got super confused with the syntax.

The use of the `*` character is so confusing here. I would see a lot of people, myself included to expect some kind of globbing from this `*`, but that not the case at all, we are just precising a prefix or suffix, which shouldn't need a `*` in the syntax.

The current approach has 4 scenarios:
```
"*": [metadata, labels]
"prefix_*": [metadata, labels]
"*_suffix": [metadata, labels]
"prefix_*_suffix": [metadata, labels]
```
This relies on an implicit behavior from the parser that requires users to have an understand of the code. Whereas we could make it more expressive and have something like:
```
labelsFromPath:
  - path: [metadata, labels]
    prefix: "lorem"
    suffix: "ipsum"
```

IMO, the more explicit a syntax is, the easier it is for an external user to understand.

Prometheus-adapter is a prime example of a very bad syntax. Its configuration was made overly complex in an attempt to make it more generic and adaptable (I guess), but the result is that it confuses everyone and discourage users from using the project.
Here is a "simple" example from the repo:
```yaml
rules:
- seriesQuery: '{__name__=~"^container_.*",container!="POD",namespace!="",pod!=""}'
  resources:
    overrides:
      namespace: {resource: "namespace"}
      pod: {resource: "pod"}
  name:
    matches: "^container_(.*)_seconds_total$"
  metricsQuery: "sum(rate(<<.Series>>{<<.LabelMatchers>>,container!="POD"}[2m])) by (<<.GroupBy>>)"
```
We need to avoid reaching that point at all cost.

rexagod:
I see. My only concern here is when moving from, 

```
"*": [metadata, labels]
"prefix_*": [metadata, labels]
"*_suffix": [metadata, labels]
"prefix_*_suffix": [metadata, labels]
```

to, 

```
labelsFromPath:
  - path: [metadata, labels]
    prefix: "lorem"
    suffix: "ipsum"
```

will require multiple `path`s, since the `lorem_*` approach allows for specifying distinct paths that can differ from each other, such as,

```
"*": [metadata, labels, foo]
"prefix_*": [metadata, labels, bar]
"*_suffix": [metadata, labels, baz]
"prefix_*_suffix": [metadata, labels, que]
```

as compared to,

```
each:
  type: Gauge
  gauge:
    path: [status, conditions]
    labelsFromPath:
      - path: [metadata, labels, foo]
      	prefix: ""
      	suffix: ""
      - path: [metadata, labels, bar]
      	prefix: "prefix"
      	suffix: ""
      <"*_suffix" and "prefix_*_suffix" as well>
      type: ["type"]
    valueFrom: ["status"]
```

which may give rise to ambiguous conclusions since there's a `path` definition outside `labelsFromPath` as well, to allow relative definitions for labels such as `type` (above).

dgrisonnet:
> which may give rise to ambiguous conclusions since there's a path definition outside labelsFromPath as well, to allow relative definitions for labels such as type (above).

I see what you mean here, although that was already the case implicitly, it will now be explicit and could be very confusing. Maybe we could call it `from` or `fromPath`?

rexagod:
@dgrisonnet Please correct me if I'm wrong but I believe transforming `labelsFromPath` to a list will be a breaking change. For instance, the following (currently the way to specify label key-values with no wildcards) will need to be adjust to conform to the newer YAML list-based format.

```
labelsFromPath:
  status: [conditions, "0", status]
```

could either be,

```
labelsFromPath:
  - path : [conditions, "0", status]
    prefix: "status"
    suffix: ""
```

or,

```
labelsFromPath:
  - path : [conditions, "0", status]
    prefix: ""
    suffix: "status"
```

dgrisonnet:
Yes it would be a breaking changes. That would also be the case for this comment on your other PR which would apply also here: https://github.com/kubernetes/kube-state-metrics/pull/2068#discussion_r1511539495

We will need to break the config entirely if we want to build solid foundation with CEL, so hopefully we will change it in the future. In the meantime we can continue with the current UX, but we will need to draw a line at one point. I would be fine with drawing it once your two PRs are merged and then stop accepting feature requests until we've migrated to CEL.

PR #2059, opened 2023-05-07
Simplify custom resource state metrics API using CEL

### LINKED ISSUE #1978: Simplify custom resource metrics API by leveraging jq or yq
A simplified API for [CustomResourceStateMetrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#multiple-metricskitchen-sink), which only supports `values`, `labels` and `labels_appended`, instead of supporting `each`, `path`, `labelFromKey`, `labelsFromPath`, `valueFrom`, `commonLabels`, `labelsFromPath` and `*`.

```bash
kind: CustomResourceStateMetricsV2
spec:
  resources:
    - groupVersionKind:
        group: myteam.io
        kind: "Foo"
        version: "v1"
      metrics:
        - name: "ready_count"
          help: "Number Foo Bars ready"
          values: jq '[.status.sub[].ready]' # valueFrom: [ready] // [2,4]
          labels:
          - jq '[.status]'
          - jq '[ .status.sub | keys | .[] | {name: .}]' # labelFromKey: type // [{"name": "type-a"}, {"name": "type-b"}]
          - jq '[{ custom_metric:"yes" }]' # custom_metric: "yes" // [{custom_metric="yes"}]
          - jq '[.metadata.labels]' # "*": [metadata, labels] // [{"bar": "baz","qux": "quxx"}]
          - jq '[.metadata.annotations]' # "**": [metadata, annotations] // [{"foo": "bar"}]
          - jq '[{ name: .metadata.name }]' # name: [metadata, name] // [{"name": "foo"}]
          - jq '[{ foo: .metadata.labels.foo }]' # foo: [metadata, labels, foo] // [{foo": "bar"}]
         labels_appended:
         - jq '[.status.sub[].active | {active: .}]' # labelsFromPath:  active: [active] // [{active": 1}, {"active": 3}]
        
```
The number of metric `ready_count` = |label_1| * |label_2| * |label_3| * ... * |label_n|. In the above case, the cardinality is 2 so it will have these 2 metrics.
```
kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", customresource_version="v1", active="1",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-a"} 2
kube_customresource_ready_count{customresource_group="myteam.io", customresource_kind="Foo", customresource_version="v1", active="3",custom_metric="yes",foo="bar",name="foo",bar="baz",qux="quxx",type="type-b"} 4
```

**Why is this needed**:
1.  easier for KSM community to maintain
2. easier for users to use and debug custom resource metrics

**Additional context**

CatherineF-dev:
**What this PR does / why we need it**: Simplify custom resource state metrics API

Fixes #1978

### DIFF
--- a/docs/design/simplify-custom-resource-metrics-api.md
+++ b/docs/design/simplify-custom-resource-metrics-api.md
@@ -1,0 +1,198 @@
+# Kube-State-Metrics - Simplify Custom Resource State Metrics API Proposal
+
+
+---
+
+Author: Catherine Fang (CatherineF-dev@), Han Kang (logicalhan@)
+
+Date: 7. May 2023
+
+Target release: v
+
+---
+
+
+## Glossary
+- CR: custom resource, similar to an instance of a class
+- CRD: custom resource definition, similar to a class
+
+## Problem Statement
+
+### Background
+Current [Custom Resource State Metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#multiple-metricskitchen-sink) supports 8+ operations to extract metric value and labels from custom resource.
+- each
+- path
+- labelFromKey
+- labelsFromPath
+- valueFrom
+- commonLabels
+- labelsFromPath
+- *.
+- ...
+
+### Problem 
+1. Custom resource metrics API isn't scalable and it's a little hard to maintain. 
+  1.1 The maintaining work is O(8) and there are several bugs around these 8 operations. For example, Crash on nonexistent metric paths in custom resources (#1992).
+  1.2 More additional operations might be added to satisfy other needs.
+2. Custom resource metrics API with existing 8 operations is not complete, which means some cases aren't covered. For example, it doesn't support querying number of CRs under one CRD.
+
+## Goal
+
+- Simplify 8 operations into one operation to reduce maintaining work.
+- A complete API, so that can support more cases. For example, querying number of CRs under one CRD.
+
+## Proposal
+
+Use common expression language ([cel](https://kubernetes.io/docs/reference/using-api/cel/)) to extract fields from custom resource as metric labels or metric value.
+
+
+```
+kind: CustomResourceStateMetricsV2
+spec:
+  resources:
+    - groupVersionKind:
+        group: myteam.io
+        kind: "Foo"
+        version: "v1"
+      mode: for_loop # or merged
+      metrics:
+        - name: "ready_count"
+          help: "Number Foo Bars ready"
+          values:  x.cel_selection_1 // [2, 4]
+          labels:
+          - x.cel_selection_2 // [{"cr_name": "bar"}], it will be copied into 2 same items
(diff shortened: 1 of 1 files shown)


docs/design/simplify-custom-resource-metrics-api.md
```diff
@@ -0,0 +1,198 @@
...
+kind: CustomResourceStateMetricsV2
+spec:
+  resources:
+    - groupVersionKind:
+        group: myteam.io
+        kind: "Foo"
+        version: "v1"
+      mode: for_loop # or merged
+      metrics:
+        - name: "ready_count"
```
mrueg, line 186: how is the metrics type defined?

rexagod: I presume we'll still need fields with hardcoded values, eg., `type`.

rexagod: Noting here that there is still the need for non-path-based constructs, like so:
- `type` (used to check OpenMetrics conformance, and if `value` field is applicable)
- `path` (relative paths when used with `labelsFrom*`)
	- non-relative
	- relative
	- multi-relative (relative to different paths) (KSM does not supports this right now)
- `commonLabels`
- [`list`](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#stateset) (maybe replaced by a CEL-based alternative)

Something we might want to persue under the configuration revamp is to keep such constructs to a minimum (and try to replace most of these using CEL, wherever possible), as we may find ourselves implementing custom logic and introducing complexity as such constructs grow over time.

rexagod (commented):
Thank you for working on this!

rexagod:
While this is easily one of the features we'd like to see implemented, tested, and polished enough throughout `v2.x` releases to finally make its way into `v3`, I just want to mention the primary concern around this, before we commit to the move, was if CEL could replace the existing native metrics. If we could verify the same first, I believe we could then move forward confidently which this change.

docs/design/simplify-custom-resource-metrics-api.md
```diff
@@ -0,0 +1,198 @@
...
+- Simplify 8 operations into one operation to reduce maintaining work.
+- A complete API, so that can support more cases. For example, querying number of CRs under one CRD.
+
+## Proposal
+
+Use common expression language ([cel](https://kubernetes.io/docs/reference/using-api/cel/)) to extract fields from custom resource as metric labels or metric value.
+
+
+```
+kind: CustomResourceStateMetricsV2
```
chrischdi, line 50:
```suggestion
kind: CustomResourceStateMetrics
apiVersion: v2
```

I'd propose to use the common versioning keys :-) We could even provide conversion from the old to the new version if we want

mrueg:
Closing this as custom-resource state metrics will be phased out once resource-state-metrics is available.

Issue #2051, opened 2023-04-20
customresource GitRepo cannot be scraped


ulikl:
**What happened**:

I try to get metrics for the following custom resources:
```
$ kubectl api-resources 
...
gitrepos                                                        fleet.cattle.io/v1alpha1                     true         GitRepo
```

During the execution the kind "GitRepo" is replaced for the kube api call with "gitrepoes", for which the api call fails as this kind does not exists. 
There is an "e" too much  in  gitrepo**e**s.

Here the logs
```
I0420 14:48:28.600525       1 reflector.go:257] Listing and watching fleet.cattle.io/v1alpha1, Kind=GitRepo from pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169
I0420 14:48:28.600792       1 round_trippers.go:466] curl -v -XGET  -H "Accept: application/json" -H "User-Agent: kube-state-metrics/v2.8.0 (linux/amd64) kubernetes/" -H "Authorization: Bearer <masked>" 'https://10.43.0.1:443/apis/fleet.cattle.io/v1alpha1/gitrepoes?limit=500&resourceVersion=0'
I0420 14:48:28.601511       1 round_trippers.go:553] GET https://10.43.0.1:443/apis/fleet.cattle.io/v1alpha1/gitrepoes?limit=500&resourceVersion=0 404 Not Found in 0 milliseconds

W0420 14:48:44.174514       1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: failed to list fleet.cattle.io/v1alpha1, Kind=GitRepo: the server could not find the requested resource
E0420 14:48:44.174541       1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.1/tools/cache/reflector.go:169: Failed to watch fleet.cattle.io/v1alpha1, Kind=GitRepo: failed to list fleet.cattle.io/v1alpha1, Kind=GitRepo: the server could not find the requested resource
```

and the scrape is empty.

with "Kind: gitrepos" I get

```
# initial read ok
[...]

PR #2065, opened 2023-05-12
customresourcestate: Use pluralize logic from k8s/apimachinery

mrueg:

kubernetes and kubebuilder use a different logic to generate / guess plurals. The goal of this PR is to use the same logic as Kubernetes.

Kubernetes: https://github.com/kubernetes/apimachinery/blob/master/pkg/api/meta/restmapper.go#L126
Kubebuilder: https://github.com/gobuffalo/flect

See also:
https://github.com/kubernetes-sigs/kubebuilder/issues/3402
https://github.com/kubernetes-sigs/kubebuilder/pull/3408

None
None

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -5,7 +5,6 @@ go 1.19
 require (
 	github.com/dgryski/go-jump v0.0.0-20211018200510-ba001c3ffce0
 	github.com/fsnotify/fsnotify v1.6.0
-	github.com/gobuffalo/flect v1.0.2
 	github.com/google/go-cmp v0.5.9
 	github.com/oklog/run v1.1.0
 	github.com/prometheus/client_golang v1.15.1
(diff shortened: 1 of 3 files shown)


logicalhan:
But the pluralization logic in upstream k8s is super broken...

logicalhan (commented):
What does this solve?

Kubebuilder didn't opt to use upstream k8s pluralization logic because it was so wonky to begin with, so arguably this makes things worse.

logicalhan:
For instance, `mess` gets pluralized to `meses` using upstream k8s pluralization logic.

logicalhan:
> For instance, `PodSchedulingHints` gets pluralized to `PodSchedulingHintses` using upstream k8s pluralization logic.

mrueg:
@logicalhan I am aware that it's broken. At least it's somewhat deterministically broken instead of relying on flect to have a better heuristic (and also sometimes getting things wrong https://github.com/kubernetes/kube-state-metrics/issues/2051).

What are better options?
Should we query the Kubernetes API during startup instead?

A-Hilaly:
@mrueg how about using https://github.com/gertd/go-pluralize?

rexagod:
That's doable, but there will always be a room for risk unless machinery and kubebuilder decide on the same pluralizing logic, which ideally should become the de-facto standard for generating resource names.

mrueg:
@dgrisonnet https://github.com/kubernetes/kube-state-metrics/issues/2093 I think moving to GRV might make this obsolete

dgrisonnet:
Oh right, I need to catch up on that. Thanks for the update

PR #2067, opened 2023-05-15
feat: allow field KV general matching

rexagod:
**What this PR does / why we need it**: Currently all matching support is limited to map-list representations; this adds support for a more general "key:value" based matching expression for any `interface{}` field (non-map-list-like).

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -522,6 +522,9 @@ Examples:
 
 # if the value to be matched is a number or boolean, the value is compared as a number or boolean  
 [status, conditions, "[value=66]", name]  # status.conditions[1].name = "b"
+
+# For generally matching against a field in an object schema, use the following syntax:
+[metadata, "name=foo"] # if v, ok := metadata[name]; ok && v == "foo" { return v; } else { /* ignore */ }
 ```
 
 ### Wildcard matching of version and kind fields
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -623,6 +623,16 @@ func compilePath(path []string) (out valuePath, _ error) {
 				part: part,
 				op: func(m interface{}) interface{} {
 					if mp, ok := m.(map[string]interface{}); ok {
+						kv := strings.Split(part, "=")
+						if len(kv) == 2 /* k=v */ {
+							key := kv[0]
+							val := kv[1]
+							if v, ok := mp[key]; ok {
+								if v == val {
+									return v
+								}
+							}
+						}
 						return mp[part]
 					} else if s, ok := m.([]interface{}); ok {
 						i, err := strconv.Atoi(part)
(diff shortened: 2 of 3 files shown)


docs/customresourcestate-metrics.md
```diff
@@ -480,4 +480,7 @@ Examples:
 
 # if the value to be matched is a number or boolean, the value is compared as a number or boolean  
 [status, conditions, "[value=66]", name]  # status.conditions[1].name = "b"
+
+# For generally matching against a field in an object schema, use the following syntax:
+[metadata, annotations, "bar:baz"] # metadata.annotations["bar:baz"] = "baz" (if present, ignored otherwise)
```
mrueg, line 485: should this read as
```suggestion
[metadata, annotations, "bar:baz"] # metadata.annotations["bar"] = "baz" (if present, ignored otherwise)
```

mrueg (commented):
Do you an example use case for this?

rexagod:
> Do you an example use case for this?

So right now we don't support generating metrics from a map for a `key:value` match. This is useful in cases where there are multiple instances (say, `deployment`s) with key-value pairs that may or may not be present on them, in which case we'd want to make sure that even through the key is same as the one specified in the `path`, the value matches as well.

```yaml
# instance A
# ...
labels:
 foo: "bar"
# ...
---
# instance B
# ...
labels:
 foo: "baz"
# ...
---
# config
# ...
labelsFromPath:
  foo: [metadata, labels, "foo:bar"]
# ...
```

rexagod:
Closing in favor of the [newer model](https://github.com/kubernetes/kube-state-metrics/pull/2068#issuecomment-1591437858).

rexagod:
Reopening in favor of https://github.com/kubernetes/kube-state-metrics/pull/2068#issuecomment-1598669259.

docs/customresourcestate-metrics.md
dgrisonnet, line 485: To be in line with how we specify values, what would you think about:
```
[metadata, annotations, "[key=bar]", "[value=baz]"]
```
I feel like it would be easier for users if we stick to the existing UX we have for values and just extend it to keys.

docs/customresourcestate-metrics.md
```diff
@@ -480,4 +480,7 @@ Examples:
 
 # if the value to be matched is a number or boolean, the value is compared as a number or boolean  
 [status, conditions, "[value=66]", name]  # status.conditions[1].name = "b"
+
+# For generally matching against a field in an object schema, use the following syntax:
+[metadata, annotations, "bar=baz"] # metadata.annotations["bar=baz"] = "baz" (if present, ignored otherwise)
```
mrueg, line 485: Can we do a "name=a" here similar to a couple of lines above?

rexagod: Addressed, PTAL.

mrueg:
/approve

/hold 
for @dgrisonnet to take a look

dgrisonnet (approved):
/lgtm
/unhold

PR #2068, opened 2023-05-15
feat: Support list expansions

rexagod:
**What this PR does / why we need it**: Allow CRS configuration to take in path values for lists, denoted by "*". This means it's possible to specify `[..., <list>, "*", foo, ...]` in the configuration to dynamically generate multiple metrics that reflect different states of `foo` in various list elements.

### DIFF
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -552,8 +552,11 @@ path: [status, namespaces, "*", status]
 labelsFromPath:
   # this can be combined with the wildcard prefixes feature introduced in #2052
   "available_*": [available]
+  "lorem_*": [pending]
+# this can be combined with dynamic valueFrom expressions introduced in #2068
+valueFrom: [lorem_resourceB] # or [available_resourceB]
 # outputs:
-...{...,available_resourceA="10",available_resourceB="20",...} ...
+...{...,available_resourceA="10",available_resourceB="20",lorem_resourceA="0",lorem_resourceB="6"...} 6
 ```
 
 ### Wildcard matching of version and kind fields
--- a/docs/customresourcestate-metrics.md
+++ b/docs/customresourcestate-metrics.md
@@ -531,6 +531,29 @@ Examples:
 
 # For generally matching against a field in an object schema, use the following syntax:
 [metadata, "name=foo"] # if v, ok := metadata[name]; ok && v == "foo" { return v; } else { /* ignore */ }
+
+# expand a list
+[spec, order, "*", value] # syntax
+# for the object snippet below
+...
+status:
+  namespaces:
+  - namespace: "foo"
+    status:
+      available:
+        resourceA: '10'
+        resourceB: '20'
+      pending:
+        resourceA: '0'
+        resourceB: '6'
+...
+# resolves to: [status, namespaces, <range len(namespaces)>, status] (a multi-dimensional array)
+path: [status, namespaces, "*", status]
+labelsFromPath:
+  # this can be combined with the wildcard prefixes feature introduced in #2052
+  "available_*": [available]
+# outputs:
+...{...,available_resourceA="10",available_resourceB="20",...} ...
 ```
 
 ### Wildcard matching of version and kind fields
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -228,6 +228,44 @@ type compiledGauge struct {
 	labelFromKey string
 }
 
+func underscoresToIndices(extractedValueFrom string, it interface{}) interface{} {
+	// `it` is the search space.
+	m, isResolvableMap := it.(map[string]interface{})
+	arr, isResolvableArr := it.([]interface{})
+	if !isResolvableMap && !isResolvableArr {
+		return nil
+	}
+	// `extractedValueFrom` is the search term.
+	// Split `extractedValueFrom` by underscores.
+	terms := strings.Split(extractedValueFrom, "_")
+	resolvedTerm := interface{}(terms[0])
+	for _, term := range terms[1:] {
+		if isResolvableMap {
+			t, ok := m[term]
+			if !ok {
+				return resolvedTerm
+			}
+			resolvedTerm = t
+			if _, isResolvableMap = t.(map[string]interface{}); isResolvableMap {
+				m = t.(map[string]interface{})
+			}
+		} else if isResolvableArr {
+			for _, el := range arr {
+				t, ok := el.(map[string]interface{})
+				if !ok {
+					continue
+				}
+				if v, ok := t[term]; ok {
+					resolvedTerm = v
+					m = t
+					break
+				}
+			}
+		}
+	}
+	return resolvedTerm
+}
+
 func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error) {
 	onError := func(err error) {
 		errs = append(errs, fmt.Errorf("%s: %v", c.Path(), err))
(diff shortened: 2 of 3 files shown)


docs/customresourcestate-metrics.md
```diff
@@ -480,4 +480,7 @@ Examples:
 
 # if the value to be matched is a number or boolean, the value is compared as a number or boolean  
 [status, conditions, "[value=66]", name]  # status.conditions[1].name = "b"
+
+# expand a list
```
mrueg, line 484: Can you explain a bit more what it does?

CatherineF-dev:
A related https://github.com/kubernetes/kube-state-metrics/issues/1978

I think cel can support this.

dgrisonnet:
I agree with Catherine, we shouldn't add more complexity to the existing config but rather focus on the new model.

rexagod:
So, are we sunsetting the older model?

CatherineF-dev:
Reuse the wheel (cel, a query model) can reduce our maintaining work on how to extract fields. 
Then we can spend efforts on other things, instead of putting many efforts on implementing a querying language.

CatherineF-dev:
cel should be a `complete` language, which means it can extract every field.

cel is more powerful than current APIs.

rexagod:
@dgrisonnet I'm assuming going forward with the newer model will mean we'll deprecated the older one. So, with that in mind, are PRs such as this and others like: #2067, #2052 obsolete and should be closed?

dgrisonnet:
I need to catch up on the discussions, a lot happened when I was away, but my general feeling is that we are starting to make the configuration way too complex. My fear is that this is going to become too confusing for the users and I am also worries about the current state of the parser. Without a proper lexer/parser combination this is slowly becoming unmaintainable.

IMO with the state of things, we should rather focus on building new foundations for the config rather than making the existing one more complex since it will just make the migration process harder. wdyt?

rexagod:
I couldn't agree more.

The current CRS featureset, in addition to being complex and hardly maintainable for developers, doesn't have a clear, semantical, or easy-to-understand UX at the moment. Off the top of my mind, I can think of a few examples that go on to show a couple of such instances. I personally would not at all expect the user to be familiar with the following intricacies.

- Relative paths can only be defined **once**, and the fact that technically there's two `labelsFromPath` possible but the other one does not support it.

	```yaml
          each:
            type: Info
            info:
              path: [spec, selector, labels]
              labelsFromPath:
                "*": [matchLabels]
          labelsFromPath:
            name: [metadata, name]
   ```
- The outer `lablesFromPath` operates on the subset that satified the inner `labelsFromPath`. In the configuration snippet below, the outer `labelsFromPath` will only generate `name` labels for all objects that satisfied the inner `labelsFromPath`, i.e., a non-nil `foo` label.

	```yaml
	 	  each:
            type: Info
            info:
              path: [metadata]
              labelsFromPath:
                "foo_label": [labels, foo]
          labelsFromPath:
            name: [metadata, name]
	```
- The difference in each of the configurations below, which looks more like a side-effect of another change.

	```yaml
		  each:
            type: Info
            info:
              path: [users]
              labelsFromPath:
                user: []
    ```
	outputs: `kube_customresource_metric_name{..., user="user1"} 1` and `kube_customresource_metric_name{..., user="user2"} 1` while,
	
	```yaml
		  each:
            type: Info
            info:
              labelsFromPath:
                user: [users]
    ```
	outputs: `kube_customresource_metric_name{..., user="[user1 user2]"} 1`.

TBH only reason I raised the aforementioned PRs was because I wasn't sure about the direction in which we were going to move forward over the last month regarding this, and assumed that doing so would entail discussions spanning a good while, so things would basically be done the same way as they were till now.

But, I'm enthusiastic and happy to see that there's a consensus on improving the current approach, and the discussions are yeilding actionable results. It'd be nice to see the proposed solution replacing as many configuration constructs as possible, in order to achieve a more accessible and deterministic UX for the user, thanks to @CatherineF-dev. 🎉

rexagod:
While it's great to see we've got the ball rolling on the CEL implementation, I think in the meantime, it makes sense to continue supporting and improving the current configuration constructs until that is in place and we start with the deprecation cycle for the current constructs that are covered by it.

logicalhan:
> While it's great to see we've got the ball rolling on the CEL implementation, I think in the meantime, it makes sense to continue supporting and improving the current configuration constructs until that is in place and we start with the deprecation cycle for the current constructs that are covered by it.

I disagree, depending on what we introduce it may become increasingly difficult to deprecate or replace functionality. Let's figure out the plan and then move forward, in the same direction, together.

docs/customresourcestate-metrics.md
```diff
@@ -480,4 +480,7 @@ Examples:
...
+[spec, order, "*", value] # spec.order[*].value = true
```
dgrisonnet, line 485: Do you have an example of use-case for that?

As far as I can tell, the best practice would be to have one metric per value in the list no?

rexagod: With `*`, the list is made available for further filtering, so for the YAML snippet below, 

```yaml
status:
  namespaces:
    - namespace: "foo"
      status:
        available:
          resourceA: '10'
          resourceB: '20'
        pending:
          resourceA: '0'
          resourceB: '6'
```
and the configuration below,
```yaml
path: [status, namespaces, "*", status]
labelsFromPath:
  "available_*": [available]
```
would yield the metric,
```yaml
...{...,available_resourceA="10",available_resourceB="20"} ...
```

rexagod: With a non-wildcard string, and the configuration below,
```yaml
path: [status, namespaces, "*", status]
labelsFromPath:
  "label": [available]
```
would yield the metric,
```yaml
...{...,label0="10",label1="20"} ...
```

pkg/customresourcestate/registry_factory.go
```diff
@@ -671,22 +707,95 @@ func famGen(f compiledFamily) generator.FamilyGenerator {
 	}
 }
 
+func resolveWildcard(path valuePath, object map[string]interface{}) []valuePath {
+	if path == nil {
+		return nil
+	}
+	fn := func(i *int) bool {
```
mrueg, line 714: Can we move this function outside and give it a better name like "findWildcard" or similar?

rexagod:
(pulled in https://github.com/kubernetes/kube-state-metrics/pull/2052 to combine all patches for the dynamic `valueFrom`)

docs/customresourcestate-metrics.md
```diff
@@ -552,8 +552,11 @@ path: [status, namespaces, "*", status]
 labelsFromPath:
   # this can be combined with the wildcard prefixes feature introduced in #2052
   "available_*": [available]
+  "lorem_*": [pending]
+# this can be combined with dynamic valueFrom expressions introduced in #2068
+valueFrom: [pending_resourceB]
 # outputs:
-...{...,available_resourceA="10",available_resourceB="20",...} ...
+...{...,available_resourceA="10",available_resourceB="20",lorem_resourceA="0",lorem_resourceB="6"...} 6
```
rexagod, line 559: `valueFrom` now allows users to utilise the generated wildcard labels as metric values.

docs/customresourcestate-metrics.md
rexagod: Done below. 👍🏼

rexagod:
Up for reviews.

rexagod:
/hold
Let me improve on this a bit.

rexagod:
/hold cancel
Up for reviews.

rexagod:
FYI This is up for reviews (or merging, if this looks good).

docs/customresourcestate-metrics.md
```diff
@@ -525,6 +531,32 @@ Examples:
...
+        resourceB: '20'
+      pending:
+        resourceA: '0'
+        resourceB: '6'
+...
+# resolves to: [status, namespaces, <range len(namespaces)>, status] (a multi-dimensional array)
+path: [status, namespaces, "*", status]
+labelsFromPath:
+  # this can be combined with the wildcard prefixes feature introduced in #2052
+  "available_*": [available]
```
dgrisonnet, line 554: Some afterthought, wouldn't it have been better to just do it in a sprintf fashion? We would then be able to convert to https://github.com/google/cel-go/issues/638 once we switch to CEL

dgrisonnet (commented):
Some new thoughts, I will try to have an actual look at the code later this week

eugenepaniot:
We're interested in this feature and hope to see it in the next release soon!

pkg/customresourcestate/registry_factory.go
```diff
@@ -237,9 +287,41 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+					if ok {
+						var resolvedKeyArr []string
+						for _, dynamicPath := range dynamicPaths {
+							resolvedKeyArr = append(resolvedKeyArr, dynamicPath.part)
+						}
+						// resolvedKey will map to unresolved key "foo_*"'s corresponding valid path string, i.e., "spec_fooObj_*".
+						resolvedKey := strings.Join(resolvedKeyArr, "_")
+						extractedValueFrom = resolvedKey + extractedValueFrom[lastUnderscoreIndex:]
+					}
+				}
```
dgrisonnet, line 310: Why do we do that?

As far as I can tell this is made to propagate the wildcard to the value no? Like:
```
"foo_*": ["spec", "fooObj"]
```
yields the following path:
```
spec_fooObj_*
```
no?

pkg/customresourcestate/registry_factory.go
```diff
@@ -554,11 +636,35 @@ type pathOp struct {
...
+				return nil
+			}
+			if idx < 0 || idx >= len(tobj) {
+				return nil
+			}
+			return tobj[idx]
+		default:
+			return nil
+		}
+	}
```
dgrisonnet, line 658: could you make that its own function so that the code is easier to read?

pkg/customresourcestate/registry_factory.go
```diff
@@ -674,14 +780,85 @@ func famGen(f compiledFamily) generator.FamilyGenerator {
 	}
 }
 
+func findWildcard(path valuePath, i *int) bool {
```
dgrisonnet, line 783: I am not a fan of modifying the index and returning whether we find the wildcard or not. A more idiomatic approach would be similar to the search functions in the strings package where if found you return the index, otherwise `-1`

pkg/customresourcestate/registry_factory.go
```diff
@@ -674,14 +780,85 @@ func famGen(f compiledFamily) generator.FamilyGenerator {
...
+		l = len(list) // number of elements in the list
+		pathCopyStart := make(valuePath, i)
+		copy(pathCopyStart, path[:i])
+		pathCopyWildcard := make(valuePath, len(path)-i-1)
+		copy(pathCopyWildcard, path[i+1:])
+		for k := 0; k < l; k++ {
+			pathCopyStart = append(pathCopyStart, pathOp{part: strconv.Itoa(k)})
+			pathCopyStart = append(pathCopyStart, pathCopyWildcard...)
+			expandedPaths = append(expandedPaths, pathCopyStart)
+		}
```
dgrisonnet, line 827: since we know the size of the path from the get go, it would be less expensive in terms of allocations to just allocate one slice and then fill it correctly

PR #2076, opened 2023-05-22
chore: Bump dependencies for next release

mrueg:
Bumps dependencies to prepare for next release.

@logicalhan @rexagod 

Fixes #


go.mod
```diff
@@ -10,20 +10,20 @@ require (
...
+	github.com/stretchr/testify v1.8.3
 	gopkg.in/yaml.v3 v3.0.1
-	k8s.io/api v0.26.4
-	k8s.io/apimachinery v0.26.4
-	k8s.io/client-go v0.26.4
-	k8s.io/component-base v0.26.4
+	k8s.io/api v0.26.5
+	k8s.io/apimachinery v0.26.5
+	k8s.io/client-go v0.26.5
+	k8s.io/component-base v0.26.5
```
rexagod, line 23: Just to be clear, we are going to bump these (and other `k8s.io/*` deps) to `0.27.x` in the next major release (`v3`)? Also, is the contrary also true, i.e., to upgrade to the next minor `k8s.io/*` version, we need to create a major release?

mrueg: We will bump them in v2.10.0 to 0.27.x, for k8s deps we should raise minor early in the development cycle.

rexagod: Ah, so we're pointing to the latest patch `k8s` versions in this release, and only then are we going to point them to the next minor version (in the next-to-next release)?

mrueg: Yes, will bump it after v2.9 is released

rexagod: ACK, can you PTAL at [this](https://github.com/kubernetes/kube-state-metrics/pull/2076#pullrequestreview-1436982645) as well?

mrueg:
> GitHub won't let me review lines outside of the changeset, so mentioning here that we might want to update [`kube-openapi`](https://github.com/kubernetes/kube-state-metrics/pull/2076/files#diff-33ef32bf6c23acb95f5902d7097b7a1d5128ca061167ec0716715b0b9eeaa5f6R84) to `v0.0.0-20230501164219-8b0f38b5fd1f`.

This is an indirect dependency, so I prefer not to update it.

PR #2096, opened 2023-06-13
feat: Adds new metric `kube_pod_service_account`

swarup-stripe:

It adds a new pod metric `kube_pod_service_account` to expose the service account (`p.Spec.ServiceAccountName`) for a particular pod. This is needed to see which user has created a pod and view metrics for pods created by a particular user.

Increases

### DIFF
--- a/pkg/app/server_test.go
+++ b/pkg/app/server_test.go
@@ -235,6 +235,7 @@ func TestFullScrapeCycle(t *testing.T) {
 # HELP kube_pod_overhead_cpu_cores The pod overhead in regards to cpu cores associated with running a pod.
 # HELP kube_pod_overhead_memory_bytes The pod overhead in regards to memory associated with running a pod.
 # HELP kube_pod_runtimeclass_name_info The runtimeclass associated with the pod.
+# HELP kube_pod_service_account The service account for a pod.
 # HELP kube_pod_owner [STABLE] Information about the Pod's owner.
 # HELP kube_pod_restart_policy [STABLE] Describes the restart policy in use by this pod.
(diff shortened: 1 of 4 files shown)


pkg/app/server_test.go
```diff
@@ -334,6 +336,7 @@ kube_pod_info{namespace="default",pod="pod0",uid="abc-0",host_ip="1.1.1.1",pod_i
 kube_pod_labels{namespace="default",pod="pod0",uid="abc-0"} 1
 kube_pod_owner{namespace="default",pod="pod0",uid="abc-0",owner_kind="",owner_name="",owner_is_controller=""} 1
 kube_pod_restart_policy{namespace="default",pod="pod0",uid="abc-0",type="Always"} 1
+kube_pod_service_account{namespace="default",pod="pod0",uid="abc-0",service_account=""} 1
```
CatherineF-dev, line 339: QQ: can this metric kube_pod_service_account be used in alerts? Or only informational purpose.

swarup-stripe: @CatherineF-dev my use case is to have this for for informational purposes so our users can query metrics by their service account. All pods have the service account in the podspec according to [the docs](https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#use-the-default-service-account-to-access-the-api-server) so I think it could be used for alerts. However if it's a question about cardinality, I'm not too sure.

swarup-stripe:
@CatherineF-dev wanted to follow-up and see if we can get this PR merged

swarup-stripe:
@fpetkovski Bumping this - it's a pretty small change and would be really useful for our use-case! Happy to clarify if necessary

mrueg:
/lgtm

Thanks for your contribution!

PR #2097, opened 2023-06-14
move pod owner-ref to all pod metrics

### LINKED ISSUE #1993: Gather `ownerReferences` for all object types
As far as I see a metrics like `kube_pod_owner` is created for very few object types: pods, replicasets and jobs. There are more object types being owned by some other objects or controllers. We are also using a custom CRDs which  own deployments and HPAs.

**Why is this needed**:
Having owner references of all kubernetes objects would allow connecting objects together in grafana dashboards. As for now this is limited to selected objects only.

I would like to have owner references for all object types, including custom CRDs. This could be opt-in feature - some command line flag, maybe something similar to `--resources` for default object types and some field in custom resource configuration for CRDs.
This way it would not break backward compatibility nor introduce new metrics for those space concerned.

gabeduke:

This is a Draft PR with a suggested approach to solving #1993 

This PR adds the owner meta to all of the pod metrics. This is useful in situations where a user would like to roll up the pods by their owner (for instance to associate container restarts with a deployment in a multi-tenant namespace). Alternatively to adding the refs to _all_ pod metrics, it could be added to just those which could benefit from the rollup. I would like feedback from the project maintainers as to if the additional labels would cause issues regarding cardinality or otherwise.

The functionality was already written in the `pod_info` metric, I just moved it into the `wrapPodFunc` so it would be exposed to all of the pod metrics.

If this approach is accepted I will fix the tests and clean things up a bit.

I don't believe this would add much cardnality but would like feedback in this area.

Partially Fixes #1993

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -505,20 +505,6 @@ func createPodContainerStatusWaitingFamilyGenerator() generator.FamilyGenerator 
 		"",
 		wrapPodFunc(func(p *v1.Pod) *metric.Family {
 			ms := make([]*metric.Metric, len(p.Status.ContainerStatuses))
-			//createdBy := metav1.GetControllerOf(p)
-			//createdByKind := ""
-			//createdByName := ""
-			//if createdBy != nil {
-			//	if createdBy.Kind != "" {
-			//		createdByKind = createdBy.Kind
-			//	}
-			//	if createdBy.Name != "" {
-			//		createdByName = createdBy.Name
-			//	}
-			//}
-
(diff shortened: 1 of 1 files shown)


CatherineF-dev:
Could you create a new metric with label uid to join with pod metrics?

Because 
1. some important pod metrics are STABLE and their cardinality is high. Though it won't break alerts, but it would increase cpu/memory usage.
2. in some monitoring systems (stackdriver), it has limit on number of metric labels.

PR #2114, opened 2023-07-04
feat: Add webhooks client config service metrics

dgrisonnet:

In order to better identify, classify, and debug webhook latency issues, it is important to have a metric that would point to the resource it is responsible for. However, it is not possible to have that dimension in the metrics exposed by Kubernetes because of the unbound cardinality that such a label would have.

The name of the webhook could be an alternative since it usually contains some information about the resource that the webhook targets, however this is not very practical to use in multi-tenants environments.

A solution for these kind of platform is to tie a specific webhook to a namespace in order to be able to know which tenant manages it and take actions depending on that. This is achievable by leveraging the client config information of webhooks configured via WebhookConfiguration resources since Services are namespaced objects.

With these new metrics, users will be able to split the alerting severity of webhook latency / rejection rate per namespace on top of being able to do it based on the webhook name. This is key in environment where administrators don't have control over the webhooks installed by the various tenants. 

O(webhooks)


mrueg:
/hold 
for others to review, feel free to cancel if nobody else gets around to it
/lgtm

mrueg:
/hold cancel

2 weeks have passed, let's get this change in.

Issue #536, opened 2018-09-11
Pod labels in metrics


abinet:
Currently `namespace`, `pod` are default labels provided in the metrics.
I would like to print some additional pod labels in the metrics's output to have an ability to group on later in prometheus / grafana. Is it possible?

brancz:
You just need to join metrics, for example to get the `node` label from `kube_pod_info` onto the resource requests metric of a container i n a pod, you can just do:

```
kube_pod_container_resource_requests_memory_bytes * on(namespace, pod) group_left(node) kube_pod_info
```

@andyxning @zouyee @mxinden as this is a request we have on a weekly basis, what do you think of adding a new page in the documentation to add examples like the above? (only for metrics within kube-state-metrics though, mixing metrics from different sources would open a whole other can of worms in terms of support questions by users)

andyxning:
@brancz Yes, IIRC, we have discussed this in an issue. But that is talked about mix metrics from cAdvisor and kube-state-metrics. I totally agree the opinion about documenting join usage like the above.

mxinden:
@brancz sounds good to me. As we had good experience with community contributed documentation in  https://github.com/coreos/prometheus-operator, how about adding a `help-wanted` label?

Maybe @abinet you want to give it a shot? In case this fixed your use case.

Otherwise I can also step in.

andyxning:
> sounds good to me. As we had good experience with community contributed documentation in https://github.com/coreos/prometheus-operator, how about adding a help-wanted label?

Is the different promethues application usage documentation under prometheus-operator repo the authoritative one? If then, it's ok to add usage doc there and add a link in kube-state-metrics to point to that doc. If the answer is no, then we need to make a new directory to add this under kube-state-metrics reop.

brancz:
I think Max was just suggesting to label this issue with help wanted and let the community get started on a documentation page like that :)

andyxning:
Got it.

abinet:
thanks guys to pointing on that. sorry for silly question.

brancz:
Not a silly question at all as I said it's a common request, would you like to start the above mentioned docs page @abinet ? :slightly_smiling_face:

abinet:
@brancz yes, let me try

ravvereddy:
@brancz
I would like to configure alerts only for specific namespaces and  pods with labels for example app=python-helloworld, app=youtube, datastream=mobile-phone etc..

[...]

Issue #1980, opened 2023-02-07
Add an option to allow adding Kubernetes Labels as Prometheus Metric Labels per metric-series


kostz:
**What would you like to be added**:
Add an option to tag metrics with kubernetes labels

**Why is this needed**:
The case I'm working on involves otel-collector scraping prometheus metrics from KSM and exporting them into the commercial backend. The thing is that commercial backend charges for metric cardinality, so reporting metrics with such labels as pod_name and then do joins would cost a lot.

One of the possibility to avoid this excessive costs is to shrink cardinality by aggregating before exporting metrics (well, with acceptable loosing of visibility of keeping data for each pod) and it looks very promising to have this labelling on KSM and then remove extra labels (like pod_name, uid, etc) on otel-collector's processor.

**Describe the solution you'd like**
Have a feature toggle similar to `metric-allowlist` which will not add labels to `kube-pod-labels` metrics, but add labels for related metric timeseries, like `kube_pod_container_resource_requests`

**Additional context**
Follow up for https://github.com/kubernetes/kube-state-metrics/issues/1758

dgrisonnet:
My concerns from https://github.com/kubernetes/kube-state-metrics/issues/1758#issuecomment-1192710597 still applies.

You could today remove labels via the otel-collector relabeling config but more complex aggregation similar to what Prometheus has with recording rules is not possible: https://github.com/open-telemetry/opentelemetry-collector-contrib/blob/main/receiver/prometheusreceiver/README.md#unsupported-features. Maybe there is a way to do post-processing in Otel, but that is out of kube-state-metrics' scope.

dgrisonnet:
This is what you are most likely looking for: https://github.com/open-telemetry/opentelemetry-collector-contrib/issues/4968

kostz:
@dgrisonnet , sorry, my bad of using the 'aggregation' term, that may lead to confusion about caching something and doing a real aggregation in KSM, that's not I'm asking for, I'm asking just about having an option to tag time series points with tags taking them from pod specs.

Honestly I didn't get a point about enormous memory footprint you're projecting for this feature, KSM will still deliver metrics almost 1:1, but adding additional (limited) tags taking them from pod specs

Also this feature may affect not the whole set of metrics, but just some of them (well, we're using less then 10% IIRC)

kostz:
[...]

Issue #1758, opened 2022-06-10
Add an option to allow adding Kubernetes Labels as Prometheus Metric Labels per metric-series 


sherifabdlnaby:
I created this Issue for discussing and tracking PR #1689 
Below is a transcript of the PR 👇🏻
------

# What this PR does / why we need it:

This PR Introduce a new flag `--per-metric-labels-allowlist` that its syntax works similar to `--metric-labels-allowlist` but instead of being a filter to add labels to `kube_X_labels`, it is a filter for K8S' labels that will be added to **each metric** time series of a resource.

## Motivation

Motivation for this change can be better described from a Platform Team POV, who is responsible for the observability stack and provides it to multiple teams and/or multiple tenants.

**The goal is to make it easier to define queries, alerts, and rules without the need for complex joins. Making the barrier for smaller less experienced teams smaller.** As well as alleviate pressure from Prometheus Server that's constantly doing joins for each alert rule.

#### Use Case 1
1. A Development Team wants to create alerts for the multiple components of their applications.
2. **Different Components have different alerts, severities, thresholds**. (example, web pods, background consumers, different kinds of jobs) and as components live in the same namespace; filtering with a namespace is not feasible.
3. You'll now have to use joins with `kube_X_labels` to filter for the specific resources. Complex queries become more complex, especially ones that had joins already.

#### Use Case 2
1. Platform Team defining **general, default rules** for every namespace.
2. Platform team expects resources to have a set of standard labels, that defines something like `alerting.xxx/severity`, `alerting.xxx/slack-channel` along side the `app.kubernetes.io/name`, `app.kubernetes.io/component` ones.
3. Complex Rules are defined to join with these labels and generate alerts with variables based on labels defined by teams. Queries become even more complex as you want to join with **more than one label**.

### Complex Queries Example

`Deployment has not matched the expected number of replicas.`

- Using only 1 label for joins

   ```prometheus
  (
    kube_deployment_spec_replicas{} * on (deployment) group_left(label_app_kubernetes_io_name) kube_deployment_labels{ label_app_kubernetes_io_name="emojiapp", namespace=~"emoji"}
    !=
    kube_deployment_status_replicas_available{} * on (deployment) group_left(label_app_kubernetes_io_name) kube_deployment_labels{ label_app_kubernetes_io_name="emojiapp", namespace=~"emoji"}
  ) 
  and 
  (
[...]

PR #2129, opened 2023-07-26
feat: Append existing resource labels to all related metrics

alen-z:
⚠️ Disclaimer: Please guide me with your comments to make this PR better. I haven't touched any code for years and it's my first serious interaction with Go. Thank you for your understanding in advance.

KSM is great, but to make it even better (fit our use case) here's the proposal on how to expand resource related metrics with custom labels. Labels are expected to already be defined as part of the resource. 

Now supported resources:
- Pod
- Service
- ...

Issue is that Prometheus `relabelings` can't be used for KSM because Prometheus scrapes only KSM container which provides metrics for many resources scattered around the cluster.

Ultimately, this should allow us to get custom labels in Alertmanager without any default alert rule changes in Prometheus. Having custom labels can significantly improve alert routing based on more detailed labels.

Example:
Set `--metric-labels-append=regex=meta.example.com/(.*),product` and get:
```
kube_pod_container_info{container="container1",container_id="docker://ab123",image="k8s.gcr.io/hyperkube1",image_id="docker://sha256:aaa",image_spec="k8s.gcr.io/hyperkube1_spec",meta_example_com_owner="team1",namespace="ns1",pod="pod1",product="frontend1",uid="uid1"} 1
```

**Proposed phases**

- [ ] Get feedback for Pod metrics and labels. Polish to the degree so it can be expanded to other resources.
- [ ] Expand this method to all relevant resources.
- (optional) Include `replacement` mechanism when `regex` is used. Similar to Prometheus relabelings. This should allow us to rename matched labels.

Increases. Adding labels to exising metrics.

Fixes https://github.com/prometheus-operator/kube-prometheus/issues/887
Fixes https://github.com/kubernetes/kube-state-metrics/issues/536


alen-z:
Hi @dashpole, thank you for taking this one further.

Hey @dgrisonnet, nice to meet you. Currently there's support for Pod and Service resources. I'll be extending the code to all resources, but until then please let me know if there's anything else that needs to be done or done differently to comply with your expectations. New commits soon...

alen-z:
Planning to allocate some time to finish this PR soon...

alen-z:
> Are there any gaps in the existing mechanism to expose labels value that you would want to be covered here?

Hi @dgrisonnet, yes. With current state, as far as I understand, we can not get arbitrary/custom Prometheus metric labels into _all_ metrics that kube-state-metrics provides — we want to include labels in metrics from e.g. existing Pod or Service labels.

Having custom labels in only one kube-state-metrics metric (`--metric-labels-allowlist `) does not work for us. Here's why...

Specific use case is: We have certain set of metadata related to our product (kind of service directory metadata) that we deploy along with our Kubernetes resources in form of resource manifest labels. Those manifest labels contain information about teams that own the product, communication channel etc. — now, we want to route native Prometheus alerts based on this custom owner label to proper teams! This is only possible if, among other metrics, kube-state-metrics has owner label in its metrics.

This is only one use case, there might be others related to visualization of Prometheus data based on filters or for pulling out some statistics grouped by custom label...

Please let me know if there is different approach to our use case that I might have missed.

dgrisonnet:
>  With current state, as far as I understand, we can not get arbitrary/custom Prometheus metric labels into all metrics that kube-state-metrics provides 

That's totally intentional because having these labels in all the metrics would increase the cardinality of the metrics tremendously and would cause kube-state-metrics to generate even more timeseries than it is currently. Leading to higher memory consumption from Prometheus and potentially higher cost as well for users using a SAS monitoring solution.

Your request is very similar to these two ones that were opened some time ago:
- https://github.com/kubernetes/kube-state-metrics/issues/1758
- https://github.com/kubernetes/kube-state-metrics/issues/1980

And both your use case and their requests can be solved today by using joins in the promQL queries of your alerts on `kube_*_labels` to surface the labels that are necessary. And I am not convinced that the convenience of having the labels in all the metrics is worth the cost of exposing these info everywhere when we already have them in one place.

dgrisonnet:
IMO your needs are outside of ksm scope and need a middleware to be achieved such as the Prometheus agent.

dgrisonnet:
Still, having something like that is out of scope for kube-state-metrics. The only purpose of kube-state-metrics is to expose metrics about Kubernetes objects with a 1:1 mapping. Aggregation and relabeling should be done by other tools.
In this micro services era, it is very important to stick to original scopes.

> but since this is an off by default feature, it would not break any existing users.

This is not a valid reason to add a feature to a project. At the end of the day, we would have to maintain that new functionality and most likely extend it for future needs. No extending the project past its original purpose is intentional to keep it maintainable and avoid making it too complex to use for the users.

alen-z:
Hi @dgrisonnet 👋 

> metrics would increase the cardinality

I would let cluster operators decide what they want to do and provide them with the mechanisms to do it. This goes along what @DanTulovsky is saying with "People who choose to use this will need to understand the ramifications". Nothing would change by default for existing users.

> using joins in the promQL queries

I'd say yes in theory, you are right, but the reality is many alert definitions are used from [here,](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/prometheus/rules-1.14) not being applicable to change (use join).

> we already have them in one place (metrics)

Which is not a very convenient place for us and I'd argue for number of other people and use cases too.

> The only purpose of kube-state-metrics is to expose metrics about Kubernetes objects with a 1:1 mapping. Aggregation and relabeling should be done by other tools.

It'd still expose information about objects with 1:1 mapping, only to describe them better with their own labels that already exist? Aggregation without it is hard, not impossible, but hard in reality.

--

Ultimately, @dgrisonnet and @rexagod, I'm interested should we keep this PR open and continue the work here or should we close and fork it privately to support our use case, without impacting this project? I suspected this might create some controversy to be honest, so not really surprised to meet the skepticism :) Good discussion.

Looking forward to hear from you.

dgrisonnet:
> I guess I don't understand how this is different from --metric-labels-allowlist, which allows * and comes with a huge warning. Why allow the former, but not this one?

In the current case, we add this information only once per resource. That's already very expensive if you expose all the labels of all the resources in your clusters hence the warning, but if you were to do that for all the metrics exported by kube-state-metrics, the additional cost would be tremendous.

dgrisonnet:
> I would let cluster operators decide what they want to do and provide them with the mechanisms to do it. This goes along what @DanTulovsky is saying with "People who choose to use this will need to understand the ramifications". 

Ultimately that's the goal with kube-state-metrics. Empowering cluster operators to tune the project to match their needs without needing any knowledge about the codebase. That's why we've released support for CRD and more recently have been discussing making kube-state-metrics configuration based: https://github.com/kubernetes/kube-state-metrics/issues/2165.

But, on top of the additional resource consumption from storing the labels information multiple times, adding them to all the metrics would go against the current state of kube-state-metrics where we have metrics about specific parts of Kubernetes object. Each one of them have one unique purpose and that's how we currently keep kube-state-metrics simple to use and consume. Because of that, it would be wrong to add unrelated information to the various metrics.

So with the current state of kube-state-metrics, I don't think it is wise to add a mechanism that goes against the current design. Additionally, if we commit to this feature that goes against the best practices that we've put in place, we will never be able to move away from it without breaking users. However, I am not against making that possible once we fully empower users to tweak kube-state-metrics however they want.

> I'd say yes in theory, you are right, but the reality is many alert definitions are used from [here,](https://github.com/prometheus-community/helm-charts/tree/main/charts/kube-prometheus-stack/templates/prometheus/rules-1.14) not being applicable to change (use join).

Charts are hard to extend which is why https://github.com/prometheus-operator/kube-prometheus is not providing them but instead raw manifests that are generating from Jsonnet, which are way easier to extend. I would recommend looking into it if you want to update the rules to match your needs.

> It'd still expose information about objects with 1:1 mapping, only to describe them better with their own labels that already exist?

So then why make an exception for labels and not also add annotations, containers, statuses, ...? This is a far fetched example and I know why you specifically want labels, but that would break a boundary that we've set between the information.

> Aggregation without it is hard, not impossible, but hard in reality.

I totally hear that, but I don't see a way we could improve that today without breaking the current UX of kube-state-metrics.
But this for sure, re-emphasize the idea that we should empower users to extend ksm however they want. I will for sure keep that in mind while designing the future UX.

>  I'm interested should we keep this PR open and continue the work here or should we close and fork it privately to support our use case, without impacting this project?

As you can probably tell, I am not a fan of adding this feature to ksm as it would be considered out of scope right now and I'd prefer leaving any kind of aggregation to a third party.
So I would personally reject the feature for now and maybe revive it in the future if we go forward with https://github.com/kubernetes/kube-state-metrics/issues/2165.

alen-z:
I'd prefer much stronger arguments, but closing the PR.

dgrisonnet:
I mean the arguments are pretty strong to me:
1. increase memory consumption for duplicated data
2. manipulation on the metrics is outside kube-state-metrics' scope and should be done at the aggregation layer
3. this feature goes against the best practices
4. it changes the expected experience behind kube-state-metrics' metrics
5. will complexify and already very complex configuration around labels

Although I said that this feature might be reconsidered in the future if we empower our users even more, I would still discourage anyone from doing that since duplicating data is a waste of resource.

PR #2133, opened 2023-08-01
feat: Add CSI volume attributes to PV info metric

machadovilaca:

This PR adds a new volume attributes field for CSI PV in `kube_persistentvolume_info` metric

- adds a new label for `kube_persistentvolume_info`
- label value should not change often

Fixes #


mrueg:
Can you update https://github.com/kubernetes/kube-state-metrics/blob/main/docs/persistentvolume-metrics.md to include the new attribute as well?

internal/store/persistentvolume.go
```diff
@@ -219,6 +221,13 @@ func persistentVolumeMetricFamilies(allowAnnotationsList, allowLabelsList []stri
 				case p.Spec.PersistentVolumeSource.CSI != nil:
 					csiDriver = p.Spec.PersistentVolumeSource.CSI.Driver
 					csiVolumeHandle = p.Spec.PersistentVolumeSource.CSI.VolumeHandle
+
+					for k, v := range p.Spec.PersistentVolumeSource.CSI.VolumeAttributes {
+						if len(attributes) != 0 {
+							attributes += ","
+						}
+						attributes += fmt.Sprintf("%s=%s", k, v)
+					}
```
dgrisonnet, line 230: This seems very much unbounded and would require a similar protection mechanism as the one we have for labels and annotations to avoid cardinality explosions.

dgrisonnet (changes requested):

internal/store/persistentvolume.go
machadovilaca: I tried to reduce the possible cardinality of labels. Please check if this would work better. Labels are already unique per PV since one of the labels is the name of the volume and I expect both the mounter and the map options to rarely be changed

internal/store/persistentvolume.go
```diff
@@ -219,6 +220,14 @@ func persistentVolumeMetricFamilies(allowAnnotationsList, allowLabelsList []stri
...
 					csiDriver = p.Spec.PersistentVolumeSource.CSI.Driver
 					csiVolumeHandle = p.Spec.PersistentVolumeSource.CSI.VolumeHandle
+
+					for k, v := range p.Spec.PersistentVolumeSource.CSI.VolumeAttributes {
+						if k == "mapOptions" {
+							csiMapOptions = v
+						} else if k == "mounter" {
+							csiMounter = v
+						}
+					}
```
dgrisonnet, line 230: These attributes are CSI drivers specific and are not defined in any CSI spec right?

machadovilaca: Yes, they are... Is there any way we can make this more generic while avoiding big cardinality increases?

machadovilaca: wdyt about creating a new metric, not enabled by default, that would contain the unbound list of volume attributes?
and then anyone could make it active through `MetricOptInList`

machadovilaca:
@dgrisonnet created a new opt-in metric, maybe this is enough to avoid very specific information

also, I rebased with main but the bot did not remove the label, am I missing something?

internal/store/persistentvolume.go
```diff
@@ -349,6 +352,45 @@ func persistentVolumeMetricFamilies(allowAnnotationsList, allowLabelsList []stri
...
+				}
+
+				var csiMounter, csiMapOptions string
+				for k, v := range p.Spec.PersistentVolumeSource.CSI.VolumeAttributes {
+					if k == "mapOptions" {
+						csiMapOptions = v
+					} else if k == "mounter" {
+						csiMounter = v
+					}
```
dgrisonnet, line 374: could you document where these attributes are coming from for future reference?

docs/persistentvolume-metrics.md
```diff
@@ -10,6 +10,7 @@
...
+| kube_persistentvolume_csi_attributes     | Gauge       | CSI attributes of the Persistent Volume.                                                                                  |                         | `persistentvolume`=&lt;persistentvolume-name&gt; <br> `csi_mounter`=&lt;csi-mounter&gt; <br> `csi_map_options`=&lt;csi-map-options&gt;                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                               | EXPERIMENTAL |
```
dgrisonnet, line 13: Could you mention in the description that it is not enabled by default?

internal/store/persistentvolume.go
machadovilaca: added: "// storage attributes handled by external CSI drivers"

docs/persistentvolume-metrics.md
machadovilaca: added, updated to: "CSI attributes of the Persistent Volume, disabled by default, manage with [--metric-opt-in-list](./cli-arguments.md))"

dgrisonnet (approved):
/lgtm
/approve

PR #2134, opened 2023-08-01
feat: add volumemode to PVC info metric

machadovilaca:

This PR adds a label to `kube_persistentvolumeclaim_info` to track the volume mode set in the PVC, either "Block" or "Filesystem"

- adds a label that is never updated for a given PVC

Fixes #


dgrisonnet (commented):
The new metric look reasonable to me. My approval is pending Manuel's comment resolution: https://github.com/kubernetes/kube-state-metrics/pull/2134#issuecomment-1674885538

PR #2145, opened 2023-08-14
feat: disable labels and annotations metrics when metric-annotations-…

### LINKED ISSUE #1829: do not create annotation and label metrics if none are allowed
Currently kube-state-metrics creates object_annotations and object_labels metrics for all objects even if the objects have no configured allow list.

**Why is this needed**:
This creates many metrics which contain no useful information, especially for pods and replicasets. For example:
```
kube_pod_annotations{instance="...:8080", job="monitoring/kube-state-metrics", namespace="namespace", pod="podname-bc94dfc78-fcf4j", uid="a482250c-6f6e-4533-a596-3eff484d8969"}
```

If the annotation or label has no configured allowed values (--metric-annotations-allowlist, --metric-labels-allowlist) no object_annotations or object_labels metrics should be created.

opeco17:

`kube-state-metrics` provides a lot of `kube_*_annotations` and `kube_*_labels` by default, and it consumes memory significantly.

```
kube_pod_annotations{instance="...:8080", job="monitoring/kube-state-metrics", namespace="namespace", pod="podname-bc94dfc78-fcf4j", uid="a482250c-6f6e-4533-a596-3eff484d8969"}
```

By disabling those metrics by default, `kube-state-metrics` can reduce memory usage.

Those metrics are provided when `kube_*_annotations` and `kube_*_labels` are configured.

decrease

Fixes #1829

### DIFF
--- a/internal/store/certificatesigningrequest.go
+++ b/internal/store/certificatesigningrequest.go
@@ -49,6 +49,9 @@ func csrMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 			basemetrics.ALPHA,
 			"",
 			wrapCSRFunc(func(j *certv1.CertificateSigningRequest) *metric.Family {
+				if len(allowAnnotationsList) == 0 {
+					return &metric.Family{}
+				}
 				annotationKeys, annotationValues := createPrometheusLabelKeysValues("annotation", j.Annotations, allowAnnotationsList)
 				return &metric.Family{
 					Metrics: []*metric.Metric{
@@ -68,6 +71,9 @@ func csrMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 			basemetrics.STABLE,
 			"",
 			wrapCSRFunc(func(j *certv1.CertificateSigningRequest) *metric.Family {
+				if len(allowLabelsList) == 0 {
+					return &metric.Family{}
+				}
 				labelKeys, labelValues := createPrometheusLabelKeysValues("label", j.Labels, allowLabelsList)
 				return &metric.Family{
 					Metrics: []*metric.Metric{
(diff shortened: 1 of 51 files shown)


mrueg:
@dgrisonnet @opeco17 Is this a breaking change to the previous behaviour?

dgrisonnet:
It is indeed a breaking change. I was a bit too quick with merging this one.

Although this is a breaking change, I wonder if users really rely on these empty metrics since the information inside of them are already present in other metrics such as the more commonly used `_info`.

That said, I looked at the options' doc and was expecting that we are saying that we are not exposing labels/annotations metrics when the allowlist are empty. But in the end we say:
```
By default the metric contains only name and namespace labels.
```

dgrisonnet:
These metrics being EXPERIMENTAL, I would personally move ahead with this change and update the doc accordingly.

What is your opinion on that @mrueg @rexagod?

mrueg:
This works for me, we should mention it in the changelogs though as a breaking change.

rexagod:
https://github.com/kubernetes/kube-state-metrics/pull/2145#issuecomment-1688375789 makes sense. Even if this is a breaking change, we aren't doing anything that goes against the laid out commitments. +1.

cykl:
@dgrisonnet 

> Although this is a breaking change, I wonder if users really rely on these empty metrics 

I will put an end to the suspense. At least one user really relied on these metrics, me. 😁 

> since the information inside of them are already present in other metrics such as the more commonly used _info.

That doesn't sound right to me. AFAIK The only way to retrieve labels associated with a kubernetes object is the kube_*_labels metrics. In my case, some labels are required to aggregate some metrics or route some alerts.   

It seems a very common need, and I expect that others have been impacted too (for example search for "kubernetes prometheus cronjob" and you will find dozen of  examples relying on them).

While release notes mention it, I'm a bit surprised that kube-state-metrics introduced this breaking changes with such lightness.

RodrigoMenezes-Vantage:
Agreed with @cykl I just spends a few hours today debugging an issue on my observability stack due to this change. 

Would loved to have seen this feature go behind a flag for those that don't want the metrics to be able to turn it off instead of breaking things for those that have been relying on the metrics being there.

PR #2146, opened 2023-08-14
feat: include reason label to kube_deployment_status_condition

### LINKED ISSUE #1991: Include `reason` label to the `kube_deployment_status_condition` metric

I would like the label `reason` to be added to the `kube_deployment_status_condition` metric.

**Why is this needed**:

The `kube_deployment_status_condition` has only two properties relating to the `DeploymentCondition` object: `condition` and `status`, which is not enough to distinguish between a [Progressing Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progressing-deployment) and a [Complete Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment). The `reason` label will add sufficient information to distinguish the two states.

Simply extract and emit the `reason` label for the deployments.

**Additional context**

.

opeco17:

Include `reason` label to `kube_deployment_status_condition` metrics.

It's necessary to distinguish between [Progressing Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progressing-deployment) and [Complete Deployment ](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment).

no change

Fixes #1991

### DIFF
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -174,8 +174,8 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 					for j, m := range conditionMetrics {
 						metric := m
 
-						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						metric.LabelKeys = []string{"reason", "condition", "status"}
+						metric.LabelValues = append([]string{c.Reason, string(c.Type)}, metric.LabelValues...)
 						ms[i*len(conditionStatuses)+j] = metric
 					}
 				}
--- a/docs/deployment-metrics.md
+++ b/docs/deployment-metrics.md
@@ -9,7 +9,7 @@
 | kube_deployment_status_replicas_unavailable | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_status_replicas_updated | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
 | kube_deployment_status_observed_generation | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
-| kube_deployment_status_condition | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `condition`=&lt;deployment-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; | STABLE |
+| kube_deployment_status_condition | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; <br> `reason`=&lt;deployment-transition-reason&gt; <br> `condition`=&lt;deployment-condition&gt; <br> `status`=&lt;true\|false\|unknown&gt; | STABLE |
 | kube_deployment_spec_replicas | Gauge | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt; | STABLE |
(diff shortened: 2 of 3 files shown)


opeco17:
Hi, @dgrisonnet @logicalhan

Could you please take a look at PR whenever you can?
If additional work is necessary, please let me know.

Thank you in advance.

mrueg (approved):

internal/store/deployment.go
```diff
@@ -174,8 +174,8 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 					for j, m := range conditionMetrics {
 						metric := m
 
-						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						metric.LabelKeys = []string{"reason", "condition", "status"}
+						metric.LabelValues = append([]string{c.Reason, string(c.Type)}, metric.LabelValues...)
```
dgrisonnet, line 178: Since reason could theoretically be anything, could you provide a list of allowed reasons and only expose these? Similar to what we have for pods: https://github.com/kubernetes/kube-state-metrics/blob/main/internal/store/pod.go#L40

The list of valid deployment reasons can be found in: https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L67-L101

dgrisonnet:
Hi @opeco17, we are planning to start cutting a new release of kube-state-metrics Wednesday. Do you think you'll be able to address the comments by then?

internal/store/deployment.go
opeco17: Do you mean the output format will be like below?

As is
```
kube_deployment_status_condition{namespace="default",deployment="nginx",condition="Progressing",status="true"} 1
kube_deployment_status_condition{namespace="default",deployment="nginx",condition="Progressing",status="false"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",condition="Progressing",status="unknown"} 0
```

To be
```
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="true"} 1
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="false"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="unknown"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="MinimumReplicasUnavailable",condition="Progressing",status="true"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="MinimumReplicasUnavailable",condition="Progressing",status="false"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="MinimumReplicasUnavailable",condition="Progressing",status="unknown"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="FoundNewReplicaSet",condition="Progressing",status="true"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="FoundNewReplicaSet",condition="Progressing",status="false"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="FoundNewReplicaSet",condition="Progressing",status="unknown"} 0
...
```

opeco17:
Let me try to finish it by Wednesday as much as possible but it might be difficult...

dgrisonnet:
No pressure, we can always have it in the next release

internal/store/deployment.go
CatherineF-dev: qq: will one deployment in several conditions? Can we only keep 3 streams?

For example, only keep
```
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="true"} 1
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="false"} 0
kube_deployment_status_condition{namespace="default",deployment="nginx",reason="NewReplicaSetAvailable",condition="Progressing",status="unknown"} 0
```

CatherineF-dev: Checked cardinality keeps same. My comment has been resolved.

PR #2150, opened 2023-08-17
feat(customresourcestate): allow string Array for labelFromPath

adberger:

Some CRs expose information as Arrays of type String and not Array of type Maps (e.g. namespaces in https://capsule.clastix.io/docs/general/crds-apis/#tenantstatus).
AFAIK it's currently not possible to add these as labels to a metric, this PR should fix this.

If I understand this correctly, the cardinality should not increase.

https://github.com/kubernetes/kube-state-metrics/discussions/2149

### DIFF
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -192,7 +192,7 @@ func newCompiledMetric(m Metric) (compiledMetric, error) {
 		}
 		valueFromPath, err := compilePath(m.StateSet.ValueFrom)
 		if err != nil {
-			return nil, fmt.Errorf("each.gauge.valueFrom: %w", err)
+			return nil, fmt.Errorf("each.stateSet.valueFrom: %w", err)
 		}
 		return &compiledStateSet{
 			compiledCommon: *cc,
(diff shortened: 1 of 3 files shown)


pkg/customresourcestate/registry_factory.go
```diff
@@ -192,7 +192,7 @@ func newCompiledMetric(m Metric) (compiledMetric, error) {
 		}
 		valueFromPath, err := compilePath(m.StateSet.ValueFrom)
 		if err != nil {
-			return nil, fmt.Errorf("each.gauge.valueFrom: %w", err)
+			return nil, fmt.Errorf("each.stateSet.valueFrom: %w", err)
```
mrueg, line 195: Good catch!

mrueg:
@adberger do you want to create a PR for https://github.com/kubernetes/kube-state-metrics/pull/2150#discussion_r1299726191 I think it'd be good to fix the misleading error message.

adberger:
> @adberger do you want to create a PR for [#2150 (comment)](https://github.com/kubernetes/kube-state-metrics/pull/2150#discussion_r1299726191) I think it'd be good to fix the misleading error message.

@mrueg done.

Issue #2141, opened 2023-08-10
CustomResourceStateMetrics didn't report the custom resource status data to metrics and kube-state-metrics crash if the custom resource property change


chihshenghuang:
**What happened**:
There are 2 issues:
1. kube-state-metrics didn't report the status data for custom resource to metrics when deploy the custom resource at first, it will report the metrics only after kube-state-metrics pod restart. The reason is some status data will be reported later, and kube-state-metrics throw error "got nil while resolving path" at the beginning when there is no status data, and never check the status data again unless restart the pod. And it will cause other metrics can't be reported.
2. kube-state-metrics will crash if remove the property in the custom resource which is used in the valueFrom path, unless use nilIsTrue. The reason is [this function](https://github.com/kubernetes/kube-state-metrics/blob/a9bdda09b92ef7508aa46069b7e2043850107ccf/pkg/customresourcestate/registry_factory.go#L451) might return value as nil, then trying to access a property from nil value at [here](https://github.com/kubernetes/kube-state-metrics/blob/a9bdda09b92ef7508aa46069b7e2043850107ccf/pkg/customresourcestate/registry_factory.go#L281C41-L281C53) 

**What you expected to happen**:
1. kube-state-metrics should detect the custom resource status data change, and report the status data to metrics.
2. kube-state-metrics don't panic and crash, which can be fixed by the PR I sent #2140

**How to reproduce it (as minimally and precisely as possible)**:
1. For kube-state-metrics can't report status data issue
- Use the following config file:
```yaml
apiVersion: v1
data:
  config.yaml: |-
    kind: CustomResourceStateMetrics
    spec:
      resources:
        - groupVersionKind:
            group: autoscaling.k8s.io
            kind: "VerticalPodAutoscaler"
            version: "v1"
          labelsFromPath:
            verticalpodautoscaler: [metadata, name]
            namespace: [metadata, namespace]
            target_api_version: [spec, targetRef, apiVersion]
            target_kind: [spec, targetRef, kind]
            target_name: [spec, targetRef, name]
          metrics:
            - name: "verticalpodautoscaler_status_recommendation_containerrecommendations_target_memory"
              help: "Kubernetes labels converted to Prometheus labels."
              each:
                type: Gauge
                gauge:
                  path: [status, recommendation, containerRecommendations]
                  valueFrom: [target, memory]
                  labelsFromPath:
                    container: [containerName]
                commonLabels:
                  resource: "memory"
                  unit: "byte"
[...]

PR #2154, opened 2023-08-18
fix: custommetrics: always extract the headers but only write it when we have metrics

### LINKED ISSUE #2142: CustomResource: no metrics if no resource exists for CRD when ksm starts
**What happened**:

There are two issues when no resources exist for a CRD at the time when kube-state-metrics starts:

1. If a resource for a custom resource gets created **after** kube-state-metrics starts:
	- no `HELP` or `TYPE` will get exported
2. In a special case, there will also no metrics get exported for it.

* Not requiring a restart to expose the metrics

* Applied a CRD
* Started KSM with configuration for the CRD
* Sleep for 15s
* Created a CR of the above CRD type
* The metrics don't get exposed
* Restart KSM
* Metrics will get exposed

Used Configuration:

```yaml
# This file was auto-generated via: make generate-metrics-config
spec:
  resources:
  - groupVersionKind:
      group: cluster.x-k8s.io
      kind: Cluster
      version: v1beta1
    labelsFromPath:
      name:
      - metadata
      - name
      namespace:
      - metadata
      - namespace
      uid:
      - metadata
      - uid
    metricNamePrefix: capi_cluster
    metrics:
    - name: created
      help: Unix creation timestamp.
      each:
        gauge:
          path:
          - metadata
          - creationTimestamp
        type: Gauge
    - name: status_condition
      help: The condition of a cluster.
      each:
        stateSet:
          labelName: status
          labelsFromPath:
            type:
            - type
          list:
          - 'True'
          - 'False'
          - Unknown
          path:
          - status
          - conditions
          valueFrom:
          - status
        type: StateSet
```

Note: in this case: after creating the resource:

* The metric `capi_cluster_created` will get exposed, but no `HELP` or `TYPE`.
* The metric for `capi_cluster_status_condition` will not get exposed

I think the issue got introduced in #1851

If the method `hasResources` which is called here:

https://github.com/kubernetes/kube-state-metrics/blob/cc06755239faf600cc3468737a321204a1432fcf/internal/store/builder.go#L548

get's changed to always return true instead of actually tring to list CRs:

https://github.com/kubernetes/kube-state-metrics/blob/cc06755239faf600cc3468737a321204a1432fcf/internal/store/builder.go#L627

it meets the expectation and metrics get exposed as soon as the first CR for the CRD gets created.

**Environment**:
- kube-state-metrics version: v2.9.2
- Kubernetes version (use `kubectl version`): v1.27.3
- Cloud provider or hardware configuration: kind
- Other info:

chrischdi:

This PR removes the `hasResources` check to initialise all custom resource metric headers at startup.
By doing this, there will be no delay in adding custom resource metrics on startup (because of trying to list resources and timeout after 10s and then doing the same for the next bunch of CRs)

Instead the metrics writer now checks if there are are metrics in the first element of the store for the given header and outputs depending on that.

Does not affect the cardinality.

Fixes #2142

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -38,10 +38,6 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
-	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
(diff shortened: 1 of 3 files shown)


chihshenghuang:
i can confirm this PR fixed the first issue i mentioned at #2141 , thanks @chrischdi

mrueg:
/lgtm 

/hold 
for others to review.

rexagod:
/lgtm
This makes so much sense, thank you! 🎉

PR #2180, opened 2023-08-31
docs: Update job-metrics.md

### LINKED ISSUE #2179: docs state that kube_job_created is a gauge, but seems to be unix timestamp
**What happened**:

https://github.com/kubernetes/kube-state-metrics/blob/main/docs/job-metrics.md states that kube_job_created is a gauge, but when querying the metrics in prometheus, it seems to be a unix timestamp. I am not sure if the docs is incorrect or the metrics is incorrect at this point.

The docs to accurately represent the truth. 

check the kube_job_created metric

**Environment**:
- kube-state-metrics version: v2.8.2
- Kubernetes version (use `kubectl version`): 1.25
- Cloud provider or hardware configuration: GCP
- Other info:

rfsburger:
Added description for kube_job_created

Adding a description for kube_job_created to reduce these kind of questions popping up: https://github.com/kubernetes/kube-state-metrics/issues/2179

does not change cardinality

Fixes #2179


rexagod:
Gauge values can very well be timestamps as well. Explicitly stating so for every gauge-metric with a timestamp value, and maintaining a similar point in documentation for future metrics seems a bit redundant, since between counters, gauges, info, and state-set metrics (types supported in KSM), it only makes sense for gauges to hold timestamps (as these can possibly be non monotonic consecutively, so counters can't hold them).

/close

PR #2193, opened 2023-09-13
docs: updated VPA CustomResourceStateMetrics docs

### LINKED ISSUE #2041: Missing replacement config for VPA collector in CRM

I would like to have a complete replacement example for VPA CRM configuration [here](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/customresourcestate-metrics.md#verticalpodautoscaler) instead of a simple example defining how to create vpa annotation metrics when most useful metrics (basically any of the resource) are complex to define for people starting out with CRM

**Why is this needed**:

Current documentation is not really usable. It more or less feels like users have to do the replacement themselves, are forced to learn how CRM works and dig into the code of KSM when a sufficient documentation could help users get started faster.

Have a full example of a crm config to have the same vpa metrics as before.
Some metrics might change (because the VPA collected does some calculation to change the unit that is currently not supported by CRM) but at least having the same information (labelset and values) would help people upgrade to 2.9.0.

**Additional context**
Coming from this https://github.com/kubernetes/kube-state-metrics/issues/1718#issuecomment-1497213264

sherifkayad:
docs PR to address #2041

N.A. .. Just a docs PR

Fixes #2041

### DIFF
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -325,14 +325,9 @@ kube_customresource_verticalpodautoscaler_status_recommendation_containerrecomme
 
 #### VerticalPodAutoscaler
 
-In v2.9.0 the `vericalpodautoscalers` resource was removed from the list of default resources. In order to generate metrics for `verticalpodautoscalers`, you can use the following Custom Resource State config:
+In v2.9.0 the `verticalpodautoscalers` resource was removed from the list of default resources. In order to generate Memory / CPU recommendation metrics for `verticalpodautoscalers`, you can use the following Custom Resource State config:
 
 ```yaml
-# Using --resource=verticalpodautoscalers, we get the following output:
-# HELP kube_verticalpodautoscaler_annotations Kubernetes annotations converted to Prometheus labels.
-# TYPE kube_verticalpodautoscaler_annotations gauge
-# kube_verticalpodautoscaler_annotations{namespace="default",verticalpodautoscaler="hamster-vpa",target_api_version="apps/v1",target_kind="Deployment",target_name="hamster"} 1
-# A similar result can be achieved by specifying the following in --custom-resource-state-config:
 kind: CustomResourceStateMetrics
 spec:
   resources:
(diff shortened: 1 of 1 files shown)


docs/customresourcestate-metrics.md
```diff
@@ -253,11 +253,6 @@ kube_customresource_ref_info{customresource_group="myteam.io", customresource_ki
 In v2.9.0 the `vericalpodautoscalers` resource was removed from the list of default resources. In order to generate metrics for `verticalpodautoscalers`, you can use the following Custom Resource State config:
```
drisbee, line 253: Line 253 should be `verticalpodautoscalers` instead of `vericalpodautoscalers`

sherifkayad: resolved

docs/customresourcestate-metrics.md
```diff
@@ -272,19 +267,51 @@ spec:
         target_kind: [spec, targetRef, kind]
         target_name: [spec, targetRef, name]
       metrics:
-        - name: "annotations"
-          help: "Kubernetes annotations converted to Prometheus labels."
+        - name: "vpa_containerrecommendations_target"
```
chihshenghuang, line 270: Suggest to follow the previous VPA metrics naming format "verticalpodautoscaler_status_recommendation_containerrecommendations_target_cpu". Append resource (cpu/memory) at the end of metrics name due to this is changed to be customized metrics name, we can't have duplicate metrics name at line 270 and line 282 which both are "vpa_containerrecommendations_target".

sherifkayad: @chihshenghuang I slightly disagree due to two reasons:
- The name of the metric will be super long as there is already a prefix appended automatically by the custom resource (check the doc further for the full name)
- The previous implemention had already the same metric with different labels / selectors for CPU / Memory and I think that to make it close to the original implementation (prior to v2.9.0), it should be like that.

Remember this is just an example and people can freely choose what to do / how to name the metric. I can suggest adding a line or two in the docs that this is also possible of course.

sherifkayad:
Fixed MD linting issues

/test all

sherifkayad:
@dashpole @dgrisonnet can you please re-trigger the testing pipeline again?

docs/customresourcestate-metrics.md
```diff
@@ -272,19 +267,53 @@ spec:
...
+
+# HELP kube_customresource_vpa_containerrecommendations_target VPA container recommendations for cpu.
+kube_customresource_vpa_containerrecommendations_target{container="hamster",customresource_group="autoscaling.k8s.io",customresource_kind="VerticalPodAutoscaler",customresource_version="v1",namespace="default",resource="cpu",target_api_version="autoscaling.k8s.io/v1",target_kind="Deployment",target_name="hamster",unit="core",verticalpodautoscaler="hamster-vpa"} 0.1
 ```
 
-The above configuration was tested on [this](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/examples/hamster.yaml) VPA configuration, with an added annotation (`foo: 123`).
+To ensure that the VPA Metrics will be captured seamlessly, make sure that the Cluster Role / Role assigned to the Kube State Metrics has permissions to `list` & `watch` VPA resources e.g.:
+
+```yaml
```
k0nstantinv, line 308: seems like the Cluster Role/ Role assigned to KSM requires additional permissions to CRDs

```
- verbs:
     - list
     - watch
   apiGroups:
     - apiextensions.k8s.io
   resources:
     - customresourcedefinitions
```

otherwise `verticalpodautoscalers` still won't be available 

```
E1030 12:35:19.531362       1 reflector.go:148] pkg/mod/k8s.io/client-go@v0.27.5/tools/cache/reflector.go:231: Failed to watch apiextensions.k8s.io/v1, Resource=customresourcedefinitions: failed to list apiextensions.k8s.io/v1, Resource=customresourcedefinitions: customresourcedefinitions.apiextensions.k8s.io is forbidden: User "system:serviceaccount:monitoring:kube-prometheus-stack-kube-state-metrics" cannot list resource "customresourcedefinitions" in API group "apiextensions.k8s.io" at the cluster scope
```

sherifkayad: @k0nstantinv why do you need CRDs there? .. It's only regarding the VPA stuff and there I am not sure if you need anything other than the `verticalpodautoscalers` resource in the `autoscaling.k8s.io` group ..

sherifkayad: Now I get where my confusion comes from. I am using the Helm Chart directly to install KSM and hence, I just needed to add the VPA stuff .. I re-committed with the CRD RBAC permissions also added. Please have another look.

k0nstantinv: Thanks! looks ok

k0nstantinv (commented):
looks ok now

sherifkayad:
> looks ok now

Can you please re-trigger the tests?

k0nstantinv:
> > looks ok now
> 
> Can you please re-trigger the tests?

I’m not sure I’m able to do that :)

k0nstantinv:
@CatherineF-dev seems like we need re-trigger tests again, the doc was updated

korjek:
guys, let's move on to have this merged! This would save so much time for me (was lucky enough to bump into this PR after spending some time trying to figure out the correct config) and maybe for other folks if it's merged.

docs/customresourcestate-metrics.md
```diff
@@ -272,19 +267,55 @@ spec:
...
+                container: [containerName]
+          commonLabels:
+            resource: "cpu"
+            unit: "core"
+```
+
+The above configuration was tested on [this](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/examples/hamster.yaml) VPA configuration, outputting the following metrics:
+
+```
+# HELP kube_customresource_vpa_containerrecommendations_target VPA container recommendations for memory.
```
mrueg, line 299: As you can see here, this will generate two different help outputs for the same title. This should either be postfixed with _cpu / _memory (if you want two help outputs) or 
```suggestion
# HELP kube_customresource_vpa_containerrecommendations_target VPA container recommendations for CPU and memory.
```

docs/customresourcestate-metrics.md
```diff
@@ -272,19 +267,55 @@ spec:
...
 ```
 
-The above configuration was tested on [this](https://github.com/kubernetes/autoscaler/blob/master/vertical-pod-autoscaler/examples/hamster.yaml) VPA configuration, with an added annotation (`foo: 123`).
+In RBAC-enabled clusters and to ensure that the VPA Metrics will be captured seamlessly, make sure that the Cluster Role / Role assigned to the Kube State Metrics has permissions to `list` & `watch` VPA resources (as well as `customresourcedefinitions`) e.g.:
+
+```yaml
+- verbs:
+    - list
+    - watch
+  apiGroups:
```
mrueg, line 312: Usually permissions are set per apigroup/resource Otherwise you would give permission to an (unlikely, but possible) customresourcedefinitions.autoscaling.k8s.io or verticalpodautoscalers.apiextensions.k8s.io which you did not intend to provide.

docs/customresourcestate-metrics.md
korjek: using the same metric name both for cpu and memory will result in two HELP messages for the same type (ie `# HELP kube_customresource_vpa_containerrecommendations_target VPA container recommendations for memory` and `# HELP kube_customresource_vpa_containerrecommendations_target VPA container recommendations for cpu`). As a result, client (for example telegraf) fails to fetch metrics with `Error in plugin: error reading metrics for http://kube-state-metrics.kube-system.svc.cluster.local:8080/metrics: reading text format failed: text format parsing error in line 11485: second HELP line for metric name "kube_customresource_vpa_containerrecommendations_target"`.

So +1 for appending resource at the end (ie `- name: "vpa_containerrecommendations_target_cpu"`).
And if we do this, probably `commonLabels` can be removed

docs/customresourcestate-metrics.md
korjek, line 299: `# TYPE kube_customresource_vpa_containerrecommendations_target gauge`
is missing in the output

sherifkayad:
I am not really sure what needs to be done to merge this one. Has been open for a very long time and I would love to see it closed.

korjek:
indeed would be great to have this merged, so it's much easier to find correct documentation

sherifkayad:
I re-opened the PR and will work on the review comments to append a postfix to the metrics

docs/customresourcestate-metrics.md
sherifkayad: done! please re-check!

docs/customresourcestate-metrics.md
sherifkayad: done! please re-check!

sherifkayad: done! please re-check!

docs/customresourcestate-metrics.md
sherifkayad: As per the docs already in the file, that's how it needs to be .. please check the section above titled `### RBAC-enabled Clusters`!

rexagod:
This doesn't seem to add all of the missing VPA metrics?

sherifkayad:
@rexagod These are the recommendtion metrics for cpu and memory .. what else is missing?

PR #2204, opened 2023-09-20
chore: resolve sync concerns

rexagod:
* defer to the later go version when merging back to main (and not the
  older version introduced in the release branch).
* resolve merge conflicts.

Signed-off-by: Pranshu Srivastava <rexagod@gmail.com>

***

To resolve https://github.com/kubernetes/kube-state-metrics/pull/2203#discussion_r1332045207 (force-pushes are disable for `release-*`).

/cc @dgrisonnet


dgrisonnet:
I don't think we should do that since the Kubernetes 1.27 deps are based on go 1.20.z.

I would prefer having the release-2.10 diverge away from master and stick to go 1.20.z whilst master is on 1.21.z

rexagod:
Makes sense, but wouldn't `release-2.10` conflict with `main`? My plan was to resolve this on `release-2.10` itself and create a release from the last commit before this one, so merging back into `main` wouldn't be an issue.

dgrisonnet:
Normally you should never have to rebase a release branch on main since it is meant to diverge away. I am not sure why the release process is like that here, but I don't think it will work well going forward if we have to continue keeping go version up-to-date on release branch.

rexagod:
/close

Since the superseding PR was closed as well.

PR #2192, opened 2023-09-12
chore: Cut `v2.10.1`

rexagod:
**What this PR does / why we need it**: `v2.10.1` release prep.

```markdown
## v2.10.1 / 2023-09-13

### Note
- This release addresses a regression introduced in https://github.com/kubernetes/kube-state-metrics/pull/2105.

* [BUGFIX] Remove FieldSelector from non-namespaced resources by @mrueg and @dgrisonnet in https://github.com/kubernetes/kube-state-metrics/pull/2190
```

***

### TODOs

- [ ] Release v2.10.1.
- [ ] `kpromo` changes.

PR #2190, opened 2023-09-12
fix: Remove FieldSelector from non-namespaced resources 

dgrisonnet:
This PR takes over the initial work started by @mrueg in https://github.com/kubernetes/kube-state-metrics/pull/2189 and fix the additional cluster-scoped resources that were missing.

**What this PR does / why we need it**:

This should resolve the issue with namespace-denylist as discovered in https://github.com/kubernetes/kube-state-metrics/issues/2187

Regression introduced in https://github.com/kubernetes/kube-state-metrics/commit/c3c552821af3432afdaedec607f43ba1efba1be0

I'm not sure if the fix is complete or other resources are affected. I'm also not sure if this is the right fix or we should simply keep the field selector but fail if the field doesn't exist.

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2187

PR #2213, opened 2023-10-04
revert: Revert sync-based commits

rexagod:
This reverts commit 98ade63, reversing changes made to d68bad2.

***

`main` was merged into `release-2.10` to make the whole "merge back to `main`" process predictable, so there's no conflicts, while creating a `v2.10.1` release from the `HEAD~2` commit (the last commit before the merge `main` into `release-2.10` happened). But none of that applies anymore since we will be letting release branches diverge going forward, allowing for easier patch releases, be it CVEs, critical fixes, etc. without the need to merge back into main. Whether we'll do the same for minor version releases as well, is up for discussion at this moment.

Also note that the last commit regarding lint fixes was done owing to the `markdownlint` check that got in during the merge, which is also reverted in this PR.

cc @dgrisonnet


dgrisonnet:
Wait why do we have are we building with go 1.21 (https://github.com/kubernetes/kube-state-metrics/blob/release-2.10/Dockerfile) on release-2.10, but the go version is marked to be 1.20.8 (https://github.com/kubernetes/kube-state-metrics/blob/release-2.10/Makefile#L19)?

dgrisonnet:
An additional reason for not merging back

dgrisonnet:
@rexagod should we perhaps revert everything since the v2.10.0 cut and then I cherry-pick https://github.com/kubernetes/kube-state-metrics/pull/2190 properly and you do https://github.com/kubernetes/kube-state-metrics/pull/2192 again?

rexagod:
Seems like https://github.com/kubernetes/kube-state-metrics/pull/2213#issuecomment-1748866190 came in from 435aa5a0. Closing this in favor of https://github.com/kubernetes/kube-state-metrics/pull/2213#issuecomment-1748890182.

PR #2217, opened 2023-10-11
fix: Support scraping pods metrics that is still in scheduling status and has no assigned node

mickeyzzc:

When daemonset is used to scrape the metrics of pods, it can only scrape the metrics of pods assigned to nodes, but in scheduling, for example, pods in pending state cannot be scraped. 
This results in pending pods with no indicators to observe and alert. 
This pr adds a configuration parameter that, when enabled, captures pending pods from apiservice.

### DIFF
--- a/README.md
+++ b/README.md
@@ -250,7 +250,7 @@ The downside of using an auto-sharded setup comes from the rollout strategy supp
 
 For pod metrics, they can be sharded per node with the following flag:
 
-* `--node`
+* `--node=$(NODE_NAME)`
 
 Each kube-state-metrics pod uses FieldSelector (spec.nodeName) to watch/list pod metrics only on the same node.
 
@@ -274,6 +274,21 @@ spec:
             fieldRef:
               apiVersion: v1
               fieldPath: spec.nodeName
+```
+
+To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:
+```
+apiVersion: apps/v1
+kind: Deployment
+spec:
+  template:
+    spec:
+      containers:
+      - image: registry.k8s.io/kube-state-metrics/kube-state-metrics:IMAGE_TAG
+        name: kube-state-metrics
+        args:
+        - --resources=pods
+        - --node=""
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
--- a/README.md
+++ b/README.md
@@ -277,6 +277,7 @@ spec:
 ```
 
 To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:
+
 ```
 apiVersion: apps/v1
 kind: Deployment
--- a/examples/daemonsetsharding/deployment-no-node-pods.yaml
+++ b/examples/daemonsetsharding/deployment-no-node-pods.yaml
@@ -3,9 +3,9 @@ kind: Deployment
 metadata:
   labels:
     app.kubernetes.io/component: exporter
-    app.kubernetes.io/name: kube-state-metrics-global
+    app.kubernetes.io/name: kube-state-metrics-pods
     app.kubernetes.io/version: 2.10.0
-  name: kube-state-metrics-global
+  name: kube-state-metrics-pods
   namespace: kube-system
 spec:
   replicas: 1
--- a/examples/daemonsetsharding/deployment-no-node-pods.yaml
+++ b/examples/daemonsetsharding/deployment-no-node-pods.yaml
@@ -3,9 +3,9 @@ kind: Deployment
 metadata:
   labels:
     app.kubernetes.io/component: exporter
-    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/name: kube-state-metrics-global
     app.kubernetes.io/version: 2.10.0
-  name: kube-state-metrics-pods
+  name: kube-state-metrics-global
   namespace: kube-system
 spec:
   replicas: 1
--- a/examples/daemonsetsharding/deployment-no-node-pods.yaml
+++ b/examples/daemonsetsharding/deployment-no-node-pods.yaml
@@ -1,0 +1,59 @@
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: exporter
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: 2.10.0
+  name: kube-state-metrics-pods
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kube-state-metrics
+  template:
+    metadata:
+      labels:
+        app.kubernetes.io/component: exporter
+        app.kubernetes.io/name: kube-state-metrics
+        app.kubernetes.io/version: 2.10.0
+    spec:
+      automountServiceAccountToken: true
+      containers:
+      - args:
+        - --resources=pods
+        - --node=""
+        - --enable-no-node-scrape
+        image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.10.0
+        livenessProbe:
+          httpGet:
+            path: /healthz
+            port: 8080
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        name: kube-state-metrics
+        ports:
+        - containerPort: 8080
+          name: http-metrics
+        - containerPort: 8081
+          name: telemetry
+        readinessProbe:
+          httpGet:
+            path: /
+            port: 8081
+          initialDelaySeconds: 5
+          timeoutSeconds: 5
+        securityContext:
+          allowPrivilegeEscalation: false
+          capabilities:
+            drop:
+            - ALL
+          readOnlyRootFilesystem: true
+          runAsNonRoot: true
+          runAsUser: 65534
+          seccompProfile:
+            type: RuntimeDefault
+      nodeSelector:
+        kubernetes.io/os: linux
+      serviceAccountName: kube-state-metrics
--- a/examples/daemonsetsharding/deployment-no-node-pods.yaml
+++ b/examples/daemonsetsharding/deployment-no-node-pods.yaml
@@ -24,7 +24,6 @@ spec:
       - args:
         - --resources=pods
         - --node=""
-        - --enable-no-node-scrape
         image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.10.0
         livenessProbe:
           httpGet:
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -46,7 +46,6 @@ type Options struct {
 	Namespaces               NamespaceList   `yaml:"namespaces"`
 	NamespacesDenylist       NamespaceList   `yaml:"namespaces_denylist"`
 	Node                     NodeType        `yaml:"node"`
-	NoNodeScrape             bool            `yaml:"no_node_scrape"`
 	Pod                      string          `yaml:"pod"`
 	Port                     int             `yaml:"port"`
 	Resources                ResourceSet     `yaml:"resources"`
--- a/pkg/options/options.go
+++ b/pkg/options/options.go
@@ -46,6 +46,7 @@ type Options struct {
 	Namespaces               NamespaceList   `yaml:"namespaces"`
 	NamespacesDenylist       NamespaceList   `yaml:"namespaces_denylist"`
 	Node                     NodeType        `yaml:"node"`
+	NoNodeScrape             bool            `yaml:"no_node_scrape"`
 	Pod                      string          `yaml:"pod"`
 	Port                     int             `yaml:"port"`
 	Resources                ResourceSet     `yaml:"resources"`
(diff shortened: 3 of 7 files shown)


examples/daemonsetsharding/deployment-no-node-pods.yaml
```diff
@@ -0,0 +1,59 @@
...
+        app.kubernetes.io/component: exporter
+        app.kubernetes.io/name: kube-state-metrics
+        app.kubernetes.io/version: 2.10.0
+    spec:
+      automountServiceAccountToken: true
+      containers:
+      - args:
+        - --resources=pods
+        - --node=""
+        - --enable-no-node-scrape
```
CatherineF-dev, line 27: I think we can remove `- --enable-no-node-scrape`. If node == "", then it will collect pod metrics whose don't have  assigned nodes.

CatherineF-dev:
Cool! Thanks for it!
We also found Daemonset sharding loses pod metrics who doesn't have assigned nodes. Just found it can be fixed by this PR.

examples/daemonsetsharding/deployment-no-node-pods.yaml
CatherineF-dev: QQ: could you provide a config example on `only namespace is selected regardless of whether it is assigned to node.`?

Is it --node="" && --namespace="kube-system"

mickeyzzc: Yes, let me run a test to confirm.

mickeyzzc: I just ran a test with this PR version.
My cluster currently has the following pods pending
image
When I use the following configuration，it can scrape to pending pods under namespace media
```
          args:
            - '--resources=pods'
            - '--node='
            - '--enable-no-node-scrape'
            - '--namespaces=media'
```
When I use the following configuration，it can scrape to all pods under namespace media
```
          args:
            - '--resources=pods'
            - '--node='
            - '--namespaces=media'
```

CatherineF-dev: We need to change this:

```
func (n *NodeType) GetNodeFieldSelector() string {
	if string(*n) != "" {
		return fields.OneTermEqualSelector("spec.nodeName", string(*n)).String()
	}
	return EmptyFieldSelector()
}
```
to 
```
func (n *NodeType) GetNodeFieldSelector() string {
	return fields.OneTermEqualSelector("spec.nodeName", string(*n)).String()
}
```

https://github.com/kubernetes/kube-state-metrics/blob/5c71459a614eaf8370bed2f406899f2269edff66/pkg/options/types.go#L111C20-L116

CatherineF-dev: Oh, I realized the issue.

If we change the above, it will break for `--namespaces=media`. Because we assume collecting all nodes when node = "".

mickeyzzc: Yes, it is. I think the current PR version is flexible. So this configuration combination was added.

CatherineF-dev: One solution is to 1. change node type from string to list 2. add an assert that #nodes <= 1:

```
o.cmd.Flags().StringVar((*string)(&o.Node), "node"
```
to 
```
o.cmd.Flags().Var(&o.Namespaces, "namespaces", # similar to namespaces.
```
So that we can differentiate between nil and "". 

https://github.com/kubernetes/kube-state-metrics/blob/5c71459a614eaf8370bed2f406899f2269edff66/pkg/options/options.go#L139C2-L139C2

CatherineF-dev: QQ: what happened during upgrade? I thought it only had one small breaking change:

Before: --node= works
Now: --node= works

One breaking change: 
Before: --node="" means collecting all nodes.
Now: --node="" means collecting empty node.

But I think most people will not set --node if they don't use this field. We can update release note about it. Since it's an experimental feature, it could have this fix.

I think this way is unified. I didn't realize collecting empty node metrics, and we should fix it.

mickeyzzc: I'm testing the code, just a little confused to make sure. I will submit the revised version later.

examples/daemonsetsharding/deployment-no-node-pods.yaml
```diff
@@ -0,0 +1,58 @@
+apiVersion: apps/v1
```
mrueg, line 1: This file should be generated from jsonnet See: https://github.com/kubernetes/kube-state-metrics/blob/main/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet

mickeyzzc: > This file should be generated from jsonnet See: https://github.com/kubernetes/kube-state-metrics/blob/main/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet

fix

examples/daemonsetsharding/deployment-no-node-pods.yaml
```diff
@@ -0,0 +1,58 @@
...
+kind: Deployment
+metadata:
+  labels:
+    app.kubernetes.io/component: exporter
+    app.kubernetes.io/name: kube-state-metrics
+    app.kubernetes.io/version: 2.10.0
+  name: kube-state-metrics
```
CatherineF-dev, line 8: nit: kube-state-metrics -> kube-state-metrics-global

We have another deployment with the same name. In the future, we can let ` kube-state-metrics-global` to collect all other metrics which can't be sharded by node.

pkg/options/options.go
```diff
@@ -46,6 +46,7 @@ type Options struct {
 	Namespaces               NamespaceList   `yaml:"namespaces"`
 	NamespacesDenylist       NamespaceList   `yaml:"namespaces_denylist"`
 	Node                     NodeType        `yaml:"node"`
+	NoNodeScrape             bool            `yaml:"no_node_scrape"`
```
CatherineF-dev, line 49: nit: we don't need this now.

pkg/options/types.go
```diff
@@ -105,14 +106,52 @@ func (r *ResourceSet) Type() string {
 }
 
 // NodeType represents a nodeName to query from.
-type NodeType string
+type NodeType map[string]struct{}
+
+func (n *NodeType) Set(value string) error {
+	s := *n
```
CatherineF-dev, line 112: Make `s` a new map.

Because if `n` has 10 node names with 2 duplicated.  Length of `cols := make([]string, 0, len(n))` is 10 instead of 9.

pkg/options/types.go
```diff
@@ -105,14 +106,52 @@ func (r *ResourceSet) Type() string {
...
-	if string(*n) != "" {
-		return fields.OneTermEqualSelector("spec.nodeName", string(*n)).String()
+	if nil == n || len(*n) == 0 {
+		klog.InfoS("Using node type is nil")
+		return EmptyFieldSelector()
 	}
-	return EmptyFieldSelector()
+	pattern := "[^a-zA-Z0-9_,-]+"
+	re := regexp.MustCompile(pattern)
+	result := re.ReplaceAllString(n.String(), "")
```
CatherineF-dev, line 147: QQ: why do we need `re.ReplaceAllString(n.String(), "")`

CatherineF-dev: nit: remove `re.ReplaceAllString(n.String(), "")`. If node name isn't correct, keep it as before. Then users can know node name isn't correct.

pkg/options/types_test.go
```diff
@@ -162,13 +162,25 @@ func TestNodeFieldSelector(t *testing.T) {
 		Wanted string
 	}{
 		{
```
CatherineF-dev, line 164: Add a test that it can differentiate between nil and "".

CatherineF-dev:
Cool! Seems we can make it without an extra flag.

pkg/options/types_test.go
mickeyzzc: > ```
> `spec.nodeName=""`  # ""
> `spec.nodeName=` # nil
> ```
> 
> This might work, since we can't use " spec.nodeName="" "

I thought so too, so I added regulars to prevent this, as well as modifications to the configuration.

mickeyzzc: > ```
> `spec.nodeName=""`  # ""
> `spec.nodeName=` # nil
> ```
> 
> This might work, since we can't use " spec.nodeName="" "

I thought so too, so I added regulars to prevent this, as well as modifications to the configuration.

pkg/options/types.go
mickeyzzc: > QQ: why we need `re.ReplaceAllString(n.String(), "")`

There are two considerations for regularization here: 
1. Handle the difference between "" and nil 
2. Correction of node name

pkg/options/types_test.go
mickeyzzc: > Add a test that it can differentiate between nil and "".

I fixed it locally. Submit it later along with other questions
image

pkg/options/types.go
mickeyzzc: > Make `s` a new map.
> 
> Because if `n` has 10 node names with 2 duplicated. Length of `cols := make([]string, 0, len(n))` is 10 instead of 9.

Is it possible to configure the function NewOptions？
```
// NewOptions returns a new instance of `Options`.
func NewOptions() *Options {
	return &Options{
		Resources:            ResourceSet{},
		MetricAllowlist:      MetricSet{},
		MetricDenylist:       MetricSet{},
		MetricOptInList:      MetricSet{},
		Node:                 NodeType{},
		AnnotationsAllowList: LabelsAllowList{},
		LabelsAllowList:      LabelsAllowList{},
	}
}
```

examples/daemonsetsharding/deployment-no-node-pods.yaml
```diff
@@ -0,0 +1,58 @@
...
+    app.kubernetes.io/component: exporter
+    app.kubernetes.io/name: kube-state-metrics-global
+    app.kubernetes.io/version: 2.10.0
+  name: kube-state-metrics-global
+  namespace: kube-system
+spec:
+  replicas: 1
+  selector:
+    matchLabels:
+      app.kubernetes.io/name: kube-state-metrics
```
CatherineF-dev, line 14: nit: not sure whether we need to rename it to kube-state-metrics-global.

It has two kube-state-metrics (one Daemonset and one Deployment). If both of them are using the same selector, some wrong pods might be selected. But not sure about it.

pkg/options/types.go
CatherineF-dev: QQ: could you give two examples for these two cases?

We shouldn't change node name, which users are not aware of.

examples/daemonsetsharding/deployment-no-node-pods.yaml
mickeyzzc: I spent some time testing a few more scenarios to give a multi-point combination of cases out.

mickeyzzc: Pods will be clearer, I made some combinations, and then came up with some names, it seems that pods will be better, we can make supplementary explanations in the documentation

mickeyzzc: two kube-state-metrics (one Daemonset and one Deployment).
I tested it directly on my side repeatedly and found no duplication.

mickeyzzc:
> /triage accepted

@dashpole Do I need your approval to merge again?

dashpole:
@mickeyzzc I just did triage, to ensure the PR has an appropriate assignee.  You will need approval from the reviewers and assignees on the PR to merge.

pkg/options/types.go
CatherineF-dev: Ping~

Keep node name as same from input.

mickeyzzc:
> @mickeyzzc I just did triage, to ensure the PR has an appropriate assignee. You will need approval from the reviewers and assignees on the PR to merge.

@mrueg Can you help with the merger?

mrueg:
To my understanding, when someone runs the daemonset for pod metrics, you suggest to have an additional deployment with --node="" to track unassigned pods?

Isn't it simpler to include pod metrics with --node="" in the regular [ksm deployment](https://github.com/kubernetes/kube-state-metrics/blob/main/examples/daemonsetsharding/deployment.yaml) that is already there additionally to the DS or does it make sense to have an additional deployment? (I'm not sure how users use this in terms of their metric storage system. @CatherineF-dev might be able to provide insights)

And before we can merge, can you update the README so users know how to apply this? 
https://github.com/kubernetes/kube-state-metrics#daemonset-sharding-for-pod-metrics

CatherineF-dev:
> Isn't it simpler to include pod metrics with --node="" in the regular [ksm deployment](https://github.com/kubernetes/kube-state-metrics/blob/main/examples/daemonsetsharding/deployment.yaml) that is already there additionally to the DS or does it make sense to have an additional deployment?

Yes, this PR want to fix ` --node=""` to not select any nodes. Without this CL, --node="" collects all nodes.

CatherineF-dev:
I left one comment here https://github.com/kubernetes/kube-state-metrics/pull/2217/files#r1399988722

mickeyzzc:
> And before we can merge, can you update the README so users know how to apply this?
> https://github.com/kubernetes/kube-state-metrics#daemonset-sharding-for-pod-metrics

@mrueg completed

pkg/options/types.go
mrueg: @mickeyzzc can you address this comment?

README.md
```diff
@@ -276,6 +276,21 @@ spec:
               fieldPath: spec.nodeName
 ```
 
+To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:
+```
```
mrueg, line 280: @mickeyzzc can you add an empty line before the codeblock to fix the lint? Thanks!

README.md
```diff
@@ -276,6 +276,21 @@ spec:
               fieldPath: spec.nodeName
 ```
 
+To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:
```
mrueg, line 279:
```suggestion
To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:

```

mrueg:
/lgtm 

Thank you for your contribution!

PR #2222, opened 2023-10-19
feat: Add new metric kube_pod_scheduler

### LINKED ISSUE #2207: schedulerName in kube_pod_* metrics group
It would be great if we will have schedulerName in exported kube_info_metrics.
**Why is this needed**:
When we have more than one scheduler in the cluster, we need to aggregate some data by scheduler.
It could be a metric called kube_pod_scheduler_name with labels set {pod, namespace, uid, schedulerName (or just name)}. Value can be taken from .spec.schedulerName.

adinhodovic:

Adds a new pod metric `kube_pod_scheduler_name` that allows seeing which scheduler scheduled the pod. Allows aggregation by scheduler.

Increases.

Fixes #2207


internal/store/pod.go
```diff
@@ -1708,6 +1709,27 @@ func createPodServiceAccountFamilyGenerator() generator.FamilyGenerator {
...
+func createPodSchedulerNameFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGeneratorWithStability(
+		"kube_pod_scheduler_name",
+		"The scheduler name for a pod.",
+		metric.Gauge,
+		basemetrics.ALPHA,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			m := metric.Metric{
+				LabelKeys:   []string{"scheduler_name"},
```
dgrisonnet, line 1721: I am wondering if having something like `kube_pod_scheduler{name="foo"}` could be better to remove the redundancy with the metric name its label.

adinhodovic: Think that makes sense.  Updated.

dgrisonnet (approved):
/lgtm
/approve

PR #2234, opened 2023-10-30
feat: Support filtering annotations allowlist by "*"

xonvanetta:
Same as #1823. Thought this was done on both allowlist since they are the same type but this wasn't the case. Added support for both now.
>Support filtering label allowlist by "*", which will expand to the enabled resources, while inferring their values based on its value(s).


mfld (approved):

CatherineF-dev:
Robot asked to me 

/assign @mrueg

internal/store/builder.go
```diff
@@ -219,32 +219,41 @@ func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.Registry
 	}
 }
 
-// WithAllowAnnotations configures which annotations can be returned for metrics
-func (b *Builder) WithAllowAnnotations(annotations map[string][]string) {
-	if len(annotations) > 0 {
-		b.allowAnnotationsList = annotations
+func (b *Builder) allowLabels(labels map[string][]string) (map[string][]string, error) {
```
mrueg, line 222: Can you rename this function to `allowList` (if it's used for annotations and labels) or similar and add a descriptive comment on what it does?

Also please rename the labels argument to something more generic

internal/store/builder_test.go
```diff
@@ -113,3 +113,86 @@ func TestWithAllowLabels(t *testing.T) {
 		}
 	}
 }
+
+func TestWithAllowAnnotations(t *testing.T) {
+	tests := []struct {
+		Desc                 string
+		AnnotationsAllowlist map[string][]string
+		EnabledResources     []string
+		Wanted               LabelsAllowList
```
mrueg, line 122: should this be
```suggestion
		Wanted               AnnotationsAllowList
```

?

xonvanetta: Well, It reuses the same type from the test above [here](https://github.com/kubernetes/kube-state-metrics/pull/2234/files/e4d2bd48c8cfa950941dd51149a75ffbbb8f44eb#diff-c8cf0574c6fe66c4a3d8522108993b75d0421ec94af5eab80c3ed25d01d64cd1R26). I could declare a new type and point it to `options.LabelsAllowList`, while at it should I change [this](https://github.com/kubernetes/kube-state-metrics/blob/82fd15ccc4a4c716721fcb135b1f7ef129025a3f/pkg/options/options.go#L32) to a more generic also? 

I noticed that these two tests test the same function in the end, I wrote tests first then went to implement, remove the test or keep it?

mrueg: Ah I didn't know AnnotatiosAllowList don't exist. I think we can keep it for this change like this. :)

PR #2254, opened 2023-12-01
build(deps): Bump go-restful/v3 to v3.11.0 to fix vul issues

### LINKED ISSUE #2253: Twistlock PRISMA-2022-0227
As discussed in this issue: https://github.com/kubernetes/kube-state-metrics/issues/1999#issuecomment-1652439596

Description of CVE: github.com/emicklei/go-restful/v3 module prior to v3.10.0 is vulnerable to Authentication Bypass by Primary Weakness. There is an inconsistency in how go-restful parses URL paths. This inconsistency could lead to several security check bypass in a complex system.

I think its still appearing because of this line: https://github.com/kubernetes/kube-state-metrics/blob/main/go.mod#L35

Is it possibly a false positive then?

CatherineF-dev:
**What this PR does / why we need it**: #2253

Fixes #2253

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -32,7 +32,7 @@ require (
 	github.com/cespare/xxhash/v2 v2.2.0 // indirect
 	github.com/coreos/go-systemd/v22 v22.5.0 // indirect
 	github.com/davecgh/go-spew v1.1.2-0.20180830191138-d8f796af33cc // indirect
-	github.com/emicklei/go-restful/v3 v3.9.0 // indirect
+	github.com/emicklei/go-restful/v3 v3.11.0 // indirect
 	github.com/evanphx/json-patch v5.6.0+incompatible // indirect
 	github.com/go-kit/log v0.2.1 // indirect
 	github.com/go-logfmt/logfmt v0.5.1 // indirect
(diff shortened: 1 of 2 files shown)


CatherineF-dev:
Ok! Thx! I didn't notice this issue

CatherineF-dev:
We need to bump k8s dependency instead of this library directly. https://github.com/kubernetes/kube-state-metrics/commit/89403d239a47b75625d5939c8f26c73f4dc34245

mrueg:
I also don't know if this is necessary at all and we shouldn't consider this a false positive that is safe to ignore.

See:
https://github.com/kubernetes/kubernetes/pull/119865#issuecomment-1739082300
```
To be crystal clear, Kubernetes is not affected at all by the go-restful security issue because we apply authentication / authorization directly to the incoming http.Request object / URL, not using go-restful filters.

The primary motivation for picking up this version update is to silence security scanners incorrectly flagging Kubernetes as impacted by the go-restful vulnerability. If PRISMA-2022-0227 decides to remark 3.11.0+ as vulnerable because the default behavior triggers the go-restful bug, picking up this update will not do anything to silence those scanners.
```

Issue #2223, opened 2023-10-20
repeated adding and deleting CustomResourceDefinitions causes duplicate metric entries  


k15r:
**What happened**:

This is part of our kubestate customresource configuration:
```
       - errorLogV: 0
         groupVersionKind:
           group: operator.kyma-project.io
           kind: Keda
           version: '*'
         metrics:
         - each:
             stateSet:
               labelName: state
               list:
               - Ready
               - Processing
               - Error
               - Deleting
               - Warning
               path:
               - status
               - state
             type: StateSet
           errorLogV: 0
           help: status of Keda CR
           labelsFromPath:
             name:
             - metadata
             - name
             namespace:
             - metadata
             - namespace
           name: keda_status
```

after adding and deleting the corresponding CRD and on CR its kind this is a part of the `/metrics` response of kubestatemetrics:

```
# HELP kube_customresource_keda_status status of Keda CR
# TYPE kube_customresource_keda_status stateset
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Deleting"} 0
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Error"} 0
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Processing"} 1
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Ready"} 0
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Warning"} 0
# HELP kube_customresource_keda_status status of Keda CR
# TYPE kube_customresource_keda_status stateset
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Deleting"} 0
kube_customresource_keda_status{customresource_group="operator.kyma-project.io",customresource_kind="Keda",customresource_version="v1alpha1",name="default",namespace="kyma-system",state="Error"} 0
[...]

PR #2257, opened 2023-12-06
fix: fix the issue that duplicated CR metrics are generated when CRD is deleted/applied multiple times

CatherineF-dev:
**What this PR does / why we need it**: https://github.com/kubernetes/kube-state-metrics/issues/2223

```
# It returns same metrics data many times when CRD is deleted/applied multiple times
 curl localhost:8089/metrics
# HELP cr_creationtimestamp
# TYPE cr_creationtimestamp gauge
cr_creationtimestamp{customresource_group="example.com",customresource_kind="Bar",customresource_version="v1",name="mybar"} 1.701828773e+09
# HELP cr_resourceversion
# TYPE cr_resourceversion gauge
cr_resourceversion{customresource_group="example.com",customresource_kind="Bar",customresource_version="v1",name="mybar"} 909919
# HELP cr_creationtimestamp
# TYPE cr_creationtimestamp gauge
cr_creationtimestamp{customresource_group="example.com",customresource_kind="Bar",customresource_version="v1",name="mybar"} 1.701828773e+09
# HELP cr_resourceversion
# TYPE cr_resourceversion gauge
cr_resourceversion{customresource_group="example.com",customresource_kind="Bar",customresource_version="v1",name="mybar"} 909919
```

Mitigation for https://github.com/kubernetes/kube-state-metrics/issues/2223

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -121,6 +121,7 @@ func (b *Builder) WithEnabledResources(r []string) error {
 	b.enabledResources = append(b.enabledResources, sortedResources...)
 	// It's an mitigation for #2223. unique should be removed after #2223 is fixed.
 	b.enabledResources = unique(b.enabledResources)
+	sort.Strings(b.enabledResources)
 	return nil
 }
 
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -119,7 +119,20 @@ func (b *Builder) WithEnabledResources(r []string) error {
 	sort.Strings(sortedResources)
 
 	b.enabledResources = append(b.enabledResources, sortedResources...)
+	b.enabledResources = unique(b.enabledResources)
 	return nil
+}
+
+func unique(arr []string) []string {
+	mp := make(map[string]bool)
+	for _, s := range arr {
+		mp[s] = true
+	}
+	result := []string{}
+	for key, _ := range mp {
+		result = append(result, key)
+	}
+	return result
 }
 
 // WithFieldSelectorFilter sets the fieldSelector property of a Builder.
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -119,6 +119,7 @@ func (b *Builder) WithEnabledResources(r []string) error {
 	sort.Strings(sortedResources)
 
 	b.enabledResources = append(b.enabledResources, sortedResources...)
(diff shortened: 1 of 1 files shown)


internal/store/builder.go
```diff
@@ -119,9 +119,22 @@ func (b *Builder) WithEnabledResources(r []string) error {
 	sort.Strings(sortedResources)
 
 	b.enabledResources = append(b.enabledResources, sortedResources...)
+	b.enabledResources = unique(b.enabledResources)
```
mrueg, line 122: Does it make sense to have a map[string] bool here?

dgrisonnet:
I don't think this will fix the inherent issue. It seems that the mechanism detecting the deletion of CRD matched by the ksm config is not working properly. Normally when that happens, the internal store should be deleted alongside its informer because we don't want to expose metrics about this resource anymore.

dgrisonnet:
To be clearer, the scenario where a specific resource appears twice in ksm code should never happen, so we should understand why it happened and fix it so that it appears only once and is correctly cleaned up when needed.

CatherineF-dev:
got it, agree. Will debug again to find root cause.

CatherineF-dev:
It should be related to https://github.com/kubernetes/kube-state-metrics/pull/1851 where adding PollForCacheUpdates function.

CatherineF-dev:
It didn't delete removed CR https://github.com/rexagod/kube-state-metrics/blob/25a1d8da057cf761d614c59a52785335d34082d1/internal/discovery/discovery.go#L232.


How about creating a new issue and merging this PR as a mitigation?
I think it needs some changes around https://github.com/kubernetes/kube-state-metrics/pull/1851.

cc @dgrisonnet

internal/store/builder.go
CatherineF-dev: I think using unique is fine as a mitigation.

It didn't remove deleted CR https://github.com/rexagod/kube-state-metrics/blob/25a1d8da057cf761d614c59a52785335d34082d1/internal/discovery/discovery.go#L232. After this is fixed, we can remove unique.

logicalhan (commented):
/triage accepted
/assign @dgrisonnet

korjek:
any updates here?
It leads to loosing the metrics, which is not good.

PR #2270, opened 2023-12-13
fix: fallback to `gauge` for `protofmt`-based negotiations

### LINKED ISSUE #2248: Info and StateSet types cause errors when scrapped by Prometheus
**What happened**:
When kube-state-metrics is configured with custom resources that produce Info and/or StateSet type of metrics, and Prometheus is configured with `--enable-feature=native-histograms`, prometheus is unable to scrape the kube-state-metrics target.

Filed an issue with prometheus here https://github.com/prometheus/prometheus/issues/13172 but seems like a possible issue with KSM as well.

Successful scraping of metrics. 

Run prometheus with `--enable-feature=native-histograms` and setup kube-state-metrics with a metric of type: Info or Stateset. 

See a https://github.com/fluxcd/flux2-monitoring-example/issues/14 or https://github.com/defenestration/flux2-monitoring-example/pull/1 for an example setup.

With the `native-histograms` feature disabled, prometheus seems to be able to scrape successfully. So a bug is also filed with prometheus. 

I guess the suggestion would be to include a setting on the custom resource to produce a promehteus-compatible type on the /metrics page for these custom resource types. 

**Environment**:

* kube-state-metrics version: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.10.1
* Kubernetes version (use `kubectl version`):  v1.27.7+k3s2
* Cloud provider or hardware configuration:  kind
* Other info:

rexagod:
**What this PR does / why we need it**: Fallback to `gauge` metric type when the negotiated content-type is `protofmt`-based, since Prometheus' protobuf machinery does not recognize all OpenMetrics types (`info` and `statesets` in this context).

**How does this change affect the cardinality of KSM**: None.

Fixes #2248

### DIFF
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -20,6 +20,8 @@ import (
 	"fmt"
 	"strings"
 
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
+
 	"github.com/gobuffalo/flect"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/klog/v2"
@@ -148,7 +150,7 @@ type Generator struct {
 type Metric struct {
 	// Type defines the type of the metric.
 	// +unionDiscriminator
-	Type MetricType `yaml:"type" json:"type"`
+	Type metric.Type `yaml:"type" json:"type"`
 
 	// Gauge defines a gauge metric.
 	// +optional
@@ -170,9 +172,16 @@ type ConfigDecoder interface {
 func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscoverer) (func() ([]customresource.RegistryFactory, error), error) {
 	var customResourceConfig Metrics
 	factoriesIndex := map[string]bool{}
+
+	// Decode the configuration.
 	if err := decoder.Decode(&customResourceConfig); err != nil {
 		return nil, fmt.Errorf("failed to parse Custom Resource State metrics: %w", err)
 	}
+
+	// Override the configuration with any custom overrides.
+	configOverrides(&customResourceConfig)
+
+	// Create a factory for each resource.
 	fn := func() (factories []customresource.RegistryFactory, err error) {
 		resources := customResourceConfig.Spec.Resources
 		// resolvedGVKPs will have the final list of GVKs, in addition to the resolved G** resources.
@@ -205,4 +214,15 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 		return factories, nil
 	}
 	return fn, nil
+}
+
+// configOverrides applies overrides to the configuration.
+func configOverrides(config *Metrics) {
+	for i := range config.Spec.Resources {
+		for j := range config.Spec.Resources[i].Metrics {
+
+			// Override the metric type to lowercase, so the internals have a single source of truth for metric type definitions.
+			config.Spec.Resources[i].Metrics[j].Each.Type = metric.Type(strings.ToLower(string(config.Spec.Resources[i].Metrics[j].Each.Type)))
+		}
+	}
 }
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -222,6 +222,7 @@ func configOverrides(config *Metrics) {
 		for j := range config.Spec.Resources[i].Metrics {
 
 			// Override the metric type to lowercase, so the internals have a single source of truth for metric type definitions.
+			// This is done as a convenience measure for users, so they don't have to remember the exact casing.
 			config.Spec.Resources[i].Metrics[j].Each.Type = metric.Type(strings.ToLower(string(config.Spec.Resources[i].Metrics[j].Each.Type)))
 		}
 	}
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -72,7 +72,7 @@ func compileCommon(c MetricMeta) (*compiledCommon, error) {
 func compileFamily(f Generator, resource Resource) (*compiledFamily, error) {
 	labels := resource.Labels.Merge(f.Labels)
 
-	if f.Each.Type == MetricTypeInfo && !strings.HasSuffix(f.Name, "_info") {
+	if f.Each.Type == metric.Info && !strings.HasSuffix(f.Name, "_info") {
 		klog.InfoS("Info metric does not have _info suffix", "gvk", resource.GroupVersionKind.String(), "name", f.Name)
 	}
 
@@ -153,7 +153,7 @@ type compiledMetric interface {
 // newCompiledMetric returns a compiledMetric depending on the given metric type.
 func newCompiledMetric(m Metric) (compiledMetric, error) {
 	switch m.Type {
-	case MetricTypeGauge:
+	case metric.Gauge:
 		if m.Gauge == nil {
 			return nil, errors.New("expected each.gauge to not be nil")
 		}
@@ -172,7 +172,7 @@ func newCompiledMetric(m Metric) (compiledMetric, error) {
 			NilIsZero:      m.Gauge.NilIsZero,
 			labelFromKey:   m.Gauge.LabelFromKey,
 		}, nil
-	case MetricTypeInfo:
+	case metric.Info:
 		if m.Info == nil {
 			return nil, errors.New("expected each.info to not be nil")
 		}
@@ -185,7 +185,7 @@ func newCompiledMetric(m Metric) (compiledMetric, error) {
 			compiledCommon: *cc,
 			labelFromKey:   m.Info.LabelFromKey,
 		}, nil
-	case MetricTypeStateSet:
+	case metric.StateSet:
 		if m.StateSet == nil {
 			return nil, errors.New("expected each.stateSet to not be nil")
 		}
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -152,7 +152,7 @@ type compiledMetric interface {
 
 // newCompiledMetric returns a compiledMetric depending on the given metric type.
 func newCompiledMetric(m Metric) (compiledMetric, error) {
-	switch m.Type {
+	switch metric.Type(strings.ToLower(string(m.Type))) {
 	case metric.Gauge:
 		if m.Gauge == nil {
 			return nil, errors.New("expected each.gauge to not be nil")
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -152,7 +152,7 @@ type compiledMetric interface {
 
 // newCompiledMetric returns a compiledMetric depending on the given metric type.
 func newCompiledMetric(m Metric) (compiledMetric, error) {
-	switch metric.Type(strings.ToLower(string(m.Type))) {
+	switch m.Type {
 	case metric.Gauge:
 		if m.Gauge == nil {
 			return nil, errors.New("expected each.gauge to not be nil")
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -44,8 +44,8 @@ type Type string
 type TypeN int
 
 // String returns the string representation of the metric type.
-func (i TypeN) String() Type {
-	return MetricTypeNMap[i]
+func (i TypeN) String() string {
+	return string(TypeNMap[i])
 }
 
 // NString returns the string representation of the metric type as an integer.
@@ -55,7 +55,7 @@ func (i TypeN) NString() string {
 
 // Supported metric types.
 const (
-	// GaugeN defines a OpenMetrics gauge.
+	// GaugeN defines an OpenMetrics gauge.
 	GaugeN TypeN = iota
 
 	// InfoN defines an OpenMetrics info.
@@ -64,14 +64,14 @@ const (
 	// StateSetN defines an OpenMetrics stateset.
 	StateSetN
 
-	// CounterN defines a OpenMetrics counter.
+	// CounterN defines an OpenMetrics counter.
 	CounterN
 )
 
 // Supported metric types.
 var (
 
-	// Gauge defines a OpenMetrics gauge.
+	// Gauge defines an OpenMetrics gauge.
 	Gauge Type = "gauge"
 
 	// Info defines an OpenMetrics info.
@@ -80,19 +80,19 @@ var (
 	// StateSet defines an OpenMetrics stateset.
 	StateSet Type = "stateset"
 
-	// Counter defines a OpenMetrics counter.
+	// Counter defines an OpenMetrics counter.
 	Counter Type = "counter"
 
-	// MetricTypeNMap is a map of MetricTypeN to MetricType.
-	MetricTypeNMap = map[TypeN]Type{
+	// TypeNMap is a map of MetricTypeN to MetricType.
+	TypeNMap = map[TypeN]Type{
 		GaugeN:    Gauge,
 		InfoN:     Info,
 		StateSetN: StateSet,
 		CounterN:  Counter,
 	}
 
-	// MetricTypeMap is a map of MetricType to MetricTypeN.
-	MetricTypeMap = map[Type]TypeN{
+	// TypeMap is a map of MetricType to MetricTypeN.
+	TypeMap = map[Type]TypeN{
 		Gauge:    GaugeN,
 		Info:     InfoN,
 		StateSet: StateSetN,
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -40,11 +40,6 @@ var (
 // Type represents the type of the metric. See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
 type Type string
 
-// String returns the string representation of the metric type.
-func (t Type) String() string {
-	return string(t)
-}
-
 // Supported metric types.
 var (
 
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -40,33 +40,10 @@ var (
 // Type represents the type of the metric. See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
 type Type string
 
-// TypeN represents the type of the metric as an integer.
-type TypeN int
-
 // String returns the string representation of the metric type.
-func (i TypeN) String() string {
-	return string(TypeNMap[i])
+func (t Type) String() string {
+	return string(t)
 }
-
-// NString returns the string representation of the metric type as an integer.
-func (i TypeN) NString() string {
-	return strconv.Itoa(int(i))
-}
-
-// Supported metric types.
-const (
-	// GaugeN defines an OpenMetrics gauge.
-	GaugeN TypeN = iota
-
-	// InfoN defines an OpenMetrics info.
-	InfoN
-
-	// StateSetN defines an OpenMetrics stateset.
-	StateSetN
-
-	// CounterN defines an OpenMetrics counter.
-	CounterN
-)
 
 // Supported metric types.
 var (
@@ -82,22 +59,6 @@ var (
 
 	// Counter defines an OpenMetrics counter.
 	Counter Type = "counter"
-
-	// TypeNMap is a map of MetricTypeN to MetricType.
-	TypeNMap = map[TypeN]Type{
-		GaugeN:    Gauge,
-		InfoN:     Info,
-		StateSetN: StateSet,
-		CounterN:  Counter,
-	}
-
-	// TypeMap is a map of MetricType to MetricTypeN.
-	TypeMap = map[Type]TypeN{
-		Gauge:    GaugeN,
-		Info:     InfoN,
-		StateSet: StateSetN,
-		Counter:  CounterN,
-	}
 )
 
 // Metric represents a single time series.
--- a/pkg/metric/metric.go
+++ b/pkg/metric/metric.go
@@ -37,21 +37,68 @@ var (
 	}
 )
 
-// Type represents the type of a metric e.g. a counter. See
-// https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
+// Type represents the type of the metric. See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
 type Type string
 
-// Gauge defines a OpenMetrics gauge.
-var Gauge Type = "gauge"
+// TypeN represents the type of the metric as an integer.
+type TypeN int
 
-// Info defines an OpenMetrics info.
-var Info Type = "info"
+// String returns the string representation of the metric type.
+func (i TypeN) String() Type {
+	return MetricTypeNMap[i]
+}
 
-// StateSet defines an OpenMetrics stateset.
-var StateSet Type = "stateset"
+// NString returns the string representation of the metric type as an integer.
+func (i TypeN) NString() string {
+	return strconv.Itoa(int(i))
+}
 
-// Counter defines a OpenMetrics counter.
-var Counter Type = "counter"
+// Supported metric types.
+const (
+	// GaugeN defines a OpenMetrics gauge.
+	GaugeN TypeN = iota
+
+	// InfoN defines an OpenMetrics info.
+	InfoN
+
+	// StateSetN defines an OpenMetrics stateset.
+	StateSetN
+
+	// CounterN defines a OpenMetrics counter.
+	CounterN
+)
+
+// Supported metric types.
+var (
+
+	// Gauge defines a OpenMetrics gauge.
+	Gauge Type = "gauge"
+
+	// Info defines an OpenMetrics info.
+	Info Type = "info"
+
+	// StateSet defines an OpenMetrics stateset.
+	StateSet Type = "stateset"
+
+	// Counter defines a OpenMetrics counter.
+	Counter Type = "counter"
+
+	// MetricTypeNMap is a map of MetricTypeN to MetricType.
+	MetricTypeNMap = map[TypeN]Type{
+		GaugeN:    Gauge,
+		InfoN:     Info,
+		StateSetN: StateSet,
+		CounterN:  Counter,
+	}
+
+	// MetricTypeMap is a map of MetricType to MetricTypeN.
+	MetricTypeMap = map[Type]TypeN{
+		Gauge:    GaugeN,
+		Info:     InfoN,
+		StateSet: StateSetN,
+		Counter:  CounterN,
+	}
+)
 
 // Metric represents a single time series.
 type Metric struct {
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -19,6 +19,11 @@ package metricsstore
 import (
 	"fmt"
 	"io"
+	"strings"
+
+	"github.com/prometheus/common/expfmt"
+
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
 )
 
 // MetricsWriterList represent a list of MetricsWriter
@@ -82,13 +87,22 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 	return nil
 }
 
-// SanitizeHeaders removes duplicate headers from the given MetricsWriterList for the same family (generated through CRS).
-// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
-func SanitizeHeaders(writers MetricsWriterList) MetricsWriterList {
+// SanitizeHeaders sanitizes the headers of the given MetricsWriterList.
+func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWriterList {
 	var lastHeader string
 	for _, writer := range writers {
 		if len(writer.stores) > 0 {
 			for i, header := range writer.stores[0].headers {
+				// If the requested content type was proto-based, replace the type with "gauge", as "info" and "statesets" are not recognized by Prometheus' protobuf machinery.
+				if strings.HasPrefix(contentType, expfmt.ProtoType) &&
+					strings.HasPrefix(header, "# HELP") &&
+					(strings.HasSuffix(header, " "+string(metric.Info)) || strings.HasSuffix(header, " "+string(metric.StateSet))) {
+					typeStringWithoutTypePaddedIndex := strings.LastIndex(header, " ")
+					typeStringWithoutType := header[:typeStringWithoutTypePaddedIndex]
+					writer.stores[0].headers[i] = typeStringWithoutType + " " + string(metric.Gauge)
+				}
+				// Removes duplicate headers from the given MetricsWriterList for the same family (generated through CRS).
+				// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
 				if header == lastHeader {
 					writer.stores[0].headers[i] = ""
 				} else {
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -106,17 +106,16 @@ func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWrite
 				} else if strings.HasPrefix(header, "# HELP") {
 					lastHeader = header
 
-					// If the requested content type was proto-based, replace the type with "gauge", as "info" and "statesets" are not recognized by Prometheus' protobuf machinery,
-					// else replace them by their respective string representations.
-					if strings.HasPrefix(contentType, expfmt.ProtoType) &&
-						(strings.HasSuffix(header, metric.InfoN.NString()) || strings.HasSuffix(header, metric.StateSetN.NString())) {
-						writer.stores[0].headers[i] = header[:len(header)-1] + string(metric.Gauge)
-					}
-
-					// Replace all remaining type enums with their string representations.
-					n := int(header[len(header)-1]) - '0'
-					if n >= 0 && n < len(metric.TypeNMap) {
-						writer.stores[0].headers[i] = header[:len(header)-1] + string(metric.TypeNMap[metric.TypeN(n)])
+					// If the requested content type was proto-based, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
+					if strings.HasPrefix(contentType, expfmt.ProtoType) {
+						infoTypeString := metric.Info.String()
+						stateSetTypeString := metric.StateSet.String()
+						if strings.HasSuffix(header, infoTypeString) {
+							writer.stores[0].headers[i] = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
+						}
+						if strings.HasSuffix(header, stateSetTypeString) {
+							writer.stores[0].headers[i] = header[:len(header)-len(stateSetTypeString)] + string(metric.Gauge)
+						}
 					}
 				}
 			}
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -100,7 +100,7 @@ func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWrite
 				// Skip this step if we encounter a repeated header, as it will be removed.
 				if header != lastHeader && strings.HasPrefix(header, "# HELP") {
 
-					// If the requested content type was proto-based, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
+					// If the requested content type was proto-based (such as FmtProtoDelim, FmtProtoText, or FmtProtoCompact), replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
 					if strings.HasPrefix(contentType, expfmt.ProtoType) {
 						infoTypeString := string(metric.Info)
 						stateSetTypeString := string(metric.StateSet)
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -101,25 +101,45 @@ func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWrite
 
 				// Removes duplicate headers from the given MetricsWriterList for the same family (generated through CRS).
 				// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
-				if header == lastHeader {
-					writer.stores[0].headers[i] = ""
-				} else if strings.HasPrefix(header, "# HELP") {
-					lastHeader = header
+				// Skip this step if we encounter a repeated header, as it will be removed.
+				if header != lastHeader && strings.HasPrefix(header, "# HELP") {
 
 					// If the requested content type was proto-based, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
 					if strings.HasPrefix(contentType, expfmt.ProtoType) {
 						infoTypeString := metric.Info.String()
 						stateSetTypeString := metric.StateSet.String()
 						if strings.HasSuffix(header, infoTypeString) {
-							writer.stores[0].headers[i] = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
+							header = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
+							writer.stores[0].headers[i] = header
 						}
 						if strings.HasSuffix(header, stateSetTypeString) {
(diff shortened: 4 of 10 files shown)


rexagod:
Need to test this.

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,13 +87,22 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
-// SanitizeHeaders removes duplicate headers from the given MetricsWriterList for the same family (generated through CRS).
-// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
-func SanitizeHeaders(writers MetricsWriterList) MetricsWriterList {
+// SanitizeHeaders sanitizes the headers of the given MetricsWriterList.
+func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWriterList {
 	var lastHeader string
 	for _, writer := range writers {
 		if len(writer.stores) > 0 {
 			for i, header := range writer.stores[0].headers {
+				// If the requested content type was proto-based, replace the type with "gauge", as "info" and "statesets" are not recognized by Prometheus' protobuf machinery.
```
CatherineF-dev, line 96: Use a good function name to replace comments.

Wrap these into a function.
```
				if strings.HasPrefix(contentType, expfmt.ProtoFmt) &&
					strings.HasPrefix(header, "# TYPE") &&
					(strings.HasSuffix(header, " "+string(metric.Info)+"\n") || strings.HasSuffix(header, " "+string(metric.StateSet)+"\n")) {
					typeStringWithoutTypePaddedIndex := strings.LastIndex(header, " ")
					typeStringWithoutType := header[:typeStringWithoutTypePaddedIndex]
					writer.stores[0].headers[i] = typeStringWithoutType + string(metric.Gauge)
				}
```

rexagod:
Verified using the following configurations. Error logs were reported without this patch, but none when this was applied.

<details>
<summary>Details</summary>


### Prometheus error log
```
ts=2023-12-13T16:01:02.547Z caller=scrape.go:1418 level=debug component="scrape manager" scrape_pool=kube-state-metrics target=http://127.0.0.1:8080/metrics msg="Append failed" err="invalid metric type \"info\""
```
### KSM CRS Configuration
```
	  ...
      metrics:
        - name: "foo_info"
          help: "foo_help"
          each:
            type: Info
            info:
              path: [metadata]
              labelsFromPath:
                "name": [name]
```
### Prometheus Configuration
```
global:
  evaluation_interval: 15s
  scrape_interval:     15s
  scrape_protocols: [PrometheusProto]
scrape_configs:
  - job_name: 'kube-state-metrics'
    static_configs:
      - targets: ['127.0.0.1:8080']
```

### Generated CRS Metric (from Prometheus UI)

```
kube_customresource_foo_info{customresource_group="contoso.com", customresource_kind="MyPlatform", customresource_version="v1alpha1", instance="127.0.0.1:8080", job="kube-state-metrics", name="test-dotnet-app"} 1
```

</details>

pkg/metricshandler/metrics_handler.go
```diff
@@ -208,7 +209,7 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 
-	m.metricsWriters = metricsstore.SanitizeHeaders(m.metricsWriters)
+	m.metricsWriters = metricsstore.SanitizeHeaders(string(gotContentType), m.metricsWriters)
```
mrueg, line 212: Can we have a boolean here "isOpenMetricsFmt" instead? I think on txt format we would want to fall back as well.

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,13 +87,22 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+				if strings.HasPrefix(contentType, expfmt.ProtoFmt) &&
+					strings.HasPrefix(header, "# HELP") &&
```
mrueg, line 98: Have you considered using a regular expression here?

pkg/metrics_store/metrics_writer.go
rexagod: Wrapping both of the behaviors defined here might impact the readability. Additionally, for eg., if those functions use a helper function, that would mean context switching between three different functions to figure out what's going on. In contrast, the function length doesn't exceed the one enforced by the linters, but we could consider that once that happens.

I think we'd want to include a rule for it in the `golangci` config if we really want to consistently check for patterns such as this.

pkg/metrics_store/metrics_writer.go
rexagod: I was on the fence between using a regex and the in-built `strings` for this, since the latter seemed more readable, and the exposition format isn't prone to change. But LMK and I'll revert.

pkg/metricshandler/metrics_handler.go
rexagod: Enabling native histograms sets the `scrape_protocols` to `[ PrometheusProto, OpenMetricsText1.0.0, OpenMetricsText0.0.1, PrometheusText0.0.4 ]`. [`ProtoFmt`](https://github.com/prometheus/common/blob/658f673b7580e72a7d0c0fac26a67858b782fbd5/expfmt/expfmt.go#L25) should catch all [`Proto` formats](https://github.com/prometheus/prometheus/blob/69abd6d9f65a6358025461510237366f1d939c70/config/config.go#L451), including any [`Proto`-text ones](https://github.com/prometheus/common/blob/658f673b7580e72a7d0c0fac26a67858b782fbd5/expfmt/expfmt.go#L33-L35) as well.

I have restricted this behavior to "only Prometheus-proto" rather than "anything other than OpenMetricsText" so we can be sure if anything other than these two starts causing problems for us in the future, and we don't end up `s/info|statesets/gauge` for it since that may be due to an entirely different bug.

pkg/metrics_store/metrics_writer.go
mrueg: I was mainly wondering how costly the operation is. Might be worth looking at our benchmarks in CI

rexagod: Benchmarking `regex` v/s `strings` showed the following results.

### `regex` snippet

```go
if strings.HasPrefix(contentType, expfmt.ProtoType) {
	pattern := `^# HELP.*\n# TYPE.* (info|stateset)`
	matched, err := regexp.MatchString(pattern, header)

	// Fail silently.
	if err == nil && matched {
		typeStringWithoutTypePaddedIndex := strings.LastIndex(header, " ")
		typeStringWithoutType := header[:typeStringWithoutTypePaddedIndex]
		writer.stores[0].headers[i] = typeStringWithoutType + " " + string(metric.Gauge)
	}
}
```

### `strings` snippet

```go
if strings.HasPrefix(contentType, expfmt.ProtoType) &&
	strings.HasPrefix(header, "# HELP") &&
	(strings.HasSuffix(header, " "+string(metric.Info)) || strings.HasSuffix(header, " "+string(metric.StateSet))) {
	typeStringWithoutTypePaddedIndex := strings.LastIndex(header, " ")
	typeStringWithoutType := header[:typeStringWithoutTypePaddedIndex]
	writer.stores[0].headers[i] = typeStringWithoutType + " " + string(metric.Gauge)
}
```

### Benchmarks

`strings` seems to be ~95% faster than `regex`.

```
┌[rexagod@nebuchadnezzar] [/dev/ttys003] [2248 ⚡]
└[~/repositories/oss/kube-state-metrics]> benchstat regex.txt strings.txt
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │   regex.txt    │             strings.txt             │
                                                  │     sec/op     │   sec/op     vs base                │
SanitizeHeaders/text-format_unique_headers-10          559.2µ ± 0%   560.7µ ± 0%   +0.27% (p=0.000 n=10)
SanitizeHeaders/text-format_duplicate_headers-10       528.2µ ± 0%   528.6µ ± 1%   +0.08% (p=0.035 n=10)
SanitizeHeaders/proto-format_unique_headers-10       388.639m ± 2%   3.981m ± 0%  -98.98% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   356906.4µ ± 0%   590.2µ ± 0%  -99.83% (p=0.000 n=10)
geomean                                                14.23m        913.5µ       -93.58%

                                                  │   regex.txt    │               strings.txt               │
                                                  │      B/op      │    B/op     vs base                     │
SanitizeHeaders/text-format_unique_headers-10         0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
SanitizeHeaders/text-format_duplicate_headers-10      0.000 ± 0%     0.000 ± 0%         ~ (p=1.000 n=10) ¹
SanitizeHeaders/proto-format_unique_headers-10      737.8Mi ± 0%     0.0Mi ± 0%  -100.00% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   733.2Mi ± 0%     0.0Mi ± 0%  -100.00% (p=0.000 n=10)
geomean                                                          ²               ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean

                                                  │   regex.txt   │               strings.txt                │
                                                  │   allocs/op   │  allocs/op   vs base                     │
SanitizeHeaders/text-format_unique_headers-10        0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
SanitizeHeaders/text-format_duplicate_headers-10     0.000 ± 0%      0.000 ± 0%         ~ (p=1.000 n=10) ¹
SanitizeHeaders/proto-format_unique_headers-10      4.101M ± 0%     0.000M ± 0%  -100.00% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   4.100M ± 0%     0.000M ± 0%  -100.00% (p=0.000 n=10)
geomean                                                         ²                ?                       ² ³
¹ all samples are equal
² summaries must be >0 to compute geomean
³ ratios must be >0 to compute geomean
```

The benchmarking suite used is present [here](https://github.com/kubernetes/kube-state-metrics/pull/2270/commits/640960336ac58d655a83f81ccbe789bbced58c3f#diff-a2fba22b7063f6a5f98f7c677327f9d87bc5f07d84cd567009a1f41ce6799adeR268).

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,13 +87,22 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
-// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
-func SanitizeHeaders(writers MetricsWriterList) MetricsWriterList {
+// SanitizeHeaders sanitizes the headers of the given MetricsWriterList.
+func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWriterList {
 	var lastHeader string
 	for _, writer := range writers {
 		if len(writer.stores) > 0 {
 			for i, header := range writer.stores[0].headers {
+				// If the requested content type was proto-based, replace the type with "gauge", as "info" and "statesets" are not recognized by Prometheus' protobuf machinery.
+				if strings.HasPrefix(contentType, expfmt.ProtoType) &&
```
rexagod, line 97: FYI I changed this from `ProtoFmt` to `ProtoType`.

```go
	ProtoType                = `application/vnd.google.protobuf`
	ProtoFmt                 = ProtoType + "; proto=" + ProtoProtocol + ";"
```

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,13 +87,22 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+					strings.HasPrefix(header, "# HELP") &&
```
dgrisonnet, line 98: Since we control the header value in https://github.com/kubernetes/kube-state-metrics/blob/0bc4b19ac5cf7039fa877a43a85646367f3d63a4/pkg/metric_generator/generator.go#L79, we could just check for the suffixes, we shouldn't need to check if `# HELP` is there

dgrisonnet:
How about a separate approach.

The current storage is built based on the assumption that the returned types never changes (which used to hold true until recently). Now that we support both the Prometheus text format and the OpenMetrics format, this isn't valid anymore.

The current approach you have is expensive because we do a search in the header for each metric to replace the type.

How about we just don't add it to the generated header stored in the cache anymore (from https://github.com/kubernetes/kube-state-metrics/blob/0bc4b19ac5cf7039fa877a43a85646367f3d63a4/pkg/metric_generator/generator.go#L93), and rather add it close to when the format is negotiated? Maybe in a new function before SanitizeHeader. We would then just need to surface the metrics family type there somehow and then make a decision on the type added to the header depending on the format.

rexagod:
A couple of notes around the latest patch.

* Manipulate headers only when they are not duplicates (since they will be removed before any operations).
* The earlier approach used `strings.Has*` and `strings.LastIndex` functions, with respective complexities of O(1) and O(n). This patch removes the dependency on the latter so the whole operation is lazily run in constant time.

### Benchmarks after the first optimization (comparing the first commit v/s second one on this PR)

```
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │   PUSHED    │               STAGED                │
                                                  │   sec/op    │   sec/op     vs base                │
SanitizeHeaders/text-format_unique_headers-10       561.4µ ± 0%   561.2µ ± 0%        ~ (p=0.811 n=10)
SanitizeHeaders/text-format_duplicate_headers-10    529.1µ ± 0%   281.6µ ± 0%  -46.77% (p=0.000 n=10)
SanitizeHeaders/proto-format_unique_headers-10      4.047m ± 0%   3.985m ± 0%   -1.54% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   591.4µ ± 0%   279.1µ ± 0%  -52.81% (p=0.000 n=10)
geomean                                             918.2µ        647.5µ       -29.48%
```

### Benchmarks after the both optimizations (comparing the first commit v/s second one on this PR)

```
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │    PUSHED    │                 STAGED                 │
                                                  │    sec/op    │    sec/op      vs base                 │
SanitizeHeaders/text-format_unique_headers-10       561.39µ ± 0%    31.79µ ±  0%   -94.34% (p=0.000 n=10)
SanitizeHeaders/text-format_duplicate_headers-10    529.06µ ± 0%    31.82µ ±  0%   -93.98% (p=0.000 n=10)
SanitizeHeaders/proto-format_unique_headers-10       4.047m ± 0%   34.068m ± 65%  +741.84% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   591.35µ ± 0%    96.74µ ± 33%   -83.64% (p=0.000 n=10)
geomean                                              918.2µ         240.3µ         -73.83%

```

> How about we just don't add it to the generated header [...]

I'm doing something similar here, where instead of entirely removing the multi-length type strings, we append their enum values instead (unit-length). This means we won't need to search for the last " " anymore to trim that off, while still retaining the type data within the header required to convert the header back to the expected representation.

However, either way (removing the type entirely, or appending a known-length suffix) we see that the tests need to be reformatted dramatically (lots of repeated copy-pasting), so I was wondering I'll hold off on doing that until we have an approach that we can go forward with.

rexagod:
Posting baseline benchmarks, forgot to do this earlier.

* Baseline (same as `main`) v/s changes in this PR.
```
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │   BASELINE   │                   HEAD                    │
                                                  │    sec/op    │     sec/op      vs base                   │
SanitizeHeaders/text-format_unique_headers-10       251.62µ ± 0%     31.17µ ±  0%     -87.61% (p=0.000 n=10)
SanitizeHeaders/text-format_duplicate_headers-10    248.52µ ± 4%     31.22µ ±  0%     -87.44% (p=0.000 n=10)
SanitizeHeaders/proto-format_unique_headers-10       250.5µ ± 0%   34093.4µ ± 49%  +13509.75% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10   247.63µ ± 1%     95.04µ ± 41%     -61.62% (p=0.000 n=10)
geomean                                              249.6µ          237.0µ            -5.05%
```

* `main` + skip empty headers v/s changes in this PR (`main` outperforms this PR when protobuf sanitization is the only thing different in this PR).

```
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │   BASELINE   │                   HEAD                    │
                                                  │    sec/op    │     sec/op      vs base                   │
SanitizeHeaders/text-format_unique_headers-10       249.42µ ± 0%     31.14µ ±  0%     -87.51% (p=0.000 n=10)
SanitizeHeaders/text-format_duplicate_headers-10     42.73µ ± 0%     31.12µ ±  0%     -27.16% (p=0.000 n=10)
SanitizeHeaders/proto-format_unique_headers-10       251.1µ ± 0%   36649.4µ ± 64%  +14492.68% (p=0.000 n=10)
SanitizeHeaders/proto-format_duplicate_headers-10    42.74µ ± 0%     93.45µ ± 38%    +118.64% (p=0.000 n=10)
geomean                                              103.4µ          240.0µ          +132.09%
```

To summarize, we optimized the flow with the [first one](https://github.com/kubernetes/kube-state-metrics/pull/2270/files#diff-60450a33adea08c953656dd1e78a80e9f3b279bbc7656dedf31fd1a0c7fc1196R94-R97) quite a bit (to the point where it negated the 1.3x more impact on `SanitizeHeaders`, and gave us a 0.05x boost), but if both `main` and this PR have that in common, with the only feature differentiating this PR being the `proto` sanitization, that impacts the performance compared to the baseline.

Please note that these are the stats compared to the baseline, and not the `regex` implementation (that's present [here](https://github.com/kubernetes/kube-state-metrics/pull/2270#discussion_r1427005713)).

rexagod:
Added in a couple of O(1) checks from the first commit back in. The benchmarks may differ slightly, but the overall inference should be the same.

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
+}
+
+// NString returns the string representation of the metric type as an integer.
+func (i TypeN) NString() string {
+	return strconv.Itoa(int(i))
+}
+
+// Supported metric types.
+const (
+	// GaugeN defines a OpenMetrics gauge.
```
mrueg, line 58:
```suggestion
	// GaugeN defines an OpenMetrics gauge.
```

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
+	GaugeN TypeN = iota
+
+	// InfoN defines an OpenMetrics info.
+	InfoN
+
+	// StateSetN defines an OpenMetrics stateset.
+	StateSetN
+
+	// CounterN defines a OpenMetrics counter.
```
mrueg, line 67:
```suggestion
	// CounterN defines an OpenMetrics counter.
```

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
+	CounterN
+)
 
-// StateSet defines an OpenMetrics stateset.
-var StateSet Type = "stateset"
+// Supported metric types.
+var (
+
+	// Gauge defines a OpenMetrics gauge.
```
mrueg, line 74:
```suggestion
	// Gauge defines an OpenMetrics gauge.
```

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
+	Gauge Type = "gauge"
+
+	// Info defines an OpenMetrics info.
+	Info Type = "info"
+
+	// StateSet defines an OpenMetrics stateset.
+	StateSet Type = "stateset"
+
+	// Counter defines a OpenMetrics counter.
```
mrueg, line 83:
```suggestion
	// Counter defines an OpenMetrics counter.
```

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
-// Gauge defines a OpenMetrics gauge.
-var Gauge Type = "gauge"
+// TypeN represents the type of the metric as an integer.
+type TypeN int
 
-// Info defines an OpenMetrics info.
-var Info Type = "info"
+// String returns the string representation of the metric type.
+func (i TypeN) String() Type {
+	return MetricTypeNMap[i]
```
mrueg, line 48: Shouldn't String() return a String() and not a Type?

pkg/metricshandler/metrics_handler.go
mrueg: Thanks for the clarification! I misunderstood, that makes sense to me. :)

pkg/metric/metric.go
```diff
@@ -37,21 +37,68 @@ var (
...
+
+	// StateSet defines an OpenMetrics stateset.
+	StateSet Type = "stateset"
+
+	// Counter defines a OpenMetrics counter.
+	Counter Type = "counter"
 
-// Counter defines a OpenMetrics counter.
-var Counter Type = "counter"
+	// MetricTypeNMap is a map of MetricTypeN to MetricType.
```
mrueg, line 86: I find this very confusing and not intuitive to use. What's the intent behind having two maps here?

mrueg:
Thanks for running those benchmarks, this is super helpful to get confidence.

pkg/metric/metric.go
rexagod: Ah, right.

pkg/metric/metric.go
rexagod: I introduced these maps to provide a way to convert `types` to and back from their respective `typeN` representations.

rexagod:
Good for reviews.

internal/store/certificatesigningrequest_test.go
```diff
@@ -29,13 +29,13 @@ import (
 func TestCsrStore(t *testing.T) {
 	const metadata = `
 		# HELP kube_certificatesigningrequest_labels [STABLE] Kubernetes labels converted to Prometheus labels.
-		# TYPE kube_certificatesigningrequest_labels gauge
+		# TYPE kube_certificatesigningrequest_labels 0
```
mrueg, line 32: This seems unusual to me. Is this supported in prometheus?

mrueg: I understand that you're doing this only in the test. I think it would be helpful to have a conversion function as well and keep the metadata enduser friendly (e.g. I copy and paste something into my testbench).

rexagod: +1, but AFAICT we can only deduce the `content-type` once the request is made (`ServeHTTP`), which is where `SanitizeHeaders` is being invoked from currently, for this reason.

Moving the call further down in the stack, say, to where the headers are being generated wouldn't provide us enough information (`content-type`) to "sanitize" the headers.

To circumvent this, the current workflow is:
* append the `typeN` representation of the `metric.type` to the `# TYPE` header metadata,
* once a request is made, deduce if it was `expfmt.ProtoType`-based,
* * if so, `s/(infoN|statesetN)/gaugeN`
* finally, `s/xN/x` (convert `typeN` back to `metric.type`)

rexagod: That being said, the whole `s/gauge/0` etc., in the tests seems odd indeed, but I'm not quite sure if there's a better way around this.

rexagod: I just realized that we really don't need to search for anything, as opposed to my initial [train of thought](https://github.com/kubernetes/kube-state-metrics/pull/2270#issuecomment-1857791232). Let me patch this again.

rexagod:
/hold

Need to add a test too, will push shortly.

pkg/metrics_store/metrics_writer.go
```diff
@@ -101,25 +101,45 @@ func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWrite
...
-				if header == lastHeader {
-					writer.stores[0].headers[i] = ""
-				} else if strings.HasPrefix(header, "# HELP") {
-					lastHeader = header
+				// Skip this step if we encounter a repeated header, as it will be removed.
+				if header != lastHeader && strings.HasPrefix(header, "# HELP") {
 
 					// If the requested content type was proto-based, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
 					if strings.HasPrefix(contentType, expfmt.ProtoType) {
 						infoTypeString := metric.Info.String()
```
mrueg, line 109: looks like you only need those lengths here, it might be easier to store them in a global var? (Not sure if the compiler will optimize this)

mrueg: Ah nvm, you're comparing as well. Should be all fine.

mrueg:
/hold
(for @rexagod to unhold)

/lgtm

Changeset looks good, did you test this with your local setup that scrapes with protobuf?

rexagod:
Pasting the requested information below.

<details>
<summary>Testing locally on a Prometheus Protobuf setup.</summary>

Screenshot 2023-12-20 at 17 05 11
Screenshot 2023-12-20 at 17 07 14

</details>

<details>
<summary>Benchmarking this PR against `main`.</summary>

```
goos: darwin
goarch: arm64
pkg: k8s.io/kube-state-metrics/v2/pkg/metrics_store
                                                  │    BASELINE     │                 HEAD                  │
                                                  │     sec/op      │    sec/op     vs base                 │
SanitizeHeaders/text-format_unique_headers-10           255.0µ ± 3%   1225.8µ ± 3%  +380.70% (p=0.000 n=10)
SanitizeHeaders/text-format_duplicate_headers-10    258262.50n ± 4%    10.87n ± 0%  -100.00% (p=0.000 n=10)
SanitizeHeaders/proto-format_unique_headers-10                         2.071m ± 0%
SanitizeHeaders/proto-format_duplicate_headers-10                      18.50n ± 1%
geomean                                                 256.6µ         4.754µ        -98.58%
```

</details>

rexagod:
/cc @dgrisonnet

(for another review before we merge)

pkg/customresourcestate/config.go
```diff
@@ -206,3 +215,14 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 	}
 	return fn, nil
 }
+
+// configOverrides applies overrides to the configuration.
+func configOverrides(config *Metrics) {
+	for i := range config.Spec.Resources {
+		for j := range config.Spec.Resources[i].Metrics {
+
+			// Override the metric type to lowercase, so the internals have a single source of truth for metric type definitions.
```
dgrisonnet, line 224: Could you perhaps add a comment on why this is necessary, just so we remember it in the future?

pkg/customresourcestate/registry_factory.go
```diff
@@ -152,8 +152,8 @@ type compiledMetric interface {
 
 // newCompiledMetric returns a compiledMetric depending on the given metric type.
 func newCompiledMetric(m Metric) (compiledMetric, error) {
-	switch m.Type {
-	case MetricTypeGauge:
+	switch metric.Type(strings.ToLower(string(m.Type))) {
```
dgrisonnet, line 155: do we really need to convert to lowercase here since the config should've already been overriden?

pkg/metric/metric.go
```diff
@@ -37,21 +37,29 @@ var (
...
 )
 
-// Type represents the type of a metric e.g. a counter. See
-// https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
+// Type represents the type of the metric. See https://github.com/OpenObservability/OpenMetrics/blob/main/specification/OpenMetrics.md#metric-types.
 type Type string
 
-// Gauge defines a OpenMetrics gauge.
-var Gauge Type = "gauge"
+// String returns the string representation of the metric type.
```
mrueg, line 43: Are we able to remove this one and just use built-in string(type) conversion?

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,20 +87,59 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+				// Skip this step if we encounter a repeated header, as it will be removed.
+				if header != lastHeader && strings.HasPrefix(header, "# HELP") {
+
+					// If the requested content type was proto-based, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
+					if strings.HasPrefix(contentType, expfmt.ProtoType) {
+						infoTypeString := metric.Info.String()
+						stateSetTypeString := metric.StateSet.String()
+						if strings.HasSuffix(header, infoTypeString) {
+							header = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
+							writer.stores[0].headers[i] = header
```
dgrisonnet, line 113: I am not a big fan of modifying the cache here. Ideally, the cache should only be modified when generating metrics after getting an event from an informer.

Also, we shouldn't rely on the assumption that only one format will be served at all time. We never know if a user might have two different backends scraping kube-state-metrics with different formats, so we would just endlessly modify the cache. IMO, it would be better to just remove the type from the cache and generate it here.

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,20 +87,59 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+	// Remove all empty headers.
+	for _, writer := range writers {
+		if len(writer.stores) > 0 {
+			for i := len(writer.stores[0].headers) - 1; i >= 0; i-- {
+				if writer.stores[0].headers[i] == "" {
+					writer.stores[0].headers = append(writer.stores[0].headers[:i], writer.stores[0].headers[i+1:]...)
+				}
+			}
+		}
+	}
```
dgrisonnet, line 142: why do we do that? The order of the headers is very important so modifying it doesn't like a good idea

pkg/metrics_store/metrics_writer.go
```diff
@@ -82,20 +87,59 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
 				}
+
+				// Update the last header.
+				lastHeader = header
 			}
 		}
 	}
+
+	// Remove all empty headers.
+	for _, writer := range writers {
```
mrueg, line 134: This sounds like something we should optimize by merging it into the loop above (at least the writers to avoid looping through that twice)

pkg/metrics_store/metrics_writer.go
rexagod: `G**` resolutions can [result in duplicated families](https://github.com/rexagod/kube-state-metrics/blob/main/pkg/metrics_store/metrics_writer.go#L95). I noticed that in such cases we modified the headers to an empty string, but since that is not a valid metric byte line, I added some functionality to strip those and enforced that using some [tests here](https://github.com/kubernetes/kube-state-metrics/pull/2270/files#diff-a2fba22b7063f6a5f98f7c677327f9d87bc5f07d84cd567009a1f41ce6799adeR298-R315).

pkg/metrics_store/metrics_writer.go
rexagod: +1, everything within `ServeHTTP` will be computed for all requests made to it (from one or more metric backends).

Unfortunately, owing the fact that we do not operate on structured metrics in this context, but rather textual representations, I'm not sure if we have a type-independent way of determining the type from the data we receive, other than parsing that text and looking for certain substrings to determine the sanitization behavior.

Quoting https://github.com/kubernetes/kube-state-metrics/pull/2270#discussion_r1432015068,

> +1, but AFAICT we can only deduce the content-type once the request is made (ServeHTTP, and the headers are already generated), which is where SanitizeHeaders is being invoked from currently, for this reason.

> Moving the call further down in the stack, say, to where the headers are being generated wouldn't provide us enough information (content-type) to "sanitize" the headers.

That being said, wouldn't the cache still be modified, since we'd be appending the `# TYPE` metadata to the headers' cache in the case we are able to do so, as compared to editing (`s/typeA/typeB`) it currently?

dgrisonnet: I looked into the code again and I think we could separate the type from the generated headers in the metrics stores: https://github.com/kubernetes/kube-state-metrics/blob/90bd24f27c618ce916db4cd9da34d854d49db5b6/pkg/metrics_store/metrics_store.go#L34-L42.

We could create a new list of types for the metrics families in the metrics store and similarly to how we extract and generate the headers from the metrics families, we could extract the types https://github.com/kubernetes/kube-state-metrics/blob/90bd24f27c618ce916db4cd9da34d854d49db5b6/internal/store/builder.go#L512 and store them in the metrics store.

That will not be a problem when writing the headers because the type is the last element of the headers: https://github.com/kubernetes/kube-state-metrics/blob/90bd24f27c618ce916db4cd9da34d854d49db5b6/pkg/metric_generator/generator.go#L90-L95 so we would just need to append it to the existing headers at runtime.

pkg/metrics_store/metrics_writer.go
dgrisonnet: That sounds like a bug where we don't account for the version anywhere besides in the `customresource_version` label. So we create identical stores but for different resources no?

Do you mind to take that to a separate PR? It sounds like a different issue than the original one.

pkg/metrics_store/metrics_writer.go
rexagod: > That being said, wouldn't the cache still be modified, since we'd be appending the # TYPE metadata to the headers' cache in the case we are able to do so, as compared to editing (s/typeA/typeB) it, that is done currently?

Sorry for dragging this on further, but I'm not able to make out how we will circumvent the original concern, i.e., [modifying the cache at runtime](https://github.com/kubernetes/kube-state-metrics/pull/2270#discussion_r1432727642) ([`ServeHTTP`](https://github.com/kubernetes/kube-state-metrics/pull/2270#discussion_r1451795203)), since appending to the header metadata would still fall under that category, and would be similar to what we are doing currently: https://github.com/kubernetes/kube-state-metrics/pull/2270/files#diff-60450a33adea08c953656dd1e78a80e9f3b279bbc7656dedf31fd1a0c7fc1196R108-R109 would effectively become `writer.stores[0].headers[i] += string(metric.Gauge)`, nonetheless, IIUC, we'd still be modifying the header cache at runtime.

I believe we'd still have to deduce the `content-type` at runtime, and modify cache (at runtime) to satisfy the payload expectations. I've provided a comparison of both approaches below.

- [proposed] fetching the metric type from the store and appending the type based on the current `context-type`, in contrast to,
- [current] modifying the last `n` type bytes from the `#TYPE` header based on the current `context-type`.

Although both approaches modify cache at runtime, no fetch operations would be necessary in the latter since we maintain the default header value in the `#TYPE` header (which also translates to less than or equal to the maximum number of appends being done in the previous approach).

PS. PLMK if I missed or misunderstood anything in interpreting your comment above.

dgrisonnet: The type of the metric is not dynamic, it will always be either Gauge, Counter, StateSet or Info. What is dynamic is the type that we will serve to the server depending on the content-type. The one we store doesn't have to change.

My suggestion would be to:
1. remove the type from the header in cache: https://github.com/kubernetes/kube-state-metrics/blob/90bd24f27c618ce916db4cd9da34d854d49db5b6/pkg/metric_generator/generator.go#L93
2. add a slice of types to the metrics store
3. update WriteAll to add the correct type dynamically: https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metrics_store/metrics_writer.go#L61-L71

for example something like:
```diff
 // MetricsWriterList represent a list of MetricsWriter
@@ -59,6 +63,18 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 	}
 
 	for i, help := range m.stores[0].headers {
+
+		if strings.HasPrefix(contentType, expfmt.ProtoType) {
+			switch m.stores[0].types[i] {
+			case metric.Info, metric.StateSet:
+				help += string(metric.Gauge)
+			default:
+				help += string(m.stores[0].types[i])
+			}
+		} else {
+			help += string(m.stores[0].types[i])
+		}
+
 		if help != "" && help != "\n" {
 			help += "\n"
 		}
```

rexagod: Resolved over call, I'll open up a new PR to move types to their own cache, since that should be cleaner.

rexagod:
@dgrisonnet I believe all concerns have been addressed, can you PTAL once more?

schahal:
Hi folks, is this change targeted to be in the release-2.11 cut (which has the k8s 1.28 support of the client libs)?

dgrisonnet:
Yes @schahal

dgrisonnet:
@rexagod didn't we discuss potentially removing OpenMetrics format support?

rexagod:
@dgrisonnet Yes, I believe we talked about dropping metric-standard lock-ins, and implementing a one-fits-all mechanism (probably on top of gauges) that allows us to generate an exposition format that is ingestible by most metric backends, if not all.

rexagod:
(friendly ping @mrueg @dgrisonnet to /lgtm)

remram44:
2.11.0 introduced a regression, as metrics include `stateset` entries which are refused by Prometheus. This worked correctly on 2.10 so I'm stuck until this MR is merged.

pkg/metricshandler/metrics_handler.go
```diff
@@ -178,15 +178,16 @@ func (m *MetricsHandler) Run(ctx context.Context) error {
...
+// ServeHTTP implements the http.Handler interface. It writes all generated metrics to the response body.
+// Note that all operations defined within this procedure are performed at every request.
 func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	m.mtx.RLock()
 	defer m.mtx.RUnlock()
 	resHeader := w.Header()
 	var writer io.Writer = w
 
 	contentType := expfmt.NegotiateIncludingOpenMetrics(r.Header)
+	gotContentType := contentType
```
dgrisonnet, line 190: Why do we use a copy of the content type here? If the content type of the request was OpenMetrics and we fallback to Text, I would expect us to propagate Text and not OpenMetrics since that is the format we are going to serve.

rexagod: It seems out of all the `content-type`s (as mentioned below):
- `FmtUnknown`
- `FmtText`
- `FmtProtoDelim`
- `FmtProtoText`
- `FmtProtoCompact`
- `FmtOpenMetrics_1_0_0`
- `FmtOpenMetrics_0_0_1`

We fallback to `FmtText` if the case is anything other than the last two:
```go
if contentType != expfmt.FmtOpenMetrics_1_0_0 && contentType != expfmt.FmtOpenMetrics_0_0_1 {
	contentType = expfmt.FmtText
}
```

But IIUC, since we now have a workaround to handle other `protobuf`-based types as well, we can drop this fallback behavior.

rexagod: Sent a patch.

dgrisonnet: My original comment was confusing sorry. I meant if the content type of the request is **anything other** (I wrote the opposite above sry) than OpenMetrics, we would want to propagate the new FmtProtoText down to the parser, when what you did before was to propagate the content type of the request that you stored in `gotContentType`.

So my suggestion was just to propagate `contentType` directly to the parser.

rexagod:
It should, yes.

LaikaN57:
> Hi folks, is this change targeted to be in the release-2.11 cut (which has the k8s 1.28 support of the client libs)?

@dgrisonnet Looks like this fix got merged in too late for v2.11.0. How do we feel about cutting a new release soon™ to get people unblocked? (Thanks in advance.)

LaikaN57:
> > Hi folks, is this change targeted to be in the release-2.11 cut (which has the k8s 1.28 support of the client libs)?
> 
> @dgrisonnet Looks like this fix got merged in too late for v2.11.0. How do we feel about cutting a new release soon™ to get people unblocked? (Thanks in advance.)

@rexagod bump ^

rexagod:
I'll put https://github.com/kubernetes/kube-state-metrics/pull/2335#issuecomment-2021195000 up for review today.

rexagod:
Sharing here for visibility: https://github.com/kubernetes/kube-state-metrics/pull/2335#issuecomment-2032148989.

seruman:
hi @rexagod,

i'm probably missing something but should not it also fallback to gauge on `text/plain`?

with `--enable-feature=native-histograms` prometheus (2.52.0) sends;
```
Accept: application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.5,application/openmetrics-text;version=1.0.0;q=0.4,application/openmetrics-text;version=0.0.1;q=0.3,text/plain;version=0.0.4;q=0.2,*/*;q=0.1
```

but handler falls-back to `text/plain` here;

https://github.com/kubernetes/kube-state-metrics/blob/d3f0c1853f11b387d7dc2c89bd5e52760f65079c/pkg/metricshandler/metrics_handler.go#L200-L206

thus exporter responds with type `info` that makes prometheus unhappy ; `invalid metric type "info"`.

PR #2104, opened 2023-06-27
refactor: move discovery and util under package app/customresource to make packages highly cohesive

CatherineF-dev:
**What this PR does / why we need it**:

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)*

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2103

PR #2271, opened 2023-12-13
refactor: make ResolveGVKToGVKPs shorter and use if-else

CatherineF-dev:
**What this PR does / why we need it**: refactor  ResolveGVKToGVKPs a little

Fixes #


rexagod:
I feel changes like https://github.com/kubernetes/kube-state-metrics/pull/2104 and the one here should be acted upon only if the linter finds out there's something wrong with the current way of implementing the targeted behaviors.

If there is something wrong that the linter misses to catch, consider adding a rule or a lint package for in the `golangci-lint` configuration, since keeping lookout for and catching every such instance manually will lead to almost greater degrees of bikeshedding.

PR #2273, opened 2023-12-13
[PDNM] chore: comment bench results on PR

rexagod:
**What this PR does / why we need it**: Comment benchmark outputs on the PR, for better visibility, and to make sure we don't miss out on catching potential bottlenecks.

**How does this change affect the cardinality of KSM**: None.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -166,6 +166,28 @@ jobs:
       run: |
         make test-benchmark-compare
 
+    - name: Comment benchmark results
+      uses: actions/github-script@v4
+      with:
+        script: |
+          const fs = require('fs');
+          const data = fs.readFileSync('benchstat.txt', 'utf8');
+          github.issues.createComment({
+            issue_number: context.issue.number,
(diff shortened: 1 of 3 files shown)


mrueg:
Hah you read my mind. I think commenting will be a bit too noisy. We can provide a nicer table output in the github actions check format though

rexagod:
Do you think it'd be fine to keep the bench results implicit? I thought we may want to post them on the PR (and remove older ones when a new one is posted), to make the benchmarks more apparent in a manner that they are never missed.

For the noisy part, I was wondering we could do something like (the click-to-expand element),

### KSM Benchmark Output

<details>
<summary>Details</summary>

```
foo
```

</details>

I'm neutral on this. LMK if you think we'd be better off without this, and I'll create a PR for improving the format in the CI logs themselves.

mrueg:
I'm afraid of too much noise if someone force pushes several times. 
Do you think it would be possible to only comment if benchmarks change let's say more than 5% to the baseline?

Issue #2274, opened 2023-12-15
Kubernetes-Security-Slam-2023


SD-13:
Tasks we are gonna do

- [ ] Run tests for Clomonitor

jescalada:
I'd like to tackle Task 7!

jescalada:
Please take a look at the CLOMonitor .yaml PR here:
https://github.com/cncf/clomonitor/pull/1380

Thank you!

SD-13:
Hey @mrueg @dgrisonnet @rexagod I want to know whether kube-state-metrics is generating SBOM as part of the release pipeline. Where to look for the release pipeline?

dalehenries:
I looked into adding the OpenSSF Best Practices badge to the README, but I think a maintainer would need to first request the badge at https://www.bestpractices.dev/

mrueg:
> Hey @mrueg @dgrisonnet @rexagod I want to know whether kube-state-metrics is generating SBOM as part of the release pipeline. Where to look for the release pipeline?

We're currently not generating it. The release process is documented here: https://github.com/kubernetes/kube-state-metrics/blob/main/RELEASE.md If this is something that can be attached to a github release, it should be triggered by a release creating and execute a github action ideally that attaches the sbom

rexagod:
I think https://github.com/advanced-security/gh-sbom (SBOM generation) coupled with https://github.com/anchore/sbom-action (SBOM pushes) should help accomplish the SBOM workflow.

rexagod:
FYI Appended some open questions to the issue description.

ricardoapl:
I think the following tasks are already done:
- Check for Binary Artifacts (task 8) (no binaries found in the repo)
- Review the code review (task 9) (all changesets reviewed)
- Dangerous Workflow (task 10) (no dangerous workflow patterns detected)
- Dependency update tool (task 13) (update tool detected, dependabot)

I think the following tasks are still missing something:
- Token Permissions (task 16)

![Screenshot 2024-04-24 at 09-41-32 Kube State Metrics](https://github.com/kubernetes/kube-state-metrics/assets/48807108/556c3e7c-88e4-4c52-9caf-62b12d320353)

What do you think about publishing the OpenVEX data with the remaining release artifacts?

@SD-13 do you mind if I assign **Ensure SBOMs are generated by Kubernetes BOM (task 3)** to me?

SD-13:
@ricardoapl Please feel free to assign it to you!

[...]

PR #2276, opened 2023-12-15
feat: Generate OpenVX data for every release

shafeeqes:
This PR adds a github workflow action to generate OpenVX data for every release.

#SecuritySlam

Fixes part of #2274


.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,19 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
+        name: Run vexctl
+        with:
+            product: pkg:generic/hello@${{ env.RELEASE_VERSION }}
```
mrueg, line 19: does this need to be adjusted as well?

.github/workflows/openvex.yaml
```diff
@@ -0,0 +1,19 @@
...
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*.*.*'
+jobs:
+  vexctl:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v2
```
mrueg, line 13: can you use the lastest version of checkout here?

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,19 @@
...
+  workflow_dispatch:
+  push:
+    tags:
+      - 'v*.*.*'
+jobs:
+  vexctl:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
```
rexagod, line 13: (nit) We use `v4` without pinning to a specific minor or patch version throughout the repository, so IMO it should be safe to do the same here, since we'd notice any bugs linked to it anyway.

```suggestion
        uses: actions/checkout@v4
```

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,19 @@
...
+      - name: Set env
```
rexagod, line 14:
```suggestion
      - name: Set environment variables
```

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,19 @@
...
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
```
rexagod, line 16: Can we pin this to the latest major as well?

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,19 @@
...
+    tags:
+      - 'v*.*.*'
+jobs:
+  vexctl:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@b4ffde65f46336ab88eb53be808477a3936bae11 #v4.1.1
+      - name: Set env
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
```
rexagod, line 15: I'm afraid running `openvex` for all releases would generate unnecessary noise, since we generally only maintain the latest release branches. Can we restrict this to that only? cc @mrueg

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,19 @@
...
+      - uses: openvex/generate-vex@31b415924ea0d72ed5f2640f1dee59dea6c2770b
+        name: Run vexctl
```
rexagod, line 17: (nit)

```suggestion
        # Refer: https://github.com/openvex/vexctl#operational-model
        name: Run vexctl
```

rexagod (commented):
Thank you for picking this up! 🎉

.github/workflows/openvex.yml
rexagod: I see `openvex` has not tagged any release yet. Can we pin this to the latest `c59881b41451d7ccba5c3b74cd195382b8971fcd` commit then?

rexagod: I've raised an issue regarding releases: https://github.com/openvex/generate-vex/issues/2.

shafeeqes: > I see `openvex` has not tagged any release yet. Can we pin this to the latest `c59881b41451d7ccba5c3b74cd195382b8971fcd` commit then?

Done.

.github/workflows/openvex.yml
mrueg: This is restricted only to a pushed tag. See line 7

mrueg: I think it might be better to run on release creation. https://docs.github.com/en/actions/using-workflows/events-that-trigger-workflows#release

.github/workflows/openvex.yml
```diff
@@ -0,0 +1,20 @@
...
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Set environment variables
+        run: echo "RELEASE_VERSION=${GITHUB_REF#refs/*/}" >> $GITHUB_ENV
+      - uses: openvex/generate-vex@c59881b41451d7ccba5c3b74cd195382b8971fcd
+        # Refer: https://github.com/openvex/vexctl#operational-model
+        name: Run vexctl
+        with:
+            product: pkg:golang/kubernetes/kube-state-metrics@${{ env.RELEASE_VERSION }}
```
mrueg, line 20:
```suggestion
            product: pkg:golang/k8s.io/kube-state-metrics/v2@${{ env.RELEASE_VERSION }}
```

should this be the golang package name?

mrueg: See also: https://github.com/package-url/purl-spec/blob/master/PURL-TYPES.rst#golang

.github/workflows/openvex.yml
rexagod: +1, better to run on releases than tags.

.github/workflows/openvex.yml
shafeeqes: Why the v2? Is it not covered by the env?

mrueg: v2 is because it expects the full name of the golang package.

mrueg:
/lgtm

Thanks! Right now this will output in Github Actions log so it should not block any release.

PR #2277, opened 2023-12-15
docs: add OpenSSF Scorecard to README

dalehenries:
Adds the OpenSSF Scorecard to the README

#SecuritySlam

Related to #2274

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,33 +31,31 @@ are deleted they are no longer visible on the `/metrics` endpoint.
 
 ## Table of Contents
 
-- [Overview](#overview)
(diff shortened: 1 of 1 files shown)


mrueg:
Thanks for your contribution @dalehenries !

rexagod:
Thank you for your work here, @dalehenries!

The [OpenSSF practices](https://github.com/ossf/scorecard#scorecard-checks) are something we can surely utilize within the repository. However, to better visualize and trigger the checks (IIUC these are only presented as raw `json` currently), can we add an [action](https://github.com/ossf/scorecard#scorecard-github-action) for it here as well?

PR #2278, opened 2023-12-15
chore: add Security Insights (task 11)

dalehenries:
Adds SECURITY-INSIGHTS.yml

#SecuritySlam

Task 11 from #2274

### DIFF
--- a/SECURITY-INSIGHTS.yml
+++ b/SECURITY-INSIGHTS.yml
@@ -1,0 +1,31 @@
+header:
+  schema-version: "1.0.0"
+  expiration-date: "2023-12-15T19:10:00.000Z"
+  project-url: https://github.com/kubernetes/kube-state-metrics
+  changelog: https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md
+  license: https://github.com/kubernetes/kube-state-metrics/blob/main/LICENSE
+project-lifecycle:
+  status: active
+  bug-fixes-only: false
+  core-maintainers:
+    - github:dgrisonnet
+    - github:mrueg
+    - github:rexagod
+contribution-policy:
+  accepts-pull-requests: true
+  accepts-automated-pull-requests: true
+  contributing-policy: https://github.com/kubernetes/kube-state-metrics/blob/main/CONTRIBUTING.md
+  code-of-conduct: https://github.com/kubernetes/kube-state-metrics/blob/main/code-of-conduct.md
+distribution-points:
+  - https://github.com/kubernetes/kube-state-metrics/releases
+security-contacts:
+  - type: website
+    value: https://github.com/kubernetes/kube-state-metrics/blob/main/SECURITY_CONTACTS
+vulnerability-reporting:
+  accepts-vulnerability-reports: true
+  security-policy: https://github.com/kubernetes/kube-state-metrics/blob/main/SECURITY.md
+dependencies:
+  third-party-packages: true
+  dependencies-lists:
+    - https://github.com/kubernetes/kube-state-metrics/blob/main/go.mod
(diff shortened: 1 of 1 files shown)


SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,31 @@
+header:
+  schema-version: "1.0.0"
+  expiration-date: "2023-12-15T19:10:00.000Z"
```
mrueg, line 3: What's the expiry date used for?

dalehenries: According to the [specification](https://github.com/ossf/security-insights-spec/blob/main/specification.md#specification), it's a required field. The description is:

> The date this file should no longer be considered valid, and it can be at most one year from the date it was last reviewed.

mrueg: Thanks, so should it be 2024 then? Otherwise it would be expired right now

dalehenries: I'm sure there are some other fields I could have added, but I only wanted to add the ones I was fairly confident were correct.  I looked through the specification and at examples from other projects, so hopefully this is a good starting point.

dalehenries: Good point!  I'll fix that.

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,31 @@
...
+    - github:dgrisonnet
+    - github:mrueg
+    - github:rexagod
+contribution-policy:
+  accepts-pull-requests: true
+  accepts-automated-pull-requests: true
+  contributing-policy: https://github.com/kubernetes/kube-state-metrics/blob/main/CONTRIBUTING.md
+  code-of-conduct: https://github.com/kubernetes/kube-state-metrics/blob/main/code-of-conduct.md
+distribution-points:
+  - https://github.com/kubernetes/kube-state-metrics/releases
```
rexagod, line 20:
```suggestion
  - https://github.com/kubernetes/kube-state-metrics/releases
  - https://github.com/kubernetes/k8s.io/blob/main/registry.k8s.io/images/k8s-staging-kube-state-metrics/images.yaml
```

SECURITY-INSIGHTS.yml
```diff
@@ -0,0 +1,31 @@
...
+  - type: website
+    value: https://github.com/kubernetes/kube-state-metrics/blob/main/SECURITY_CONTACTS
+vulnerability-reporting:
+  accepts-vulnerability-reports: true
+  security-policy: https://github.com/kubernetes/kube-state-metrics/blob/main/SECURITY.md
+dependencies:
+  third-party-packages: true
+  dependencies-lists:
+    - https://github.com/kubernetes/kube-state-metrics/blob/main/go.mod
+    - https://github.com/kubernetes/kube-state-metrics/blob/main/Dockerfile
```
rexagod, line 31: PTAL added some fields. :)

```suggestion
    - https://github.com/kubernetes/kube-state-metrics/blob/main/Dockerfile
documentation:
    - https://github.com/kubernetes/kube-state-metrics/tree/main/docs
security-testing:
- tool-type: dast
  tool-name: govulncheck
  tool-version: latest
  tool-url: https://go.googlesource.com/vuln
  tool-rulesets:
  - built-in
  integration:
    ci: true
  comment: |
    Detects vulnerabilities as a result of the affected call-paths being invoked directly in the repository, while reducing false positives by ignoring dormant call-paths for package dependencies.
```

rexagod (commented):
Thank you for adding this!

mrueg:
/lgtm 

Thanks for your contribution!

PR #2299, opened 2024-01-11
docs: Add guide on how to debug CustomStateMetrics feature quickly

CatherineF-dev:

More and more bugs are opened for CustomStateMetrics feature. This will help users to debug CustomStateMetrics by themselves quicker.

Fixes #

### DIFF
--- a/README.md
+++ b/README.md
@@ -406,6 +406,14 @@ Then curl the metrics endpoint
 
 To run the e2e tests locally see the documentation in [tests/README.md](./tests/README.md).
 
+##### Debug CustomStateMetrics feature
+```
+minikube start
+...
+go run main.go --custom-resource-state-only --custom-resource-state-config-file ksm-2223/custom-resource-config-file.yaml --kubeconfig ~/.kube/config
+```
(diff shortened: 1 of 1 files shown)


rexagod:
The change seems fairly straight-forward. I'd suggest adding a section demonstrating the usage of https://github.com/go-delve/delve to catch contexts at breakpoints set by the user.

CatherineF-dev:
Could we merge this? Not everyone knows minikube.

You can add how to use some debugging tools in another PR.

README.md
```diff
@@ -406,6 +406,16 @@ Then curl the metrics endpoint
 
 To run the e2e tests locally see the documentation in [tests/README.md](./tests/README.md).
 
+##### Debug CustomStateMetrics feature
```
dgrisonnet, line 409: Instead of adding it to the README, I think it would be better to put it in its own doc under https://github.com/kubernetes/kube-state-metrics/tree/main/docs, or maybe in a new directory there.

I suggested reworking the structure of our doc directory in the past, we might want to revive that effort.

PR #2303, opened 2024-01-13
fix: adjust "volume name" label for kube_persistentvolumeclaim_*

### LINKED ISSUE #2288: `kube_persistentvolume_*` and `kube_persistentvolumeclaim_*` should expose the volume name under the same label

`kube_persistentvolume_*` metrics and `kube_persistentvolumeclaim_*` metrics should expose the "name of the volume" inside the same metric label.

**Why is this needed**:

It's a bit difficult at the moment to combine the Persistent Volume and Persistent Volume Claim metrics together to get, for instance, the "size" of the PVC in a specific namespace:

* `kube_persistentvolume_*` metrics have a [`persistentvolume` label](https://github.com/kubernetes/kube-state-metrics/blob/532fb2dba7e73d2675c9bcdd69a8821798cd694c/internal/store/persistentvolume.go#L45) containing the name of the underlying persistent volume.
* `kube_persistentvolumeclaim_*` on the other hand, expose [the name of the persistent volume in the `volumename` label](https://github.com/kubernetes/kube-state-metrics/blob/532fb2dba7e73d2675c9bcdd69a8821798cd694c/internal/store/persistentvolumeclaim.go#L107-L108)

It's then difficult to combine both metrics together, and bring the PVC labels into the PV metric (or vice-versa) (see https://github.com/prometheus/prometheus/issues/2204)

Rename one of the two label to have the same label as the other metric.

**Additional context**

I was trying to get the "size" of the PVCs with a query like this, but it doesn't work because it's not possible to join on different labels:

```
avg(kube_persistentvolume_capacity_bytes) by (persistentvolume)
* on (volumename) group_left (namespace, persistentvolumeclaim)
avg(kube_persistentvolumeclaim_info) by (namespace, persistentvolumeclaim, volumename)
```

multani:

This changes the name of the volume name label, reported by the `kube_persistentvolumeclaim_*` metrics, to be the same label as the one in the `kube_persistentvolume_*` metrics.

This makes it easier to combine the 2 related set of metrics together to being labels and values from one metric into another.

Does not change cardinality

Fixes #2288


multani:
Sorry for the noise, I was trying to understand what was the problem with the commit message :facepalm:

Should be good now :)

CatherineF-dev:
Also update https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md

multani:
> Also update https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md

Done! :+1:

docs/persistentvolumeclaim-metrics.md
```diff
@@ -4,7 +4,7 @@
...
-| kube_persistentvolumeclaim_info                            | Gauge       |                                                                                                                           |                         | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt;<br>`volumename`=&lt;volumename&gt;<br>`volumemode`=&lt;volumemode&gt; | STABLE       |
+| kube_persistentvolumeclaim_info                            | Gauge       |                                                                                                                           |                         | `namespace`=&lt;persistentvolumeclaim-namespace&gt; <br> `persistentvolumeclaim`=&lt;persistentvolumeclaim-name&gt; <br> `storageclass`=&lt;persistentvolumeclaim-storageclassname&gt;<br>`persistentvolume`=&lt;volumename&gt;<br>`volumemode`=&lt;volumemode&gt; | STABLE       |
```
CatherineF-dev, line 7: This metric is STABLE, so can't change label name.

CatherineF-dev: Have you tried updating your promQL to support join on different label names?

multani: As mentioned in [the parent issue](https://github.com/kubernetes/kube-state-metrics/issues/2288), AFAIK it's difficult to [join metrics on different label names](https://github.com/prometheus/prometheus/issues/2204).

We are already relabelling the `kube_persistentvolumeclaim_annotations` metric to "fix" the label, but my hope with this pull request would be to bring this change upstream.

Is there a simpler way to join on different label names that would not require this change? (this discussion belongs more in the parent issue perhaps...)

mrueg:
> Also update https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md
> ## Next:

Has anything changed in the release process I'm not aware of? We usually do not update the changelog within the PR.

docs/persistentvolumeclaim-metrics.md
multani: @CatherineF-dev is there any other options to move this change forward?

multani:
/remove-lifecycle stale

Still looking for feedback on how to move this forward (cf. https://github.com/kubernetes/kube-state-metrics/pull/2303#discussion_r1525927488)

PR #2314, opened 2024-02-13
Clean up `select {}`

CatherineF-dev:

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)* N/A

Fixes #


CatherineF-dev:
cc @mrueg to review.

Does this `select {}` have some special purposes?

mrueg:
AFAIK the select here is used to block similar to an infinite for loop.

rexagod:
Yep, closing.

PR #2316, opened 2024-02-18
Update version to reflect the latest patch release

rexagod:
Reflect patch release changes on `main`, which was not the case with `2.10.1` since that diverged and was never merged back into `main` (around backporting concerns). I believe in the future we can merge everything back into `main` to reflect the latest version changes in main, but diverge the release branches starting from the pre-merge commit to make sure backports are clean and maintainable (these will not be merged back into the default branch).


tools/tools.go
```diff
@@ -20,6 +20,7 @@ limitations under the License.
 package tools
 
 import (
+	_ "github.com/Kunde21/markdownfmt/v3"
```
dgrisonnet, line 23:
```suggestion
	_ "github.com/Kunde21/markdownfmt/v3/cmd/markdownfmt"
```

rexagod:
/close
Superseded by #2317.

PR #2324, opened 2024-02-19
chore: lint README.md

rexagod:
To primarily avoid conflicts and allow for automations.

PR #2325, opened 2024-02-19
chore: automate release workflow

rexagod:
Blocked by (and rebased on) https://github.com/kubernetes/kube-state-metrics/pull/2316 and https://github.com/kubernetes/kube-state-metrics/pull/2324. Please merge them first, preferably in that order.

***

Test run: https://github.com/rexagod/kube-state-metrics/pull/9.


scripts/generate-release-notes.sh
```diff
@@ -0,0 +1,21 @@
...
+  fi
+  SED=gsed
+fi
+
+# Get the latest tag from the git repository.
+latest_tag=$(git describe --tags --abbrev=0)
+
+# Replace the first line of the CHANGELOG.md file with the expected metadata.
+# NOTE: This is done to ensure the latest release header is consistent, and will replace the manual placeholder.
+$SED -i "1s/.*/## v$latest_tag / $(date +'%Y-%m-%d')/" CHANGELOG.md
```
rexagod, line 21: Also modified the PR template to reflect this change: https://github.com/kubernetes/kube-state-metrics/pull/2325/files#diff-18813c86948efc57e661623d7ba48ff94325c9b5421ec9177f724922dd553a35.

scripts/update-compatibility-matrix.sh
```diff
@@ -0,0 +1,53 @@
+#!/bin/bash
```
mrueg, line 1: Can we do this a bit more programmatically instead of search and replacing values?
We could use a README.md.tmpl template file + a config and https://github.com/hairyhenderson/gomplate or a similar tool.

assu-2000: @mrueg  agree with you! Since the project already has commit convention rules (`feat`, `fix`, `chore` ...) and commits are signed, for changelogs one possible approach could be to explore a solution provided by Google [https://github.com/googleapis/release-please-action](https://github.com/googleapis/release-please-action) and adapt it to the project's changelog file.  
It works almost the same, requiring adjustments to the message format and the addition of the commit author's handle by customizing the release YAML file and another custom JSON file , for example, mapping the type `fix` to a section like `[BUGFIX]` according to the KSM `CHANGELOG` file format.  
of course, this is just one idea to explore for the CI/CD release pipeline.

PR #2329, opened 2024-02-27
feat: add field selector for node store ListWatch if node argument given

gqcn:

> If node argument given, it can list and watch node resource using specified node name, instead of full amount of nodes from api-server, which could cause big cpu pressure for api-server if nodes are in big amount.

> does not change cardinality.

Fixes #

> no issue associated

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -18,11 +18,11 @@ package store
 
 import (
 	"context"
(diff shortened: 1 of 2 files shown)


gqcn:
It seems that the EasyCLA in CI did not pass. I'll commit a new PR for this, close this one.

PR #2347, opened 2024-03-21
fix(pod autosharding): transition from labelselector to fieldselector

pkoutsovasilis:

This PR is a minor change for pod autosharding mode where it substitutes the LabelSelector with a FieldSelector. Specifically, at the moment in this mode, we detect and extract the labels of pod-owner StatefulSet based on which a LabelSelector is constructed and used in NewFilteredListWatchFromClient. However, if a label of the former StatefulSet is changed for any reason, e.g. an arbitrary operator manages and injects in the labels the hash of the whole statefulset to decide whether it has changed during a reconcile, as the pods won't restart (k8s design) no more events will arrive and thus shards won't be updated properly. Instead of relying on LabelSelector this PR replaces it with a FieldSelector that targets the owner StatefulSet by its name, thus it will always receive updates. This last bit, is aligned with the current `if ss.Name != statefulSetName` in [`AddFunc`](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metricshandler/metrics_handler.go#L123) and [`UpdateFunc`](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metricshandler/metrics_handler.go#L145)

* does not change cardinality

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2355

### DIFF
--- a/pkg/metricshandler/metrics_handler.go
+++ b/pkg/metricshandler/metrics_handler.go
@@ -111,12 +111,12 @@ func (m *MetricsHandler) Run(ctx context.Context) error {
 	}
 	statefulSetName := ss.Name
 
-	labelSelectorOptions := func(o *metav1.ListOptions) {
-		o.LabelSelector = fields.SelectorFromSet(ss.Labels).String()
+	fieldSelectorOptions := func(o *metav1.ListOptions) {
+		o.FieldSelector = fields.OneTermEqualSelector("metadata.name", statefulSetName).String()
 	}
 
 	i := cache.NewSharedIndexInformer(
-		cache.NewFilteredListWatchFromClient(m.kubeClient.AppsV1().RESTClient(), "statefulsets", m.opts.Namespace, labelSelectorOptions),
+		cache.NewFilteredListWatchFromClient(m.kubeClient.AppsV1().RESTClient(), "statefulsets", m.opts.Namespace, fieldSelectorOptions),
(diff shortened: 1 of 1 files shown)


pkoutsovasilis:
@dgrisonnet @logicalhan could you please have a review on this PR? 🙂

CatherineF-dev:
/lgtm 

1. `metadata.name` is more invariant compared to `labels`. Though I am not sure why it used `labels` in the beginning. 

2. Most k/k codes are using metadata.name instead labels. https://github.com/search?q=repo%3Akubernetes%2Fkubernetes+OneTermEqualSelector%28%22metadata.name%22&type=code&p=1

3. Tests are done to verify that it can work.

CatherineF-dev:
cc @dgrisonnet to approve

diranged:
I'm not sure what the status is here - but this is a really huge pain point for us right now. Any time we ship any release that updates labels on any deployments/statefulsets/etc, our `kube-state-metrics` pods get into a bad state and start sending invalid data.. which invariably leads to our ops teams getting paged with incorrect alerts about pods being in bad states.

CatherineF-dev:
In asking an approval for this PR. I only have LGTM permission so far.

rexagod:
/approve

@CatherineF-dev feel free to send a PR to add yourself to approvers.

PR #2357, opened 2024-04-03
kube_secret_labels add secret type

huizluo:
**What this PR does / why we need it**: Sometimes we need to use the type attribute of a secret for statistical calculations. This PR (Pull Request) adds a type attribute to kube_secret_labels.

Fixes #


mrueg:
It looks like unittests are failing, can you fix them as well?

mrueg:
> It looks like unittests are failing, can you fix them as well?

@huizluo would you be able to address this issue or is it okay to take over the PR?

mrueg:
Having another look: We already expose a `kube_secret_type` metric, so I don't think we should add it to the standardized _labels metric.

/close

Issue #2353, opened 2024-03-27
sharding with a deployment with '--resources=pods' and '--node=""' does not fetch pending pods


carlosrmendes:
**What happened**: a deployment with `--resources=pods` and `--node=""` does not fetch pending pods

**What you expected to happen**: pending pods be fetched by ksm

**How to reproduce it (as minimally and precisely as possible)**:
- create a ksm deploy with `--resources=pods` and `--node=""` and `-v=10`
- force another pod to be in a pending state
- check the logs for the api request like `https://<kubernetes_api_ip>:443/api/v1/pods?allowWatchBookmarks=true&fieldSelector=spec.nodeName%3D%22%22&resourceVersion=3998681&timeout=5m23s&timeoutSeconds=323&watch=true`
- 'kubect proxy' and try to make the same request
- make the same request without the `&fieldSelector=spec.nodeName%3D%22%22` and confirm that on the response, the pending pods does not have any `nodeSelectot` key with an empty value

* kube-state-metrics version: v2.10.1
* Kubernetes version (use `kubectl version`): k3s v1.24.10

CatherineF-dev:
Thanks for reporting this. 

Could you help paste your pending pod yaml, if it's available? Want to see nodeName value on the yaml

CatherineF-dev:
Wondering whether apiserver fieldSelector supports `-nodeName` which means nodeName doesn't exist

CatherineF-dev:
From https://kubernetes.io/docs/concepts/overview/working-with-objects/field-selectors,

one solution might be 
1. Daemonset: status.phase=Running && nodeName = x
2. Deployment: status.phase!=Running

CatherineF-dev:
I tried again and verified `--field-selector spec.nodeName=` can get unscheduled pods.

--node="" is translated into `spec.nodeName=`

1. nothing: get all pods
2. --node="A":  `--field-selector spec.nodeName=A ` get pods on node A
3. --node="": `--field-selector spec.nodeName= ` get pods with no node assigned
```
kubectl get pods --field-selector spec.nodeName= -A

NAMESPACE   NAME                          READY   STATUS    RESTARTS   AGE
default     helloworld-7f5cd44b59-zhgzj   0/1     Pending   0          87s
```

```
kube_pod_status_scheduled{namespace="default",pod="helloworld-7f5cd44b59-zhgzj",uid="26018553-ceb6-427f-be1a-088032edac7a",condition="false"} 1
```

[...]

PR #2359, opened 2024-04-03
docs: update doc around fetching unscheduled pods using --node=

CatherineF-dev:

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)* n/a

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2353

### DIFF
--- a/README.md
+++ b/README.md
@@ -296,6 +296,12 @@ spec:
         - --node=""
 ```
 
+| KSM --node parameter | FieldSelector | Note |
+| ---                  | --- | --- |
+| --node="node1"       | kubectl get pods --field-selector spec.nodeName="node1" | Fetch all pods running on node1 |
+| --node=              | kubectl get pods --field-selector spec.nodeName= | Fetch all pods without an assigned node |
+|                      | kubectl get pods | Fetch all pods no matter whether they have an assigned node or not|
+
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
 
 ### Setup
--- a/README.md.tpl
+++ b/README.md.tpl
@@ -297,6 +297,12 @@ spec:
         - --node=""
 ```
 
+| KSM --node parameter | FieldSelector | Note |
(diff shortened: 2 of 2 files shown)


README.md
```diff
@@ -296,6 +296,12 @@ spec:
         - --node=""
 ```
 
+| KSM --node parameter | FieldSelector |
```
mrueg, line 299: Can we have a human readable column as well that explains what this setting does?

CatherineF-dev: Updated

mrueg:
Sorry, I forgot: The change needs to be made in README.md.tpl and then you'd need to run the make target to regenerate the README file with gomplate.

README.md
```diff
@@ -296,6 +296,12 @@ spec:
         - --node=""
 ```
 
+| KSM --node parameter | FieldSelector | Note |
+| ---                  | --- | --- |
+| --node="node1"       | kubectl get pods --field-selector spec.nodeName="node1" | Fetch all pods running on node1 |
+| --node=""            | kubectl get pods --field-selector spec.nodeName= | Fetch all pods without an assigned node |
```
dgrisonnet, line 302: Similar to my comment in https://github.com/kubernetes/kube-state-metrics/pull/2373#issuecomment-2066910515, I don't think that specifying `--node=""` is very user friendly.

We could either implicitly do it, or put it in a dedicated flag such as `--fetch-unscheduled-pods`

CatherineF-dev: Fixed the documentation to `--node=` instead of `--node=""`. Will it be clearer? Or we prefer `--fetch-unscheduled-pods`

mrueg:
Let's add this once https://github.com/kubernetes/kube-state-metrics/pull/2388 is in.

We can refactor the behavior later.

CatherineF-dev:
This PR is not needed due to https://github.com/kubernetes/kube-state-metrics/pull/2388

PR #2373, opened 2024-04-17
fix:  `.` is removed in nodeName

CatherineF-dev:

Fixes # an issue raised in slack chat `We see the fieldSelector  get created with the node-name … but it’s missing the .`

### DIFF
--- a/pkg/options/types.go
+++ b/pkg/options/types.go
@@ -18,7 +18,6 @@ package options
 
 import (
 	"errors"
-	"regexp"
 	"sort"
 	"strings"
 
@@ -145,9 +144,7 @@ func (n *NodeType) GetNodeFieldSelector() string {
 		klog.InfoS("Using node type is nil")
 		return EmptyFieldSelector()
 	}
-	pattern := "[^a-zA-Z0-9_,-\\.]+"
-	re := regexp.MustCompile(pattern)
-	result := re.ReplaceAllString(n.String(), "")
+	result := n.String()
 	klog.InfoS("Using node type", "node", result)
 	return fields.OneTermEqualSelector("spec.nodeName", result).String()
 
--- a/pkg/options/types.go
+++ b/pkg/options/types.go
@@ -18,7 +18,6 @@ package options
 
 import (
 	"errors"
-	"regexp"
 	"sort"
 	"strings"
 
@@ -145,9 +144,7 @@ func (n *NodeType) GetNodeFieldSelector() string {
 		klog.InfoS("Using node type is nil")
 		return EmptyFieldSelector()
 	}
-	pattern := "[^a-zA-Z0-9_,-]+"
-	re := regexp.MustCompile(pattern)
-	result := re.ReplaceAllString(n.String(), "")
+	result := n.String()
 	klog.InfoS("Using node type", "node", result)
 	return fields.OneTermEqualSelector("spec.nodeName", result).String()
 
--- a/pkg/options/types.go
+++ b/pkg/options/types.go
@@ -18,6 +18,7 @@ package options
 
 import (
 	"errors"
+	"regexp"
 	"sort"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


diranged:
This should fix https://github.com/kubernetes/kube-state-metrics/issues/2374. Thank you, I hope this can get merged quickly!

LaikaN57 (approved):
LGTM. Alternatively, it looks like the `,` in the `pattern` might have been a typo. We could just replace that with a `.`.

mrueg:
There was a conversation here: https://github.com/kubernetes/kube-state-metrics/pull/2217/files#r1356982673 https://github.com/kubernetes/kube-state-metrics/pull/2217/files#r1357024577

Do we still need this to differentiate between "" and nil?

CatherineF-dev:
Good point. Updated to only add `\\.` 

Tested: https://go.dev/play/p/K5GtruNWjO0

mrueg:
https://kubernetes.io/docs/concepts/overview/working-with-objects/names/#dns-subdomain-names

We can reuse https://github.com/kubernetes/apimachinery/blob/0ee3e6150890f56b226a3fe5a95ba33b1b2bf7c7/pkg/util/validation/validation.go#L177 probably.

pkg/options/types.go
```diff
@@ -145,7 +145,7 @@ func (n *NodeType) GetNodeFieldSelector() string {
 		klog.InfoS("Using node type is nil")
 		return EmptyFieldSelector()
 	}
-	pattern := "[^a-zA-Z0-9_,-]+"
+	pattern := "[^a-zA-Z0-9_,-\\.]+"
```
diranged, line 148: I would add some tests that verify that different hostnames/fqdns work.

CatherineF-dev: Removed this regexp.

CatherineF-dev:
Updated. Need review again.

dgrisonnet:
I don't get why we have this `NodeType` structure from https://github.com/kubernetes/kube-state-metrics/pull/2217/. Pods in scheduling state could be handled separately, even maybe with its own CLI option.

Also one thing I am curious about is what would happen once the pod is scheduled on the node. Do we continue to produce a stale metrics with the state of the pod when it wasn't schedule? Or do we have a mechanism to check that this pod shouldn't be watched by this shard anymore?

diranged (approved):
I approve - but I'm not sure that means anything. :)

CatherineF-dev:
> Do we continue to produce a stale metrics with the state of the pod when it wasn't schedule? 


No. I tested for both daemonset (watch scheduled pods) and deployment (watch not scheduled pods).


```
# Test for the deployment

# 1. deploy a pod
kubectl apply -f https://raw.githubusercontent.com/CatherineF-dev/k8s-hello-world/main/unschedule_pod.yaml

kubectl get pods --field-selector spec.nodeName= -A
NAMESPACE   NAME                          READY   STATUS    RESTARTS   AGE
default     hello-node-7d78fd84d9-rqbjh   0/1     Pending   0          38s

# 2. can see metrics when this pod is not scheduled

curl localhost:8080/metrics | grep hello-node | grep schedule

kube_pod_status_scheduled{namespace="default",pod="hello-node-7d78fd84d9-7grpc",uid="1c9982aa-e99b-46a0-97c0-b03bf1b94b47",condition="true"} 0
kube_pod_status_scheduled{namespace="default",pod="hello-node-7d78fd84d9-7grpc",uid="1c9982aa-e99b-46a0-97c0-b03bf1b94b47",condition="false"} 1
kube_pod_status_scheduled{namespace="default",pod="hello-node-7d78fd84d9-7grpc",uid="1c9982aa-e99b-46a0-97c0-b03bf1b94b47",condition="unknown"} 0
kube_pod_scheduler{namespace="default",pod="hello-node-7d78fd84d9-7grpc",uid="1c9982aa-e99b-46a0-97c0-b03bf1b94b47",name="default-scheduler"} 1

# 3. can't see metrics when this pod is scheduled 
curl localhost:8080/metrics | grep hello-node | grep schedule
```

CatherineF-dev:
@dgrisonnet need review again.

mrueg:
Closing this in favor of https://github.com/kubernetes/kube-state-metrics/pull/2388 

Feel free to reopoen if anything changes.

Issue #2206, opened 2023-09-26
kube_cronjob_next_schedule_time in the past when cronjob timezone specified


paulclark222:
**What happened**:
Defined a timezone for a cronjob: 
`timeZone: "Europe/London"`
Now kube_cronjob_next_schedule_time is always returning the **last** schedule time rather than the **next** one.

**What you expected to happen**:
kube_cronjob_next_schedule_time should be the **next** scheduled time.

**How to reproduce it (as minimally and precisely as possible)**:
Define a cronjob with non-UTC timezone (may only happen on zones east of UTC). Wait for the job to execute and then check kube_cronjob_next_schedule_time - you will see it's in the past.

* kube-state-metrics version: 2.10
* Kubernetes version (use `kubectl version`): 1.27
* Cloud provider or hardware configuration: Azure AKS

CatherineF-dev:
The source code is here: https://github.com/kubernetes/kube-state-metrics/blob/b91dd311397047f15aae4f6dacba0708f24cf446/internal/store/cronjob.go#L238-L263

Could you dive into it and help debug if you are available?

paulclark222:
I'm not a go developer, but I suspect that what's happening is that j.Spec.Schedule doesn't include the timezone, so the cron library is interpreting it in local time (UTC). So if this code were to run just after the cronjob had run (in UTC+1), it would say the next scheduled time was 1 hour away.

The cronjob library expects the timezone to be specified like this:
`schedule, err := cron.ParseStandard("TZ=America/Los_Angeles 0 2 * * 0")`
or
`loc, _ := time.LoadLocation("Europe/Paris")
 cronProcess := cron.New(cron.WithLocation(loc))`

I'm not sure if this is being done.

CatherineF-dev:
Hi @paulclark222, j.Spec does have a field named timezone, which hasn't been used

https://github.com/kubernetes/api/blob/master/batch/v1/types.go#L637

CatherineF-dev:
Could you ramp up golang and try to fix this issue, if you are available and have enough time?

dgrisonnet:
The code for the metric predates the support for timezone, so it is not taken into account. We would need to had support for it in ksm.

/help

mhmtsvr:
I encountered the same issue after configuring the timezone. Any recent updates or ETA on a resolution? Thanks!

dehoej:
+1
Really an issue when you are trying to create alerts if cronjobs are running longer than expected.

[...]

Issue #500, opened 2018-07-24
kube_cronjob_next_schedule_time has time in the past


balous:
**Is this a BUG REPORT or FEATURE REQUEST?**:

> Uncomment only one, leave it on its own line: 
>
/kind bug
> /kind feature

**What happened**:
I've got a cron job with:
- lastScheduleTime: 2018-07-24T00:00:00Z
- schedule: 0 2 * * *
- creationTimestamp: 2018-06-06T13:55:15Z

For this cronjob, current value of *kube_cronjob_next_schedule_time* metric is 1532397600 (now it is 1532433916, so it is in the past). When formated for my timezone (+0200), the time is 2018-07-24 02:00:00 which is equal to the time the job was run tonight.

**What you expected to happen**:
*kube_cronjob_next_schedule_time* should contain time in the future (i.e. 2018-07-25 02:00:00).

**How to reproduce it (as minimally and precisely as possible)**:
It happens to all my cron jobs so I suppose it is enough to create a cronjob and check *kube_cronjob_next_schedule_time*'s value.

**Anything else we need to know?**:
I've also played with another job, which was created 47 days ago and was suspended since the beginning. I unsuspended it, checked the metric and got 47.75 days in the past.

Also note that the cluster's hosts timezone is set to CEST.

**Environment**:
- Kubernetes version (use `kubectl version`): N/A
- Kube-state-metrics image version 1.3.1, 1.2.0 and 1.1.0 (not tested the earlier versions)

balous:
I poked around a bit and it seems that K8S API is reporting timestamps with UTC timezone designator (the Z at the end).

brancz:
Interesting, so you are saying, that the Kubernetes API is not respecting your timezone settings, and as a result kube-state-metrics is using the same "wrong" timezone. That seems plausible, we should open a new issue on Kubernetes in that case.

balous:
I think that kubernetes API works correctly - I don't need the API to keep the timezone, I just need that the status contains correct timestamp.

I.e. the job is specified to run at 2AM and the server is in CEST. The job have been started at 2AM CEST. Just the API shows 0AM UTC, which represents the same point in time so there is no problem in it.

brancz:
That would mean we need to teach kube-state-metrics the timezone that the cronjob is working off of, which would either mean we need to read it from somewhere (if that information would actually be correct, I don't know) or allow it to be passed as a flag. What do you think?

balous:
[...]

PR #2376, opened 2024-04-22
feat: Add timezone to kube_cronjob_info / Make kube_cronjob_next_schedule_time timezone-aware

mrueg:
Cronjob is not timezone aware.
Adds one label to the kube_cronjob_info metric.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2206


internal/store/cronjob.go
```diff
@@ -96,11 +96,15 @@ func cronJobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
+				timeZone := "local"
+				if j.Spec.TimeZone != nil {
+					timeZone = *j.Spec.TimeZone
+				}
 				return &metric.Family{
 					Metrics: []*metric.Metric{
 						{
-							LabelKeys:   []string{"schedule", "concurrency_policy"},
-							LabelValues: []string{j.Spec.Schedule, string(j.Spec.ConcurrencyPolicy)},
+							LabelKeys:   []string{"schedule", "concurrency_policy", "timezone"},
```
dgrisonnet, line 106: wouldn't it be better in its own metric rather than in the info metric?

mrueg: There's a 1:1 relationship, why would you prefer a dedicated metric for it?

mrueg: @dgrisonnet any feedback here?

dgrisonnet: `_info` is not a very intuitive suffix which is why I don't really like these metrics where we just end up dumping a bunch of labels that aren't really related between one another.

I don't know what was the historical reasoning for putting labels into these metrics, but I can also see some dedicated metrics for values part of objects spec: https://github.com/kubernetes/kube-state-metrics/blob/569e820c65ecfad6c158d5b5eb1ac4fd94cb1b80/internal/store/pod.go#L1212-L1231

To me, info should be limited to the default labels. Any other information that we want to expose should be in its own dedicated metric. What is your opinion on that?

mrueg: I agree on having a separate metric for anything that does not have a 1:n relationship within the object.  For 1:1 relationships, additional time series make it difficult to correlate between multiple labels vs. simply extracting the ones you need: 
`max(resource_info) by (label1, label2, label3)`.
Trying to do the same with multiple metrics can get very annoying and more error-prone.

To me, the _info metric should describe all single-value keys of the object that are static and non-numeric.

rexagod: Hmm, there seems to be a naming discrepancy between per-property gauges and per-object info metrics, so there's no clear pattern we can adhere to.

That being said, accumulating all cardinally-bound fields as label-sets into a per-object `_info` metric opens the room for metrics with a large number of labels. The metric itself can grow to become unintentionally cardinal and dropping unwanted labels will entail relabelling efforts. Splitting this into per-field metric will limit the cardinality and allow for a more granular control.

Users should be able to add the timezone label to any exported metrics using a `foo_metric * on(job) group_left(timezone) kube_cronjob_timezone`.

mrueg: I tried to define it here:
`the _info metric should describe all single-value keys of the object that are static over its lifetime and non-numeric`

rexagod: I meant a documentation/standard of sorts in the repository that enforces this. Not sure if there is one.

mrueg: I can add that in a follow-up PR.

mrueg:
@dgrisonnet @rexagod anything left to do to get this change in?

internal/store/cronjob.go
mrueg: I'm currently working on a tool that will generate ideal metric names but it will take some time.

rexagod: I believe the tool's definition on an incorrect or correct metric representation for a nested object needs to weigh in the point Damien raised above.

Basically, are we in consensus on what is the standard approach for representing a nested value (since more granularity leads to more joins, but more abstractions lead to more cardinal label-sets (which is comparatively more harmful))?

der-eismann:
@dgrisonnet @rexagod Sorry for the ping, but this feature is pretty important to us since we can't set the timezone on AWS' Bottlerocket OS. Is there any way to accelerate the review process?

chlunde:
@dgrisonnet maybe we could merge this without the info label, and then have a separate PR for the label?

Also linking to https://github.com/kubernetes/kube-state-metrics/issues/500, I wrote more or less the same PR to fix _that_ issue, which I believe have the same root cause

mrueg:
@dgrisonnet if you agree with https://github.com/kubernetes/kube-state-metrics/pull/2528 and your time permits, can you help me get this merged? :)

dgrisonnet (approved):
/lgtm
/approve

PR #2381, opened 2024-04-24
ci: generate sbom on release

ricardoapl:

Generate an SBOM for each new release of kube-state-metrics

Does not change cardinality

Part of https://github.com/kubernetes/kube-state-metrics/issues/2274

### DIFF
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -3,7 +3,7 @@ name: Generate SBOM with Kubernetes BOM
 on:
   release:
     types:
-      - published
+      - released
 
 permissions:
   contents: read
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -16,7 +16,7 @@ jobs:
       contents: write
 
     env:
-      OUTPUT: kube-state-metrics-${{ github.event.release.tag_name }}.spdx
+      OUTPUT: sbom.spdx
       TAG: ${{ github.event.release.tag_name }}
 
     steps:
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -16,8 +16,8 @@ jobs:
       contents: write
 
     env:
-      OUTPUT: kube-state-metrics-${{ github.ref_name }}.spdx
-      TAG: ${{ github.ref_name }}
+      OUTPUT: kube-state-metrics-${{ github.event.release.tag_name }}.spdx
+      TAG: ${{ github.event.release.tag_name }}
 
     steps:
       - name: Fetch source code into GITHUB_WORKSPACE
--- a/.github/workflows/sbom.yaml
+++ b/.github/workflows/sbom.yaml
@@ -1,0 +1,41 @@
+name: Generate SBOM with Kubernetes BOM
+
+on:
+  release:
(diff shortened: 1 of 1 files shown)


ricardoapl:
/cc @puerco 

Please take a look if you can

Thank you

.github/workflows/sbom.yaml
```diff
@@ -0,0 +1,41 @@
...
+        uses: actions/checkout@1d96c772d19495a3b5c517cd2bc0cb401ea0529f # v4.1.3
+
+      - name: Install Kubernetes BOM
+        uses: kubernetes-sigs/release-actions/setup-bom@841d76a188a7c121231a863572e27012805715a2 # v0.1.4
+
+      - name: Generate SBOM
+        run: |
+          bom generate \
+            --dirs=. \
+            --image=registry.k8s.io/kube-state-metrics/kube-state-metrics:$TAG \
```
mrueg, line 33: Does this assume that the tag is already in the registry and built during release?

ricardoapl: Yes, it assumes the container image has been tagged already with the same tag used for GitHub Release.

Is this safe to assume?

.github/workflows/sbom.yaml
```diff
@@ -0,0 +1,41 @@
...
+jobs:
+  sbom:
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    env:
+      OUTPUT: kube-state-metrics-${{ github.ref_name }}.spdx
+      TAG: ${{ github.ref_name }}
```
ricardoapl: This wouldn't work with event `release`, but should be resolved with 7fc6c5dd2e6596d6534587497c2821c090bf9555

See https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release

.github/workflows/sbom.yaml
mrueg: See our RELEASE.md, we usually create a pre-release and then have another PR to get the container images promoted from the staging registry.

ricardoapl: As far as I could understand, by the time a pre-release is marked as a regular release, the image was already promoted with the proper tag:

> Once the PR to promote the image is merged, mark the pre-release as a regular release.

https://github.com/kubernetes/kube-state-metrics/blob/main/RELEASE.md

Therefore, using `types: [released]` instead of `types: [published]` should resolve this, because the workflow will only trigger when:

> A release was published, or a pre-release was changed to a release.

https://docs.github.com/en/webhooks/webhook-events-and-payloads?actionType=released#release

Should be resolved with 2a23510

What do you think?

.github/workflows/sbom.yaml
```diff
@@ -16,7 +16,7 @@ jobs:
       contents: write
 
     env:
-      OUTPUT: kube-state-metrics-${{ github.event.release.tag_name }}.spdx
+      OUTPUT: sbom.spdx
```
ricardoapl, line 19: CLOMonitor expects asset name to contain `sbom`

https://clomonitor.io/docs/topics/checks/#software-bill-of-materials-sbom

mrueg:
/hold cancel

Let's get this change in.

Issue #2374, opened 2024-04-17
Node selection for fully qualified node-names fails (--node=ip-xx-xx-xx-xx.myzone.com)


diranged:
**What happened**:

I’m trying to use the `kube-state-metrics` pods in the DaemonSet mode with `--resources=pods` and `--node=$(NODE_NAME)`… in my local testing on a Kind environment, it worked fine. However when I run it in a real EKS cluster to test, I get an odd behavior. We see the fieldSelector  get created with the node-name … but it’s missing the `.`'s:

eg:
```
│   containers:                                                                                                                                                                                                                             │   - args:                                                                                                                                                                                                                                
│     - -v=7                                                                                                                                                                                                                               
│     - --resources=pods                                                                                                                                                                                                                    
│     - --node="$(NODE_NAME)"                                                                                                                                                                                                               
│     - --port=8080                                                                                                                                                                                                                         
│     env:                                                                                                                                                                                                                                  
│     - name: NODE_NAME                                                                                                                                                                                                                     
│       valueFrom:                                                                                                                                                                                                                          
│         fieldRef:                                                                                                                                                                                                                         
│           apiVersion: v1                                                                                                                                                                                                                  
│           fieldPath: spec.nodeName                                                                                                                                                                                                        
```

and then we see this:
```
│ I0417 20:56:30.604141       1 server.go:339] "Started kube-state-metrics self metrics server" telemetryAddress=":8081"                                                                                                                    
│ I0417 20:56:30.604284       1 builder.go:520] "FieldSelector is used" fieldSelector="spec.nodeName=ip-100-80-189-206us-west-2computeinternal"                                                                                             
│ I0417 20:56:30.604321       1 builder.go:282] "Active resources" activeStoreNames="pods"                                                                                                                                                  
│ I0417 20:56:30.604332       1 reflector.go:289] Starting reflector *v1.Pod (0s) from pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229                                                                                        
│ I0417 20:56:30.604342       1 reflector.go:325] Listing and watching *v1.Pod from pkg/mod/k8s.io/client-go@v0.29.3/tools/cache/reflector.go:229                                                                                           
│ I0417 20:56:30.604381       1 server.go:73] levelinfomsgListening onaddress:8080                                                                                                                                                          
[...]

PR #2388, opened 2024-05-09
fix: use --track-unscheduled-pods to select unscheduled pods in Daemonset sharding

CatherineF-dev:

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2353

Tested using minikube

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ spec:
               fieldPath: spec.nodeName
 ```
 
-To track metrics for unassigned pods, you need to add an additional deployment and set `--enable-no-node-scrape`, as shown in the following example:
+To track metrics for unassigned pods, you need to add an additional deployment and set `--fetch-unscheduled-pods`, as shown in the following example:
 
 ```
 apiVersion: apps/v1
@@ -295,7 +295,7 @@ spec:
         name: kube-state-metrics
         args:
         - --resources=pods
-        - --enable-no-node-scrape
+        - --fetch-unscheduled-pods
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ spec:
               fieldPath: spec.nodeName
 ```
 
-To track metrics for unassigned pods, you need to add an additional deployment and set `--enable-unscheduled-pods-fetching`, as shown in the following example:
+To track metrics for unassigned pods, you need to add an additional deployment and set `--track-unscheduled-pods`, as shown in the following example:
 
 ```
 apiVersion: apps/v1
@@ -295,7 +295,7 @@ spec:
         name: kube-state-metrics
         args:
         - --resources=pods
-        - --enable-unscheduled-pods-fetching
+        - --track-unscheduled-pods
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ spec:
               fieldPath: spec.nodeName
 ```
 
-To track metrics for unassigned pods, you need to add an additional deployment and set `--node=""`, as shown in the following example:
+To track metrics for unassigned pods, you need to add an additional deployment and set `--node=`, as shown in the following example:
 
 ```
 apiVersion: apps/v1
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ spec:
               fieldPath: spec.nodeName
 ```
 
-To track metrics for unassigned pods, you need to add an additional deployment and set `--fetch-unscheduled-pods`, as shown in the following example:
+To track metrics for unassigned pods, you need to add an additional deployment and set `--enable-unscheduled-pods-fetching`, as shown in the following example:
 
 ```
 apiVersion: apps/v1
@@ -295,7 +295,7 @@ spec:
         name: kube-state-metrics
         args:
         - --resources=pods
-        - --fetch-unscheduled-pods
+        - --enable-unscheduled-pods-fetching
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
--- a/README.md
+++ b/README.md
@@ -282,7 +282,7 @@ spec:
               fieldPath: spec.nodeName
 ```
 
-To track metrics for unassigned pods, you need to add an additional deployment and set `--node=`, as shown in the following example:
+To track metrics for unassigned pods, you need to add an additional deployment and set `--enable-no-node-scrape`, as shown in the following example:
 
 ```
 apiVersion: apps/v1
@@ -295,7 +295,7 @@ spec:
         name: kube-state-metrics
         args:
         - --resources=pods
-        - --node=""
+        - --enable-no-node-scrape
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,7 +48,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
+      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,7 +48,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
+      --track-unscheduled-pods           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,6 +48,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+      --enable-no-node-scrape                      This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,7 +48,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-      --fetch-unscheduled-pods                     This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
+      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,7 +48,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-      --enable-no-node-scrape                      This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
+      --fetch-unscheduled-pods                     This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
--- a/docs/developer/cli-arguments.md
+++ b/docs/developer/cli-arguments.md
@@ -48,7 +48,6 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
-      --track-unscheduled-pods           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
   -h, --help                                       Print Help text
       --host string                                Host to expose metrics on. (default "::")
       --kubeconfig string                          Absolute path to the kubeconfig file
@@ -82,6 +81,7 @@ Flags:
       --telemetry-port int                         Port to expose kube-state-metrics self metrics on. (default 8081)
       --tls-config string                          Path to the TLS configuration file
       --total-shards int                           The total number of shards. Sharding is disabled when total shards is set to 1. (default 1)
+      --track-unscheduled-pods                     This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
       --use-apiserver-cache                        Sets resourceVersion=0 for ListWatch requests, using cached resources from the apiserver instead of an etcd quorum read.
   -v, --v Level                                    number for the log level verbosity
       --vmodule moduleSpec                         comma-separated list of pattern=N settings for file-filtered logging
--- a/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
@@ -377,7 +377,7 @@
       local c = ksm.deployment.spec.template.spec.containers[0] {
         args: [
           '--resources=pods',
-          '--enable-no-node-scrape',
+          '--fetch-unscheduled-pods',
         ],
         name: shardksmname,
       };
@@ -410,7 +410,7 @@
       local c = ksm.deployment.spec.template.spec.containers[0] {
         args: [
           '--resources=pods',
-          '--enable-no-node-scrape',
+          '--fetch-unscheduled-pods',
         ],
       };
       local shardksmname = ksm.name + "-no-node-pods";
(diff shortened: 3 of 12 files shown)


mrueg:
Will this supersede https://github.com/kubernetes/kube-state-metrics/pull/2373 ?

CatherineF-dev:
> Will this supersede https://github.com/kubernetes/kube-state-metrics/pull/2373 ?

Yes.

pkg/options/types.go
```diff
@@ -145,11 +144,11 @@ func (n *NodeType) GetNodeFieldSelector() string {
 		klog.InfoS("Using node type is nil")
 		return EmptyFieldSelector()
 	}
-	pattern := "[^a-zA-Z0-9_,-]+"
-	re := regexp.MustCompile(pattern)
-	result := re.ReplaceAllString(n.String(), "")
-	klog.InfoS("Using node type", "node", result)
-	return fields.OneTermEqualSelector("spec.nodeName", result).String()
+	nodeName := n.String()
+	if nodeName == "" {
```
mrueg, line 148: Can you add a comment here, why we're handling this case specially?

CatherineF-dev: Added.

mrueg:
/hold 
for others to review.
/lgtm

logicalhan (commented):
tests?

diranged:
@logicalhan Can we just skip that for now? We've been waiting for 6 weeks to https://github.com/kubernetes/kube-state-metrics/issues/2374 resolved, and we're having _daily_ pain points with KSM right now in our environment. While I appreciate adding tests, we're literally removing code and making it simpler here.. can we bypass this and just get a fix in ASAP?

logicalhan:
> @logicalhan Can we just skip that for now? We've been waiting for 6 weeks to https://github.com/kubernetes/kube-state-metrics/issues/2374 resolved, and we're having _daily_ pain points with KSM right now in our environment. While I appreciate adding tests, we're literally removing code and making it simpler here.. can we bypass this and just get a fix in ASAP?

How do you know the fix works without tests?

diranged:
> > @logicalhan Can we just skip that for now? We've been waiting for 6 weeks to #2374 resolved, and we're having _daily_ pain points with KSM right now in our environment. While I appreciate adding tests, we're literally removing code and making it simpler here.. can we bypass this and just get a fix in ASAP?
> 
> How do you know the fix works without tests?

I think this is one of those vanity test situations... I don't see it worth blocking on this just to get a code coverage line. If you have a test proposal, would you put it up? Otherwise, can we get this feature fixed since it's been broken?

logicalhan:
> I think this is one of those vanity test situations... I don't see it worth blocking on this just to get a code coverage line. If you have a test proposal, would you put it up? Otherwise, can we get this feature fixed since it's been broken?

This bug exists because we didn't have test coverage what the heck are you talking about. The answer is no.

logicalhan:
If you want it so badly @diranged then you should write some tests. We accept pulls.

diranged:
> If you want it so badly @diranged then you should write some tests. We accept pulls.

Given that I'm not deeply familiar with the code - that seems like a poor idea. I wish you would reconsider here how much you want to stand on the 'needs tests' statement for such a simple PR.

mrueg:
@CatherineF-dev do you have plans to add tests to this PR to get it into a mergable state?

CatherineF-dev:
Okay! Work in Progress

tests/e2e.sh
```diff
@@ -97,24 +98,45 @@ function kube_state_metrics_up() {
...
+    kubectl apply -f ./tests/e2e/testdata/pods.yaml
+    sleep 10
+    kubectl get deployment -n default
+    curl -s "http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state-metrics-shard:http-metrics/proxy/metrics" >${KUBE_STATE_METRICS_LOG_DIR}/daemonset-scraped-metrics
+    curl -s "http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state-metrics:http-metrics/proxy/metrics" >${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-metrics
+    curl -s "http://localhost:8001/api/v1/namespaces/kube-system/services/kube-state-metrics-no-node-pods:http-metrics/proxy/metrics" >${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-no-node-metrics
+
+    cat ${KUBE_STATE_METRICS_LOG_DIR}/daemonset-scraped-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
+    cat ${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
+    cat ${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-no-node-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
```
LaikaN57, line 116:
```suggestion
    cat ${KUBE_STATE_METRICS_LOG_DIR}/daemonset-scraped-metrics ${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-metrics ${KUBE_STATE_METRICS_LOG_DIR}/deployment-scraped-no-node-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
```

Fixes:
```
In ./tests/e2e.sh line 114:
    cat ${KUBE_STATE_METRICS_LOG_DIR}/daemonset-scraped-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
    ^-- SC21[29](https://github.com/kubernetes/kube-state-metrics/actions/runs/9390258474/job/25859669104?pr=2388#step:5:30) (style): Consider using { cmd1; cmd2; } >> file instead of individual redirects.

For more information:
  https://www.shellcheck.net/wiki/SC2129 -- Consider using { cmd1; cmd2; } >>...In ./tests/e2e.sh line 114:
    cat ${KUBE_STATE_METRICS_LOG_DIR}/daemonset-scraped-metrics >> ${KUBE_STATE_METRICS_LOG_DIR}/all-pod-metrics
    ^-- SC21[29](https://github.com/kubernetes/kube-state-metrics/actions/runs/9390258474/job/25859669104?pr=2388#step:5:30) (style): Consider using { cmd1; cmd2; } >> file instead of individual redirects.

For more information:
  https://www.shellcheck.net/wiki/SC2129 -- Consider using { cmd1; cmd2; } >>...
```

https://github.com/kubernetes/kube-state-metrics/actions/runs/9390258474/job/25859669104?pr=2388#step:5:28

pkg/options/types.go
```diff
@@ -141,15 +140,17 @@ func (n *NodeType) Type() string {
 
 // GetNodeFieldSelector returns a nodename field selector.
 func (n *NodeType) GetNodeFieldSelector() string {
-	if nil == n || len(*n) == 0 {
+	if nil == n {
```
LaikaN57, line 143: To make the test pass and keep most of the logic, I think we can just revert this line.

```suggestion
	if nil == n || len(*n) == 0 {
```

I was trying to look at the usage of this function in [pkg/app/server.go#L208](https://github.com/kubernetes/kube-state-metrics/blob/7995d5fd23bcff7ae24ab6849f7c393d262fb025/pkg/app/server.go#L208).

... However, I still don't fully understand what the intended behavior is when `--node` is not provided.

LaikaN57: After reading https://github.com/kubernetes/kube-state-metrics/pull/2359 I believe this is the correct fix.

data.yaml
```diff
@@ -1,7 +1,7 @@
 # The purpose of this config is to keep all versions in a single file and make them machine accessible
 
 # Marks the latest release
-version: "2.12.0"
+version: "2.12.1"
```
mrueg, line 4: Please do not bump a version in a bugfix.

tests/e2e/testdata/pods.yaml
```diff
@@ -0,0 +1,75 @@
...
+        app: pendingpod2
+    spec:
+      nodeSelector:
+        debug-node: "non"
+      containers:
+      - command:
+        - /agnhost
+        - netexec
+        - --http-port=8080q
+        image: registry.k8s.io/e2e-test-images/agnhost:2.39
```
mrueg, line 30: where is this image from? will the cluster attempt to download it?

data.yaml
CatherineF-dev: Ok, I was testing whether e2e is using latest KSM or not.

tests/e2e/testdata/pods.yaml
```diff
@@ -0,0 +1,75 @@
...
+      labels:
+        app: pendingpod2
+    spec:
+      nodeSelector:
+        debug-node: "non"
+      containers:
+      - command:
+        - /agnhost
+        - netexec
+        - --http-port=8080q
```
LaikaN57, line 29: typo?

```
$ docker run -it --rm -p 8080:8080 registry.k8s.io/e2e-test-images/agnhost:2.39 netexec --http-port=8080q
Error: invalid argument "8080q" for "--http-port" flag: strconv.ParseInt: parsing "8080q": invalid syntax
```

LaikaN57, line 29:
```suggestion
        - --http-port=8080
```

tests/e2e/testdata/pods.yaml
LaikaN57: > where is this image from?

The question is a bit unclear but, [kubernetes/kubernetes (/test/images/agnhost)](https://github.com/kubernetes/kubernetes/tree/master/test/images/agnhost#image) seams to be the source.

> will the cluster attempt to download it?

Again, the question is a bit unclear but, no end-user clusters will not download this; it is just used for e2e testing AFAIK.

mrueg: Apologies for not being precise, the reason I was asking mainly is:
* Will this image be downloaded to a node? (= can I run e2e tests in an airgapped environment, or when developing on a plane?)
* Is the image source stable (= will CI integration tests break)?

LaikaN57:
@CatherineF-dev Is there anything I can do to help here? We had good momentum last week and I am looking to help continue that!

How do you feel about my comments for fixing the unit tests? (ie. adding back `|| len(*n) == 0` check)

I naively feel that fixing that will help out with e2e tests but I'm sure you have way more context working on this and other related code / architecture.

CatherineF-dev:
I was looking at the E2E test failure https://github.com/kubernetes/kube-state-metrics/actions/runs/9417050145/job/25941445490?pr=2388

docs/developer/cli-arguments.md
```diff
@@ -46,6 +46,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+      --enable-no-node-scrape                      This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
```
mrueg, line 49: Suggestion: Move this closer to the `--node` argument and rename from no_node to unscheduled_pods or similar?

```suggestion
      --node-sharding-unscheduled-pods-only                     This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is scraped from the Kubernetes API. This is experimental.
```

README.md
```diff
@@ -293,7 +293,7 @@ spec:
         name: kube-state-metrics
         args:
         - --resources=pods
-        - --node=""
+        - --fetch-unscheduled-pods
```
mrueg, line 296: If you don't want to prefix it by --node (no strong opinion here), can we use something like: 
```suggestion
        - --enable-unscheduled-pods-fetching
```
Using gerund instead of infinitive style (not to mix both)

docs/developer/cli-arguments.md
```diff
@@ -46,6 +46,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+      --fetch-unscheduled-pods                     This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
```
mrueg, line 49: Should the description make it more clear that this is only useful if you're doing sharding?

pkg/options/types_test.go
```diff
@@ -162,37 +162,38 @@ func TestNodeFieldSelector(t *testing.T) {
...
+			t.Errorf("Test error for Desc: %s. Want: %+v. Got: %+v.", test.Desc, test.Wanted, actual)
+		}
+	}
+	tests1 := []struct {
+		Desc   string
+		Node   NodeType
+		Wanted string
+	}{
+		{
+			Desc:   "empty node name",
```
mrueg, line 189: Can we add a few more tests here? What happens if I set --node and the --fetch-unscheduled-pods argument at the same time? Should it warn about this?

CatherineF-dev:
@LaikaN57 fixed all tests now.

pkg/options/types.go
```diff
@@ -106,56 +105,21 @@ func (r *ResourceSet) Type() string {
...
-func (n *NodeType) Set(value string) error {
-	s := *n
-	cols := strings.Split(value, ",")
-	for _, col := range cols {
-		col = strings.TrimSpace(col)
-		if len(col) != 0 {
-			s[col] = struct{}{}
+// GetNodeFieldSelector returns a nodename field selector.
+func (n *NodeType) GetNodeFieldSelector(noNodeAssigned bool) string {
+	if noNodeAssigned {
```
LaikaN57, line 112:
```suggestion
func (n *NodeType) GetNodeFieldSelector(fetchUnscheduledPods bool) string {
	if fetchUnscheduledPods {
```

nit: Function argument name was a bit unclear. It was describing the effect of the argument not the cause/condition it represented.

pkg/options/types.go
```diff
@@ -106,56 +105,21 @@ func (r *ResourceSet) Type() string {
 }
 
 // NodeType represents a nodeName to query from.
-type NodeType map[string]struct{}
+type NodeType string
```
LaikaN57, line 108: Because this is no longer a complex type should we just remove this type alias so we don't have to cast in other places?

LaikaN57:
Thanks @CatherineF-dev! I am OOO until Monday, but I did a quick review of the new code. Just 2 small nits but nothing blocking. I can review more fully on Monday if others do not approve by then. 🤞 I relly appreciate all the work you put into this!

docs/developer/cli-arguments.md
```diff
@@ -46,6 +46,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of no scheduled pods is scraped. This is experimental.
```
mrueg, line 49:
```suggestion
      --enable-unscheduled-pods-fetching           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
```

jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -377,18 +377,27 @@
       local c = ksm.deployment.spec.template.spec.containers[0] {
         args: [
           '--resources=pods',
-          '--node=""',
+          '--enable-unscheduled-pods-fetching',
         ],
+        name: shardksmname,
       };
-      local shardksmname = ksm.name + "-pods";
+      local shardksmname = ksm.name + "-no-node-pods";
```
mrueg, line 384:
```suggestion
      local shardksmname = ksm.name + "-unscheduled-pods-fetching";
```

CatherineF-dev:
@mrueg need a review again. Rebased to HEAD.

LaikaN57:
> tests?

@logicalhan Tests now implemented thanks to all of @CatherineF-dev's hard work 🙏 Looks like we are looking for review again now. Thanks!

pkg/options/types.go
rexagod: This seems necessary since there seem to be methods defined on it below.

pkg/options/types.go
```diff
@@ -106,56 +105,21 @@ func (r *ResourceSet) Type() string {
...
 
-// Set converts a comma-separated string of nodename into a slice and appends it to the NodeList
-func (n *NodeType) Set(value string) error {
```
rexagod, line 112: Let's keep the [`flag.Flag`](https://pkg.go.dev/flag#Value) behavior.

tests/e2e/testdata/pods.yaml
rexagod: I believe the image is stable ([scheduled releases](https://github.com/kubernetes/kubernetes/commits/master/test/images/agnhost/VERSION)), and safe for us to use in the given context. LMK if I missed anything though.

pkg/options/options.go
```diff
@@ -136,6 +136,7 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
 
 	o.cmd.Flags().BoolVar(&o.CustomResourcesOnly, "custom-resource-state-only", false, "Only provide Custom Resource State metrics (experimental)")
 	o.cmd.Flags().BoolVar(&o.EnableGZIPEncoding, "enable-gzip-encoding", false, "Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.")
+	o.cmd.Flags().BoolVar(&o.EnableUnscheduledPodsFetching, "enable-unscheduled-pods-fetching", false, "This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.")
```
dgrisonnet, line 139: Thinking about it again, would it really make sense for any user to want to opt-out of collection metrics from unscheduled pods?

As a user, I would expect to collect the same metrics whether I am using uuid sharding, node sharding or even an unsharded installation of ksm. So based on that I am wondering if we should really make this a flag and instead just enable that by default when node sharding is enabled.

LaikaN57: I think the issue is that as soon as you node shard, the current code will only select node-bound pods. Therefore you would run an additional deployment of KSM with this option enabled specifically. I don't think there is an easy way in the current codebase for us to filter pods on the current node and unbound pods at the same time. But I will leave this to the experts, this was just my current understanding.

dgrisonnet: Ah indeed you are correct, I forgot we had to differentiate it from the other shards.

mrueg: So we should still set it to true by default and disable when sharding?

pkg/options/types.go
```diff
@@ -106,56 +105,21 @@ func (r *ResourceSet) Type() string {
...
-			s[col] = struct{}{}
+// GetNodeFieldSelector returns a nodename field selector.
+func (n *NodeType) GetNodeFieldSelector(fetchUnscheduledPods bool) string {
+	if fetchUnscheduledPods {
+		if string(*n) != "" {
+			klog.Warningf("spec.nodeName=%s will not be used, because --enable-unscheduled-pods-fetching is set", string(*n))
 		}
+		klog.InfoS("Using spec.nodeName= to select unscheduable pods without node")
+		return "spec.nodeName="
 	}
```
dgrisonnet, line 118: It feels odd to handle this here since fetchUnscheduledPods is not related to the node object. I think we should take care of this scenario in the caller `RunKubeStateMetrics`

pkg/options/types.go
rexagod: It seems https://github.com/kubernetes/kube-state-metrics/pull/2388#discussion_r1655039540 was resolved without a fix? Can you PTAL?

docs/developer/cli-arguments.md
```diff
@@ -46,6 +46,7 @@ Flags:
       --custom-resource-state-config-file string   Path to a Custom Resource State Metrics config file (experimental)
       --custom-resource-state-only                 Only provide Custom Resource State metrics (experimental)
       --enable-gzip-encoding                       Gzip responses when requested by clients via 'Accept-Encoding: gzip' header.
+      --track-unscheduled-pods           This configuration is used in conjunction with node configuration. When this configuration is true, node configuration is empty and the metric of unscheduled pods is fetched from the Kubernetes API Server. This is experimental.
```
dgrisonnet, line 49: format

rexagod (changes requested):
PTAL https://github.com/kubernetes/kube-state-metrics/pull/2388#discussion_r1655039540.

pkg/options/types.go
LaikaN57: > Let's keep the [`flag.Flag`](https://pkg.go.dev/flag#Value) behavior.

Did you mean [`flag.Value`](https://pkg.go.dev/flag#Value)?

rexagod: Right, sorry, I meant that.

mrueg:
Heads up: This is currently listed for the v2.13 milestone. In order to get a release out soon, I'll drop all open PRs by Wednesday from that milestone.

CatherineF-dev:
@rexagod have fixed the change request.  Thx! Need a lgtm

CatherineF-dev:
@rexagod fixed.

dgrisonnet:
LGTM

pkg/options/options.go
```diff
@@ -156,7 +157,7 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
 	o.cmd.Flags().StringVar(&o.TLSConfig, "tls-config", "", "Path to the TLS configuration file")
 	o.cmd.Flags().StringVar(&o.TelemetryHost, "telemetry-host", "::", `Host to expose kube-state-metrics self metrics on.`)
 	o.cmd.Flags().StringVar(&o.Config, "config", "", "Path to the kube-state-metrics options config file")
-	o.cmd.Flags().Var(&o.Node, "node", "Name of the node that contains the kube-state-metrics pod. Most likely it should be passed via the downward API. This is used for daemonset sharding. Only available for resources (pod metrics) that support spec.nodeName fieldSelector. This is experimental.")
+	o.cmd.Flags().StringVar((*string)(&o.Node), "node", "", "Name of the node that contains the kube-state-metrics pod. Most likely it should be passed via the downward API. This is used for daemonset sharding. Only available for resources (pod metrics) that support spec.nodeName fieldSelector. This is experimental.")
```
mrueg, line 160: Is the type conversion needed?

jeffmccune:
@CatherineF-dev @mrueg @dgrisonnet FYI, I ran into some very easy to overlook syntax errors in this CL that eagle-eyed `jsonnet` complained loudly about when trying to customize kube-prometheus.

 Please take a look at https://github.com/kubernetes/kube-state-metrics/pull/2454 for a quick fix.

PR #2395, opened 2024-05-22
ci: least privileged GITHUB_TOKEN permissions

ricardoapl:

The `GITHUB_TOKEN` is an automatically generated secret to make authenticated calls to the GitHub API. GitHub recommends setting minimum token permissions for the `GITHUB_TOKEN`, otherwise attackers may use a compromised token with write access to, for example, push malicious code into the project.

Found in [CLOMonitor Checks](https://clomonitor.io/projects/cncf/kube-state-metrics#kube-state-metrics_security) and [OpenSSF Scorecard Report](https://securityscorecards.dev/viewer/?uri=github.com/kubernetes/kube-state-metrics).

Does not change cardinality

Part of #2274

### DIFF
--- a/.github/workflows/openvex.yml
+++ b/.github/workflows/openvex.yml
@@ -4,6 +4,10 @@ on:
   workflow_dispatch:
   release:
     types: [published]
+
+permissions:
+  contents: read
+
 jobs:
   vexctl:
     runs-on: ubuntu-latest


mrueg:
If https://github.com/kubernetes/kube-state-metrics/pull/2398 is getting merged, will this workflow work with only read permissions?

ricardoapl:
> If #2398 is getting merged, will this workflow work with only read permissions?

That's a good point, the workflow will not work with read only permissions then.

I suggest we keep default top-level permissions to read-only, but add `contents: write` permissions at job-level to #2398.

Although perhaps unlikely, there's a chance that a new job is added to this workflow, and its permissions could be left undefined because of human error.

What do you think?

mrueg:
Sounds good to me, let's mark this read-only for now.

/lgtm

PR #2396, opened 2024-05-22
chore: pin dependencies in GitHub Actions by hash

ricardoapl:

> Pinned dependencies reduce several security risks:
> - They ensure that checking and deployment are all done with the same software, reducing deployment risks, simplifying debugging, and enabling reproducibility.
> - They can help mitigate compromised dependencies from undermining the security of the project (in the case where you've evaluated the pinned dependency, you are confident it's not compromised, and a later version is released that is compromised).
> 
> (...)

Found in [OpenSSF Scorecard Report](https://securityscorecards.dev/viewer/?uri=github.com/kubernetes/kube-state-metrics).

Does not change cardinality

Part of #2274

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -29,10 +29,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
     - name: Check out code into the Go module directory
-      uses: actions/checkout@v4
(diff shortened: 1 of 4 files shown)


ricardoapl:
There are 4 more dependencies that we should probably pin by hash, but those will require a little bit more effort:

image

I suggest we do those in a separate pull request.

PR #2415, opened 2024-06-07
feat: Promote kube_statefulset_ordinals_start from ALPHA -> STABLE

pwschuurman:
Promote `kube_statefulset_ordinals_start` metrics from ALPHA -> STABLE
 * KEP: https://github.com/kubernetes/enhancements/issues/3335

Does not change the cardinality of KSM


mrueg:
I don't have any strong opinions here, I believe we haven't used beta before and went directly to stable. 
Any thoughts on this from people that care more about metrics stability? @dgrisonnet @rexagod @CatherineF-dev @logicalhan

dgrisonnet:
/triage accepted

I think this should wait for the next major since the current stability model only used `Experimental` and `Stable`.

pwschuurman:
> I think this should wait for the next major since the current stability model only used Experimental and Stable.

Thanks. @dgrisonnet Just to clarify, would you be OK with this PR being updated to use Stable instead of Beta? Or are you suggesting that this PR (promoting `kube_statefulset_ordinals_start`) should wait until the next major release?

pwschuurman:
Hi @dgrisonnet @mrueg, any guidance on how to proceed with this PR? Sounds like the options are:
1. Merge as-is with Stable
2. Wait for next major release and promote from Alpha -> Beta

dgrisonnet:
I think we should wait for either the next major to update to the correct stability level at the time or wait for the feature to become stable before making the metric stable here.

My reasoning is that at the moment there is only Experimental (Alpha) and Stable in kube-state-metrics and there is always the possibility that a Beta feature in k/k never reaches Stable and gets deprecated/removed.

pwschuurman:
Great, thanks @dgrisonnet. https://github.com/kubernetes/enhancements/issues/3335 is on track to be Stable in 1.31. I'll follow-up after 1.31 is released.

atiratree:
@pwschuurman I suppose, this can proceed for 1.32 now, right?

dgrisonnet:
@pwschuurman can you please update the PR to mark the metric as Stable?

mrueg:
@pwschuurman any interest in continuing on this one before we release v2.14?

pwschuurman:
@mrueg Yes, interested. Will update this today.

pwschuurman:
@mrueg @dgrisonnet Updated PR and description

mrueg:
/lgtm

Thanks

PR #2418, opened 2024-06-12
enhancement: add `livez` endpoint

rexagod:
Add a `livez` endpoint to identify network outages. This helps in restarting the binary if such as case is observed.


pkg/app/server.go
```diff
@@ -400,8 +402,23 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
...
 	mux.Handle("/debug/pprof/trace", http.HandlerFunc(pprof.Trace))
 
+	// Add metricsPath
 	mux.Handle(metricsPath, promhttp.InstrumentHandlerDuration(durationObserver, m))
 
+	// Add livezPath
+	mux.Handle(livezPath, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+
+		// Query the Kube API to make sure we are not affected by a network outage.
+		got := client.CoreV1().RESTClient().Get().AbsPath("/apis/").Do(context.Background())
```
mrueg, line 412: Should we query Kubernetes' API readyz endpoint https://kubernetes.io/docs/reference/using-api/health-checks/ instead to detect an outage of the kubernetes apiserver?

rexagod: +1, API (`/apis`) discoverability used here is a subset of `/livez`. I'll make the changes.

mrueg: Should this be `/livez` or rather `/readyz` ? 

The Kubernetes API Server could be healthy (/livez = true) but not be able to accept client request (/ready = false).

rexagod: `/livez` knows when to restart the container, and thus knows if there's been an outage. Besides, it makes sense to point the our `/livez` to the cluster's `/livez` to ensure the same thing.

`readinessProbe` is currently set to the telemetry metrics' availability in our [`jsonnet` config](https://github.com/kubernetes/kube-state-metrics/blob/6de105ebbe0eeb5d97d9e6adf1cc83314434e8cc/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet#L198-L201), which seems okay. A more robust approach would be coupling that with the cluster's `/readyz` and exposing that under a dedicated `/readyz` (we don't have a dedicated endpoint just yet). This would mean that (a) the cluster components are ready, and (b) the binary itself is ready. I can open another PR for that.

mrueg: Sounds good to me! Let's open another PR for that

mrueg: One question: We already have a /healthz path, what would this /livez path then be used for?

rexagod: `/healthz` will send out a `200` if the binary is running, `/readyz` will send out a `200` if the exposition machinery is working as expected (we are ready to serve metrics), and `/livez` will send out a `200` if none of the collectors are affected by any outages (collectors depend on the Kube API to gather data).

A `/healthz` endpoint would be more suitable for a `startupProbe`, which is especially useful if we believe the binary takes a considerable time to start.

mrueg: Thanks for clarifying! Could you add this to the README and update our jsonnet as well?

README.md
```diff
@@ -342,6 +342,13 @@ Note that your GCP identity is case sensitive but `gcloud info` as of Google Clo
 
 After running the above, if you see `Clusterrolebinding "cluster-admin-binding" created`, then you are able to continue with the setup of this service.
 
+#### Probes
+
+The following probes are available, and follow the [Kubernetes best practices](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes):
+
+* `livenessProbe`: Checks if the application is not affected by an outage, and is able to access the Kube API by querying the cluster's `/livez` endpoint.
+* `readinessProbe`: Checks if the application is ready to serve metrics by querying its own telemetry metrics.
```
mrueg, line 350: Do we have one already?

README.md
```diff
@@ -342,6 +342,13 @@ Note that your GCP identity is case sensitive but `gcloud info` as of Google Clo
 
 After running the above, if you see `Clusterrolebinding "cluster-admin-binding" created`, then you are able to continue with the setup of this service.
 
+#### Probes
```
mrueg, line 345:
```suggestion
#### Health endpoints
```

rexagod: I'd prefer the former since per-port endpoints are not consistent across the binary and its telemetry expositions. Also, we don't have a dedicated `/readyz` but the `readinessProbe` makes use of the telemetry port's `/metrics` to determine if we are able to serve requests.

rexagod: Friendly ping. :)

rexagod: Added a section for endpoints as well, PTAL.

README.md
```diff
@@ -342,6 +342,21 @@ Note that your GCP identity is case sensitive but `gcloud info` as of Google Clo
...
+The following probes are available, and follow the [Kubernetes best practices](https://kubernetes.io/docs/concepts/workloads/pods/pod-lifecycle/#container-probes):
+
+* `livenessProbe`: Checks if the application is not affected by an outage, and is able to access the Kube API by querying the cluster's `/livez` endpoint.
+* `readinessProbe`: Checks if the application is ready to serve metrics by querying its own telemetry metrics.
+
+#### Healthcheck Endpoints
+
+The following healthcheck endpoints are available, some of which are used to determine the result of the aforementioned probes:
+
+* `/livez`: Returns a 200 status code if the application is not affected by an outage. This is used to determine the result of the liveness probe.
```
mrueg, line 356:
```suggestion
* `/livez`: Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to use this as a liveness probe.
```

README.md
```diff
@@ -342,6 +342,21 @@ Note that your GCP identity is case sensitive but `gcloud info` as of Google Clo
...
+* `/metrics`: Returns a 200 status code if the application is able to serve metrics. While this is available for both ports, only self metrics are used to determine the result of the readiness probe.
```
mrueg, line 357:
```suggestion
* `/metrics`: Returns a 200 status code if the application is able to serve metrics. While this is available for both ports, we recommend to use the telemetry metrics endpoint as a readiness probe.
```

README.md
```diff
@@ -342,6 +342,21 @@ Note that your GCP identity is case sensitive but `gcloud info` as of Google Clo
...
+* `/healthz`: Returns a 200 status code if the application is running.
```
mrueg, line 358:
```suggestion
* `/healthz`: Returns a 200 status code if the application is running. We recommend to use this as a startup probe.
```

README.md
mrueg: Should we remove this paragraph? I don't think it's specific to kube-state-metrics and having the endpoints documented is good enough. :)

rexagod: Done, PTAL.

jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
```diff
@@ -193,7 +193,7 @@
       },
       livenessProbe: { timeoutSeconds: 5, initialDelaySeconds: 5, httpGet: {
         port: 8080,
-        path: '/healthz',
+        path: '/livez',
       } },
       readinessProbe: { timeoutSeconds: 5, initialDelaySeconds: 5, httpGet: {
         port: 8081,
```
mrueg, line 199: Can you change the path below to /metrics ?

mrueg:
/lgtm 

thanks!

PR #2433, opened 2024-06-26
feat: add kube_ingress_status metric

### LINKED ISSUE #1366: Support for ingress status hostname
It would be great to have ability to collect metrics of ingress status from **status.loadBalancer.ingress[]** for ALB loadbalancer hostnames:
```yaml
kind: Ingress
metadata:
  annotations:
    alb.ingress.kubernetes.io/healthcheck-path: /health
    alb.ingress.kubernetes.io/inbound-cidrs: 10.0.0.0/8
    alb.ingress.kubernetes.io/listen-ports: '[{"HTTP":80}]'
    alb.ingress.kubernetes.io/scheme: internal
    kubernetes.io/ingress.class: public-alb
  generation: 1
  name: test
  namespace: kube-system
spec:
  backend:
    serviceName: my-service
    servicePort: 8080
status:
  loadBalancer:
    ingress:
    - hostname: internal-9a4003c3-kube-system-test-bbc4-1234567890.us-east-1.elb.amazonaws.com
```

isuyyy:
Collects metrics of ingress status. / The info is needed when we use ALB.

Increases.

Fixes #1366


internal/store/ingress.go
```diff
@@ -202,6 +202,28 @@ func ingressMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				}
 			}),
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_ingress_status",
+			"Ingress status.",
+			metric.Gauge,
+			basemetrics.STABLE,
```
richabanker, line 209: Perhaps should start with an `EXPERIMENTAL` metric https://github.com/kubernetes/kube-state-metrics/blob/main/docs/developer/guide.md#add-new-metrics ?
cc @CatherineF-dev

internal/store/ingress.go
```diff
@@ -202,6 +202,28 @@ func ingressMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
+			"",
+			wrapIngressFunc(func(i *networkingv1.Ingress) *metric.Family {
+				ms := []*metric.Metric{}
+				for _, ingress := range i.Status.LoadBalancer.Ingress {
+					for _, port := range ingress.Ports {
+						ms = append(ms, &metric.Metric{
+							LabelKeys:   []string{"ip", "hostname", "port", "protocol"},
```
richabanker, line 216: Do you need both the `ip` & the `hostname` ?

richabanker: Wondering since IPs can keep on changing so cardinality might explode. Just using hostname might suffice?

internal/store/ingress.go
isuyyy: Fixed here https://github.com/kubernetes/kube-state-metrics/pull/2433/commits/c88e3830358a860a0c5dd833c012d4a4ac7ee3fd

internal/store/ingress.go
isuyyy: At first, I just wanted to include all the fields, but as you mentioned, the hostname will suffice.
Made the changes here: https://github.com/kubernetes/kube-state-metrics/pull/2433/commits/622505e6ce1e0df535e61d69e636a5962c70ce50

richabanker:
/assign @dgrisonnet 
for approval

internal/store/ingress.go
mrueg: If we're removing IPs could we end up with duplicate metrics if multiple LB VIPs serve the same hostname (not sure if this is likely at all, maybe with multiple regionally deployed LBs)?

internal/store/ingress.go
```diff
@@ -202,6 +202,28 @@ func ingressMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				}
 			}),
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_ingress_status",
```
mrueg, line 206:
```suggestion
			"kube_ingress_status",
```
I'm a bit torn on this name.
Maybe:
kube_ingress_status_loadbalancer

internal/store/ingress_test.go
```diff
@@ -231,7 +233,42 @@ func TestIngressStore(t *testing.T) {
...
+						Ingress: []networkingv1.IngressLoadBalancerIngress{
+							{
+
+								IP:       "1.2.3.4",
+								Hostname: "www.example.com",
+								Ports: []networkingv1.IngressPortStatus{
+									{
+										Port:     8888,
+										Protocol: "TCP",
+										Error:    nil,
```
mrueg, line 257: Should we expose this error somehow as well? 

I assume it can potentially cause high cardinality, but it will be interesting to know if there's an error on the IngressPortStatus?

internal/store/ingress.go
logicalhan: I don't think we should include IPs (due to the cardinality concern that Richa pointed out).

> If we're removing IPs could we end up with duplicate metrics if multiple LB VIPs serve the same hostname (not sure if this is likely at all, maybe with multiple regionally deployed LBs)?

Won't we have unique instance IDs?

mrueg: With instance IDs you mean in this object https://pkg.go.dev/k8s.io/api/networking/v1#Ingress ?

I don't know if that is the case, I assume there's must be some reason why the object is so convoluted...

https://pkg.go.dev/k8s.io/api/networking/v1#IngressLoadBalancerStatus contains a list of https://pkg.go.dev/k8s.io/api/networking/v1#IngressLoadBalancerIngress (both IP and Hostname are optional, without having looked into the code, I assume you need to set either one or the other?)
which then contains a list of https://pkg.go.dev/k8s.io/api/networking/v1#IngressPortStatus

dgrisonnet: I think we should keep both `ip` and `hostname` otherwise this metric won't make much sense since we won't be able to distinguish between the load balancers.
If cardinality is a concern we can always have this metric disabled by default, but here I don't think it is a problem because the number of timeseries for this metric will be limited to the number of Endpoints in the cluster. This is just the theoretical worst case scenario, but in practice it will only be a subset of that since it only includes the endpoints of the loadbalancer. and since the number of Endpoints in a cluster is limited due to scalability limits, this metric will be bounded.

Disclaimer: I am not too knowledgeable about Ingress so for what I know it might just be bounded to the number of Services instead of Endpoints which is even better. It all depends on where the <ip/hostname>:<port> pair come from.

Issue #2044, opened 2023-04-12
Duplicated custom resource metrics when exposed for builtin type


grzesuav:
**What happened**:

Wanted to expose additional information for storageclass, expanded the config

```yaml
    kind: CustomResourceStateMetrics
    spec:
      resources:
        - groupVersionKind:
            group: storage.k8s.io
            kind: StorageClass
            version: v1
          metricNamePrefix: kube_storageclass
          metrics:
            - name: "parameters"
              help: "StorageClass parameters"
              each:
                type: Info
                info:
                  labelsFromPath:
                    skuName: [parameters, skuName]
                    storageclass: [metadata, name]
```

I noticed that they are exposed twice, which makes it impossible to be scraped by prometheus (metrics are duplicated)

```
# HELP kube_storageclass_parameters StorageClass parameters
# TYPE kube_storageclass_parameters info
kube_storageclass_parameters{customresource_group="storage.k8s.io",customresource_kind="StorageClass",customresource_version="v1",storageclass="standard"} 1
# HELP kube_storageclass_parameters StorageClass parameters
# TYPE kube_storageclass_parameters info
kube_storageclass_parameters{customresource_group="storage.k8s.io",customresource_kind="StorageClass",customresource_version="v1",storageclass="standard"} 1
```

**What you expected to happen**:

Metrics are exposed once

**How to reproduce it (as minimally and precisely as possible)**:

Use the config provided with k-s-m and kind cluster

**Anything else we need to know?**:
Full logs
```
I0412 11:54:47.701990       1 wrapper.go:98] "Starting kube-state-metrics"
I0412 11:54:47.702368       1 builder.go:192] "The internal resource store already exists and is overridden by a custom resource store with the same name, please make sure it meets your expectation" registryName="storageclasses"
I0412 11:54:47.702518       1 server.go:186] "Used default resources"
I0412 11:54:47.702577       1 types.go:184] "Using all namespaces"
I0412 11:54:47.702660       1 server.go:219] "Metric allow-denylisting" allowDenyStatus="Including the following lists that were on allowlist: kube_deployment_labels, kube_pod_labels, kube_storageclass_labels, kube_storageclass_parameters"
W0412 11:54:47.702744       1 client_config.go:618] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.
I0412 11:54:47.703382       1 server.go:364] "Tested communication with server"
[...]

PR #2435, opened 2024-06-28
fix: support empty groups again, e.g. Node

### LINKED ISSUE #2434: custom metrics for resources w/ group="" broken, e.g. Node
**What happened**:
Custom metrics for resources where group is empty, e.g. `Node` resource do not work. This was a regression between v2.8.2 and v2.9.0

Custom metrics should work even with an empty group.

Run kube-state-metrics v2.9.0, compare vs v2.8.2

```yaml
resources:
  - groupVersionKind:
      kind: Node
      version: v1
      group: ""
    labelsFromPath:
      node: [metadata, name]
    metricNamePrefix: demo
    metrics:
      - name: timestamp
        each:
          type: Gauge
          gauge:
            path: [metadata]
            valueFrom: [creationTimestamp]
```

Error:
```
"failed to resolve GVK" err="group is required in the defined GVK /v1, Kind=Node" gvk="_v1_Node"
```

This breakage was introduced accidently in https://github.com/kubernetes/kube-state-metrics/pull/1851

Specifically here:
https://github.com/kubernetes/kube-state-metrics/blob/dc86adec127688c7e02a71bb4d60b1f9f98f0f61/internal/discovery/discovery.go#L123-L125

This was previously reported https://github.com/kubernetes/kube-state-metrics/discussions/2333

Rolling KSM back to v2.8.2 and the custom metric DOES work, but that's not an option in production environment due to the need for RBAC features introduced 

**Environment**:
* kube-state-metrics version: v2.10.1
* Kubernetes version (use `kubectl version`): v1.25.1
* Cloud provider or hardware configuration: private
* Other info:

Screenshot of the metrics working in v2.8.2:
screenshot

robbat2:
PR#1851 introduced accidental breakage for custom metrics on resources with an empty empty, e.g. Node.

Fix it, and add tests to catch the next breakage.

Reference: https://github.com/kubernetes/kube-state-metrics/pull/1851
Reference: https://github.com/kubernetes/kube-state-metrics/issues/2434
Signed-off-by: Robin H. Johnson <rjohnson@coreweave.com>

Fixes breakage introduced in v2.9.0

Does not change cardinality.

Fixes #2434


dgrisonnet:
All custom resources have an API group. Resources who don't are native Kubernetes resources for which kube-state-metrics already provides a safe set of metrics that is frequently updated to support new users requests. The ability to extend the metrics exposed by kube-state-metrics was never meant for natives resources and that is not something we wanted to support going forward.

Some additional context about that decision was added in https://github.com/kubernetes/kube-state-metrics/issues/2044.

rexagod:
Exactly what Damien said.

Additionally, PTAL at the docs, https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-1e83b132955ba4b76949317660faf1fa222f0ee999d4fb3b4fb9dd318b7ac2beR521.

daveoy:
Could you point me in the direction of docs showing functionality such as this, then?

```
          - groupVersionKind:
              kind: Node
              version: v1
            labelsFromPath:
              node:
              - metadata
              - name
            metricNamePrefix: ""
            metrics:
            - name: last_timestamp_of_thing_that_happened
              help: last timestamp thing
              each:
                gauge:
                  path:
                  - metadata
                  - annotations
                  valueFrom:
                  - timestamp-value-annotation
                type: Gauge
```
this would twist the dimensions of a label from a node resource to a timestamp gauge value in a new metric which is extremely useful, in addition to it being both possible AND encouraged for custom resources. seems like a bit of a shame to not allow users to do this sort of thing with core/builtin resources?

rexagod:
Again, native resources are supported in-tree only, by design, and thus cannot leak into the CRS (_**Custom Resource**_ State) machinery. If you believe something needs to be added for them, please open a PR, just as the docs say in the aforementioned link.

daveoy:
just so im clear, this functionality isn't available for native resources somewhere else within ksm and we've just been trying to shoehorn the custom resource version of it into native resources?

is there any plan to replicate this functionality for native resources? or are we limited to the metrics and their dimensions as provided by this project for native resources? 

again, it seems a bit of a waste of great available functionality "restricting" this sort of metrics manipulation to only custom resources

robbat2:
As @daveoy notes, one of our use cases is exporting a timestamp value from a Node annotation. This is not possible via KSM since v2.9.0. It *was* previously available and worked in in v2.8.2 - from that perspective you broke existing functionality.

Another use case we have is splitting out a very large set of Node labels. 

We have Nodes with 100+ labels, over multiple namespaces, and even with metric-labels-allowlist mechanism, this causes cardinality problems in Prometheus.

We want to export each of those namespaces of labels as a distinct metric - to reduce the complexity of `kube_node_labels`, as the metrics consumers don't need all of them, just their own namespaces.

mrueg:
> As @daveoy notes, one of our use cases is exporting a timestamp value from a Node annotation. This is not possible via KSM since v2.9.0. It _was_ previously available and worked in in v2.8.2 - from that perspective you broke existing functionality.

The custom resource metrics feature is still experimental (See: 
https://github.com/kubernetes/kube-state-metrics/blob/main/docs/README.md#metrics-from-custom-resources ) so functionality might change over time as the feature evolves.

Please do not rely on an experimental feature in production systems.

robbat2:
Here's example of what we're doing to export the labels in something beyond `kube_node_labels`
```
resources:
  - groupVersionKind:
      kind: Node
      version: v1
    labelsFromPath:
      node: [metadata, name]
    metricNamePrefix: ""
    metrics:
      - name: kube_node_ns1_info
        help: ns1 example labels
        each:
          type: Info
          info:
            path: [metadata]
            labelsFromPath:
              ns1_example_com_foo: [labels, "ns1.example.com/foo"]
              ns1_example_com_bar: [labels, "ns1.example.com/bar"]
              ns1_example_com_baz: [labels, "ns1.example.com/baz"]
      - name: kube_node_ns2_info
        help: ns2 example labels
        each:
          type: Info
          info:
            path: [metadata]
            labelsFromPath:
              ns2_example_com_a: [labels, "ns2.example.com/a"]
              ns2_example_com_b: [labels, "ns2.example.com/b"]
              ns2_example_com_c: [labels, "ns2.example.com/c"]
```

robbat2:
@mrueg  the text says `See [Custom Resource State Metrics](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md) for experimental support for custom resources.`

These are core resources, and therefore not included in that definition.

If there's a better way to export the metrics from our use cases, please do share, because right now the v2.8.2 functionality looks to be the most fitting solution.

To recap:
- export *parts* of a core resource as a custom metric series: subsets of node & pod labels & annotations as metric labels in a distinct series, node annotation values as gauge values

What else does the PR need, in terms of supporting code and tests, to show this is valid & safe functionality? I did already include unit testing & E2E testing to show it works within the scope of the existing test suites.

bh-tt:
I opened the initial [discussion](https://github.com/kubernetes/kube-state-metrics/discussions/2333) asking how to get metrics for the core ResourceQuota resources. That resource specifically is designed to be extendable to all kinds of resource types (not just the standard cpu, memory and ephemeralStorage but for example GPU cores). In that discussion I also mentioned that supporting all those different options would likely be a significant maintenance burden on the project, aside from creating a variable number of metric names.

I understand the logic of not allowing the customresource setup to use core resources, but this usecase absolutely needs support. We frequently run into problems with resourcequotas blocking pod creation as our developers migrate more stuff to kubernetes. We would like to have monitoring alerting our IT department whenever resource usage goes > 80% of the quota, but this is currently not possible with KSM.

daveoy:
Are there office hours we can join to discuss this? Would love to come to some sort of agreement on how to move forward with allowing custom metrics for core resources.

mrueg:
> Are there office hours we can join to discuss this? Would love to come to some sort of agreement on how to move forward with allowing custom metrics for core resources. @mrueg / @rexagod please let me know.

There are SIG-Instrumentation office hours that you can join if you want to discuss this further. I am unable to attend those though as they conflict with my day job.

daveoy:
I started a thread here https://kubernetes.slack.com/archives/C20HH14P7/p1720615100646059

dgrisonnet:
Based on the discussion we had on slack, are we ok with closing this?

PR #2442, opened 2024-07-03
fix: add `readyz` endpoint

rexagod:
Discourage the usage of `/metrics` for any probe, and use `/readyz` in place of the earlier telemetry metrics endpoint to secure the exposition data.

/cc @mrueg

### DIFF
--- a/README.md
+++ b/README.md
@@ -346,11 +346,11 @@ After running the above, if you see `Clusterrolebinding "cluster-admin-binding" 
 
 #### Healthcheck Endpoints
 
-The following healthcheck endpoints are available, some of which are used to determine the result of the aforementioned probes:
+The following healthcheck endpoints are available (`self` refers to the telemetry port, while `main` refers to the exposition port):
 
-* `/healthz`: Returns a 200 status code if the application is running. We recommend to use this for the startup probe.
-* `/livez`: Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to using this for the liveness probe.
-* `/readyz`: Returns a 200 status code if the application is ready to accept traffic. We recommend using this for the readiness probe.
+* `/healthz` (exposed on `main`): Returns a 200 status code if the application is running. We recommend to use this for the startup probe.
+* `/livez` (exposed on `main`): Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to using this for the liveness probe.
+* `/readyz` (exposed on `self`): Returns a 200 status code if the application is ready to accept traffic. We recommend using this for the readiness probe.
 
 Note that it is discouraged to use the telemetry metrics endpoint for any probe when proxying the exposition data.
 
--- a/README.md
+++ b/README.md
@@ -348,9 +348,11 @@ After running the above, if you see `Clusterrolebinding "cluster-admin-binding" 
 
 The following healthcheck endpoints are available, some of which are used to determine the result of the aforementioned probes:
 
-* `/livez`: Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to use this as a liveness probe.
-* `/metrics`: Returns a 200 status code if the application is able to serve metrics. While this is available for both ports, we recommend to use the telemetry metrics endpoint as a readiness probe.
-* `/healthz`: Returns a 200 status code if the application is running. We recommend to use this as a startup probe.
+* `/healthz`: Returns a 200 status code if the application is running. We recommend to use this for the startup probe.
+* `/livez`: Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to using this for the liveness probe.
+* `/readyz`: Returns a 200 status code if the application is ready to accept traffic. We recommend using this for the readiness probe.
+
+Note that it is discouraged to use the telemetry metrics endpoint for any probe when proxying the exposition data.
 
 #### Limited privileges environment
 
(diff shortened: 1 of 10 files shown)


mrueg:
Is this passing through /readyz from the apiserver to ksm? I think that's not an ideal approach. It should only be ready if metrics are available.

rexagod:
Right, an oversight on my part. Patched.

pkg/app/server.go
```diff
@@ -376,6 +378,18 @@ func buildTelemetryServer(registry prometheus.Gatherer) *http.ServeMux {
 	// Add metricsPath
 	mux.Handle(metricsPath, promhttp.HandlerFor(registry, promhttp.HandlerOpts{ErrorLog: promLogger{}}))
 
+	// Add readyzPath
+	mux.Handle(readyzPath, http.HandlerFunc(func(w http.ResponseWriter, _ *http.Request) {
+		count, err := testutil.GatherAndCount(registry)
```
mrueg, line 383: I like the approach, I am a bit concerned that we're getting a testutil into a production path.
Is this considered stable?

mrueg: "Package testutil provides helpers to test code using the prometheus package of client_golang."

README.md
```diff
@@ -346,11 +346,13 @@ After running the above, if you see `Clusterrolebinding "cluster-admin-binding"
...
+The following healthcheck endpoints are available (`self` refers to the telemetry port, while `main` refers to the exposition port):
 
-* `/livez`: Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to use this as a liveness probe.
-* `/metrics`: Returns a 200 status code if the application is able to serve metrics. While this is available for both ports, we recommend to use the telemetry metrics endpoint as a readiness probe.
-* `/healthz`: Returns a 200 status code if the application is running. We recommend to use this as a startup probe.
+* `/healthz` (exposed on `main`): Returns a 200 status code if the application is running. We recommend to use this for the startup probe.
+* `/livez` (exposed on `main`): Returns a 200 status code if the application is not affected by an outage of the Kubernetes API Server. We recommend to using this for the liveness probe.
+* `/readyz` (exposed on `self`): Returns a 200 status code if the application is ready to accept traffic. We recommend using this for the readiness probe.
```
mrueg, line 353:
```suggestion
* `/readyz` (exposed on `self`): Returns a 200 status code if the application is ready to accept requests and expose metrics. We recommend using this for the readiness probe.
```

pkg/app/server.go
```diff
@@ -410,24 +437,10 @@ func buildMetricsServer(m *metricshandler.MetricsHandler, durationObserver prome
...
-		w.Write([]byte(http.StatusText(http.StatusOK)))
-	}))
+	mux.Handle(livezPath, handleClusterDelegationForProber(client, livezPath))
 
 	// Add healthzPath
-	mux.HandleFunc(healthzPath, func(w http.ResponseWriter, _ *http.Request) {
-		w.WriteHeader(http.StatusOK)
-		w.Write([]byte(http.StatusText(http.StatusOK)))
-	})
+	mux.Handle(healthzPath, handleClusterDelegationForProber(client, healthzPath))
```
mrueg, line 443: Why is healthz depending on the health of the apiserver?

rexagod: We're making `/healthz` depend on the cluster's health _in addition to_ the binary's. As is visible in the list below, there are multiple components in the cluster that are critical to KSM's health (for eg., `crd-informer-synced` and `start-apiextensions-informers`).

```
┌[rexagod@nebuchadnezzar] [/dev/ttys002] [readyz] 
└[~/repositories/oss/kube-state-metrics]> curl "127.0.0.1:8001/healthz?verbose"
[+]ping ok
[+]log ok
[+]etcd ok
[+]poststarthook/start-kube-apiserver-admission-initializer ok
[+]poststarthook/generic-apiserver-start-informers ok
[+]poststarthook/priority-and-fairness-config-consumer ok
[+]poststarthook/priority-and-fairness-filter ok
[+]poststarthook/storage-object-count-tracker-hook ok
[+]poststarthook/start-apiextensions-informers ok
[+]poststarthook/start-apiextensions-controllers ok
[+]poststarthook/crd-informer-synced ok
[+]poststarthook/start-system-namespaces-controller ok
[+]poststarthook/bootstrap-controller ok
[+]poststarthook/rbac/bootstrap-roles ok
[+]poststarthook/scheduling/bootstrap-system-priority-classes ok
[+]poststarthook/priority-and-fairness-config-producer ok
[+]poststarthook/start-cluster-authentication-info-controller ok
[+]poststarthook/start-kube-apiserver-identity-lease-controller ok
[+]poststarthook/start-deprecated-kube-apiserver-identity-lease-garbage-collector ok
[+]poststarthook/start-kube-apiserver-identity-lease-garbage-collector ok
[+]poststarthook/start-legacy-token-tracking-controller ok
[+]poststarthook/aggregator-reload-proxy-client-cert ok
[+]poststarthook/start-kube-aggregator-informers ok
[+]poststarthook/apiservice-registration-controller ok
[+]poststarthook/apiservice-status-available-controller ok
[+]poststarthook/kube-apiserver-autoregistration ok
[+]autoregister-completion ok
[+]poststarthook/apiservice-openapi-controller ok
[+]poststarthook/apiservice-openapiv3-controller ok
[+]poststarthook/apiservice-discovery-controller ok
healthz check passed
```

pkg/app/server.go
rexagod: I've extracted the function in a manner that makes it only dependent on the official library, PTAL.

pkg/app/server.go
mrueg: That doesn't make so much sense to me that /livez is on /livez and /healthz is on /healthz of the API Server.

In particular since /healthz is a deprecated endpoint.

```
The healthz endpoint is deprecated (since Kubernetes v1.16), and you should use the more specific livez and readyz endpoints instead.
```

https://kubernetes.io/docs/reference/using-api/health-checks/#api-endpoints-for-health

As we're suggesting healthz to be for the startup probe, I think it's fine to simply return 200 on /healthz once the application is running and not change the semantics there.

rexagod: I didn't realize this was deprecated. Fair enough. 👍🏼

PR #2451, opened 2024-07-19
ci: missing line break escape on sbom workflow

ricardoapl:

Generate SBOM with Kubernetes BOM workflow fails because of missing line break escape

https://github.com/kubernetes/kube-state-metrics/actions/runs/10009598032/job/27669716084#step:4:13122

Does not change cardinality

None


ricardoapl:
Confirmed to be working

https://github.com/ricardoapl/kube-state-metrics/actions/runs/10010867410/job/27673031687

mrueg:
/lgtm

Thanks!

PR #2452, opened 2024-07-23
WIP: feat: Support metrics for GatewayAPI objects

### LINKED ISSUE #2449: Support for Gateway API metrics
Expose metrics for objects that are part of the Gateway API https://gateway-api.sigs.k8s.io/implementations/v1.1/
**Why is this needed**:

To see what objects are in the data path for the new way of doing ingress.

* Use Custom Resource State Definition similar to HPA 
or
* Implement them as native resources in Kubernetes.

**Additional context**
-

mrueg:
GatewayAPI is becoming more popular and kube-state-metrics should support providing metrics and insights into it.
This should be disabled by default, and only enabled when the cluster admin wants to use it.
I expect this to be used in a lot of clusters in future, when Ingress is becoming deprecated.

New metrics for
- GatewayClass
- Gateway
- HTTPRoute
- TLSRoute

Not planned right now:
- TCPRoute
- UDPRoute

Fixes #2449


mrueg:
Closing this as we won't support non-native resources.

PR #2456, opened 2024-07-26
fix(Dockerfile): Install tools so VERSION gets set

### LINKED ISSUE #2455: Version is not applied to the container release binary
**What happened**:
https://github.com/kubernetes/kube-state-metrics/pull/2361/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R5 requires gomplate to be installed to read the version.

The Dockerfile currently does not run make install-tools to install gomplate
We should install them during the build process in the Dockerfile to ensure that the version is correctly read.

See: https://kubernetes.slack.com/archives/CJJ529RUY/p1721936633731669
**Environment**:

* kube-state-metrics version: v2.12.0+
* Other info:

CC: @Pokom

mrueg:

We require gomplate to be installed during build time. Ensure that this happens in the Dockerfile as well, as otherwise our release containers won't include the right version in the kube-state-metrics' metrics and output.

None
Fixes #2455

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -6,7 +6,7 @@ ENV GOARCH=${GOARCH}
 WORKDIR /go/src/k8s.io/kube-state-metrics/
 COPY . /go/src/k8s.io/kube-state-metrics/
 
-RUN make build-local
+RUN make install-tools && make build-local
 
 FROM gcr.io/distroless/static:latest-${GOARCH}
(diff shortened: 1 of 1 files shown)


CatherineF-dev:
/hold 
feel free to unhold

Issue #2460, opened 2024-07-31
kube-state-metrics jsonnet format wrong


wangjc0216:
**What happened**:

I can't build correct kube-state-metrics yaml  in  [ kube-prometheus customizing](https://github.com/prometheus-operator/kube-prometheus/blob/v0.13.0/docs/customizing.md) 

**What you expected to happen**:

I found the spelling mistake in this project , I want to fix it

**How to reproduce it (as minimally and precisely as possible)**:

Just replay in the [docs](https://github.com/prometheus-operator/kube-prometheus/blob/v0.13.0/docs/customizing.md)

```
$ mkdir my-kube-prometheus; cd my-kube-prometheus
$ jb init  # Creates the initial/empty `jsonnetfile.json`
# Install the kube-prometheus dependency
$ jb install github.com/prometheus-operator/kube-prometheus/jsonnet/kube-prometheus@main # Creates `vendor/` & `jsonnetfile.lock.json`, and fills in `jsonnetfile.json`

$ wget https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/example.jsonnet -O example.jsonnet
$ wget https://raw.githubusercontent.com/prometheus-operator/kube-prometheus/main/build.sh -O build.sh
$ chmod +x build.sh

$ ./build.sh 
# print error 
RUNTIME ERROR: vendor/github.com/kubernetes/kube-state-metrics/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet:392:21-22 Expected token OPERATOR but got "{"
	vendor/kube-prometheus/components/kube-state-metrics.libsonnet:51:19-124	function <anonymous>
	vendor/kube-prometheus/main.libsonnet:136:21-64	object <anonymous>
	vendor/kube-prometheus/platforms/platforms.libsonnet:37:22-40	+:
	example.jsonnet:33:90-109	thunk from <$>
	<std>:1539:24-25	thunk from <function <anonymous>>
	<std>:1539:5-33	function <anonymous>
	example.jsonnet:33:73-110	$
	example.jsonnet:33:1-112	
	example.jsonnet:33:1-112	
	During evaluation	


```

**Anything else we need to know?**:

**Environment**:  a pc with golang env

TimothySutton81:
[...]

PR #2461, opened 2024-07-31
fix:  modify kube-state-metrics.libsonnet  spelling mistake

wangjc0216:
I found kube-state-metris jsonnet format error when I build monitoring yaml  relating to kube-prometheus project.

**Which issue(s) this PR fixes**:  https://github.com/kubernetes/kube-state-metrics/issues/2460

### DIFF
--- a/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
+++ b/jsonnet/kube-state-metrics/kube-state-metrics.libsonnet
@@ -389,15 +389,15 @@
             labels: {'app.kubernetes.io/name': shardksmname}
           },
           spec: {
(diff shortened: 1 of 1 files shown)


mrueg:
There's an earlier fix in https://github.com/kubernetes/kube-state-metrics/pull/2454 which we would prefer to accept. maybe you can collaborate on it together?

wangjc0216:
> There's an earlier fix in #2454 which we would prefer to accept. maybe you can collaborate on it together?

Thank you for your feedback. I've reviewed the changes in #2454 and I can see how it relates to my PR.

PR #2464, opened 2024-08-05
chore: Avoid naked return

mrueg:
Make the code a bit more readable for folks not familiar with golang

None
Fixes #


rexagod:
/approve

Might be also worth wrapping naked errors for context (i.e., adding `wrapcheck` to https://github.com/kubernetes/kube-state-metrics/blob/d1e717911b537035bf74ed976b4af89843a74bc4/.golangci.yml#L6)?

mrueg:
> /approve
> 
> Might be also worth wrapping naked errors for context (i.e., adding `wrapcheck` to
> 
> https://github.com/kubernetes/kube-state-metrics/blob/d1e717911b537035bf74ed976b4af89843a74bc4/.golangci.yml#L6
> )?
sounds like a useful addition.
feel free to add it! :)

Issue #2463, opened 2024-08-05
Use PartialObjectMetadata for ConfigMaps and Secrets


mrueg:
**What would you like to be added**:
We should look into allowing to fetch only PartialObjectMetadata for these configmap/secrets objects.

**Why is this needed**:
ConfigMap objects can contain up to 1M of config or binary data. The metrics generated for the configmap is only relying on metadata. 
Using PartialObjectMetadata will reduce memory consumption of KSM and require less network traffic.

For Secrets I think this will add the benefit that the secret data never is requested over the wire.

**Describe the solution you'd like**
Probably this metadatalister can be used instead of the ListWatch we use for other objects. 
https://github.com/kubernetes/client-go/blob/71959c526d543a5e4c3ca6fb808f535c2726483f/metadata/metadatalister/lister.go

**Additional context**

PR #2468, opened 2024-08-06
WIP: feat: Use PartialObjectMetadata for Configmaps and Secrets

mrueg:
This PR introduces a metadataonly client, which can be used to fetch sparse data for large objects like configmaps or confidential data like secrets.

None

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2463


mrueg:
We can't use partialobjectmetadata for secrets unfortunately, since we have metrics for the secrets type that we expose.

sftim:
> We can't use partialobjectmetadata for secrets unfortunately, since we have metrics for the secrets type that we expose.

Secret supports `type` as a field selector.

I wonder if we could make a sort of informer that does a metadata-only **watch** _per known Secret type_, plus one that watches for unknown types and doesn't use partial object metadata. Should be quite a bit more efficient overall, at the cost of turning one watch into _n_.

mrueg:
I still lack seeing the hoped memory improvements with this PR.
My test looks like the following:
```
for i in $(seq 1 100); do
        base64 /dev/urandom | head -c 1000000 > file.txt
        kubectl create configmap morecm$i --from-file="file.txt"
done
```

pkg/sharding/listwatch_test.go
```diff
@@ -25,7 +25,7 @@ import (
 )
 
 func TestSharding(t *testing.T) {
-	cm := &v1.ConfigMap{
+	cm := &v1.Pod{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      "configmap1",
```
sftim, line 30: Seems like an odd name for a Pod :wink:

mrueg: heh yeah I just changed this one to make the Test still work ;) I can change that though, good catch. :D

internal/store/builder.go
```diff
@@ -541,6 +561,46 @@ func (b *Builder) buildStores(
 	return stores
 }
 
+func (b *Builder) buildMetadataOnlyStores(
```
dgrisonnet, line 564: I might be missing something, but it seems you could call `buildStores` right away since this is an exact copy of its code.

Going even further, maybe we don't even need `buildMetadataOnlyStores` and could just call `buildStores` directly when building resource specific stores like:

```diff
-	return b.buildStoresFunc(configMapMetricFamilies(b.allowAnnotationsList["configmaps"], b.allowLabelsList["configmaps"]), &v1.ConfigMap{}, createConfigMapListWatch, b.useAPIServerCache)
+	return b.buildStoresFunc(configMapMetricFamilies(b.allowAnnotationsList["configmaps"], b.allowLabelsList["configmaps"]), &metav1.PartialObjectMetadata{}, createConfigMapListWatch, b.useAPIServerCache)
```

pkg/util/utils.go
```diff
@@ -79,6 +81,34 @@ func CreateKubeClient(apiserver string, kubeconfig string) (clientset.Interface,
...
+
+	if config == nil {
+		var err error
+		config, err = clientcmd.BuildConfigFromFlags(apiserver, kubeconfig)
+		if err != nil {
+			return nil, err
+		}
+	}
+	config.UserAgent = fmt.Sprintf("%s/%s (%s/%s) kubernetes/%s", "kube-state-metrics (metadataonly)", version.Version, runtime.GOOS, runtime.GOARCH, version.Revision)
+	config.AcceptContentTypes = "application/vnd.kubernetes.protobuf;as=PartialObjectMetadataList;g=meta.k8s.io;v=v1,application/json;as=PartialObjectMetadataList;g=meta.k8s.io;v=v1,application/json,application/vnd.kubernetes.protobuf;as=PartialObjectMetadata;g=meta.k8s.io;v=v1,application/json;as=PartialObjectMetadata;g=meta.k8s.io;v=v1"
```
dgrisonnet, line 100: Do we really need to do that ourselves? The client seems to already set the Accept header accordingly: https://github.com/kubernetes/kubernetes/blob/0b3b733c8437a946b7300a346ac948fec2e0b3ff/staging/src/k8s.io/client-go/metadata/metadata.go#L222

dgrisonnet (commented):
Few comments, but I am a fan of this change! :)

internal/store/builder.go
mrueg: The reason I created that is the different signature of the listWatchFunc (using metadata.Interface not clientset.Interface). Is there's a different way to do that, I'm all ears :)

pkg/util/utils.go
mrueg: This was mainly copying from the CreateKubeClient() func, I'm not sure if we need it here additionally. 

Will metadata.List()/Watch() be called eventually or is that overwritten by ListWatchFunc?

mrueg:
> Few comments, but I am a fan of this change! :)

Great to hear! I'm still looking for better ideas to verify that it saves network traffic and reduces memory consumption on ksm.

internal/store/builder.go
```diff
@@ -519,7 +538,8 @@ func (b *Builder) buildStores(
 		if b.fieldSelectorFilter != "" {
 			klog.InfoS("FieldSelector is used", "fieldSelector", b.fieldSelectorFilter)
 		}
-		listWatcher := listWatchFunc(b.kubeClient, v1.NamespaceAll, b.fieldSelectorFilter)
+		kubeClient := b.kubeClient
+		listWatcher := listWatchFunc(kubeClient, v1.NamespaceAll, b.fieldSelectorFilter)
```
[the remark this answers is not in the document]
mrueg: agreed, reverted this change.

internal/store/builder.go
rexagod: Since generic types only support interfaces with no methods, we might want to rely on some old school type casting here?

pkg/util/utils.go
rexagod: The former as that seems to be the underlying invocation, so we should be good to drop this.

Issue #2372, opened 2024-04-16
Some kube-state-metrics shards are serving up stale metrics


schahal:
**What happened**:

We found some kube-state-metrics shards are serving up stale metrics.

For example, this pod is running and healthy:

```
$ kubectl get pods provider-kubernetes-a3cbbe355fa7-6d9d468f59-xbfsq
NAME                                                READY   STATUS    RESTARTS   AGE
provider-kubernetes-a3cbbe355fa7-6d9d468f59-xbfsq   1/1     Running   0          87m
```

However, we see for the past hour that `kube_pod_container_status_waiting_reason` is reporting it in `CreatingContainer`:

<img width="1711" alt="Screenshot 2024-04-16 at 2 00 31 PM" src="https://github.com/kubernetes/kube-state-metrics/assets/831526/6ce3633b-b813-4379-9fe6-11c408a25044">

And to prove this is being served by KSM, we looked at the incriminating shard's (`kube-state-metrics-5`) /metrics endpoint and saw this metric is definitely stale:

```
kube_pod_container_status_waiting_reason{namespace="<redacted>",pod="provider-kubernetes-a3cbbe355fa7-678fd88bc5-76dw4",uid="<redacted>",container="package-runtime",reason="ContainerCreating"} 1
```

This is one such example, there seem to be several such situations.

**What you expected to happen**:

Expectation is that the metric(s) match reality

**How to reproduce it (as minimally and precisely as possible)**:

Unfortunately, we're not quite sure when/why it gets into this state (anecdotally, it almost always happens when we upgrade KSM, though today there was no update besides some Prometheus agents)

We can **_mitigate_** the issue by restarting all the KSM shards... e.g.,

```
$ kubectl rollout restart -n kube-state-metrics statefulset kube-state-metrics
```

... if that's any clue to determine root cause.

**Anything else we need to know?**:

1. When I originally ran into the problem, I thought it had something to do with the [Compatibility Matrix](https://github.com/kubernetes/kube-state-metrics?tab=readme-ov-file#compatibility-matrix). But starting with KSM v2.11.0, I confirmed the client libraries are updated for my version of k8s (v1.28)

2. There's nothing out of the ordinary in the KSM logs:

<details>
<summary>Click to view kube-state-metrics-5 logs</summary>

```
I0409 08:17:51.349017       1 wrapper.go:120] "Starting kube-state-metrics"
[...]

PR #2478, opened 2024-08-16
fix(discovery): configure sharding every time MetricsHandler.Run runs

wallee94:

I see the following issue in a ksm deployment with custom CRD config enabled. Whenever a new CRD event is added to the cache, ksm stops watching and metrics stop reflecting new changes.

image

In the graph, ksm recovers after updating the Statefulset with any change. The metrics are `rate(kube_state_metrics_watch_total)` and the counter `kube_state_metrics_custom_resource_state_add_events_total` without rate.

Looking into the code, the problem seems to be the validation `shardingUnchanged` in `AddFunc` ([here](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metricshandler/metrics_handler.go#L123-L144)).

Without a CRD config, `MetricsHandler.Run` runs only once, and the vars `m.curShard` and `m.curTotalShards` are initially `nil`, which makes `shardingUnchanged = false` ([here](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metricshandler/metrics_handler.go#L136)).

However, if a CRD config is present, discovery runs `MetricsHandler.Run` every time a CRD event is detected ([here](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/discovery/discovery.go#L246)). If the Statefulset number of replicas/shards didn't change, the new CRD event will cancel the old metrics handler, but won't initiate a new one because `shardingUnchanged = true` in `AddFunc`.

~~This change removes the check `shardingUnchanged` in the `AddFunc` event handler. I don't think it's necessary because, in most cases, it's only called when the informer is synced at the end of `MetricsHandler.Run`.~~

This change updates `CRDiscoverer.PollForCacheUpdates` to rebuild the metrics writers in the already running metrics handler, instead of running a new one every time a CRD event occurs.

No change in cardinality.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2372


CatherineF-dev:
Hello, this code has been around for five years. Why is it only now experiencing this issue? Could there be an underlying problem?

```
			m.mtx.RLock()
			shardingUnchanged := m.curShard == shard && m.curTotalShards == totalShards
			m.mtx.RUnlock()

			if shardingUnchanged {
				return
			}
```

wallee94:
Sorry, I mentioned it in a thread in the kube-state-metrics Slack channel and forgot to put it here.

The bug isn't exactly in metrics_handler if `Run` runs only once, which is how it works without CR. The issue is that discovery started running it multiple times in [this PR](https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R246) if there's a CR config.

Removing the validation in `AddFunc` made sense to me to reconfigure every time the index informer is initially synced. I don't think it should depend on the previous `m.curShard` and `m.curTotalShards` values of earlier runs.

CatherineF-dev:
> The issue is that discovery started running it multiple times in [this PR](https://github.com/kubernetes/kube-state-metrics/pull/1851/files#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R246) if there's a CR config.

I am thinking whether we should run it only once.

wallee94:
That's a good point, I can look into that. I think `m.Run` could probably be just `m.ConfigureSharding`, but discovery doesn't have access to the Statefulset or k8s client to calculate `shard` and `totalShards`. 

On the other hand, if `Run` runs only once, the validation in this change is usually `false` because `m.curShard` and `m.curTotalShards` are initially `nil`.

wallee94:
@CatherineF-dev I added new changes to run `m.Run` only once when the pod starts, and to reconfigure sharding in discovery instead of recreating the IndexInformer.

I deployed the change to a few clusters and it seems to be working. This is the watch rate after adding a new CRD to the cluster:
image

I see the event, then a brief drop, which is when ksm is populating the cache after the reconfigure, and then it comes back to normal. All the metrics look good as well.

internal/discovery/discovery.go
```diff
@@ -239,21 +235,8 @@ func (r *CRDiscoverer) PollForCacheUpdates(
...
-				select {
-				case <-olderContext.Done():
-					// Context cancelled, don't really need to log this though.
-				default:
-					klog.ErrorS(err, "failed to run metrics handler")
-				}
-			}
-		}()
+		// Update metric handler with the new configs.
+		m.ReconfigureSharding(ctx)
```
dgrisonnet, line 239: I don't think we need this do we? None of the values set when reconfiguring sharing are changing when a new CRD is added or removed and because we don't cancel the metrics handler anymore here, it shouldn't impact sharding at all.

wallee94: I think after [these changes above](https://github.com/kubernetes/kube-state-metrics/blob/4aced25f69e00d54ec3e03a921f8bf60d4e8a783/internal/discovery/discovery.go#L224-L232) in the `storeBuilder`, `configureSharding` has to build `metricsHandler` again, [here](https://github.com/kubernetes/kube-state-metrics/blob/4aced25f69e00d54ec3e03a921f8bf60d4e8a783/pkg/metricshandler/metrics_handler.go#L99), because some values like `storeBuilder.enabledResources` have changed

dgrisonnet: You are correct, we need to rebuild the writers and more importantly the underlying stores. But we don't have to touch the number of shards.

But do you think you could decouple https://github.com/kubernetes/kube-state-metrics/blob/4aced25f69e00d54ec3e03a921f8bf60d4e8a783/pkg/metricshandler/metrics_handler.go#L99 from `ConfigureSharding` and put it in its own method like `BuildWriters` or something?

The reasoning is that we now need to reconfigure the writers, when a CRD is created/deleted and the context is cancelled.

But there is still something that annoys me here, when we detect changes to CRDs, we shouldn't have to rebuild everything. I'll continue to look into it and see how we could address that, but in the meantime I am fine with just addressing the issue at hand.

internal/discovery/discovery.go
```diff
@@ -269,7 +252,6 @@ func (r *CRDiscoverer) PollForCacheUpdates(
 					shouldGenerateMetrics = r.WasUpdated
 				})
 				if shouldGenerateMetrics {
-					olderCancel()
```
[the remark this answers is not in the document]
wallee94: I think that context is passed by the `metrics_handler`, and it cancels the prev one here before rebuilding the stores and passing a new one:
https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/metricshandler/metrics_handler.go#L79

we should probably have that decoupled as well to have it in the `BuildWriters` method. I'll work on that

internal/discovery/discovery.go
wallee94: Agreed. It can be confusing to have a reference to shards when they didn't change. Added 👍

internal/discovery/discovery.go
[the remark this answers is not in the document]
wallee94: All good, I updated the e2e test and I think the metrics handler is definitely cancelling the prev context. The problem with duplicated metrics was the store builder adding the same enabled resources multiple times.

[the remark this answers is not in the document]
wallee94: All good, I updated the e2e test and I think the metrics handler is definitely cancelling the prev context. The problem with duplicated metrics was the store builder adding the same enabled resources multiple times.

wallee94:
A summary of changes per file to help with the review:
- `internal/discovery/discovery.go`: Use m.BuildWriters(ctx) instead of m.Run(olderContext) to rebuild writers instead of recreating the whole handler. BuildWriters cancels the old context, so we no longer need the cancelations in this file.
- `internal/store/builder.go`: WithEnabledResources had a bug and wasn't deduplicating values. The new code deduplicates.
- `pkg/metricshandler/metrics_handler.go`: adds a function BuildWriters that rebuilds the metrics writers. We use it in discovery.go when the resources in the StoreBuilder have changed.
- `tests/e2e/discovery_test.go`: Update to test custom metrics after updating a CRD. I broke this into subfunctions because it failed the cyclomatic complexity linter.

internal/store/builder.go
```diff
@@ -107,18 +107,26 @@ func (b *Builder) WithMetrics(r prometheus.Registerer) {
 
 // WithEnabledResources sets the enabledResources property of a Builder.
 func (b *Builder) WithEnabledResources(r []string) error {
+	enabledResources := map[string]bool{}
```
mrueg, line 110: Can you remove this bugfix here (as it shouldn't be part of this change)?

We have another PR that tries to solve the same issue here: https://github.com/kubernetes/kube-state-metrics/pull/2502

wallee94: Sure, makes sense. I needed it for the e2e test but I just changed the validation

pkg/metricshandler/metrics_handler.go
```diff
@@ -69,24 +69,35 @@ func New(opts *options.Options, kubeClient kubernetes.Interface, storeBuilder ks
 	}
 }
 
-// ConfigureSharding (re-)configures sharding. Re-configuration can be done
-// concurrently.
-func (m *MetricsHandler) ConfigureSharding(ctx context.Context, shard int32, totalShards int) {
+// BuildWriters builds the metrics writers, cancelling any previous context and passing a new one on every build.
+// Build can be used mutlitple times and concurrently.
```
mrueg, line 73:
```suggestion
// Build can be used multiple times and concurrently.
```

pkg/metricshandler/metrics_handler.go
```diff
@@ -69,24 +69,35 @@ func New(opts *options.Options, kubeClient kubernetes.Interface, storeBuilder ks
...
-	if totalShards != 1 {
-		klog.InfoS("Configuring sharding of this instance to be shard index (zero-indexed) out of total shards", "shard", shard, "totalShards", totalShards)
-	}
 	ctx, m.cancel = context.WithCancel(ctx)
-	m.storeBuilder.WithSharding(shard, totalShards)
 	m.storeBuilder.WithContext(ctx)
 	m.metricsWriters = m.storeBuilder.Build()
+}
+
+// ConfigureSharding configures sharding. Configuration can be used mutlitple times and
```
mrueg, line 86:
```suggestion
// ConfigureSharding configures sharding. Configuration can be used mutliple times and
```

pkg/metricshandler/metrics_handler.go
wallee94: Ah good catch. Thank you!

mrueg:
/lgtm

I'll still ping the other maintainers to review, currently everyone seems to be busy.

mrueg:
/hold 

continuing here as no other reviews came in.

PR #2485, opened 2024-08-23
fix: set kube_job_status_failed metric even when there are no job.Status.Conditions present

### LINKED ISSUE #2443: Some job metrics are missing if job has no conditions
**What happened**:
Some job metrics (`kube_job_status_failed`, `kube_job_complete`, `kube_job_failed`) are missing for `Job`s without conditions.

Metrics are present regardless if conditions exist.

1. Create this job
<details><summary>job.yaml</summary>
<p>

```yaml
apiVersion: batch/v1
kind: Job
metadata:
  name: test
spec:
  template:
    spec:
      containers:
      - image: alpine
        name: test
        command:
          - "sh"
          - "-c"
        args:
          - sleep 5 && exit 1
      restartPolicy: Never
```

</p>
</details>  

2. Observe that the `.status` is missing a `conditions` objects (before the backoffLimit is reached)
3. Observe that e.g., the `kube_job_status_failed` metric is missing (`curl localhost:8080/metrics | grep kube_job_status_failed`)

The reason for this bug is that the labels and values are only set within a for loop (if a condition of type `Failed` exists):

https://github.com/kubernetes/kube-state-metrics/blob/85d1423b5b896f4fbce8394afed20cfe17371251/internal/store/job.go#L222-L224

I can provide a fix that sets the value regardless if there are any condition with type `Failed`.

**Environment**:

* kube-state-metrics version: `v2.12.0` (master `85d1423`)
* Kubernetes version (use `kubectl version`): `v1.29.6`
* Cloud provider or hardware configuration: homelab

richabanker:
Sets the `kube_job_status_failed` metric when a job's conditions are empty. Previously, the metric was only being set while iterating through a job's conditions, thus ignoring the case when no conditions were found but the job still failed.

does not change 

Fixes #2443

### DIFF
--- a/internal/store/job_test.go
+++ b/internal/store/job_test.go
@@ -213,6 +213,28 @@ func TestJobStore(t *testing.T) {
 		{
 			Obj: &v1batch.Job{
 				ObjectMeta: metav1.ObjectMeta{
+					Name:      "FailedJobWithNoConditions",
+					Namespace: "ns1",
+				},
+				Status: v1batch.JobStatus{
+					Failed: 1,
+				},
+				Spec: v1batch.JobSpec{
+					ActiveDeadlineSeconds: &ActiveDeadlineSeconds900,
+				},
+			},
+			Want: metadata + `
+				kube_job_owner{job_name="FailedJobWithNoConditions",namespace="ns1",owner_is_controller="",owner_kind="",owner_name=""} 1
+				kube_job_info{job_name="FailedJobWithNoConditions",namespace="ns1"} 1
+				kube_job_spec_active_deadline_seconds{job_name="FailedJobWithNoConditions",namespace="ns1"} 900
+				kube_job_status_active{job_name="FailedJobWithNoConditions",namespace="ns1"} 0
+				kube_job_status_failed{job_name="FailedJobWithNoConditions",namespace="ns1",reason=""} 1
+				kube_job_status_succeeded{job_name="FailedJobWithNoConditions",namespace="ns1"} 0
+`,
+		},
+		{
+			Obj: &v1batch.Job{
+				ObjectMeta: metav1.ObjectMeta{
 					Name:       "SuccessfulJob2NoActiveDeadlineSeconds",
 					Namespace:  "ns1",
 					Generation: 1,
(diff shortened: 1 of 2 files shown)


dgrisonnet (commented):
Looks good to me, but could you please add a new test in https://github.com/kubernetes/kube-state-metrics/blob/194a46c09999771736a6e93d8ec48b38219df55e/internal/store/job_test.go to verify that jobs without failed condition have metrics generated?

richabanker:
> Looks good to me, but could you please add a new test in https://github.com/kubernetes/kube-state-metrics/blob/194a46c09999771736a6e93d8ec48b38219df55e/internal/store/job_test.go to verify that jobs without failed condition have metrics generated?

Yep, for sure. Added one, thanks!

richabanker:
Hmm, super weird, when I run `make test-unit` locally - its passing all tests but CI is complaining :/

internal/store/job_test.go
```diff
@@ -214,6 +214,30 @@ func TestJobStore(t *testing.T) {
...
+				},
+				Status: v1batch.JobStatus{
+					Failed: 1,
+				},
+				Spec: v1batch.JobSpec{
+					ActiveDeadlineSeconds: &ActiveDeadlineSeconds900,
+				},
+			},
+			Want: metadata + `
+				kube_job_annotations{job_name="FailedJobWithNoConditions",namespace="ns1"} 1
```
mrueg, line 233:
```suggestion

```

internal/store/job_test.go
```diff
@@ -214,6 +214,30 @@ func TestJobStore(t *testing.T) {
...
+				kube_job_owner{job_name="FailedJobWithNoConditions",namespace="ns1",owner_is_controller="",owner_kind="",owner_name=""} 1
+				kube_job_info{job_name="FailedJobWithNoConditions",namespace="ns1"} 1
+				kube_job_labels{job_name="FailedJobWithNoConditions",namespace="ns1"} 1
```
mrueg, line 236:
```suggestion

```

If it doesn't have labels and annotations set, these ones should not be created.

internal/store/job_test.go
richabanker: Ahh I just needed to sync to HEAD and it started failing locally too. Thanks fixed now.

dgrisonnet (approved):
/lgtm
/approve

PR #2492, opened 2024-08-30
chore: Bump to kubernetes 1.31.0

mrueg:
Uses kubernetes 1.31 client libs

**This change breaks a stable metric** - I believe this is justified as the metric label never gave a correct result.
https://kubernetes.io/blog/2024/07/19/kubernetes-1-31-upcoming-changes/#deprecation-of-status-nodeinfo-kubeproxyversion-field-for-nodes-kep-4004-https-github-com-kubernetes-enhancements-issues-4004

None
Fixes #


CatherineF-dev:
/lgtm 
/hold
feel free to unhold

mrueg:
@CatherineF-dev can you take another look? Thanks!

dgrisonnet:
I think it might be better to follow the same deprecation plan as upstream. Removing it directly might be harmless here, but at least following the same plan as Kubernetes would set a precedent for what we should do in the future when such thing happen. We could make a deprecation notice in the next release changelog and then fully remove the label once the feature is removed?

docs/metrics/cluster/node-metrics.md
```diff
@@ -3,7 +3,7 @@
...
-| kube_node_info               | Gauge       | Information about a cluster node                                                                                          |                                                                                                                                                                                          | `node`=&lt;node-address&gt; <br> `kernel_version`=&lt;kernel-version&gt; <br> `os_image`=&lt;os-image-name&gt; <br> `container_runtime_version`=&lt;container-runtime-and-version-combination&gt; <br> `kubelet_version`=&lt;kubelet-version&gt; <br> `kubeproxy_version`=&lt;kubeproxy-version&gt; <br> `pod_cidr`=&lt;pod-cidr&gt; <br> `provider_id`=&lt;provider-id&gt; <br> `system_uuid`=&lt;system-uuid&gt; <br> `internal_ip`=&lt;internal-ip&gt; | STABLE       |
```
logicalhan, line 6: I would rather preserve API stability but just enter a constant "disabled" value for the existing label.

mrueg:
I made the requested changes, please have another look.

mrueg:
/hold cancel 

@CatherineF-dev @rexagod @dgrisonnet friendly ping on this one :)

rexagod:
/lgtm
/hold

Holding for others, feel free to unhold.

mauri870:
Just a friendly reminder about the missing reviews so we can get this merged. Thanks!

mrueg:
/hold cancel

I'd say this has been waiting long enough. If people have additional concerns, they can follow up with their own PR.

PR #2507, opened 2024-09-23
#2501 Count volumes attached to a node

### LINKED ISSUE #2501: Count volumes attached to a node

1. A gauge metric reflecting number of the persistent volumes attached to a node.

**Why is this needed**:

The number of volumes that can be attached to a Node is limited. 

- [Kubernetes general limitations](https://kubernetes.io/docs/concepts/storage/storage-limits/)
- [GCP - 127 persistent disks per node](https://cloud.google.com/compute/docs/disks/add-persistent-disk)
- [DigitalOcean - 15 volumes per node](https://docs.digitalocean.com/products/volumes/details/limits/)
- [Linode - 8](https://techdocs.akamai.com/cloud-computing/docs/block-storage#limits-and-considerations) (apparently that is even less because some system disks are also counted towards that limit)

Event with higher limits e.g. with Google we can still reach them. So monitoring the number of volumes attached per node and number of volumes that can be attached per node is essential to scale out when it's required.

Here is how I tried to tackle this, but apparently it is not possible and [will not be possible](https://github.com/kubernetes/kube-state-metrics/issues/2434) to create custom resource states on the core objects. So this is NOT working:

```
        spec:
          resources:
            - groupVersionKind:
                group: ""
                kind: Node
                version: v1
              labelsFromPath:
                node: [metadata, name]
              metrics:
                - name: node_volumes_attached_count
                  help: Number of volumes attached to the node
                  type: Gauge
                  gauge:
                    path: [status, volumesAttached]
                    valueFrom:
                      count: {}
                - name: node_volumes_in_use_count
                  help: Number of volumes in use on the node
                  type: Gauge
                  gauge:
                    path: [status, volumesInUse]
                    valueFrom:
                      count: {}
```

I am not sure if this max volume limit is exposed to Kubernetes by the vendors in any way, but we definitely can count the attached volumes. Knowing the given provider limitations, we can come up with the alerting strategy based on node labels and the count of the volumes that are currently attached.

LiuQhahah:

Count volumes attached to a node

increaser two metrics: kube_node_volumes_attached_count and kube_node_volumes_in_use_count

Fixes #2501

### DIFF
--- a/internal/store/node.go
+++ b/internal/store/node.go
@@ -56,6 +56,8 @@ func nodeMetricFamilies(allowAnnotationsList, allowLabelsList []string) []genera
 		createNodeStatusCapacityFamilyGenerator(),
 		createNodeStatusConditionFamilyGenerator(),
 		createNodeStateAddressFamilyGenerator(),
+		createNodeVolumeCountGenerator(),
+		createNodeVolumeInUseGenerator(),
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mrueg:
I understand what you're trying to achieve here, I wonder though if it would be easier to attach a node label to the `kube_volumeattachment_status_attached` metric or a similar one and then do a sum by that label in promql instead of exposing counts here.

LiuQhahah:
> I understand what you're trying to achieve here, I wonder though if it would be easier to attach a node label to the `kube_volumeattachment_status_attached` metric or a similar one and then do a sum by that label in promql instead of exposing counts here.

Thanks for the suggestion, from my understanding @thommeo  wants to get the number of volumns currently attached from the perspective of node, and kube_volumeattachment_status_attached is from volumnattachment from volumeattachments to count the number of volumns he is attached to by that node attchment.

From my personal understanding and implementation point of view, I think creating a metric will be better understood, the current implementation is not so complex, so that the use of the customer will be more convenient!

rexagod:
Hello, there's tests failing for this patch, could you please address those?

LiuQhahah:
> Hello, there's tests failing for this patch, could you please address those?

The action history is too long to find the log. Could you please help retrigger it? thanks

thommeo:
It was almost finished, wasn't it? Can we reopen it or open another PR with the same change? @LiuQhahah

LiuQhahah:
HI @thommeo 
I have created another PR https://github.com/kubernetes/kube-state-metrics/pull/2810
Thanks

PR #2509, opened 2024-09-23
chore: Fix the security issue about sirupsen/logrus

kevinwangcn:
There is a security concern using the `github.com/sirupsen/logrus` with version `1.8.1`

**Supporting Material/References:**
https://github.com/sirupsen/logrus/issues/1370

**Impact**
This happens if you try to log something that is longer than 64kb without newlines through something like cmd.Stdout = logrus.Writer().
This can cause a denial of service in some cases.

The PR is going to bump up the version for this particular module from `1.8.1` to `1.9.3` which has resolved the security issue. 

Fixes #


CatherineF-dev:
Can you use `go get github.com/sirupsen/logrus@v1.9.3` instead of replace?

kevinwangcn:
> Can you use `go get github.com/sirupsen/logrus@v1.9.3` instead of replace?

Updated my PR. thanks for the review.

rexagod:
Is there a CVE for this vulnerability? Does golvulncheck validate that the affected symbols are actually being used in the codebase?

kevinwangcn:
> Is there a CVE for this vulnerability? Does golvulncheck validate that the affected symbols are actually being used in the codebase?

yes, this is going to solve a CVE issue. Attached is the scan report via Twistlock.  Please refer to it for more details.
[scan-result.csv](https://github.com/user-attachments/files/17121821/scan-result.csv)

kevinwangcn:
@rexagod  @CatherineF-dev  Could you please review my PR? Thanks a lot.

PR #2510, opened 2024-09-26
perf: use concurrent map when storing metrics

### LINKED ISSUE #995: kube-state-metrics API scraping timeout

rarruda:
**What this PR does / why we need it**: In busy/large clusters, will prevent timeouts from long living locks/concurrency issues, as the writing to the map takes overly long, blocking the metrics-reading thread and as the lock doesn't get released in a timely manner, timing out the request.

In these graphs you see:
1. VMagent scrape duration (yellow: with this patch applied, green: no patch)
2.  VMagent samples scraped: (yellow: with this patch applied, green: no patch)
3. Heatmap from selfMetrics with patch applied
4. Heatmap from selfMetrics no patch
5. Number of pods in LOG scale, max is close to 16k
6. Number of nodes in cluster
Screenshot from 2024-09-26 12-48-47

You can see that with this patch we reduce latency significantly under all scenarios in our cluster.

Inspired by previous PR at https://github.com/kubernetes/kube-state-metrics/pull/1028

**How does this change affect the cardinality of KSM**: *does not change cardinality*

Fixes #995

### DIFF
--- a/pkg/metrics_store/metrics_store.go
+++ b/pkg/metrics_store/metrics_store.go
@@ -20,6 +20,7 @@ import (
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
+
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 )
 
--- a/pkg/metrics_store/metrics_store.go
+++ b/pkg/metrics_store/metrics_store.go
@@ -20,8 +20,6 @@ import (
 	"sync"
 
 	"k8s.io/apimachinery/pkg/api/meta"
-	"k8s.io/apimachinery/pkg/types"
-
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 )
 
@@ -33,7 +31,7 @@ type MetricsStore struct {
 	// metric families, containing a slice of metrics. We need to keep metrics
 	// grouped by metric families in order to zip families with their help text in
 	// MetricsStore.WriteAll().
-	metrics map[types.UID][][]byte
+	metrics sync.Map
 
 	// generateMetricsFunc generates metrics based on a given Kubernetes object
 	// and returns them grouped by metric family.
@@ -42,9 +40,6 @@ type MetricsStore struct {
 	// later on zipped with with their corresponding metric families in
 	// MetricStore.WriteAll().
 	headers []string
-
-	// Protects metrics
-	mutex sync.RWMutex
 }
 
 // NewMetricsStore returns a new MetricsStore
@@ -52,7 +47,7 @@ func NewMetricsStore(headers []string, generateFunc func(interface{}) []metric.F
 	return &MetricsStore{
 		generateMetricsFunc: generateFunc,
 		headers:             headers,
-		metrics:             map[types.UID][][]byte{},
+		metrics:             sync.Map{},
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mrueg:
Thanks for your contribution!
/lgtm 
and 
/hold
for others to review as well.

dgrisonnet:
The changes looks good to me, thank you @rarruda for pushing that improvement :)

I'll update our perfs-tests and see how much we improve with this.

dgrisonnet:
Please refrain from merging until we have the results.

rarruda:
Note: for clusters with very little load, and/or sequential read/write access patterns (synthetic data getting generated, and only after then read), I would expect that the original implementation having one mutex for the Map would perform better.

But that's not real world performance. 

In our (real world, production) case, as seen by the graphs, the locking contention gets to be too great as too many things are happening in parallel. 

As work arounds, we also tried sharding, clustering, dividing up metrics to go do different instances, but those were really just palliative measures for the underlying lock contention. It increased complexity in our setup significantly, and felt like ugly hacks.

mrueg:
@dgrisonnet do you have any results from the perf tests you can share here already?

dgrisonnet:
I need to fix the tests, kms doesn't seem to be deploying properly in their CI cluster: https://github.com/kubernetes/perf-tests/pull/2920, but I don't have much time on my hands to look at it right now.

rarruda:
Anything I can do to help?

CatherineF-dev:
@rarruda I think the current blocker is this test failure https://github.com/kubernetes/perf-tests/pull/2920

dgrisonnet:
I don't have enough times on my hand right now to fix the perf-tests, so I'll unblock the PR based on the data you've already shared that shows the improvements provided by this change.

/unhold

PR #2525, opened 2024-10-15
WIP: Testing: discovery: check number of goroutines

dgrisonnet:

Fixes #


rexagod:
We may want to incorporate https://github.com/uber-go/goleak as well, WDYT?

dgrisonnet:
Yes that would definitely be a great addition, but I didn't have time to modify our integration that. I just want to check if https://github.com/kubernetes/kube-state-metrics/pull/2478 can cause a goroutine explosion.

PR #2528, opened 2024-10-17
docs: Add best practices for metrics

mrueg:
This PR introduces a doc for best practices around metrics.
Hoping to get some comments and an agreement on these ideas.
@rexagod @dgrisonnet @dashpole @logicalhan @CatherineF-dev 

None 
Fixes #

### DIFF
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -38,7 +38,9 @@ It is a good practice to group those together into an `_info` metric.
 An object can also have a dynamic set of properties, which are usually part of the status field.
 These change during the lifecycle of the object.
 For example a pod can be in different states like "Pending", "Running" etc.
-These should be part of a new metric that includes labels that identify the object as well as the dynamic property.
+These should be part of a "State Set" that includes labels that identify the object as well as the dynamic property.
+
+For example:
 
 ### Linked properties
 
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -18,7 +18,7 @@ Feel encouraged to report these metrics and provide a pull request to improve th
 
 ## General best practices
 
-We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeloing.
+We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeling.
 
 ## Best practices for kube-state-metrics
 
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -25,7 +25,8 @@ We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practi
 ### Avoid pre-computation
 
 kube-state-metrics should expose metrics on an individual object level and avoid any sort of pre-computation unless it is required due to for example high cardinality on objects.
-By exposing raw metrics instead of counters, kube-state-metrics allows the user to have full control on how they want to use the metrics.
+We prefer not to add metrics that can be derived from existing raw metrics. For example, we would not want to expose a metric called `kube_pod_total` as it can be computed with `count(kube_pod_info)`.
+This way kube-state-metrics allows the user to have full control on how they want to use the metrics and gives them flexibility to do specific computation.
 
 ### Static object properties
 
@@ -48,7 +49,7 @@ If an object contains a substructure that links multiple properties together (e.
 
 ### Optional properties
 
-Some Kubernetes objects have optional fields. In case there is an optional value, it is better to not expose the label at all instead of exposing a "nil" value or an empty string.
+Some Kubernetes objects have optional fields. In case there is an optional value, the label should still be exposed, ideally as an empty string.
 
 ### Timestamps
 
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -52,7 +52,7 @@ Some Kubernetes objects have optional fields. In case there is an optional value
 
 ### Timestamps
 
-Timestamps like creation time or modification time should be exposed as a value. The metric should end with `_timestamp_seconds`.
+Timestamps like creation time or modification time should be exposed as a value. The metric should end with `_timestamp_seconds`. The date value is represented in [UNIX epoch seconds](https://en.wikipedia.org/wiki/Unix_time).
 
 ### Cardinality
 
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -31,8 +31,9 @@ This way kube-state-metrics allows the user to have full control on how they wan
 ### Static object properties
 
 An object usually has a stable set of properties that do not change during its lifecycle in Kubernetes.
-This includes properties like name, namespace, uid etc.
+This includes properties like name, namespace, uid etc. that have a 1:1 relationship with the object.
 It is a good practice to group those together into an `_info` metric.
+If there is a 1:n relationship (e.g. a list of ports), it should be in a separate metric to avoid generating too many metrics.
 
 ### Dynamic object properties
 
@@ -40,8 +41,6 @@ An object can also have a dynamic set of properties, which are usually part of t
 These change during the lifecycle of the object.
 For example a pod can be in different states like "Pending", "Running" etc.
 These should be part of a "State Set" that includes labels that identify the object as well as the dynamic property.
-
-For example:
 
 ### Linked properties
 
--- a/docs/design/metrics-best-practices.md
+++ b/docs/design/metrics-best-practices.md
@@ -1,0 +1,70 @@
+# Kube-State-Metrics - Timeseries best practices
+
+---
+
+Author: Manuel Rüger (<manuel@rueg.eu>)
+
+Date: October 17th 2024
+
+---
+
+## Introduction
+
+Kube-State-Metrics' goal is to provide insights into the state of Kubernetes objects by exposing them as metrics.
+This document provides guidelines with the goal to create a good user experience when using these metrics.
+
+Please be aware that this document is introduced in a later stage of the project and there might be metrics that do not follow these best practices.
+Feel encouraged to report these metrics and provide a pull request to improve them.
+
+## General best practices
+
+We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeloing.
+
+## Best practices for kube-state-metrics
+
+### Avoid pre-computation
+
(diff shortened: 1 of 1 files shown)


docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,70 @@
...
+## General best practices
+
+We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeloing.
+
+## Best practices for kube-state-metrics
+
+### Avoid pre-computation
+
+kube-state-metrics should expose metrics on an individual object level and avoid any sort of pre-computation unless it is required due to for example high cardinality on objects.
+By exposing raw metrics instead of counters, kube-state-metrics allows the user to have full control on how they want to use the metrics.
```
SuperQ, line 28: Raw counters are usually what we want, maybe you meant "derived metrics"?

```suggestion
By exposing raw metrics instead of derived rate metrics, kube-state-metrics allows the user to have full control on how they want to use the metrics.
```

CatherineF-dev:
LGTM for Stability

docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,70 @@
...
+An object usually has a stable set of properties that do not change during its lifecycle in Kubernetes.
+This includes properties like name, namespace, uid etc.
+It is a good practice to group those together into an `_info` metric.
+
+### Dynamic object properties
+
+An object can also have a dynamic set of properties, which are usually part of the status field.
+These change during the lifecycle of the object.
+For example a pod can be in different states like "Pending", "Running" etc.
+These should be part of a new metric that includes labels that identify the object as well as the dynamic property.
```
SuperQ, line 41: Usually the term we use for this is "State Set".

Maybe including a fully example here.

```suggestion
These should be part of a "State Set" that includes labels that identify the object as well as the dynamic property.

For example:
```
kube_pod_status_phase{phase="Failed",pod="mypod"} 0
kube_pod_status_phase{phase="Pending",pod="mypod"} 0
kube_pod_status_phase{phase="Running",pod="mypod"} 1
```

File: docs/design/metrics-best-practices.md [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+
+### Linked properties
+
+If an object contains a substructure that links multiple properties together (e.g. endpoint address and port), those should be reported in the same metric.
+
+### Optional properties
+
+Some Kubernetes objects have optional fields. In case there is an optional value, it is better to not expose the label at all instead of exposing a "nil" value or an empty string.
```
SuperQ, line 49: Hmm, I'm not sure about this one.

Prometheus metric families should have consistent labels.

For example, this is considered invalid.
```
my_metric{a="foo",b="bar"} 1
my_metric{b="baz"}1
```

In the above case, you want to expose an empty string label so the metric family is consistent.
```
my_metric{a="foo",b="bar"} 1
my_metric{a="",b="baz"}1
```

docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,70 @@
...
+
+### Timestamps
+
+Timestamps like creation time or modification time should be exposed as a value. The metric should end with `_timestamp_seconds`.
```
SuperQ, line 53:
```suggestion
Timestamps like creation time or modification time should be exposed as a value. The metric should end with `_timestamp_seconds`. The date value is represented in [UNIX epoch seconds](https://en.wikipedia.org/wiki/Unix_time).
```

SuperQ (commented):
Looking really nice so far. A couple of comments.

docs/design/metrics-best-practices.md
mrueg: I probably should rewrite this section a bit.
What I'm trying to describe here is that we do not want `_total` counters e.g. a `kube_pod_total` metric that counts all pods in the cluster.
Rather we prefer to expose a `kube_pod_info` metric to allow the user either run a sum on it, or even sum it by a specific attribute (e.g. per node)

docs/design/metrics-best-practices.md
mrueg: That's good to know! I'll fix it (and will add some fixes to the next release of ksm).
Is there any upstream documentation on this that suggests labels to be empty instead of leaving them out so I can reference them in here?

mrueg:
I'm also trying to suggest we move most labels into a single metric 
```
metric_with_lots_of_labels{uid=identifier, a1=a1,a2=a2,...an=an} 1
```

instead of 

```
metric_for_a1{uid=identifier,a1=a1}
metric_for_a2{uid=identifier,a2=a2}
....
metric_for_an{uid=identifier,an=an}
```

unless there are good reasons why the latter one is better for a TSDB/Querying etc

docs/design/metrics-best-practices.md
SuperQ: Oh, yes, that makes sense.

I think I've described this as "We prefer to not add metrics that can be derived from existing raw metrics. For example `kube_total_pods` can be computed with `count(kube_pod_info)`."

docs/design/metrics-best-practices.md
SuperQ: I thought we had an exmaple in the OpenMetrics exposition spec, but I can't find it right now. I'll see if I can find the docs on this.

rexagod: The technicality here can be split into two considerations:

- The lint enforcement mentioned above comes about as a result of the [registry machinery](https://github.com/prometheus/client_golang/blob/main/prometheus/registry.go#L310-L326).
- KSM [does not register metrics](https://github.com/kubernetes/kube-state-metrics/blob/0a86b24b08431c597519e19937b07162247035e3/pkg/customresourcestate/registry_factory.go#L219) in a registry since none of these are structured (except KSM's own meta-metrics, but piped-out the in the same [non-proto manner](https://github.com/kubernetes/kube-state-metrics/blob/166921b797100dadf51a153c8b932aab73dab4ff/pkg/metricshandler/metrics_handler.go#L202)).

KSM bypasses this trait by design (a `promtool check metrics` will succeed for the invalid example above).

Screenshot 2024-10-24 at 1 01 59 PM

That being said, it's always good to be consistent with the ecosystem, given we can incorporate a solution that ensures this in the CI.

rexagod:
I believe field-specific metrics may help (a) limit cardinality ([ref](https://github.com/kubernetes/kube-state-metrics/pull/2376#discussion_r1679341244))? In the original issue, there was talk about including only [tightly bounded](https://github.com/kubernetes/kube-state-metrics/pull/2376#discussion_r1626294094) labelsets in the `_info` metrics, however, that may lead to separate metrics that may fall into the "gray area" and (b) be ambiguous to the user if they are in an `_info` metric or not, as there isn't a feasible way to guess their inclusion or exclusion from the same without knowing a fields possible values beforehand. Not to mention, (c) static values may become dynamic in the future and cause breaking changes when moved to their own metrics (referring to the metric stability spec for BETA metrics).

mrueg:
> I believe field-specific metrics may help (a) limit cardinality ([ref](https://github.com/kubernetes/kube-state-metrics/pull/2376#discussion_r1679341244))? In the original issue, there was talk about including only [tightly bounded](https://github.com/kubernetes/kube-state-metrics/pull/2376#discussion_r1626294094) labelsets in the `_info` metrics, however, that may lead to separate metrics that may fall into the "gray area" and (b) be ambiguous to the user if they are in an `_info` metric or not, as there isn't a feasible way to guess their inclusion or exclusion from the same without knowing a fields possible values beforehand. 
Not to mention, (c) static values may become dynamic in the future and cause breaking changes when moved to their own metrics (referring to the metric stability spec for BETA metrics).

This is why you need to look at the lifecycle of the object. If it's not lifetime scoped, it's dynamic. I'd rather have a few metric series with more labels than a bunch of metrics to avoid bad UX during querying where you need to do a few group lefts and joins to get to the result you want to see.

docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,73 @@
...
+
+Kube-State-Metrics' goal is to provide insights into the state of Kubernetes objects by exposing them as metrics.
+This document provides guidelines with the goal to create a good user experience when using these metrics.
+
+Please be aware that this document is introduced in a later stage of the project and there might be metrics that do not follow these best practices.
+Feel encouraged to report these metrics and provide a pull request to improve them.
+
+## General best practices
+
+We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeloing.
```
SuperQ, line 21:
```suggestion
We follow [Prometheus](https://prometheus.io/docs/practices/naming/) best practices in terms of naming and labeling.
```

docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,73 @@
...
+It is a good practice to group those together into an `_info` metric.
+
+### Dynamic object properties
+
+An object can also have a dynamic set of properties, which are usually part of the status field.
+These change during the lifecycle of the object.
+For example a pod can be in different states like "Pending", "Running" etc.
+These should be part of a "State Set" that includes labels that identify the object as well as the dynamic property.
+
+For example:
```
multani, line 44: Can examples be added here, or this paragraph to be removed?

docs/design/metrics-best-practices.md
```diff
@@ -0,0 +1,73 @@
...
+
+### Linked properties
+
+If an object contains a substructure that links multiple properties together (e.g. endpoint address and port), those should be reported in the same metric.
+
+### Optional properties
+
+Some Kubernetes objects have optional fields. In case there is an optional value, the label should still be exposed, ideally as an empty string.
```
brancz, line 52: empty string and not having the label is the same thing in Prometheus, so why do this?

brancz (approved):
lgtm!

docs/design/metrics-best-practices.md
mrueg: This might be treated differently by other monitoring systems, so rather stay explicit here and ensure we do it the same way everywhere.

mrueg: Feel free to chime in on https://github.com/kubernetes/kube-state-metrics/pull/2528#discussion_r1810892031

docs/design/metrics-best-practices.md
mrueg: Thanks, I removed it. :)

mrueg:
@dgrisonnet @CatherineF-dev @rexagod I think this is in a good state now. :)

rexagod:
/assign
/lgtm

Talked this through with Damien offline, no blockers.

PR #2535, opened 2024-10-25
fix: panic in `util.GVRFromType` for core objects

L3n41c:

Fixes a panic in the `util.GVRFromType` function when called for a `core` object.

For a pod object, for example, the `apiVersion` would be `v1`,
```go
expectedTypeSlice := strings.Split(apiVersion, "/")
```
would return a slice of a single element because `apiVersion` has no `/`.
And
```go
v := expectedTypeSlice[1]
```
would panic because of an array out-of-bound access.

Does not change.

Fixes #

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -143,11 +143,10 @@ func GVRFromType(resourceName string, expectedType interface{}) *schema.GroupVer
 		return nil
 	}
 	apiVersion := expectedType.(*unstructured.Unstructured).Object["apiVersion"].(string)
-	expectedTypeSlice := strings.Split(apiVersion, "/")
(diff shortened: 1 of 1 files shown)


CatherineF-dev:
Core resource metrics are exposed using these functions https://github.com/kubernetes/kube-state-metrics/tree/main/internal/store

mrueg:
> Core resource metrics are exposed using these functions https://github.com/kubernetes/kube-state-metrics/tree/main/internal/store

Yes, that's correct. We keep core resource metrics internal.
 I still agree that the function shouldn't panic when used externally (or we should move the whole code into internal)

/lgtm

PR #2536, opened 2024-10-29
fix(custom resource state metrics): correctly convert status condition `Unknown` to a valid value

### LINKED ISSUE #2439: Metric with type Gauge throwing errors with no reason.
**What happened**:

After adding below metric. I was able to see the metric in Prometheus portal but the kube-state-metrics pod keeps throwing following error:  
`
E0701 21:29:11.117845       1 registry_factory.go:685] "gotk_resource_status" err="[status,conditions]: []: strconv.ParseFloat: parsing \"Unknown\": invalid syntax".
`

```
#configuration
---
customResourceState:
    enabled: true
    config:
      spec:
        resources:
          - groupVersionKind:
              group: kustomize.toolkit.fluxcd.io
              version: v1
              kind: Kustomization
            metricNamePrefix: gotk
            metrics:
              - name: "resource_status"
                help: "The current state of a Flux Kustomization resource."
                each:
                  type: Gauge
                  gauge:
                    path: [status, conditions]
                    labelsFromPath:
                      type: ["type"]
                    valueFrom: ["status"]
                labelsFromPath:
                  name: [metadata, name]
                  exported_namespace: [metadata, namespace]
                  suspended: [spec, suspend]
                  prune: [spec, prune]
                  revision: [status, lastAppliedRevision]
                  source_name: [spec, sourceRef, name]
```

I expect no errors in the logs after adding that configuration

Added the metric showed in the description

 As I said before the metric works as expected but the logs are overflowing with the error described on "What happened?" section:
 
```
gotk_resource_status{container="kube-state-metrics", customresource_group="kustomize.toolkit.fluxcd.io", customresource_kind="Kustomization", customresource_version="v1", endpoint="http", exported_namespace="flux-system", instance="172.22.15.193:8080", job="kube-state-metrics", name="prometheus", namespace="monitoring", pod="kube-prometheus-stack-kube-state-metrics-5b7fb8dc6-j7tqn", prune="true", revision="main@sha1:cc6b5b452b9b09a964deaef0b4de59817fe5caed", service="kube-prometheus-stack-kube-state-metrics", source_name="flux-system", type="Ready"}

```

I followed this documentation: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md#example-for-status-conditions-on-kubernetes-controllers

**Environment**:

* kube-state-metrics version:  app.kubernetes.io/version: 2.12.0    
* Kubernetes version (use `kubectl version`): v1.27.7
* Cloud provider or hardware configuration: AKS
* Other info:  
  -  helm.sh/chart: kube-state-metrics-5.18.1

Haleygo:

According to https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md#gauge
>This will work for kubernetes controller CRs which expose status conditions according to the kubernetes api (https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#Condition):

`ConditionUnknown = "Unknown"` should be converted corrctly to a valid value, and more like `0.0` instead of `1.0`.

Does not change cardinality.

Fixes #2439

### DIFF
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -337,7 +337,7 @@ Supported types are:
 * `nil` is generally mapped to `0.0` if NilIsZero is `true`, otherwise it will throw an error
 * for bool `true` is mapped to `1.0` and `false` is mapped to `0.0`
 * for string the following logic applies
-  * `"true"` and `"yes"` are mapped to `1.0` and `"false"` and `"no"` are mapped to `0.0` (all case-insensitive)
+  * `"true"` and `"yes"` are mapped to `1.0`, `"false"`, `"no"` and `"unknown"` are mapped to `0.0` (all case-insensitive)
   * RFC3339 times are parsed to float timestamp  
   * Quantities like "250m" or "512Gi" are parsed to float using <https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go>
   * Percentages ending with a "%" are parsed to float
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -337,7 +337,8 @@ Supported types are:
 * `nil` is generally mapped to `0.0` if NilIsZero is `true`, otherwise it will throw an error
 * for bool `true` is mapped to `1.0` and `false` is mapped to `0.0`
 * for string the following logic applies
-  * `"true"` and `"yes"` are mapped to `1.0`, `"false"`, `"no"` and `"unknown"` are mapped to `0.0` (all case-insensitive)
+  * `"true"` and `"yes"` are mapped to `1.0`, `"false"` and `"no"` are mapped to `0.0` (all case-insensitive)
+  * `"unknown"` is mapped to `-1.0` (all case-insensitive)
   * RFC3339 times are parsed to float timestamp  
   * Quantities like "250m" or "512Gi" are parsed to float using <https://github.com/kubernetes/apimachinery/blob/master/pkg/api/resource/quantity.go>
   * Percentages ending with a "%" are parsed to float
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -337,8 +337,7 @@ Supported types are:
 * `nil` is generally mapped to `0.0` if NilIsZero is `true`, otherwise it will throw an error
 * for bool `true` is mapped to `1.0` and `false` is mapped to `0.0`
 * for string the following logic applies
-  * `"true"` and `"yes"` are mapped to `1.0`, `"false"` and `"no"` are mapped to `0.0` (all case-insensitive)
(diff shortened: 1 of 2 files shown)


mrueg:
Thanks for the contribution! 
I don't think 0.0 is the right way to treat an `unknown` value as it could be misleading. I think a possible option might be `-1.0` for any unknown value. (but I would love to see others chime in on that as well)

CatherineF-dev:
Agree on -1 to distinguish this unknown with others.

Haleygo:
@mrueg Thanks for reviewing! I have updated the value and doc, please check.

itaiatu (approved):
/lgtm

itaiatu:
We also encountered this issue and this small fix will solve it.

CatherineF-dev:
Hi @mrueg, could you help add this into v2.14.0 milestone? https://github.com/kubernetes/kube-state-metrics/milestone/7 

Thx

mrueg:
@dgrisonnet made me aware of https://github.com/kubernetes/kube-state-metrics/blob/0738de0be2ba1607aac8b58a0d783891664d48a9/internal/store/node_test.go#L186-L188 in which we expose the status Unknown as 0 as well. Could you change it back to 0 instead of -1 for consistency? 
Thanks and apologies for the extra work!

docs/metrics/extend/customresourcestate-metrics.md
```diff
@@ -337,7 +337,8 @@ Supported types are:
 * `nil` is generally mapped to `0.0` if NilIsZero is `true`, otherwise it will throw an error
 * for bool `true` is mapped to `1.0` and `false` is mapped to `0.0`
 * for string the following logic applies
-  * `"true"` and `"yes"` are mapped to `1.0` and `"false"` and `"no"` are mapped to `0.0` (all case-insensitive)
+  * `"true"` and `"yes"` are mapped to `1.0`, `"false"` and `"no"` are mapped to `0.0` (all case-insensitive)
+  * `"unknown"` is mapped to `-1.0` (all case-insensitive)
```
mrueg, line 341:
```suggestion
  * `"true"` and `"yes"` are mapped to `1.0`, `"false"`, `"no"` and `"unknown"` are mapped to `0.0` (all case-insensitive)
```

pkg/customresourcestate/registry_factory.go
```diff
@@ -724,14 +727,17 @@ func toFloat64(value interface{}, nilIsZero bool) (float64, error) {
...
 		normalized := strings.ToLower(value.(string))
 		if normalized == "true" || normalized == "yes" {
 			return 1, nil
 		}
 		if normalized == "false" || normalized == "no" {
 			return 0, nil
 		}
+		if normalized == "unknown" {
+			return -1, nil
+		}
```
mrueg, line 740:
```suggestion
		if normalized == "false" || normalized == "no" || normalized == "unknown" {
			return 0, nil
		}
```

mrueg:
Made the changes myself real quick to get it into the v2.14.0 release

/lgtm

PR #2545, opened 2024-11-06
chore: Improve benchmarks

mrueg:
Allows to set number of rounds the benchmark should run and removes log output from it as well. 
As the e2e tests take roughly twice the time of the benchmark, we can set it to two rounds (which is still better than running it only once).

None
Fixes #

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -173,7 +173,7 @@ jobs:
         <hr />
         The table shows the median and 95% confidence interval (CI) summaries for each benchmark comparing the HEAD and the BASE, and an A/B comparison under "vs base". The last column shows the statistical p-value with ten runs (n=10).
         The last row has the Geometric Mean (geomean) for the given rows in the table.
-        Refer to [benchstat's documentation](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) for more help.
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -164,7 +164,17 @@ jobs:
...
+        BENCHSTAT_OUTPUT_FILE=result.txt make test-benchmark-compare
+    - run: |
+        echo "\`\`\`" >> "$GITHUB_STEP_SUMMARY"
+        cat result.txt >> "$GITHUB_STEP_SUMMARY"
+        echo "\`\`\`" >> "$GITHUB_STEP_SUMMARY"
+        cat <<EOL >> "$GITHUB_STEP_SUMMARY"
+        <hr />
+        The table shows the median and 95% confidence interval (CI) summaries for each benchmark comparing the HEAD and the BASE, and an A/B comparison under "vs base". The last column shows the statistical p-value with ten runs (n=10).
+        The last row has the Geometric Mean (geomean) for the given rows in the table.
+        Refer to [benchstat's documentation](https://pkg.go.dev/golang.org/x/perf/cmd/benchstat) for more help.
```
rexagod, line 176:
```suggestion
		Refer to <a href="https://pkg.go.dev/golang.org/x/perf/cmd/benchstat">benchstat's documentation</a> for more help.
```

rexagod (changes requested):
The output is much more readable now, thank you!

PR #2547, opened 2024-11-07
fix: Remove redundant factories array since it is not written after

Henrywu573:
construction.

Cleanup by remove unnecessary factories.
 does not change cardinality
This[ git commit](https://github.com/kubernetes/kube-state-metrics/commit/25a1d8da057cf761d614c59a52785335d34082d1#diff-dc8fe36bd1edf1b4cd03bacbd94b02cd4226717fe7e1a6474c534f5b5db30227L156-L164) remove the write to the factories variable, which makes factories always be an empty array.


mrueg:
thanks, makes sense to me.

/lgtm

/hold 
for @rexagod to take a look in case I missed anything.

mrueg:
/hold cancel 

Reviews from two folks should be good enough. Thanks for your contribution!

PR #2553, opened 2024-11-08
fix: panic in `util.GVRFromType` for structured types

### LINKED ISSUE #2202: [Bug] Inconsistency between `customresource` interface and `customresource` causes panic
**Issue Description**:
After the alteration of the custom resources handling behavior in https://github.com/kubernetes/kube-state-metrics/pull/1851, there has been a breaking change.

1. `GVRFromType` is invariably invoked: https://github.com/kubernetes/kube-state-metrics/blob/76f42c19af50ff7c0390a1c57c680b07b6eb3e8a/internal/store/builder.go#L200
2. The subsequent function https://github.com/rexagod/kube-state-metrics/blob/25a1d8da057cf761d614c59a52785335d34082d1/pkg/util/utils.go#L143 puts forward a presumption that `ExpectedType()` persistently returns `unstructured.Unstructured{}`. This triggers panic if a different argument type is provided.

The change leads to a breaking compatibility with the previous interface that we use internally to enhance KSM with other metrics.

Moreover, the `availableStore` index now uses `gvr.String()` over `ResourceName` as per https://github.com/kubernetes/kube-state-metrics/blob/76f42c19af50ff7c0390a1c57c680b07b6eb3e8a/internal/store/builder.go#L203. Consequently, the output of `gvrString` is structured as `Resource=verticalpodautoscalers,apiregistration.k8s.io/v1`

**Expected Outcome**:
I anticipated `ExpectedType()` to have the capacity to return any resource type and `f.Name()` to serve as the resource index in `availableStore`.
Another solution would be to introduce a new function `StoreIndex()` or `StoreName()` in this interface.

**Reproducibility Steps**:
To reproduce this issue, simply add VPA as a custom resource and observe that the code triggers panic.

**Additional Information**:
None applicable.
**Environment Details**:
I'm using the kind cluster 1.27.1.
* Kube-state-metrics version: 2.10.0
* Kubernetes version (retrievable using `kubectl version`): 1.27.1
* Cloud provider or hardware setup: macOS ARM
* Any Other Relevant Information:

L3n41c:

Make `util.GVRFromType` able to process structured types and not only unstructured ones.

Does not change.

* Fixes #2202

### DIFF
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -204,7 +204,11 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 			if err != nil {
 				return nil, fmt.Errorf("failed to create metrics factory for %s: %w", resource.GroupVersionKind, err)
 			}
-			gvrString := util.GVRFromType(factory.Name(), factory.ExpectedType()).String()
+			gvr, err := util.GVRFromType(factory.Name(), factory.ExpectedType())
+			if err != nil {
+				return nil, fmt.Errorf("failed to create GVR for %s: %w", resource.GroupVersionKind, err)
+			}
+			gvrString := gvr.String()
 			if _, ok := factoriesIndex[gvrString]; ok {
 				klog.InfoS("reloaded factory", "GVR", gvrString)
 			}
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -208,7 +208,12 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 			if err != nil {
 				return nil, fmt.Errorf("failed to create GVR for %s: %w", resource.GroupVersionKind, err)
 			}
-			gvrString := gvr.String()
+			var gvrString string
+			if gvr != nil {
+				gvrString = gvr.String()
+			} else {
+				gvrString = factory.Name()
+			}
 			if _, ok := factoriesIndex[gvrString]; ok {
 				klog.InfoS("reloaded factory", "GVR", gvrString)
 			}
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -97,8 +97,11 @@ func CreateCustomResourceClients(apiserver string, kubeconfig string, factories 
 		if err != nil {
 			return nil, err
 		}
-		gvrString := GVRFromType(f.Name(), f.ExpectedType()).String()
-		customResourceClients[gvrString] = customResourceClient
+		gvr, err := GVRFromType(f.Name(), f.ExpectedType())
+		if err != nil {
+			return nil, err
+		}
+		customResourceClients[gvr.String()] = customResourceClient
 	}
 	return customResourceClients, nil
 }
@@ -137,15 +140,14 @@ func CreateDynamicClient(apiserver string, kubeconfig string) (*dynamic.DynamicC
 }
 
 // GVRFromType returns the GroupVersionResource for a given type.
-func GVRFromType(resourceName string, expectedType interface{}) *schema.GroupVersionResource {
+func GVRFromType(resourceName string, expectedType interface{}) (*schema.GroupVersionResource, error) {
 	if _, ok := expectedType.(*testUnstructuredMock.Foo); ok {
 		// testUnstructuredMock.Foo is a mock type for testing
-		return nil
+		return nil, nil
 	}
 	t, err := meta.TypeAccessor(expectedType)
 	if err != nil {
-		klog.ErrorS(err, "Failed to get type accessor", "expectedType", expectedType)
-		return nil
+		return nil, fmt.Errorf("Failed to get type accessor for %T: %w", expectedType, err)
 	}
 	apiVersion := t.GetAPIVersion()
 	g, v, found := strings.Cut(apiVersion, "/")
@@ -158,7 +160,7 @@ func GVRFromType(resourceName string, expectedType interface{}) *schema.GroupVer
 		Group:    g,
 		Version:  v,
 		Resource: r,
-	}
+	}, nil
 }
 
 // GatherAndCount gathers all metrics from the provided Gatherer and counts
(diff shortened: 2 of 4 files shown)


mrueg:
Thanks, makes sense to me.

/lgtm

/hold 
for others to review.

rexagod (changes requested):
+1 for the change proposed, just one nit.

pkg/util/utils.go
```diff
@@ -142,7 +142,11 @@ func GVRFromType(resourceName string, expectedType interface{}) *schema.GroupVer
 		// testUnstructuredMock.Foo is a mock type for testing
 		return nil
 	}
-	apiVersion := expectedType.(*unstructured.Unstructured).Object["apiVersion"].(string)
+	t, err := meta.TypeAccessor(expectedType)
+	if err != nil {
+		panic(err)
```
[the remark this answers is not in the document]
L3n41c: Thanks for your feedback @rexagod!
I’ve just replaced the `panic` by a log in https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/4a96a782fcbfd0b26ee88ff832f79b0af118f667.

pkg/util/utils.go
```diff
@@ -142,7 +142,12 @@ func GVRFromType(resourceName string, expectedType interface{}) *schema.GroupVer
 		// testUnstructuredMock.Foo is a mock type for testing
 		return nil
 	}
-	apiVersion := expectedType.(*unstructured.Unstructured).Object["apiVersion"].(string)
+	t, err := meta.TypeAccessor(expectedType)
+	if err != nil {
+		klog.ErrorS(err, "Failed to get type accessor", "expectedType", expectedType)
+		return nil
```
rexagod, line 148: Apologies to block this again, but since this is used in a bunch of critical paths in CRS procedures, continuing silently after an error wouldn't make sense. I'd push for changing the signature here, and surfacing this up, as pointed in my previous review.

L3n41c: @mrueg @rexagod Sorry for the delay!
I’ve just updated `GVRFromType` to make it return an `error`: https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/61134a08b649dcf33f1f54db6a9d6d3a4e52a067

mrueg: @rexagod could you take a look at these changes?

pkg/customresourcestate/config.go
```diff
@@ -204,7 +204,11 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 			if err != nil {
 				return nil, fmt.Errorf("failed to create metrics factory for %s: %w", resource.GroupVersionKind, err)
 			}
-			gvrString := util.GVRFromType(factory.Name(), factory.ExpectedType()).String()
+			gvr, err := util.GVRFromType(factory.Name(), factory.ExpectedType())
+			if err != nil {
+				return nil, fmt.Errorf("failed to create GVR for %s: %w", resource.GroupVersionKind, err)
+			}
+			gvrString := gvr.String()
```
rexagod, line 211: We probably need a similar check as [`61134a0` (#2553)](https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/61134a08b649dcf33f1f54db6a9d6d3a4e52a067#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R224) here.

pkg/util/utils.go
```diff
@@ -97,8 +97,11 @@ func CreateCustomResourceClients(apiserver string, kubeconfig string, factories
 		if err != nil {
 			return nil, err
 		}
-		gvrString := GVRFromType(f.Name(), f.ExpectedType()).String()
-		customResourceClients[gvrString] = customResourceClient
+		gvr, err := GVRFromType(f.Name(), f.ExpectedType())
```
rexagod, line 100: Here too.

pkg/util/utils.go
```diff
@@ -97,8 +97,11 @@ func CreateCustomResourceClients(apiserver string, kubeconfig string, factories
...
+		if err != nil {
+			return nil, err
+		}
+		customResourceClients[gvr.String()] = customResourceClient
```
rexagod, line 104: Here too?

pkg/customresourcestate/config.go
L3n41c: I hesitated.

The thing is that the `if` block that checks if `gvrString` is `nil` and adapts if it is that I added in [`internal/discovery/discovery.go:224`](https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/61134a08b649dcf33f1f54db6a9d6d3a4e52a067#diff-916e6863e1245c673b4e5965c98dc27bafbd72650fdb38ce65ea73ee6304e027R224-R228) has been copied from [`internal/store/builder.go:205`](https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/61134a08b649dcf33f1f54db6a9d6d3a4e52a067#diff-8de36acb4313849c6ece61305374575fa2f18bdb6113d44022ee9749785d4af2R205-R206) and [`internal/store/builder.go:564`](https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/61134a08b649dcf33f1f54db6a9d6d3a4e52a067#diff-8de36acb4313849c6ece61305374575fa2f18bdb6113d44022ee9749785d4af2R564-R565) because, in those three cases, if `util.GVRFromType(…)` returns an error, we only log an error and continue executing the function.
But, in this case and and in `pkg/util/utils.go` below, if `util.GVRFromType(…)` returns an error, the function forwards the error and exits early so that this line isn’t executed.

L3n41c: @rexagod I’ve just fixed this function to make it gracefully handle the cases where `utils.GVRFromType(…)` returns neither an error nor a valid GVR: https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/c3b1afe61c2a1b7335d4caf99f1ae228dfadbc18

pkg/util/utils.go
L3n41c: Also fixed in https://github.com/kubernetes/kube-state-metrics/pull/2553/commits/c3b1afe61c2a1b7335d4caf99f1ae228dfadbc18

rexagod:
/lgtm

Thank you for the patch, @L3n41c!

PR #2555, opened 2024-11-11
feat: Add kube_node_status_images metric

hausertt:

Collects metrics of container images present on the node. With the recent growth in image sizes, especially for self-hosted AI solutions, monitoring these metrics can help fix node-pressure.

Increases one metric: kube_node_status_images

Fixes #


internal/store/node.go
```diff
@@ -56,9 +57,43 @@ func nodeMetricFamilies(allowAnnotationsList, allowLabelsList []string) []genera
...
+		createNodeStatusImagesFamilyGenerator(),
 	}
 }
 
+func createNodeStatusImagesFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGeneratorWithStability(
+		"kube_node_status_images",
+		"Container Images on the Node",
+		metric.Gauge,
+		basemetrics.STABLE,
```
mrueg, line 69: Should be experimental

mrueg:
This is an interesting idea, though I have concerns about the number of metrics exposed.
Effectively this is #nodes * #images
A rough calculation: In a cluster with 100 nodes, 50 pods scheduled in each node and the image rention being roughly 4 versions, this would mean 2000 additional series being reported.

I'm interested in thoughts from others but would at least say this should be an opt-in metric.

internal/store/node.go
hausertt: Just in case I do it wrong: The basemetrics equivalent to experimental is ALPHA? Sorry if this question seems a bit stupid.

hausertt: Thanks. Done.

PR #2502, opened 2024-09-18
fix: de-duplication of custom resource metrics

bartlettc22:
**What this PR does / why we need it**:
Fixes an issue that causes custom resource metrics to duplicate every time a CR resource that KSM is watching changed.  The duplication was happening because the informer events (add/delete) triggered a subsequent call to `WithEnabledResources` in `store.Builder`.  This would append, but not de-duplicate the CR resources to the `enabledResources` list. This, in turn, would cause the custom resources to be be added multiple times when the store was built.

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)*
Does not change cardinality

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #2446

PR #2559, opened 2024-11-21
fix(kube_pod_tolerations): deduplicate tolerations before creating metric

### LINKED ISSUE #2390: Duplicate tolerations causing issue with prometheus >= 2.52.0
**What happened**:
Starting with version 2.52.0, Prometheus introduced a mechanism to detect duplicate series during scraping. This can lead to error logs when kube-state-metrics scrapes metrics for deployments, particularly if there are duplicate entries within the toleration array.

prometheus debug logs:
```
ts=2024-05-13T19:21:09.190Z caller=scrape.go:1777 level=debug component="scrape manager" scrape_pool=serviceMonitor/monitoring/kube-state-metrics/0 target=https://10.244.5.6:8443/metrics msg="Duplicate sample for timestamp" series="kube_pod_tolerations{namespace=\"calico-system\",pod=\"calico-kube-controllers-75c647b46c-pg9cr\",uid=\"bf944c52-17bd-438b-bbf1-d97f8671bd6b\",key=\"CriticalAddonsOnly\",operator=\"Exists\"}"
ts=2024-05-13T19:21:09.207Z caller=scrape.go:1738 level=warn component="scrape manager" scrape_pool=serviceMonitor/monitoring/kube-state-metrics/0 target=https://10.244.5.6:8443/metrics msg="Error on ingesting samples with different value but same timestamp" num_dropped=1
```

There might be a need to deduplicate the toleration entries or add an index to entries with existing duplicates.

create the following deployment and look at the metrics produced by kube-state-metrics

```
apiVersion: apps/v1
kind: Deployment
metadata:
  name: test-deployment
  labels:
    app: something
spec:
  replicas: 1
  selector:
    matchLabels:
      app: test
  template:
    metadata:
      labels:
        app: test
    spec:
      containers:
      - name: test-container
        image: nginx
      tolerations:
       - key: CriticalAddonsOnly
         operator: Exists
       - key: CriticalAddonsOnly
         operator: Exists
```

Issue report I opened on the prometheus project https://github.com/prometheus/prometheus/issues/14089

**Environment**:

* kube-state-metrics version: 2.12.0
* Kubernetes version (use `kubectl version`): 1.27.9
* Cloud provider or hardware configuration: AKS
* Other info:

RiRa12621:
duplicate metrics currently create warnings in the logs of prometheus. Tolerations can be duplicated and kube-api doesn't want to change that behavior, hence we have to deduplicate here

does not change cardinality

Fixes #2390

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1642,6 +1642,21 @@ func createPodStatusUnschedulableFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+// getUniqueTolerations takes an array
+func getUniqueTolerations(tolerations []v1.Toleration) []v1.Toleration {
+	uniqueTolerationsMap := make(map[v1.Toleration]struct{})
+	var uniqueTolerations []v1.Toleration
+
+	for _, toleration := range tolerations {
+		_, exists := uniqueTolerationsMap[toleration]
+		if !exists {
+			uniqueTolerationsMap[toleration] = struct{}{}
+			uniqueTolerations = append(uniqueTolerations, toleration)
+		}
+	}
+	return uniqueTolerations
+}
+
 func createPodTolerationsFamilyGenerator() generator.FamilyGenerator {
 	return *generator.NewFamilyGeneratorWithStability(
 		"kube_pod_tolerations",
@@ -1651,8 +1666,9 @@ func createPodTolerationsFamilyGenerator() generator.FamilyGenerator {
 		"",
 		wrapPodFunc(func(p *v1.Pod) *metric.Family {
 			var ms []*metric.Metric
+			uniqueTolerations := getUniqueTolerations(p.Spec.Tolerations)
 
-			for _, t := range p.Spec.Tolerations {
+			for _, t := range uniqueTolerations {
 				var key, operator, value, effect, tolerationSeconds string
 
 				key = t.Key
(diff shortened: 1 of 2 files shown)


mrueg:
The solution looks good to me.
Any thoughts on using slices.Sort + slices.Compact similar to https://github.com/kubernetes/kube-state-metrics/pull/2502 ? This might be simpler than building a separate function for it.

CatherineF-dev:
> The solution looks good to me. Any thoughts on using slices.Sort + slices.Compact similar to #2502 ? This might be simpler than building a separate function for it.

+1

RiRa12621:
Sure, I'll get that changed tomorrow

rgarcia89:
@RiRa12621 are you still working on that PR?

RiRa12621:
Planning to, was just overly busy at day job

remram44:
> The solution looks good to me. Any thoughts on using slices.Sort + slices.Compact similar to #2502 ? This might be simpler than building a separate function for it.

I don't know if it's easier, to use `slices.Sort` rather than a map, you need Ordered rather than Comparable, so you would need to provide a comparison function which would be quite long (you need to sort by key+operator+value+effect).

mrueg:
Good point, I think this is good to go then. 

Thanks for your contribution @RiRa12621 

/lgtm

PR #2567, opened 2024-12-02
fix: Fix CR cache for GVK all specified case

### LINKED ISSUE #2566: List/watch store shouldn't be created for non-existing CRD
**What happened**:

I was tring to use CRS to collect my CR metrics and was looking into the codes, and found something that might be a bug.

IIUC everytime there's a new Add/Delete event in CRDiscoverer, it will [re-build the CR stores and restart all available stores](https://github.com/kubernetes/kube-state-metrics/blob/main/internal/discovery/discovery.go#L254-L256). `factoryGenerator` is where it recreates CR store builders using CRS config and discovery cache, which points to [here](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/customresourcestate/config.go#L186-L217). For every Custom Resource that is defined in CRS config, `ResolveGVKToGVKPs` function is used for finding the matching CR in the discovery cache, and only those having matched entries in the cache map will be returned. However, for the case where both version and kind is specified, it will [always return GVK no matter whether it exists in the cache or not](https://github.com/kubernetes/kube-state-metrics/blob/release-2.14/internal/discovery/discovery.go#L129-L147). This is causing the following issue:

For example, in CRS config I define my resource with GVK all specified:

```
spec:
  resources:
    - groupVersionKind:
        group: jobset.x-k8s.io
        kind: "JobSet"
        version: "v1alpha2"
```

Even if JobSet CRD is **not** registered, KSM will still bring up a store to list and watch my JobSet resource which causes spamming list/watch error logs `failed to list`.

Is this intended?

From what I understand, for the above case, it shouldn't create a list/watch store for JobSet when it's not registered in API server.

Basically in `ResolveGVKToGVKPs`, the logic should be:

```
	if hasVersion && hasKind {
		for _, el := range r.Map[g][v] {
			if el.Kind == k {
				return []groupVersionKindPlural{
						{
							GroupVersionKind: schema.GroupVersionKind{
								Group:   g,
								Version: v,
								Kind:    k,
							},
							Plural: el.Plural,
						},
				}, nil
			}
		}
	}
```

```bash
# An example: https://github.com/kubernetes/kube-state-metrics/issues/2223#issuecomment-1792850276
minikube start
...
go run main.go --custom-resource-state-only --custom-resource-state-config-file ksm-2223/custom-resource-config-file.yaml --kubeconfig ~/.kube/config
```

**Environment**:

* kube-state-metrics version:
* Other info:

chelseychen:

Currently in CR discoverer, for the case where the GVK are all specified for a CR in CRS config (see below example) but CRD not registered, it will still have that CR enabled in CR factories.

```
spec:
  resources:
    - groupVersionKind:
        group: jobset.x-k8s.io
        kind: "JobSet"
        version: "v1alpha2"
```

The expected behaviour is that that resource shouldn't be enabled since the cache map doesn't contain that CRD.

This PR fixes this issue.

No change for cardinality.

Fixes #2566

### DIFF
--- a/internal/discovery/discovery_test.go
+++ b/internal/discovery/discovery_test.go
@@ -163,6 +163,23 @@ func TestGVKMapsResolveGVK(t *testing.T) {
 				},
 			},
 		},
+		{
+			desc: "fixed version and kind, no matching cache entry",
+			gvkmaps: &CRDiscoverer{
+				Map: map[string]map[string][]kindPlural{
+					"testgroup": {
+						"v1": {
+							kindPlural{
+								Kind:   "TestObject2",
+								Plural: "testobjects2",
+							},
+						},
+					},
+				},
+			},
+			gvk:  schema.GroupVersionKind{Group: "testgroup", Version: "v1", Kind: "TestObject1"},
+			want: nil,
+		},
 	}
 	for _, tc := range testcases {
 		got, err := tc.gvkmaps.ResolveGVKToGVKPs(tc.gvk)
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -127,23 +127,20 @@ func (r *CRDiscoverer) ResolveGVKToGVKPs(gvk schema.GroupVersionKind) (resolvedG
 	hasKind := k != "" && k != "*"
 	// No need to resolve, return.
 	if hasVersion && hasKind {
-		var p string
 		for _, el := range r.Map[g][v] {
 			if el.Kind == k {
-				p = el.Plural
-				break
+				return []groupVersionKindPlural{
+					{
+						GroupVersionKind: schema.GroupVersionKind{
+							Group:   g,
+							Version: v,
+							Kind:    k,
+						},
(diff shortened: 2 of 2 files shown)


internal/discovery/discovery_test.go
```diff
@@ -163,6 +163,74 @@ func TestGVKMapsResolveGVK(t *testing.T) {
...
+								Kind:   "TestObject2",
+								Plural: "testobjects2",
+							},
+						},
+					},
+				},
+			},
+			gvk:  schema.GroupVersionKind{Group: "testgroup", Version: "v1", Kind: "TestObject1"},
+			want: nil,
+		},
```
rexagod, line 182: I believe this is the only test case needed to verify the change above (the fact that the first case ends up taking whatever the `group` is). The other three seem to pass before this patch too.

rexagod (commented):
Good catch, thank you! Just one nit.

rexagod:
/lgtm

Thank you!

PR #2571, opened 2024-12-10
feat: Adds category label for unschedulable pod as part of `kube_pod_status_unschedulable` metric

ilaylow:

This PR allows us to determine if a pod's reason for being unschedulable by introducing a custom `category` label:
- `RESOURCE` - Pod is unschedulable due to insufficient resources.
- `AFFINITY` - Pod is unschedulable due to pod affinity.
- `UNKNOWN` - Pod is unschedulable due to unknown reasons.

The `category` is derived from making assumptions on the the `PodCondition.Message` value (see below).

```go
// PodCondition contains details for the current condition of this pod.
type PodCondition struct {
	...
	// Human-readable message indicating details about last transition.
	// +optional
	Message string `json:"message,omitempty" protobuf:"bytes,6,opt,name=message"`
        ...
}
```

Increases cardinality mildly due to the addition of a label with 3 dimensions.

(No ongoing issue)

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1628,8 +1628,8 @@ func createPodStatusUnschedulableFamilyGenerator() generator.FamilyGenerator {
 			for _, c := range p.Status.Conditions {
 				if c.Type == v1.PodScheduled && c.Status == v1.ConditionFalse {
 					ms = append(ms, &metric.Metric{
-						LabelKeys:   []string{},
-						LabelValues: []string{},
+						LabelKeys:   []string{"reason"},
+						LabelValues: []string{c.Message},
 						Value:       1,
 					})
 				}
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -19,6 +19,7 @@ package store
 import (
 	"context"
 	"strconv"
+	"strings"
 
 	basemetrics "k8s.io/component-base/metrics"
 	"k8s.io/utils/net"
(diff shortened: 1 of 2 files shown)


mrueg:
Thanks for the PR. We omitted the message on purpose as it can cause high cardinality due to the fact it's not limited to a specific set of messages, so this is a feature we cannot include.

ilaylow:
> Thanks for the PR. We omitted the message on purpose as it can cause high cardinality due to the fact it's not limited to a specific set of messages, so this is a feature we cannot include. 

Ah I figured that would be the case. Would you be open to me introducing categorisation to a limited set of values then? 

What I'm really looking to get out of this is a way to distinguish between whether a pod was unschedulable due to insufficient resources, or due to pod affinities etc.

ilaylow:
> Thanks for the PR. We omitted the message on purpose as it can cause high cardinality due to the fact it's not limited to a specific set of messages, so this is a feature we cannot include.

@mrueg 

I added a naive example implementation in my latest commit [99d0fb6](https://github.com/kubernetes/kube-state-metrics/pull/2571/commits/99d0fb6906940d9e798794c94752c1229b189ddc) to address the cardinality issue, could I get your thoughts on this?

dgrisonnet:
KMS is direct mapping of Kubernetes object to metrics, it exposes the data as is without any modification. Also, maintaining an updated list of categories would be quite painful from a maintainer perspective

PR #2572, opened 2024-12-11
fix: make `$VERSION` 3rd-party independant

rexagod:
3rd-party dependencies cannot be installed pre-build since Makefile variables are initialized earlier than that. At the moment, this causes `kube_state_metrics_build_info` to have a labelset of `{...,version="v"}`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -2,7 +2,7 @@ FLAGS =
 TESTENVVAR =
 REGISTRY ?= gcr.io/k8s-staging-kube-state-metrics
 TAG_PREFIX = v
-VERSION = $(shell gomplate -d config=./data.yaml --in '{{ (datasource "config").version }}')
+VERSION = $(shell grep '^version:' data.yaml | grep -oE "[0-9]+.[0-9]+.[0-9]+")
 TAG ?= $(TAG_PREFIX)$(VERSION)
 LATEST_RELEASE_BRANCH := release-$(shell echo $(VERSION) | grep -ohE "[0-9]+.[0-9]+")
 BRANCH = $(strip $(shell git rev-parse --abbrev-ref HEAD))


rexagod:
Addressed in `v2.14` by https://github.com/kubernetes/kube-state-metrics/pull/2456/files.

rexagod:
@mrueg I'm opening this again as this seems to be a regression for air-tight builders that can only utilize `/vendor`'s content, but not download, for instance, the contents of `tools` if they seemingly for a part of the production build pipeline (i.e., the lack of `gomplate` results in an empty `-ldflag` for version, which results in an empty `version` labelvalue).

I'd like to cut a `2.14.1` to address this for such builders, WDYT?

mrueg:
I see two other options here:

a) add gomplate to your path in the build infra before the actual build
b) wait for go1.24 https://tip.golang.org/doc/go1.24#tools

Would those help?

rexagod:
While the second approach sounds very cool, it can take a while for us to get there. The first one will require us to keep downstream assets (all `tools` binaries) and plug them into a `$GOBIN` or `$PATH` to pick them during `make install-tools`. I find myself leaning away from the multi-point complexity that this may introduce, in favor of making the production build process third-party dependency-free (except what's already in `/vendor`).

To correct myself, we don't necessarily need to cut a minor release for this, if this is something the wider community isn't facing (I can cherry-pick the commit to our midstream).

mrueg:
I really would like to get version information programmatically. Wouldn't it be possible for you to vendor the tools folder with it's go.mod file as well?

rexagod:
Let me look into that. But I'm not sure what you mean by "getting the version information programmatically"? IIUC we do so right now as well (in this patch), without any third-party dependencies?

rexagod:
Post-vendoring, we'll need to build the binaries from that, and add them to `PATH`. I'd prefer to keep that logic upstream, so air-tight builders and others can equally depend on that workflow without the need to resort to their own makeshift workarounds. However, this will require us to start vendoring in KSM, and modify the `Dockerfile` to build and use the vendored binary instead of `make install-tools` (or anything that makes a network request).

Personally, I'd wish to see the build process pure and independent of any non-`go` or non-`core-utils` binary, with all production dependencies defined in the root `go.mod`, while the nested ones may be used for other non-production purposes. Continuing my previous comment, I believe the `^version: \d.\d.\d` assumption in this patch is a safe one, WDYT?

rexagod:
CC'ing @mrueg for a consensus here. 🙇🏼

Makefile
```diff
@@ -2,7 +2,7 @@ FLAGS =
 TESTENVVAR =
 REGISTRY ?= gcr.io/k8s-staging-kube-state-metrics
 TAG_PREFIX = v
-VERSION = $(shell gomplate -d config=./data.yaml --in '{{ (datasource "config").version }}')
+VERSION = $(shell grep '^version:' data.yaml | grep -oE "[0-9]+.[0-9]+.[0-9]+")
```
mrueg, line 5:
```suggestion
VERSION ?= $(shell gomplate -d config=./data.yaml --in '{{ (datasource "config").version }}')
```

mrueg:
As agreed in Slack, we'll proceed with this and revert back once go install -tool from go 1.24 is available.

/lgtm

Issue #2582, opened 2025-01-02
kube_service_info should also include information about the external traffic policy of the service


jahantech:
**What would you like to be added**:
A new label to kube_service_info to include the info about external traffic policy.

**Why is this needed**:
This will help in identifying the services that requires all or limited number of nodes to be healthy behind a service type LoadBalancer.

**Describe the solution you'd like**
Addition of new label with external traffic policy information or create a new metric.

PR #2584, opened 2025-01-02
feat: Add External traffic policy to kube_service_info

jahantech:
This PR adds another label to kube_service_info metric that gives us the value for externalTrafficPolicy. SREs can use this field to verify expected number of healthy nodes for each service by combining this metric with a CSP related metric.

Increases the cardinality as this PR adds a label with external Traffic Policy value. 

Resolves: #2582


jahantech:
I am happy to create a new metric if this is not acceptable.

Thanks.

mrueg:
Thanks for your contribution! No this makes sense and it's either Cluster or Local if set so cardinality wise fine. :)

/lgtm

PR #2607, opened 2025-02-09
feat(service): add kube_service_ports metric

nahuel11500:
Adds a simple metric kube_service_ports so that we can get details about the ports exposed by services.
increases


maximee-leroyy (approved):

nahuel11500:
Hello, any thought on this ? @mrueg @dgrisonnet

Thank you

nahuel11500:
Hello, sorry to bother, any update on this ?  @dgrisonnet

Thank you

internal/store/service.go
```diff
@@ -184,6 +185,49 @@ func serviceMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_service_ports",
+			"Metric providing details about the ports exposed by services.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapSvcFunc(func(e *v1.Service) *metric.Family {
+				ms := []*metric.Metric{}
+				labelKeys := []string{"port_protocol", "port_number", "port_name", "port_target", "port_node_number", "port_app_protocol"}
```
CatherineF-dev, line 196: It has the cardinality issue.

SaschaJohn: @CatherineF-dev we've the same requirement and would appreciate if this could be merged.

maxime1907: @CatherineF-dev any update on this?

nahuel11500:
Hello, sorry to bother,

Any update on this ? Should we consider it as refused because of the cardinality issue ? @CatherineF-dev @dgrisonnet 

Thank you

internal/store/service.go
CatherineF-dev: Can you share more details on connecting more easily on them with Telepresence? Do we have other alternatives?

nahuel11500: It would allow us to expose a direct URL that corresponds to the service’s DSN inside the cluster. This means developers using Telepresence could simply click the link and be routed directly to the service.

Currently, given the metrics kube-state-metrics currently exposes, we don’t have any alternative way to achieve this. But if you can think of another way, I'll be glad to take some time to consider it. 

Thanks :)

internal/store/service.go
```diff
@@ -184,6 +185,49 @@ func serviceMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				}
 			}),
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_service_ports",
```
dgrisonnet: kube_service_spec_ports

internal/store/service.go
dgrisonnet: adding a `port` prefix is redundant, I'd advice to remove it

dgrisonnet: I agree with @CatherineF-dev this metrics will generate too many timeseries. Because of that I should we should make it opt-in. You can do that by using the `generator.NewOptInFamilyGenerator`.

nahuel11500: Alright, thanks for taking time to review, I updated the code to have those changes.

Issue #2403, opened 2024-05-28
Duplicate kube_horizontalpodautoscaler_spec_target_metric


dmitriishaburov:
**What happened**:

Duplicate `kube_horizontalpodautoscaler_spec_target_metric` causing issues with Prometheus 2.52.0+
Due to new duplicates detection mechanism, I'm seeing following errors in Prometheus:
```
ts=2024-05-28T07:15:26.730Z caller=scrape.go:1738 level=warn component="scrape manager" scrape_pool=serviceMonitor/prometheus/kube-prometheus-stack-kube-state-metrics/0 target=http://172.17.60.213:8080/metrics msg="Error on ingesting samples with different value but same timestamp" num_dropped=12
```

After checking kube-state metrics I've found that metrics for HPA are duplicated:
```
cat metrics | grep 'kube_horizontalpodautoscaler_spec_target_metric{namespace="dummy",horizontalpodautoscaler="dummy",metric_name="cpu",metric_target_type="utilization"}'
kube_horizontalpodautoscaler_spec_target_metric{namespace="dummy",horizontalpodautoscaler="dummy",metric_name="cpu",metric_target_type="utilization"} 100
kube_horizontalpodautoscaler_spec_target_metric{namespace="dummy",horizontalpodautoscaler="dummy",metric_name="cpu",metric_target_type="utilization"} 100
kube_horizontalpodautoscaler_spec_target_metric{namespace="dummy",horizontalpodautoscaler="dummy",metric_name="cpu",metric_target_type="utilization"} 100
```

HPA itself have 3 separate entries for CPU, but there's different on container:
```
k get hpa dummy -oyaml
  metrics:
  - containerResource:
      container: dummy-0
      name: cpu
      target:
        averageUtilization: 100
        type: Utilization
    type: ContainerResource
  - containerResource:
      container: dummy-1
      name: cpu
      target:
        averageUtilization: 100
        type: Utilization
    type: ContainerResource
  - containerResource:
      container: dummy-2
      name: cpu
      target:
        averageUtilization: 100
        type: Utilization
    type: ContainerResource
```

**What you expected to happen**:
kube-state-metrics not producing duplicate metrics, probably by adding container label (?)

**How to reproduce it (as minimally and precisely as possible)**:
- Create HPA as posted above
- Use Prometheus 2.52.0+ as scraper

**Environment**:

* kube-state-metrics version: v2.12.0
[...]

PR #2614, opened 2025-02-19
feat: Add new hpa metrics to prevent prometheus timeseries duplication

CountryTk:

Added 4 new hpa metrics to prevent duplicated timeseries events like described in this issue: https://github.com/kubernetes/kube-state-metrics/issues/2403

Added new metrics are:
- kube_horizontalpodautoscaler_spec_target_container_metric
- kube_horizontalpodautoscaler_spec_target_object_metric
- kube_horizontalpodautoscaler_status_target_container_metric
- kube_horizontalpodautoscaler_status_target_object_metric

Cardinality is increased because of new metrics

Fixes # https://github.com/kubernetes/kube-state-metrics/issues/2403

FYI: I've also tested this change in our prelive cluster and for us it fixed the issue


richabanker:
/triage accepted
@CatherineF-dev @rexagod could you take a look here if possible, thanks!

CountryTk:
Hey, could this please be reviewed @CatherineF-dev @rexagod

CatherineF-dev:
Ok! In reviewing

internal/store/horizontalpodautoscaler.go
```diff
@@ -50,7 +50,9 @@ var (
 	descHorizontalPodAutoscalerLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descHorizontalPodAutoscalerLabelsDefaultLabels = []string{"namespace", "horizontalpodautoscaler"}
 
-	targetMetricLabels = []string{"metric_name", "metric_target_type"}
+	targetMetricLabels    = []string{"metric_name", "metric_target_type"}
+	containerMetricLabels = []string{"metric_name", "metric_target_type", "container"}
+	objectMetricLabels    = []string{"metric_name", "metric_target_type", "full_target_name"}
```
CatherineF-dev, line 55: nit: full_target_name -> target_name, container -> container_name

internal/store/horizontalpodautoscaler.go
```diff
@@ -192,22 +248,65 @@ func createHPASpecTargetMetric() generator.FamilyGenerator {
...
+						Value:       metricValue,
+					})
+				}
+			}
+			return &metric.Family{Metrics: ms}
+		}),
+	)
+}
+
+func createHPASpecTargetMetric() generator.FamilyGenerator {
```
CatherineF-dev, line 288: nit: Unify createHPASpecTargetMetric and createHPASpecTargetObjectMetric. These two functions have  some similar codes.

For example 
- createHPASpecTargetMetric(CollectContainerResourceMetricSourceType=false)
- createHPASpecTargetMetric(CollectContainerResourceMetricSourceType=true)

CatherineF-dev:
Overall LGTM. Two small comments.

CountryTk:
> Overall LGTM. Two small comments.

Thanks for the review, I've implemented your suggestions

internal/store/horizontalpodautoscaler.go
```diff
@@ -59,7 +61,11 @@ func hpaMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 		createHPAMetaDataGeneration(),
 		createHPASpecMaxReplicas(),
 		createHPASpecMinReplicas(),
-		createHPASpecTargetMetric(),
+		createHPASpecTargetContainerMetric(),
+		createHPAStatusTargetContainerMetric(),
+		createHPAStatusTargetObjectMetric(),
+		createHPASpecTargetMetric(true),
```
CatherineF-dev, line 67: Nit: 
createHPASpecTargetObjectMetric()
createHPASpecTargetMetric()

func createHPASpecTargetObjectMetric() {
    createHPASpecTarget([]{autoscaling.ObjectMetricSourceType})
}

func createHPASpecTargetMetric() {
    createHPASpecTarget([]{autoscaling.PodsMetricSourceType, autoscaling.ResourceMetricSourceType, ...})
}

func createHPASpecTarget(autoscalingTypes) {
   if m.Type not in autoscalingTypes {
      return
   }
}

CountryTk: Made some adjustments according to how I understood this comment, not 100% sure if I did it as you wish though.

CountryTk: Hey, @CatherineF-dev could you take a look if the changes requested are what you desired? Thank you

rexagod:
Thank you for the patch.

I believe this identifies a resource-agnostic pitfall where we loop over certain nested fields without including a primary key in the overall generated metrics' label-sets.

I'll take a closer look tomorrow but so for this lgtm.

internal/store/horizontalpodautoscaler.go
```diff
@@ -180,10 +188,10 @@ func createHPASpecMinReplicas() generator.FamilyGenerator {
 	)
 }
 
-func createHPASpecTargetMetric() generator.FamilyGenerator {
+func createHPASpecTargetContainerMetric() generator.FamilyGenerator {
 	return *generator.NewFamilyGeneratorWithStability(
-		"kube_horizontalpodautoscaler_spec_target_metric",
```
rexagod, line 185: Yeah, this is going to be problematic. Here's the thing with stability guarantees in KSM:
* ALPHA: labelsets may be dropped or added to the metric,
* BETA: labelsets may only be added to the metric,
* STABLE: labelsets are frozen.

However, removing a metric itself, needs to undergo a deprecation cycle. In order to facilitate migration to patterns that avoid the recent "duplication" errors [post-Promtheus v2.52.0](https://github.com/prometheus/prometheus/issues/14089) we probably should to gate all such metrics behind a, say, `--legacy-metrics` and state that these may be duplicated, disabled by default.

@mrueg WDYT?

rexagod (commented):
Question, wouldn't the earlier `kube_horizontalpodautoscaler_spec_target_metric` still be there after this is in, and cause the error to still show up?

CatherineF-dev:
Could you help enable `allowed changes from maintainers`? 

I want to make a small change.

CountryTk:
> Could you help enable `allowed changes from maintainers`?
> 
> I want to make a small change.
> 
> Or could you apply these small changes? [96d948a](https://github.com/kubernetes/kube-state-metrics/commit/96d948a6b0f00fb53f4922d2286607e6c84bf9e9)
image
For me it shows maintainer edit access is enabled.

Anyway, I've added your suggested changes in the latest commit.




> Question, wouldn't the earlier `kube_horizontalpodautoscaler_spec_target_metric` still be there after this is in, and cause the error to still show up?

Nope because `kube_horizontalpodautoscaler_spec_target_metric` is only going to have PodsMetricSourceType, ResourceMetricSourceType and ExternalMetricSourceType now, which didn't cause those duplicated errors.

ContainerResourceMetricSourceType and ObjectMetricSourceType have separate functions with added labels to prevent duplicated data

internal/store/horizontalpodautoscaler.go
CatherineF-dev: Overall LGTM. Left a small comment on some metrics are removed.

internal/store/horizontalpodautoscaler_test.go
```diff
@@ -213,13 +224,9 @@ func TestHPAStore(t *testing.T) {
 				kube_horizontalpodautoscaler_metadata_generation{horizontalpodautoscaler="hpa1",namespace="ns1"} 2
 				kube_horizontalpodautoscaler_spec_max_replicas{horizontalpodautoscaler="hpa1",namespace="ns1"} 4
 				kube_horizontalpodautoscaler_spec_min_replicas{horizontalpodautoscaler="hpa1",namespace="ns1"} 2
-				kube_horizontalpodautoscaler_spec_target_metric{horizontalpodautoscaler="hpa1",metric_name="cpu",metric_target_type="utilization",namespace="ns1"} 80
 				kube_horizontalpodautoscaler_spec_target_metric{horizontalpodautoscaler="hpa1",metric_name="events",metric_target_type="average",namespace="ns1"} 30
-				kube_horizontalpodautoscaler_spec_target_metric{horizontalpodautoscaler="hpa1",metric_name="hits",metric_target_type="average",namespace="ns1"} 12
```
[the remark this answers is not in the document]
CountryTk: metric name "hits" and some others are now under their respective metric
`hits` is now under `kube_horizontalpodautoscaler_spec_target_object_metric` instaed of `kube_horizontalpodautoscaler_spec_target_metric`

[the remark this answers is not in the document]
CountryTk: Yes it is a breaking change but I was under the impression that it's fine since these metrics are experimental.
https://github.com/kubernetes/kube-state-metrics/issues/2403#issuecomment-2276398118
>I like the second option since the metric is still experimental so we can make changes to it. It will surely break some users but we didn't provide guarantees for this metrics. So as long as we make a smooth transition it should be fine.
The benefit of this approach is that the metrics will be specialized to the metric type they are targeting, so the labels will make sense for it and we won't run in scenarios where we have labels that are not used by some types.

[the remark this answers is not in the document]
rexagod:
```
kube_horizontalpodautoscaler_spec_target_object_metric{target_name="",horizontalpodautoscaler="hpa1",metric_name="hits",metric_target_type="average",namespace="ns1"} 12
```

@dgrisonnet I'm not sure if this is what you meant, can you PTAL?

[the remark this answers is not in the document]
CountryTk: Hi, any updates? @rexagod @dgrisonnet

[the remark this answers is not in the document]
CountryTk: @CatherineF-dev Could I get some clarification on this task? Is it blocked or can we merge it?

rexagod:
I'll bring this up in the call tomorrow.

CountryTk:
> I'll bring this up in the call tomorrow.

Hey, any updates?

CountryTk:
Is there any chance we can get this merged?

CountryTk:
@rexagod Any possibility we can get this merged?

internal/store/horizontalpodautoscaler_test.go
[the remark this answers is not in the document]
dgrisonnet: @rexagod yes that is similar to what I had in mind and we would do the same for container_resource_metrics, for resource_metrics, ... That said after reading my initial comment again, the idea of adding the fully qualified name of the object is wrong as it would make correlation more difficult.

I think we should have something like that:

```
kube_horizontalpodautoscaler_spec_target_object_metric{
  object_kind=""
  object_apiversion=""
  object_name=""
  horizontalpodautoscaler="hpa1",
  metric_name="hits",
  metric_target_type="average",
  namespace="ns1"
}

kube_horizontalpodautoscaler_spec_target_container_resource_metric{
  container_name=""
  horizontalpodautoscaler="hpa1",
  metric_name="hits",
  metric_target_type="average",
  namespace="ns1"
}
```

we could instead extend the existing metric and add `target_name`, `target_kind` and `target_apiversion` to it, but I feel like having dedicated metrics would be easier to extend and work with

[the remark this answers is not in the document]
CountryTk: @dgrisonnet I'm a bit confused here, so do I delete the metrics I created and instead add `target_name`, `target_kind` and `target_apiversion` to the original metrics (`kube_horizontalpodautoscaler_spec_target_metric` and `kube_horizontalpodautoscaler_status_target_metric`) or do I leave it as it is right now because 
> having dedicated metrics would be easier to extend and work with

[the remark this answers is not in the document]
dgrisonnet: what you did is in line with what I had in mind, I was just sharing more details about my reasoning with @rexagod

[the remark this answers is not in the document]
CountryTk: @dgrisonnet it seems to me like this PR is in a limbo of some sort
Is there any chance this can get merged or

dudicoco:
Hi @CountryTk,

Looks like this PR does not fix external metric types.
Currently when you have multiple external metrics kube-state-metrics cannot deduplicate them properly.
I believe the correct way to deduplicate them is according to the `selector.matchLabels` field.

dkulchinsky:
hoping this is still being considered

CountryTk:
> hoping this is still being considered

I hope the same.



> Hi @CountryTk,
> 
> Looks like this PR does not fix external metric types. Currently when you have multiple external metrics kube-state-metrics cannot deduplicate them properly. I believe the correct way to deduplicate them is according to the `selector.matchLabels` field.

True, the company I worked for during the time of making this PR had no external metric types which is why I didn't touch upon those, feel free to make improvements. I just hope this PR gets merged

CatherineF-dev:
Looks good to me.

internal/store/horizontalpodautoscaler.go
```diff
@@ -229,9 +291,133 @@ func createHPASpecTargetMetric() generator.FamilyGenerator {
 				}
 
 				for metricTypeIndex, metricValue := range metricMap {
+					metricLabels := targetMetricLabels
+					labelValues := []string{metricName, metricTypeIndex.String()}
+					if m.Type == autoscaling.ObjectMetricSourceType {
```
CatherineF-dev: nit: move ObjectMetricSourceType out of createHPASpecTarget.

createHPASpecTargetObjectMetric()

CountryTk: There's quite a bit of work to move it out of that function since `fullTargetName` is `ObjectMetricSourceType` specific.
I propose we leave it as it is and get this PR merged since it has been marinating for quite some time now

richabanker: Since we already have a dedicated caller `createHPASpecTargetObjectMetric()`, is it a lot of work to move the entire implementation (and the special object-specific logic in `createHPASpecTarget`) out of `createHPASpecTarget()` and into `createHPASpecTargetObjectMetric()`? It does cause some duplication but has gains regarding readability and testability for both functions. 

Relying on one-off if statements like `if m.Type == autoscaling.ObjectMetricSourceType` -> then do some special handling, makes the code brittle, increasing the risk that the custom label logic will be easy to miss or overlook during future patches...

CountryTk: I'm not opposed to that either, I've updated my PR with your proposed solution @richabanker. I added the functionality with some duplicate code into `createHPASpecTargetObjectMetric` and then removed the artifacts from `createHPASpecTarget`

brainfair:
Hi @CatherineF-dev, can it be merged and released?

nqminhdl:
Can it be merged and released?

CountryTk:
@CatherineF-dev @dgrisonnet Hey, what's the hold up here? This seems to be quite the popular issue, let's merge it

dkulchinsky:
Hey 👋 is there any chance to get this merged?

PR #2626, opened 2025-03-10
WIP: feat: Introduce object limits

### LINKED ISSUE #2622: Cap / Limit number of objects ingested for native and Custom Resource Metrics
KSM should have the ability to set an upper limit on number of objects ingested.
**Why is this needed**:
We observed an event where an autoscaler by accident created 10k+ ReplicaSets which KSM tried to report on. This caused KSM to run out of memory and we lost visibility into the cluster. 

* Have a generic and a resource-level command-line option that KSM should use to limit number of items read from the Kubernetes API.
* Have metrics exposed `kube_objects_watched{group="foo", kind="bar" version="baz"}` and `kube_objects_watched_max` which shows the configuration limit to allow alerting if the threshold gets hit.
**Additional context**

mrueg:
This change allows user-controlled limits on how many objects KSM will list from the API. This is helpful to prevent resource exhaustion on KSM, in case the API creates too many resources.

The object limit it set globally and applied per resource watched.

none
Fixes #2622


mrueg:
The object limits seem to work, the metrics reporting for some reason does not report what I intended.

pkg/watch/watch.go
```diff
@@ -60,45 +77,70 @@ type InstrumentedListerWatcher struct {
...
 	}
 
-	i.metrics.ListTotal.WithLabelValues("success", i.resource).Inc()
+	list, err := meta.ExtractList(res)
+	if err != nil {
+		return nil, err
+	}
+	i.metrics.ListRequestsTotal.WithLabelValues("success", i.resource).Inc()
+
+	if i.limit != 0 {
```
rexagod: We'd want to validate against negative values, also s/int/uint preferably. I'm reviewing this on my way back from work so apologies if I missed something obvious here.

mrueg: Good point, to be honest, I have no idea why the kubernetes API made limits a signed int. https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ListOptions

pkg/options/options.go
```diff
@@ -202,5 +204,9 @@ func (o *Options) Validate() error {
 		return fmt.Errorf("value for --auto-gomemlimit-ratio=%f must be greater than 0 and less than or equal to 1", o.AutoGoMemlimitRatio)
 	}
 
+	if o.ObjectLimit < 0 {
+		return fmt.Errorf("value for --object-limit=%d must be equal or greater than 0", o.ObjectLimit)
```
rexagod, line 208: Do we want this to be greater than zero? I see in a couple of instances in `watch.go` that we check for this being greater than zero? Or maybe we want the conditions to reflect this value (0) as well, and as such, accommodate this in `watch.go` as well?

The latter would make more sense as in we keep the same expectations as `cache.ListerWatcher` allows upstream.

rexagod (commented):
Can we have an e2e test-case as well that covers `--object-limit`? Also one nit, otherwise LGTM.

pkg/options/options.go
mrueg: It throws an error because we assume that a negative object limit is invalid. If it's zero, we don't want to throw an error because this is the default and means that it's not set.

mrueg:
@rexagod I added another e2e test, please take a look if you have some spare cycles.

internal/store/builder.go
```diff
@@ -547,7 +555,7 @@ func (b *Builder) buildCustomResourceStores(resourceName string,
 	metricFamilies []generator.FamilyGenerator,
 	expectedType interface{},
 	listWatchFunc func(customResourceClient interface{}, ns string, fieldSelector string) cache.ListerWatcher,
-	useAPIServerCache bool,
+	useAPIServerCache bool, objectLimit int64,
```
dgrisonnet, line 558: @rexagod is this all good to integrate into rsm?

rexagod: +1. I don't see any issues with supporting this in RSM, tracking this here: https://github.com/rexagod/resource-state-metrics/issues/4.

rexagod:
/lgtm
/hold

For others to take a look.

Issue #2551, opened 2024-11-08
Propagate select pod labels to all metrics without requiring promql metric joins.


jacobstr:
**What would you like to be added**:

In large clusters, pod replicas often represent a "class" or "category" or worker. It's really useful to get metrics broken down by e.g. the pod's controller, or some common label that categorizes pods without the cardinality of tens of thousands of pod IDs.

The time series' `kube_pod_labels` covers some ground, but requires joins. Joins, for usability, often engender recording rules. Recording rules in turn, demand stateful prometheus servers which e.g. 4x the memory footprint in systems like prometheus running in agent mode, or grafana's alloy. Particularly, if one is forwarding the metrics to a 3rd party it's nice to minimize the footprint of in-house infra by keeping things stateless as long as possible - certainly at some point you hit a big beefy backend that you can query.

Why not skip the label join requirement entirely? And allow for configurable appending of specific kube pod labels to the underlying time series instead. Again, this works incredibly well with "stateless" prometheus remote-writers and system's like Grafana's adaptive metrics that can do some limited metric aggregation etc without a full recording rule engine.

**Why is this needed**:
* Improved UX. You can immediately get things like `kube_pod_container_status_restarts_total` categorized by a workload without having to jump into recording rules.
* Improved performance in 3rd party systems with adaptive metrics or similar solutions.

**Describe the solution you'd like**

* A flag that basically says - take this pod label and propagate it to all `kube_pod_*` time series' without requiring a label join.

**Additional context**

* Potentially, it would be desirable to do this on individual `kube_pod_` time series, that does stretch the command line arguments approach though and it might just make sense to leverage the existing allow list and simply have a secondary boolean to enable / disable the label (and I suppose annotation for symmetry...) propagation.
* Ideally this is a benign if you don't opt in, if you do I'd expect additional memory usage for each ts to propagate the additional bytes for the label. The whole theory here is that when you have 10k pods that belong to "temporal" or "flyte" the subsequent ease with which you can do aggregations using something like https://grafana.com/blog/2023/05/09/adaptive-metrics-grafana-cloud-announcement/ would make it worthwhile.
* I am indeed largely advocating for this from the POV of adaptive metrics, and if that technology was more robust maybe the problem could be pushed downstream into having Grafana allow for custom rule expression that allow for label joins... but that's kind of the heart of the ux issue here: having to do label joins affects _everything_ and is so fundamental that doing it closer-to-the source would obviate a lot of downstream work.

[...]

PR #2428, opened 2024-06-20
[WIP] fix: Add allowed labels to all metrics instead of just `_labels` metric

ronaknnathani:
**What this PR does / why we need it**:
KSM allows providing a list of labels that should be added to the prometheus metric for the resource. These labels currently are only added to the `kube_<resource>_labels` metric. However, there are several use cases where having these labels on all of the metrics for the resource would be helpful. This change adds the allowed labels to all metrics of the resource instead of just the `_labels` metric.

> [!Note]
> Currently, I have only made the change for `Pod` metrics. If this direction looks good, I can update the PR for rest of the resources.

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)*
Doesn't change the number of metrics.
Increases the labels on each prometheus metric based on allowed labels. 

**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2311

**Tests**
Tested locally against a `kind` cluster. Just showing a handful of metrics here.
```
kube_pod_container_info{namespace="kube-system",pod="kindnet-l4529",uid="5e107535-3135-48fb-8f42-043c50866177",container="kindnet-cni",image_spec="docker.io/kindest/kindnetd:v20230511-dc714da8",image="docker.io/kindest/kindnetd:v20230511-dc714da8",image_id="sha256:b18bf71b941bae2e12db1c07e567ad14e4febbc778310a0fc64487f1ac877d79",container_id="containerd://d37637e9cfc42cb4f9f606de57d5ced7fb242827e2b4d8860e6c3a28ca877136",label_app="kindnet"} 1
kube_pod_container_info{namespace="monitoring",pod="event-exporter-7ddc6ff9b-wd8f5",uid="a021852f-75f6-4e1c-975a-103396730b68",container="event-exporter",image_spec="kubernetes-event-exporter:local",image="docker.io/library/kubernetes-event-exporter:local",image_id="docker.io/library/import-2024-01-17@sha256:73a75a4829b55788efebd1ac7679256c36a09216c0384363716c3ae22e5fba5b",container_id="containerd://a8643a9cf05a6e7d708cd685822fc3799ba45655f58c5d30930cd21e545f8a22",label_app="event-exporter"} 1
...
[...]

PR #2632, opened 2025-03-20
draft: add `inject-per-series-metadata'

jacobstr:
See the discussion here https://github.com/kubernetes/kube-state-metrics/issues/2551 This frees users from having to join on the various label series, which is useful when infrastructure has agents that can't do recording rules.\

Note: this is a draft proposal because repeating this work to all other metric families will get quite tedious. I wanted to walk this implementation out a bit to propose some neatly encapsulated way to inject labels and annotations with somewhat of a wrapper func - it might not perform as well, but adding all of the slice-wrangling + kube label and annotation filtering to the business logic of each metric would be a bit messy.

**What this PR does / why we need it**: See https://github.com/kubernetes/kube-state-metrics/issues/2551 With systems like Grafana Cloud it's useful to filter / aggregate certain series by segmenting metrics according to a custom label (such as app, environment, etc). Having to first label_join the existing `kube_<resource>_label` gauges requires some upstream prometheus with a TSDB and precludes using things like prometheus in agent mode or alloy. 

By propagating the labels to individual time series we could do more effective filtering / aggregation at smart backends while having "dumb" scrapers and forwarders upstream.

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)* Increases, but the behavior is opt-in.


internal/store/pod.go
```diff
@@ -40,7 +40,12 @@ var (
 	podStatusReasons           = []string{"Evicted", "NodeAffinity", "NodeLost", "Shutdown", "UnexpectedAdmissionError"}
 )
 
-func podMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator {
+func podMetricFamilies(injectPerSeriesMetadata bool, allowAnnotationsList []string, allowLabelsList []string) []generator.FamilyGenerator {
+	mc := &MetricConfig{
```
jacobstr, line 44: Doing this instead of adding 3 arguments to each generator, I add one.   Perhaps, I suppose it makes it less "functional" as the behavior of the function is dictated by a complex configuration object.

I feel like it might be time to bite the bullet and just have a config object than growing the argument lists. This also makes it easier to add options in the future for future features / refactors.

internal/store/pod.go
```diff
@@ -1361,13 +1366,12 @@ func createPodStatusPhaseFamilyGenerator() generator.FamilyGenerator {
 
 			ms := make([]*metric.Metric, len(phases))
 
-			for i, p := range phases {
-				ms[i] = &metric.Metric{
-
+			for i, ph := range phases {
+				ms[i] = injectLabelsAndAnnos(&metric.Metric{
 					LabelKeys:   []string{"phase"},
-					LabelValues: []string{p.n},
```
jacobstr, line 1368: We were shadowing the outer `p (Pod)`.

internal/store/utils.go
```diff
@@ -38,6 +39,12 @@ var (
 	conditionStatuses  = []v1.ConditionStatus{v1.ConditionTrue, v1.ConditionFalse, v1.ConditionUnknown}
 )
 
+type MetricConfig struct {
```
jacobstr, line 42: Could be made private since it's built / used internally to this module.

internal/store/utils.go
```diff
@@ -175,6 +182,17 @@ func isPrefixedNativeResource(name v1.ResourceName) bool {
 	return strings.Contains(string(name), v1.ResourceDefaultNamespacePrefix)
 }
 
+// convenience wrapper to inject allow-listed labels and annotations to a metric if per-series injection is enabled.
+func injectLabelsAndAnnos(m *metric.Metric, metricConfig *MetricConfig, obj *metav1.ObjectMeta) *metric.Metric {
+	if !metricConfig.InjectPerSeriesMetadata {
```
jacobstr, line 187: I _think_ with the pass-by-reference + early guard clause, this should be "0-cost" for those leaving the `--inject-per-series-metadata` at it's default of `false`.

internal/store/pod.go
```diff
@@ -82,7 +87,7 @@ func podMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generat
 		createPodSpecVolumesPersistentVolumeClaimsInfoFamilyGenerator(),
 		createPodSpecVolumesPersistentVolumeClaimsReadonlyFamilyGenerator(),
 		createPodStartTimeFamilyGenerator(),
-		createPodStatusPhaseFamilyGenerator(),
+		createPodStatusPhaseFamilyGenerator(mc),
```
jacobstr, line 90: The full implementation would add this param to every metric family generator. This is also the reason I'm bundling up the configuration in a struct and passing it as a single unit.

mrueg:
Thanks for your contribution!
Unfortunately, we are not able to accept it, as it goes against the intended design of kube-state-metrics and as maintainers, we do not want to support this. 
https://github.com/kubernetes/kube-state-metrics/blob/main/docs/design/metrics-best-practices.md#avoid-pre-computation 

A solution like this has come up multiple times, for more information, please take a look at:
https://github.com/kubernetes/kube-state-metrics/pull/2428#issuecomment-2188527702
https://github.com/kubernetes/kube-state-metrics/issues/1758#issuecomment-1192710597
https://github.com/kubernetes/kube-state-metrics/pull/2129#issuecomment-1696089005

jacobstr:
@mrueg any further thoughts? I'd posit that the repeated feature requests here represent a valid need, though I realize the stance in `ksm` is not dissimilar from e.g. the kubelet's cadvisor metrics which also do not enrich "applicable" time series with configurable metadata labels where possible (e.g. all the container time series').

* Yes, what I'm asking for can be done with recording rules but my argument is based on considering the "externalized" costs of doing so. In lieu of this, I'm now running a stateful metrics agent for forwarding to enrich data when I could have a largely stateless system with half (if not more) the memory consumption, a deployment instead of a statefulset because I no longer need to persists kubernetes PVs...
* Yes, folks can and do regularly shoot themselves in the foot via misconfiguration. A conscious opt-in to specific labels ameliorates that.
* _kube-state-metrics should expose metrics on an individual object level_ was the relevant statement in [pre-computation](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/design/metrics-best-practices.md#avoid-pre-computation). You give an example of computing a count there that to me is quite different from simply hoisting the labels from the object. I'd argue materializing kubernetes object resource labels (which are attributes [kubernetes resource model](https://github.com/kubernetes/design-proposals-archive/blob/main/architecture/resource-management.md#api-topology-and-conventions)) alongside their corresponding time series' is not a form of pre-computation but is akin to surfacing the name of the resource in the TS labels.

PR #2644, opened 2025-04-01
fix: report correct reason in kube_pod_status_reason metric

### LINKED ISSUE #2612: kube_pod_status_reason is 0 for all reasons
**What happened**: 
The metric `kube_pod_status_reason` shows 0 for all reasons, even when reasons should have value of `1`. 

We use Karpenter in our clusters, and expect to be able to see when pods have a change in status based on actions Karpenter takes. In particular, we expect to see `Evicted`, `NodeLost`, and `Shutdown` reasons to show a value of `1` in clusters where consolidation is happening all the time (`consolidateAfter` value is `5m0s`). We can see in our Karpenter metrics that at any given time, some pod is being moved, and should show up with a `kube_pod_status_reason` of `Evicted` with a value of `1`.

This prometheus query: `sum(kube_pod_status_reason) by (reason)` shows `0` for every reason, and when charted, those value remain the same over any time interval.

The `kube_pod_status_phase` does not give use the information we need (specific reasons for status), and no other metric claims to provide this.

**Environment**:
Running KSM v2.13 managed via Helm chart
EKS v1.32.2
Karpenter v1.2.0

carlosmorenokm1:
This PR updates the logic for generating the kube_pod_status_reason metric. Instead of only checking p.Status.Reason, the new implementation also verifies the pod conditions and the termination reasons of container statuses. This change fixes an issue where the metric always returned 0, even when a pod had a valid status reason (such as "Evicted", "NodeLost", etc.), leading to inaccurate monitoring data. Accurately reporting these values is crucial for diagnosing pod behavior and overall cluster health.

It does not change the cardinality. The update only adjusts the value calculation for an existing metric family, so no new labels or metric series are introduced.

Fixes #2612

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -1541,15 +1541,12 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
 			ms := []*metric.Metric{}
 
 			for _, reason := range podStatusReasons {
-				metric := &metric.Metric{}
-				metric.LabelKeys = []string{"reason"}
-				metric.LabelValues = []string{reason}
-				if p.Status.Reason == reason {
-					metric.Value = boolFloat64(true)
-				} else {
-					metric.Value = boolFloat64(false)
+				m := &metric.Metric{
+					LabelKeys:   []string{"reason"},
+					LabelValues: []string{reason},
+					Value:       getPodStatusReasonValue(p, reason),
 				}
-				ms = append(ms, metric)
+				ms = append(ms, m)
 			}
 
 			return &metric.Family{
@@ -1557,6 +1554,23 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
 			}
 		}),
 	)
+}
+
+func getPodStatusReasonValue(p *v1.Pod, reason string) float64 {
+	if p.Status.Reason == reason {
+		return 1
+	}
+	for _, cond := range p.Status.Conditions {
+		if cond.Reason == reason {
+			return 1
+		}
+	}
+	for _, cs := range p.Status.ContainerStatuses {
+		if cs.State.Terminated != nil && cs.State.Terminated.Reason == reason {
+			return 1
+		}
+	}
+	return 0
 }
 
 func createPodStatusScheduledFamilyGenerator() generator.FamilyGenerator {
--- a/internal/store/pod_test.go
+++ b/internal/store/pod_test.go
@@ -2290,3 +2290,85 @@ func BenchmarkPodStore(b *testing.B) {
 		}
 	}
 }
+
+func TestGetPodStatusReasonValue(t *testing.T) {
+	reason := "TestReason"
+
+	tests := []struct {
+		name string
+		pod  *v1.Pod
+		want float64
+	}{
+		{
+			name: "matches Status.Reason",
+			pod: &v1.Pod{
+				Status: v1.PodStatus{
+					Reason: "TestReason",
+				},
+			},
+			want: 1,
+		},
+		{
+			name: "matches condition Reason",
+			pod: &v1.Pod{
+				Status: v1.PodStatus{
+					Conditions: []v1.PodCondition{
+						{
+							Reason: "TestReason",
+						},
+					},
+				},
+			},
+			want: 1,
+		},
+		{
+			name: "matches container terminated Reason",
+			pod: &v1.Pod{
+				Status: v1.PodStatus{
+					ContainerStatuses: []v1.ContainerStatus{
+						{
+							State: v1.ContainerState{
+								Terminated: &v1.ContainerStateTerminated{
+									Reason: "TestReason",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: 1,
+		},
+		{
+			name: "no match returns 0",
+			pod: &v1.Pod{
+				Status: v1.PodStatus{
+					Reason: "OtherReason",
+					Conditions: []v1.PodCondition{
+						{
+							Reason: "NotTestReason",
+						},
+					},
+					ContainerStatuses: []v1.ContainerStatus{
+						{
+							State: v1.ContainerState{
+								Terminated: &v1.ContainerStateTerminated{
+									Reason: "AnotherReason",
+								},
+							},
+						},
+					},
+				},
+			},
+			want: 0,
+		},
+	}
+
+	for _, tt := range tests {
+		t.Run(tt.name, func(t *testing.T) {
+			got := getPodStatusReasonValue(tt.pod, reason)
+			if got != tt.want {
+				t.Errorf("getPodStatusReasonValue() = %v, want %v", got, tt.want)
+			}
+		})
+	}
+}


internal/store/pod.go
```diff
@@ -1559,6 +1556,23 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
...
 
+func getPodStatusReasonValue(p *v1.Pod, reason string) float64 {
+	if p.Status.Reason == reason {
+		return 1
+	}
+	for _, cond := range p.Status.Conditions {
+		if cond.Reason == reason {
+			return 1
+		}
+	}
```
CatherineF-dev, line 1567: Should we only care about the last condition? If so, do we need to remove this part?

CatherineF-dev:
Could we update the title to be "fix: report correct reason in kube_pod_status_reason metric"

internal/store/pod.go
CatherineF-dev: Will it be a stale condition?

internal/store/pod.go
```diff
@@ -1559,6 +1556,23 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+func getPodStatusReasonValue(p *v1.Pod, reason string) float64 {
```
mrueg, line 1559: Can you add a test for this function?

internal/store/pod.go
CarlosMorenoKM: it will not be a stale condition. Kubernetes regularly updates Pod conditions, so if a condition with the corresponding reason is found, it is assumed to be current. If a stale condition were detected, that would indicate an issue in Kubernetes, not in this logic.

internal/store/pod.go
carlosmorenokm1: Done

internal/store/pod.go
CatherineF-dev: Will a pod have multiple different reasons?

camorenogi: Yes, a Pod can have different “Reasons” throughout its lifecycle. Each event or change in the Pod’s state (for example, container creation, image pulling, runtime errors, restarts, etc.) can trigger a different reason. In Kubernetes, these “Reasons” are recorded at different points in the Pod’s lifecycle, so it is entirely possible for a single Pod to go through multiple different “Reasons” as it transitions between states.

CatherineF-dev: Yes, I was thinking the case where the pod status is failed to image, then runtime errors, then restart.

Will the above metric have all of these three status?

carlosmorenokm1: Yes, if your Pod transitions through those states (e.g., failed to pull image, runtime errors, then restarts), the metric can capture each corresponding reason at the time it occurs. However, you won’t necessarily see all reasons simultaneously; rather, you’ll see them reflected as changes in the metric over the Pod’s lifecycle.

rexagod:
Hello, thank you for the PR. Regarding the second and third conditions being added here, do we see those fields exhibiting the values in `podStatusReasons` in real life? I did a quick k/k search and couldn't find an [overlap](https://github.com/kubernetes/kubernetes/blob/70540c9f43e2fb7604924a120799206c27cbbd28/staging/src/k8s.io/api/core/v1/types.go#L3375-L3408) for the field in the second condition, but I could be missing something. Happy to merge if that's not the case.

> This change fixes an issue where the metric always returned 0, even when a pod had a valid status reason (such as "Evicted", "NodeLost", etc.), leading to inaccurate monitoring data.

Not sure I get this, if the `.Status.Reason` is within the `podStatusReasons` set (valid), wouldn't this work as expected, or does this point towards the lack of fields being covered in the existing version, which are being added in this patch?

```go
if p.Status.Reason == reason {
	metric.Value = boolFloat64(true)
} else {
	metric.Value = boolFloat64(false)
}
```

rexagod:
To continue my aforementioned point, I believe we need to [expand](https://github.com/kubernetes/kube-state-metrics/issues/2612#issuecomment-2669575687) on the set of values `podStatusReasons` can take.

carlosmorenokm1:
Hi @rexagod,

In real-world environments we’ve observed that not all valid reasons show up in p.Status.Reason. For example:

When a container terminates due to OOMKilled or Completed, those values only live in containerStatus.State.Terminated.Reason, and p.Status.Reason ends up empty.

Some reasons like BackOff or image-pull errors are only recorded in pod.Status.Conditions[i].Reason.

If we only checked p.Status.Reason, the metric would still return 0 in those cases. With this patch, we also inspect pod conditions and container termination states, ensuring that any valid reason from podStatusReasons is correctly reflected in kube_pod_status_reason.

andrii-cherniuk:
Hi, there
Which version of the Helm chart includes the fix?

konstantindobroliubov:
Running kube-state-metrics 2.16.0 which should include this change.
Unfortunately, `kube_pod_status_reason{reason="Evicted"}` still returns 0 regardless of the real state of things :(
Seems like the same is true for any reason, not just "Evicted".
@carlosmorenokm1 FYI.

rexagod:
@konstantindobroliubov The change here expands the search scope for the list of reasons that we look for. Could you provide a reproducible case for which this is supposed to work but wouldn't?

konstantindobroliubov:
> Could you provide a reproducible case for which this is supposed to work but wouldn't?

Well, I didn't do anything special, just triggered Pod eviction in EKS cluster.

Manifest:

```
$ cat manifest.yaml 
apiVersion: v1
kind: Pod
metadata:
  name: test
spec:
  containers:
    - name: test
      image: busybox:1.37.0
      command: ["/bin/sh", "-c", "sleep 100000"]
```

Apply:

```
kubectl apply -f manifest.yaml -n test
```

Describe:

```
$ kubectl describe pod test -n test
Name:             test
Namespace:        test
Priority:         0
Service Account:  default
Node:             <MASKED>.eu-central-1.compute.internal/<MASKED>
Start Time:       <MASKED>
Labels:           <none>
Annotations:      <none>
Status:           Running
IP:               <MASKED>
IPs:
  IP:  <MASKED>
Containers:
  test:
    Container ID:  containerd://4fba8ea8b18c853a0ef5059bb8c773682ce67ec3299b7337bd467fa3e20e8470
    Image:         busybox:1.37.0
    Image ID:      busybox@sha256:37f7b378a29ceb4c551b1b5582e27747b855bbfaa73fa11914fe0df028dc581f
    Port:          <none>
    Host Port:     <none>
    Command:
      /bin/sh
      -c
      sleep 100000
    State:          Running
      Started:      <MASKED>
    Ready:          True
    Restart Count:  0
    Environment:    <none>
    Mounts:
      /var/run/secrets/kubernetes.io/serviceaccount from kube-api-access-cnssn (ro)
Conditions:
  Type                        Status
  PodReadyToStartContainers   True
  Initialized                 True
  Ready                       True
  ContainersReady             True
  PodScheduled                True
Volumes:
  kube-api-access-cnssn:
    Type:                    Projected (a volume that contains injected data from multiple sources)
    TokenExpirationSeconds:  3607
    ConfigMapName:           kube-root-ca.crt
    ConfigMapOptional:       <nil>
    DownwardAPI:             true
QoS Class:                   BestEffort
Node-Selectors:              <none>
Tolerations:                 node.kubernetes.io/not-ready:NoExecute op=Exists for 300s
                             node.kubernetes.io/unreachable:NoExecute op=Exists for 300s
Events:
  Type    Reason     Age   From               Message
  ----    ------     ----  ----               -------
  Normal  Scheduled  4m7s  default-scheduler  Successfully assigned test/test to <MASKED>.eu-central-1.compute.internal
  Normal  Pulled     4m4s  kubelet            Container image "busybox:1.37.0" already present on machine
  Normal  Created    4m4s  kubelet            Created container: test
  Normal  Started    4m4s  kubelet            Started container test
```

Save the Node value. Delete it (mark as deleted for EKS) and then drain it to trigger eviction:

```
$ kubectl delete node <MASKED>.eu-central-1.compute.internal
$ kubectl drain <MASKED>.eu-central-1.compute.internal --force --delete-emptydir-data --ignore-daemonsets | grep evicting
evicting pod test/test
```

Outcome: Pod is evicted.

```
$ kubectl get events -n test | grep Evicted
65s         Normal    Evicted     pod/test   Evicted pod: Forceful Termination
```

Related eviction Event:

```
$ kubectl describe events test.1856709e8c58938c -n test
Name:             test.1856709e8c58938c
Namespace:        test
Labels:           <none>
Annotations:      <none>
API Version:      v1
Count:            1
Event Time:       <nil>
First Timestamp:  <MASKED>
Involved Object:
  API Version:   v1
  Kind:          Pod
  Name:          test
  Namespace:     test
Kind:            Event
Last Timestamp:  <MASKED>
Message:         Evicted pod: Forceful Termination
Metadata:
  Creation Timestamp:  <MASKED>
  Resource Version:    163039919
  UID:                 f7e28915-08fe-4095-a2b0-fab190ceed0f
Reason:                Evicted
Reporting Component:   karpenter
Reporting Instance:
Source:
  Component:  karpenter
Type:         Normal
Events:       <none>
```

Prometheus query:

```
kube_pod_status_reason{namespace="test", pod="test", reason="Evicted"}
```

Timeseries example:

```
kube_pod_status_reason{container="kube-state-metrics",endpoint="http",instance="[<MASKED>]:8080",job="kube-state-metrics",namespace="test",pod="test",reason="Evicted",service="prometheus-kube-state-metrics",uid="7765db27-e18d-4d02-8b45-a0a2b2309281"} 0
```

All values are 0, `kube_pod_status_reason{namespace="test", pod="test", reason="Evicted"} > 0` returns empty result.
Is there any additional data I can help you with?

PR #2659, opened 2025-04-23
feat: Replace endpoints with endpointslices as default resource

mrueg:
BREAKING CHANGE: endpoints are not a default resource anymore and metrics need to be replaced with endpointslices' metrics.

https://kubernetes.io/blog/2025/04/23/kubernetes-v1-33-release/#deprecation-of-the-stable-endpoints-api.

None
Fixes #


richabanker:
/triage accepted
cc @rexagod

rexagod:
I see, this moves `endpoints` out of the default set of resources for which we generate metrics for, and users will need to use the `--resources` flag to re-enable it. While that makes sense, I'd suggest a [couple of releases](https://kubernetes.io/docs/reference/using-api/deprecation-policy/#:~:text=The%20rules%20for%20deprecating%20and,8%20months%20(whichever%20is%20longer)) indicating this (in the change-log, at runtime if `endpoints` are present in `--resources`, etc.) to warn users.

It'd additionally make sense to put this behind a flag, and enable it by default once the deprecation deadline hits. This way, users can start opt-ing into the feature before it's dropped.

CatherineF-dev:
/lgtm 

/hold 

Feel free to unhold

mrueg:
/hold cancel
since v2.17.0 is released which includes a Note that its the last release with endpoints resource enabled by default.

@rexagod @CatherineF-dev @dgrisonnet can I get a review here again? Thanks!

mrueg:
@rexagod @CatherineF-dev can I get a review for this?

mrueg:
@rexagod @CatherineF-dev @dgrisonnet can I get another review here? I would like to get this change in for the next release.

andy-townsend:
Can this get reviewed please, it spams a load of warnings in our clusters now that we'd want to get cleaned up before upgrading EKS.

Makefile
```diff
@@ -179,4 +179,4 @@ install-promtool:
 	@wget -qO- "https://github.com/prometheus/prometheus/releases/download/v${PROMETHEUS_VERSION}/prometheus-${PROMETHEUS_VERSION}.${OS}-${ARCH}.tar.gz" |\
 	tar xvz --strip-components=1 prometheus-${PROMETHEUS_VERSION}.${OS}-${ARCH}/promtool
 
-.PHONY: all build build-local all-push all-container container container-* do-push-* sub-push-* push push-multi-arch test-unit test-rules test-benchmark-compare clean e2e validate-modules shellcheck licensecheck lint lint-fix generate generate-template validate-template embedmd
+.PHONY: all build build-local all-push all-container container container-* do-push-* sub-push-* push push-multi-arch test-unit test-rules test-benchmark-compare clean e2e validate-modules shellcheck licensecheck lint lint-fix generate generate-template validate-template
```
rexagod, line 182: Unrelated but I believe `embedmd` should happen with each `generate`? We could probably drop this change from this PR and open a new one to fix this target.

toan-hf: @mrueg I would like to do it for you if you are busy on something :)

mrueg: It's happening with each generate, running         ${EMBEDMD_CLI} -w `find . -path ./vendor -prune -o -name "*.md" -print`

CatherineF-dev:
/lgtm
/hold
Users can still use endpoint if they add it back.
Feel free to unhold

mrueg:
/hold cancel

thanks for the review @CatherineF-dev

PR #2670, opened 2025-05-12
enhancement: add `pathType` to `ingress_path`

rexagod:
This avoids duplicate metrics (and thus, `PrometheusDuplicateTimestamps`) for schemas such as:
```
apiVersion: networking.k8s.io/v1
kind: Ingress
metadata:
  name: minimal-ingress
  annotations:
    nginx.ingress.kubernetes.io/rewrite-target: /
spec:
  ingressClassName: nginx-example
  rules:
  - http:
      paths:
      - backend:
          service:
            name: cart
            port:
              number: 3000
        path: /nvc/cart
        pathType: Prefix
      - backend:
          service:
            name: cart
            port:
              number: 3000
        path: /nvc/cart
        pathType: Exact
```


mrueg:
/hold cancel

due to lack of feedback from anyone else in the past weeks.

PR #2672, opened 2025-05-13
fix: Clean up CR reflectors after parent CRD deletion

rexagod:
With this patch, CRD deletions don't make CRS throw LW errors from the respective CRD's reflector anymore.

I noticed this way back, but never got around to actually patching it. That is, until a certain request downstream made me ponder it again, for which we currently have a workaround in the parent operator, which we would be better off without. Also, this workaround would more-or-less be what anyone would have to do to prevent CRS from false alerting where there's a cache race.

This reduces our urgent need for [1] to go in, as it addresses the "failed to list/watch resource" errors from client-go's `tools/cache/reflector.go` owing to which, in addition to error logs, we also saw `kube_state_metrics_{list,watch}{result="error"}` increment, leading to false alerts in the metrics backend.

Note that we still need control over client-go's error machinery, so I'll keep that issue open. See [2] for more details about this change.

[1]: https://github.com/kubernetes/kubernetes/issues/127329
[2]: https://github.com/openshift/cluster-monitoring-operator/pull/2595

***

To reproduce this, run KSM with the CRS configuration pointing to a CRD, deploy the CRD and CR, then delete them. Usually (not 100% of the time, as this is a race, but pretty frequently) this will log a "failed to watch" error, and `watch_total{result="error"}` will be incremented as well.

***

Update: The patch now consists of a single commit that cleans up reflectors for CRs for which the CRDs have been deleted, thus avoiding a memory leak. This additionally fixes the aforementioned problem statement as well.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -64,8 +64,8 @@ func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config) 
 					},
 					Plural: p,
 				}
-				r.AppendToMap(gotGVKP)
 				r.SafeWrite(func() {
+					r.AppendToMap(gotGVKP)
 					r.WasUpdated = true
 				})
 			}
@@ -89,8 +89,8 @@ func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config) 
 					},
 					Plural: p,
 				}
-				r.RemoveFromMap(gotGVKP)
 				r.SafeWrite(func() {
+					r.RemoveFromMap(gotGVKP)
 					r.WasUpdated = true
 				})
 			}
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -38,6 +38,7 @@ import (
 	policyv1 "k8s.io/api/policy/v1"
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
+	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -84,6 +85,8 @@ type Builder struct {
 	shard               int32
 	useAPIServerCache   bool
 	objectLimit         int64
+
+	GVKToReflectorStopChanMap *map[string]chan struct{}
 }
 
 // NewBuilder returns a new builder.
@@ -617,7 +620,11 @@ func (b *Builder) startReflector(
 ) {
 	instrumentedListWatch := watch.NewInstrumentedListerWatcher(listWatcher, b.listWatchMetrics, reflect.TypeOf(expectedType).String(), useAPIServerCache, objectLimit)
 	reflector := cache.NewReflectorWithOptions(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, cache.ReflectorOptions{ResyncPeriod: 0})
-	go reflector.Run(b.ctx.Done())
+	if cr, ok := expectedType.(*unstructured.Unstructured); ok {
+		go reflector.Run((*b.GVKToReflectorStopChanMap)[cr.GroupVersionKind().String()])
+	} else {
+		go reflector.Run(b.ctx.Done())
+	}
 }
 
 // cacheStoresToMetricStores converts []cache.Store into []*metricsstore.MetricsStore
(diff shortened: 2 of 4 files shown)


internal/discovery/discovery.go
```diff
@@ -111,7 +113,15 @@ func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config)
 			return
 		}
 	}()
+
+	// Start the informer.
 	go informer.Run(stopper)
+
+	// Wait for the cache to sync.
+	if !cache.WaitForCacheSync(ctx.Done(), informer.HasSynced) {
```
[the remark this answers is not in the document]
rexagod: The operation need not be transient.

Essentially, the sync isn't respected in any of the three events, but made visible in the delete clause since the reflector, which is still running based off the store implementation for that CRD, cannot list the CRD anymore as it is not made aware just as that's deleted, which is frequently transient, causing it to log this error, as well as increment the `list_total{result="error"}` signal.

I'll add a test to demonstrate the before and after cases deterministically to contribute to the overall confidence of this patch.

rexagod:
Seconding @dgrisonnet, I realized syncing caches at the first run isn't going to help us do that throughout the lifecycle.

Resyncs weren't an option either since the event could always occur before that timeout, causing the same error.

I dropped the event handlers and instead relied on indexers (latest patch, albeit a bit rough; pushed for visibility) but upon digging into cgo, it seems the issue doesn't stem from the `delete` handler's registration, instead the way the factory of informers handle deletions internally (also unrelated to our `delete` store implemention), in that the reflectors are not closed once their corresponding objects are dropped. The issue is [still persistent](https://github.com/kubernetes/kube-state-metrics/actions/runs/15551294782/job/43782287572?pr=2672#step:4:476).

https://github.com/kubernetes/kubernetes/issues/79610 is still open and tracking this particular flaw in cgo.

For now, I'll work on the alternative client-based (non-dynamic) informer solution.

pkg/app/server.go
```diff
@@ -295,6 +295,8 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 			CRDsDeleteEventsCounter: crdsDeleteEventsCounter,
 			CRDsCacheCountGauge:     crdsCacheCountGauge,
 		}
+		// storeBuilder starts reflectors for the discovered GVKs, and as such, should close them too.
+		storeBuilder.GVKToReflectorStopChanMap = &discovererInstance.GVKToReflectorStopChanMap
```
dgrisonnet, line 300: because `GVKToReflectorStopChanMap` is not initialized until `AppendToMap` is called, this will be always nil no?

internal/store/builder.go
```diff
@@ -608,7 +611,11 @@ func (b *Builder) startReflector(
 ) {
 	instrumentedListWatch := watch.NewInstrumentedListerWatcher(listWatcher, b.listWatchMetrics, reflect.TypeOf(expectedType).String(), useAPIServerCache)
 	reflector := cache.NewReflectorWithOptions(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, cache.ReflectorOptions{ResyncPeriod: 0})
-	go reflector.Run(b.ctx.Done())
+	if cr, ok := expectedType.(*unstructured.Unstructured); ok {
+		go reflector.Run((*b.GVKToReflectorStopChanMap)[cr.GroupVersionKind().String()])
+	} else {
+		go reflector.Run(b.ctx.Done())
+	}
```
dgrisonnet, line 627: do we need to make this more complex? `b.ctx.Done()` and `(*b.GVKToReflectorStopChanMap)[cr.GroupVersionKind().String()]` will point to the same channel anyway, so we could keep the same code as before

internal/store/builder.go
```diff
@@ -83,6 +84,8 @@ type Builder struct {
 	totalShards         int
 	shard               int32
 	useAPIServerCache   bool
+
+	GVKToReflectorStopChanMap *map[string]chan struct{}
```
dgrisonnet, line 89: nit
```suggestion
	GVKToReflectorStopChanMap *map[string]<-chan struct{}
```

internal/store/builder.go
rexagod: > b.ctx.Done() and (*b.GVKToReflectorStopChanMap)[cr.GroupVersionKind().String()] will point to the same channel anyway

The former should signal when the GVK's CRD is deleted, while the latter will signal on context cancellations (process interrupts). Since the native resource reflectors do not have a corresponding entry in the `b.GVKToReflectorStopChanMap`, I've kept their termination channels same as before.

pkg/app/server.go
rexagod: I've initialized this as a pointer to the map (pointer to a pointer, essentially), so at any given point of time the state of `GVKToReflectorStopChanMap` is the same in `storeBuilder` and `discovererInstance`.

dgrisonnet: Oh indeed my bad, I looked at it as if the map was a value

PR #2675, opened 2025-05-28
feat: Add HorizontalPodAutoscaler created and deletionTimestamp info into metrics

IgorIgnatevBolt:

For some reason, there is no `kube_horizontalpodautoscaler_created` metric for HPA with creation timestamp information, PR is about to add it as a similarity to other objects.

Also, add `kube_horizontalpodautoscaler_deletion_timestamp` metrics to catch the deletion timestamp field.

None

Fixes #


richabanker:
/triage accepted
@rexagod can you PTAL? Thanks

IgorIgnatevBolt:
Just a small update, also added `kube_horizontalpodautoscaler_deletion_timestamp` metrics because I need this and fixed the formatting for the table docs

brainfair:
All commits were squashed into one.

IgorIgnatevBolt:
All commits were squashed into one.

CatherineF-dev:
Hi, could you share the use case of kube_horizontalpodautoscaler_created?

If only want to the created timestamp, could we use kube_horizontalpodautoscaler_info?

IgorIgnatevBolt:
@CatherineF-dev I'd like to detect resources (for example) older than 7 days with special conditions (eq, last apply label exists or any other), and the created timestamp is very useful in expression for this because the date is defined permanently, initially I tried to use kube_horizontalpodautoscaler_info by measuring the presence of this metric in the required period, unfortunately it cannot be used because it very depends on tsdb where metrics collected and if we don't have this metric for all perion that we want to measure (like lower retention perion) then we cannot trigger alert. All other resources collected by kube-state-metrics have a created timestamp, but for some reason not hpa. I would like the ability to work equally with all resources.

IgorIgnatevBolt:
Hi @CatherineF-dev, do you need any more information about PR or anything else that can help you move forward?

mrueg:
/hold 
for further comments.
/lgtm 

Thanks for your contribution!

PR #1079, opened 2020-02-24
Rename kube_pod_deleted to kube_pod_deletion_timestamp

olivierlemasle:
**What this PR does / why we need it**:

Rename `kube_pod_deleted` to `kube_pod_deletion_timestamp`, to fix an inconsistency.

**Which issue(s) this PR fixes** :
Fixes #1068

PR #2678, opened 2025-06-02
feat: introduce deletion timestamp metric for daemonset, statefulset, deployment, service and pdb

IgorIgnatevBolt:

Some resources can be blocked by deletion from `finalizers`. To catch this and expose it to metrics, we can use the deletion timestamp metadata field.
Introduce a deletion_timestamp metric for the next resources:

- deployment `kube_deployment_deletion_timestamp`
- statefulset `kube_statefulset_deletion_timestamp`
- daemonset `kube_daemonset_deletion_timestamp`
- service `kube_service_deletion_timestamp`
- poddisruptionbudget `kube_poddisruptionbudget_deletion_timestamp`

Also formatting tables in docs

Fixes #


IgorIgnatevBolt:
All commits were squashed into one.

CatherineF-dev:
Hi, could you share more insights on use cases after these metrics are added?

Is it used for monitoring Kubernetes resources that are stuck in a terminating state?

IgorIgnatevBolt:
@CatherineF-dev Hi, yes, if the resource deletion process is stuck for some reason or blocked by the finalizer, deletiontimestamp metric can help to detect such a case and raise an alert for investigation.

CatherineF-dev:
@IgorIgnatevBolt How will we know which resource should be deleted?

IgorIgnatevBolt:
> @IgorIgnatevBolt How will we know which resource should be deleted?

Maybe I misunderstood the question, but this PR is exactly about detection for such resources that were nominated by the controller manager for deletion but not deleted for some reason, [eq blocked by finalizers](https://kubernetes.io/docs/concepts/overview/working-with-objects/finalizers/#how-finalizers-work)

```
The controller managing that finalizer notices the update to the object setting the metadata.deletionTimestamp, indicating deletion of the object has been requested.
```

IgorIgnatevBolt:
Hi @CatherineF-dev, do you need any more information about PR or anything else that can help you move forward?

docs/metrics/workload/deployment-metrics.md
```diff
@@ -1,19 +1,20 @@
...
+| kube_deployment_created                                     | Gauge       |                                                                                                                                         | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt;                                                                                          | STABLE       |
+| kube_deployment_deletion_timestamp                          | Gauge       | Unix deletion timestamp                                                                                                                 | `deployment`=&lt;deployment-name&gt; <br> `namespace`=&lt;deployment-namespace&gt;                                                                                          | EXPIREMENTAL |
```
CatherineF-dev, line 20: Should we use kube_deployment_deleted to align with kube_deployment_created?

IgorIgnatevBolt: I'd like to keep the pattern the same as for other resources like `kube_node_deletion_timestamp` or `kube_pod_deletion_timestamp`

CatherineF-dev: @richabanker how do you think?

richabanker: ah so its `kube_deployment_created` thats not following the `_timestamp` pattern :/

Looks like we have consciously made the switch to use _timestamp in the [past](https://github.com/kubernetes/kube-state-metrics/pull/1079) so maybe `kube_deployment_deletion_timestamp` is the way to go.

Additionally does it make sense to rename the  `kube_deployment_created` to also follow the same? How disruptive is that change?

IgorIgnatevBolt: I guess renaming any existing metrics would be a breaking change and require a longer release process. I'd like to keep the scope of the current PR only in the current state, with new metrics only.

IgorIgnatevBolt:
rebased, conflicts solved

IgorIgnatevBolt:
fixed typo in the depl test

mrueg:
/hold

for @CatherineF-dev to further comment. 
Renamed the title to include all additional resources in there.

/lgtm

PR #2686, opened 2025-06-18
WIP: feat: Add AuthFilter

### LINKED ISSUE #2684: Authn/Authz for metrics endpoints via controller-runtime's WithAuthenticationAndAuthorization filter
Have a kubernetes-backed authn/authz for metrics endpoints.

**Why is this needed**:
Metrics from ksm may contain sensitive information.  Currently they are openly accessible and some deployments use https://github.com/brancz/kube-rbac-proxy to protect them through Kubernetes'  Access Review.

We might be able to introduce an optional [filter](https://pkg.go.dev/sigs.k8s.io/controller-runtime@v0.18.4/pkg/metrics/filters#WithAuthenticationAndAuthorization) to protect the endpoints

**Additional context**

```
WithAuthenticationAndAuthorization provides a metrics.Filter for authentication and authorization. Metrics will be authenticated (via TokenReviews) and authorized (via SubjectAccessReviews) with the kube-apiserver
```

mrueg:
This adds an Authentication/Authorization filter through Kubernetes' TokenReview / SubjectAccessReview resources.

The client config for kube-state-metrics needs a clusterrole for

* apiGroups: authentication.k8s.io, resources: tokenreviews, verbs: create
* apiGroups: authorization.k8s.io, resources: subjectaccessreviews, verbs: create

The Prometheus client needs a clusterrole for

* nonResourceURLs: "/metrics", verbs: get

Limits access to sensitive metrics guarded by the Kubernetes API

None
Fixes #2684


docs/README.md
```diff
@@ -97,3 +97,48 @@ See [Custom Resource State Metrics](metrics/extend/customresourcestate-metrics.m
...
 Additionally, options for `kube-state-metrics` can be passed when executing as a CLI, or in a kubernetes / openshift environment. More information can be found here: [CLI Arguments](developer/cli-arguments.md)
+
+## Protecting /metrics endpoints
+
+Kube-State-Metrics' metrics can contain sensitive information about the state of the cluster, which you as an operator might want to additionally protect from unauthorized access.
+To do this, you need to enable the `--auth-filter` flag on kube-state-metrics. With this, kube-state-metrics will only accept authenticated and authorized requests to the /metrics endpoints.
+Kube-state-metrics uses Kubernetes' RBAC mechanisms for this.
+The clients scraping the endpoint, need to use a token which can be provided by a ServiceAccount that has the following set up:
+
+A ClusterRole providing access like this:
```
CatherineF-dev, line 109: Do we need to mention the con is that it increases apiserver load on validating token every scrape interval?  Or this is ignorable?

If it's not ignorable, will we add cache?

docs/README.md
```diff
@@ -97,3 +97,48 @@ See [Custom Resource State Metrics](metrics/extend/customresourcestate-metrics.m
...
+
+```
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: metrics-reader
```
CatherineF-dev, line 115: nit: kube-state-metrics-reader

CatherineF-dev:
/lgtm
/hold

Just one small comment

docs/README.md
mrueg: Good thought, I can add it. Maybe we want some ratelimiting on the metrics endpoints for that in general. 

Caching authentication is a more difficult problem. I'd avoid it if possible.

mrueg: Added it to the paragraph.

tests/e2e/auth-filter_test.go
```diff
@@ -0,0 +1,140 @@
...
+	klog.InfoS("waiting for first metrics to become available")
+	testMetric := `kube_configmap_info`
+
+	token, err := exec.Command("kubectl", "create", "token", "kube-state-metrics", "-n", "kube-system").Output()
+	if err != nil {
+		t.Fatalf("failed to create token %v", err)
+	}
+	klog.InfoS("token created")
+
+	err = wait.PollUntilContextTimeout(context.TODO(), 1*time.Second, 20*time.Second, true, func(_ context.Context) (bool, error) {
```
rexagod, line 94: Can we have a block above this to ensure that fetching the metric fails without the token?

tests/e2e/testdata/metrics-reader_role.yaml
rexagod, line 1: Is a cluster role necessary (since the role is only being attached to the KSM SA)?

tests/e2e/auth-filter_test.go
```diff
@@ -0,0 +1,140 @@
...
+		out, err := exec.Command("curl", "-H", "Authorization: Bearer "+string(token), "localhost:8080/metrics").Output() // nolint:gosec
```
rexagod, line 95: I'd suggest making sure this works for both ports since we want to ensure that behavior stays the same in the future.

tests/e2e/auth-filter_test.go
mrueg: This is what I'm testing in https://github.com/kubernetes/kube-state-metrics/pull/2686/files/3c8a087421b14c2a8f0d3d74b79f694d03769049#diff-99b6c1efb0421b6b40b2602dabdec035f253a896578a180d135d57a81548ca6cR108

mrueg: Do you feel like there's a need to test it twice?

tests/e2e/testdata/metrics-reader_role.yaml
mrueg: If I understand it correctly, nonResourceURLs are a clusterrole-only feature and won't work with roles.

tests/e2e/auth-filter_test.go
mrueg: That's a good idea. I'll add that.

tests/e2e/auth-filter_test.go
rexagod: The aforementioned link seems to be a 404 now.

tests/e2e/testdata/metrics-reader_role.yaml
rexagod: [Ah, right](https://github.com/kubernetes/kubernetes/blob/5e33c00e37832214a0c240ca4bcb18f7a6f1e5e4/pkg/apis/rbac/types.go#L60), which makes sense since `Service`s are cluster-scoped too.

tests/e2e/auth-filter_test.go
mrueg: https://github.com/kubernetes/kube-state-metrics/pull/2686/files#diff-99b6c1efb0421b6b40b2602dabdec035f253a896578a180d135d57a81548ca6cR108

tests/e2e/auth-filter_test.go
```diff
@@ -0,0 +1,140 @@
...
+		if err != nil {
+			return false, err
+		}
+		if string(out) == "" {
+			return false, nil
+		}
+		// Note: we use count to make sure that only one metrics handler is running
```
rexagod:
```suggestion
```

tests/e2e/auth-filter_test.go
```diff
@@ -0,0 +1,140 @@
...
+		if strings.Contains(string(out), testMetric) {
+			// klog.InfoS("metrics available", "metric", string(out))
```
rexagod:
```suggestion
```

tests/e2e/auth-filter_test.go
```diff
@@ -0,0 +1,140 @@
...
+			klog.InfoS("metrics fetched with bearer token")
+			// Signal the process to exit, since we know the metrics are being generated as expected.
+
+			out, err := exec.Command("curl", "localhost:8080/metrics").Output()
```
rexagod, line 106: I'm not entirely sure if this belongs here. The `if` block checks for the presence of the metric, and that should be it? We probably shouldn't tie the non-token metrics-retrieval check to `if`'s evaluation, and move it outside of the polling block entirely IMO.

Also, shouldn't the error condition below check for the contrary, i.e., this should always throw if `auth-filter` is enabled?

rexagod: The `err` will probably not throw at all (on a non-200, seeing the tests are green), but a `Contains` check should fix things, just so its robust.

tests/e2e/auth-filter_test.go
mrueg: Added it.

tests/e2e/auth-filter_test.go
mrueg: hmmm I was making it following the happy path, first check if the bearer token auth worked and then we check that it doesn't work without it.

mrueg: I see what you mean,  the err check is just a "good" habbit I guess here.

rexagod:
/lgtm

Thank you for looking into this!

PR #2692, opened 2025-06-23
chore: Prep `v2.16` release

rexagod:
Pending PRs to be included in this release:

- [ ] https://github.com/kubernetes/kube-state-metrics/pull/2686
***

Maintainers should have edit access, so feel free to edit this section to add any PRs you'd want to see to go in.

PR #2693, opened 2025-06-23
fix: Fix `examples` target

rexagod:
Fixes a typo, needed for https://github.com/kubernetes/kube-state-metrics/pull/2692.


mrueg:
/lgtm

Thanks for catching this.

PR #2702, opened 2025-07-08
feat: Add actual pod resources metrics

### LINKED ISSUE #2665: Expose actual pod CPU/memory request from status.containerStatuses.resources (Kubernetes 1.33)
Expose a new set of metrics in kube-state-metrics that report the actual CPU and memory requests and limits from status.conatinerStatuses[i].resources (introduced in Kubernetes 1.33).

**Why is this needed**:
Starting from Kubernetes 1.33, conatiners always report their actual active resource requests and limits under status.conatinerStatuses[i].resources.
Currently, kube-state-metrics only exposes values from spec.resources, which reflect the desired future state rather than the actual current.

Add new metrics like:
* kube_pod_container_actual_resource_requests,
* kube_pod_container_actual_resource_limits

These metrics should expose values from status.conatinerStatuses[i].resources and reflect the current actual CPU and memory requests and limits of each container.

**Additional context**

[KEP-1287: In-place Pod Vertical Scaling](https://github.com/kubernetes/enhancements/issues/1287)

wardady:
Starting from Kubernetes 1.33, `kube_pod_container_resource_requests` and `kube_pod_container_actual_resource_limits` reflect the desired state of the container resource requests/limits.
This PR adds 2 new metric families: `kube_pod_container_actual_resource_requests` and `kube_pod_container_actual_resource_limits` that expose values from `status.containerStatuses[i].resources` and reflect the current actual requests and limits of each container.

The PR adds 2 new metric families with the same cardinality as very similar already existing metric families.

Fixes #2665


wardady:
@dgrisonnet @logicalhan I adjusted the failing test and now it passes locally. Could you please approve the CI run?

wardady:
@dgrisonnet @logicalhan would appreciate any review or feedback on this PR.

CatherineF-dev:
When will kube_pod_container_resource_requests and kube_pod_container_actual_resource_requests have different values?

mrueg:
> When will kube_pod_container_resource_requests and kube_pod_container_actual_resource_requests have different values?

I assume this might be the case with https://kubernetes.io/blog/2025/05/16/kubernetes-v1-33-in-place-pod-resize-beta/ ?

wardady:
@mrueg, right, that is the case with in-place pod resize. @CatherineF-dev For example, when you try to apply an in-place pod resize patch, which changed requests of a container to the value above node capacity or available resources:
```sh
kubectl patch pod  pod-name --subresource resize --patch '{"spec":{"containers":[{"name":"container-name", "resources":{"requests":{"memory":<above-node-capacity>}}}]}}'
```
In this case, the value of `kube_pod_container_resource_requests` will be different from `kube_pod_container_actual_resource_requests` (`kube_pod_container_resource_requests` reports an updated value in this case, even though the container resources were not patched)

wardady:
Hey, just checking up if anything else is missing to get the review for this PR?

mrIncompetent:
As I've opened https://github.com/kubernetes/kube-state-metrics/pull/2773 here my 2 cents:
- `kube_pod_container_status_resource_*` would be more consistent with the other metric names for container status metrics (see `kube_pod_container_status_last_terminated_exitcode` , etc.)
- Extract the `ResourceList` handling into a function to avoid duplicating this code for the 6th time - unless you have differences (which could be valid given resizing is only supported for CPU & Memory, but likely doesn't hurt)
- Add support for init containers as sidecars can be resized too

mrIncompetent:
@CatherineF-dev @mrueg Do you happen to have some cycles to provide some feedback?

mrIncompetent:
Ping @CatherineF-dev @mrueg

mrIncompetent:
Ping @CatherineF-dev @mrueg

wardady:
@CatherineF-dev @mrueg just checking if you have some availability to provide feedback on this PR? Happy to continue working on it.

PR #2703, opened 2025-07-08
Continue running kube-state-metrics when config file doesnt exist at startup

### LINKED ISSUE #2700: kube-state-metrics should support running if config file doesn't exist at startup
Currently, [k-s-m expects the file provided in the config parameter to exist at startup](https://github.com/kubernetes/kube-state-metrics/blob/56d3b561e6954e0055ea0d6f2d7034f6d898b6c6/internal/wrapper.go#L50C1-L56C47). We are planning to take dependency on the config file by mounting it as an optional configmap which users can use to optionally override k-s-m configuration.

**Why is this needed**:
We are planning to take dependency on the config file but mounting it as an optional configmap. In this case, the users can create a configmap with options at any time and the k-s-m pod should be able to do a hot reload([already supported in k-s-m](https://github.com/kubernetes/kube-state-metrics/blob/56d3b561e6954e0055ea0d6f2d7034f6d898b6c6/internal/wrapper.go#L58)).

k-s-m should continue to run when file doesnt exist and watch for file creation([file create watch functionality already exists today](https://github.com/kubernetes/kube-state-metrics/blob/56d3b561e6954e0055ea0d6f2d7034f6d898b6c6/internal/wrapper.go#L58))

**Additional context**

rashmichandrashekar:
This PR enables k-s-m should to run when config file doesnt exist.
Currently, [k-s-m expects the file provided in the config parameter to exist at startup](https://github.com/kubernetes/kube-state-metrics/blob/56d3b561e6954e0055ea0d6f2d7034f6d898b6c6/internal/wrapper.go#L50C1-L56C47) and if not, errors out. We are planning to take dependency on the config file by mounting it as an optional configmap which users can use to optionally override k-s-m configuration.

Does not change cardinality

Fixes #2700

### DIFF
--- a/internal/wrapper.go
+++ b/internal/wrapper.go
@@ -43,17 +43,19 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 	}
 
 	ctx, cancel := context.WithCancel(context.Background())
+
 	if file := options.GetConfigFile(*opts); file != "" {
 		cfgViper := viper.New()
 		cfgViper.SetConfigType("yaml")
 		cfgViper.SetConfigFile(file)
-		if err := cfgViper.ReadInConfig(); err != nil {
+		var err error
+		if err = cfgViper.ReadInConfig(); err != nil {
 			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
-				klog.ErrorS(err, "Options configuration file not found", "file", file)
+				klog.InfoS("Options configuration file not found at startup", "file", file)
 			} else {
 				klog.ErrorS(err, "Error reading options configuration file", "file", file)
+				klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 			}
-			klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 		}
 		cfgViper.OnConfigChange(func(e fsnotify.Event) {
 			klog.InfoS("Changes detected", "name", e.Name)
@@ -65,13 +67,15 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 		})
 		cfgViper.WatchConfig()
 
-		// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
-		configFile, err := os.ReadFile(filepath.Clean(file))
-		if err != nil {
-			klog.ErrorS(err, "failed to read options configuration file", "file", file)
-		}
+		if err == nil {
+			// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
+			configFile, err := os.ReadFile(filepath.Clean(file))
+			if err != nil {
+				klog.ErrorS(err, "failed to read options configuration file", "file", file)
+			}
 
-		yaml.Unmarshal(configFile, opts)
+			yaml.Unmarshal(configFile, opts)
+		}
 	}
 	if opts.CustomResourceConfigFile != "" {
 		crcViper := viper.New()
--- a/internal/wrapper.go
+++ b/internal/wrapper.go
@@ -52,6 +52,10 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 		if err = cfgViper.ReadInConfig(); err != nil {
 			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
 				klog.InfoS("Options configuration file not found at startup", "file", file)
+			} else if _, err = os.Stat(filepath.Clean(file)); os.IsNotExist(err) {
+				// Adding this check in addition to the above since viper.ConfigFileNotFoundError is not working as expected due to this issue -
+				// https://github.com/spf13/viper/issues/1783
+				klog.InfoS("Options configuration file not found at startup", "file", file)
 			} else {
 				klog.ErrorS(err, "Error reading options configuration file", "file", file)
 				klog.FlushAndExit(klog.ExitFlushTimeout, 1)
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -125,29 +125,33 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 	})
 	storeBuilder := store.NewBuilder()
 	storeBuilder.WithMetrics(ksmMetricsRegistry)
-
 	got := options.GetConfigFile(*opts)
 	if got != "" {
-		configFile, err := os.ReadFile(filepath.Clean(got))
-		if err != nil {
-			return fmt.Errorf("failed to read opts config file: %v", err)
-		}
-		// NOTE: Config value will override default values of intersecting options.
-		err = yaml.Unmarshal(configFile, opts)
-		if err != nil {
-			// DO NOT end the process.
-			// We want to allow the user to still be able to fix the misconfigured config (redeploy or edit the configmaps) and reload KSM automatically once that's done.
-			klog.ErrorS(err, "failed to unmarshal opts config file")
-			// Wait for the next reload.
-			klog.InfoS("misconfigured config detected, KSM will automatically reload on next write to the config")
-			klog.InfoS("waiting for config to be fixed")
-			configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(0)
-			<-ctx.Done()
+		if _, err := os.Stat(filepath.Clean(got)); os.IsNotExist(err) {
+			klog.InfoS("config file does not exist, ignoring", "file", got)
 		} else {
-			configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(1)
-			configSuccessTime.WithLabelValues("config", filepath.Clean(got)).SetToCurrentTime()
-			hash := md5HashAsMetricValue(configFile)
-			configHash.WithLabelValues("config", filepath.Clean(got)).Set(hash)
+			configFile, err := os.ReadFile(filepath.Clean(got))
+			if err != nil {
+				return fmt.Errorf("failed to read opts config file: %v", err)
+			}
+			// NOTE: Config value will override default values of intersecting options.
+			err = yaml.Unmarshal(configFile, opts)
+
+			if err != nil {
+				// DO NOT end the process.
+				// We want to allow the user to still be able to fix the misconfigured config (redeploy or edit the configmaps) and reload KSM automatically once that's done.
+				klog.ErrorS(err, "failed to unmarshal opts config file")
+				// Wait for the next reload.
+				klog.InfoS("misconfigured config detected, KSM will automatically reload on next write to the config")
+				klog.InfoS("waiting for config to be fixed")
+				configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(0)
+				<-ctx.Done()
+			} else {
+				configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(1)
+				configSuccessTime.WithLabelValues("config", filepath.Clean(got)).SetToCurrentTime()
+				hash := md5HashAsMetricValue(configFile)
+				configHash.WithLabelValues("config", filepath.Clean(got)).Set(hash)
+			}
 		}
 	}
 


internal/wrapper.go
```diff
@@ -43,17 +43,23 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
 		cfgViper.SetConfigType("yaml")
 		cfgViper.SetConfigFile(file)
-		if err := cfgViper.ReadInConfig(); err != nil {
+		var err error
+		if err = cfgViper.ReadInConfig(); err != nil {
 			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
-				klog.ErrorS(err, "Options configuration file not found", "file", file)
+				klog.InfoS("Options configuration file not found at startup", "file", file)
+			} else if _, err = os.Stat(filepath.Clean(file)); os.IsNotExist(err) {
+				// Adding this check in addition to the above since viper.ConfigFileNotFoundError is not working as expected due to this issue -
```
mrueg: Make this a TODO, so it gets replaced whenever upstream viper is fixed.

```suggestion
				// TODO: Remove this check once viper.ConfigFileNotFoundError is working as expected, see this issue -
```

pkg/app/server.go
```diff
@@ -125,29 +125,33 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
...
+			configFile, err := os.ReadFile(filepath.Clean(got))
+			if err != nil {
+				return fmt.Errorf("failed to read opts config file: %v", err)
+			}
+			// NOTE: Config value will override default values of intersecting options.
+			err = yaml.Unmarshal(configFile, opts)
+
+			if err != nil {
+				// DO NOT end the process.
+				// We want to allow the user to still be able to fix the misconfigured config (redeploy or edit the configmaps) and reload KSM automatically once that's done.
```
mrueg, line 142: I'm not sure if this is the behaviour we want from ksm. I'd rather prefer it failing if there's a bad config supplied.

rashmichandrashekar: Thanks @mrueg. That behavior was not introduced in this PR, but I am happy to make the change and return error if needed. Although, I am unsure of the original intention of why this was implemented this way. 
PS - I think the whitespace changes are making git render them as changes. Please select _Hide Whitespace_

rashmichandrashekar: Gentle ping :) @mrueg - Pls let me know your thoughts, thanks!

rashmichandrashekar:
Gentle ping to all reviewers. 
@dgrisonnet @logicalhan @mrueg @rexagod - Please take a look and provide your feedback, thanks!

pkg/app/server.go
CatherineF-dev: >  I'm not sure if this is the behaviour we want from ksm. I'd rather prefer it failing if there's a bad config supplied.

Agree

rashmichandrashekar:
@rexagod - Could you pls point me to the where the metrics need to be added? Thanks!

pkg/app/server.go
richabanker: But if this behavior existed before, and we fail now if a bad config was supplied, is that considered breaking to the users? (Though +1 to prefer failing in case of bad config)

rashmichandrashekar: Agree that it might be a breaking change.

rexagod:
I'll have to second other maintainers as well, failing on configuration absence is the behaviour we keep for CRS at the moment too.

```
[kube-state-metrics] go run . --custom-resource-state-config-file="x.yaml"                                                                  main
E0717 23:11:29.242328    1152 wrapper.go:84] "Error reading Custom resource configuration file" err="open x.yaml: no such file or directory" file="x.yaml"
exit status 1
```

rashmichandrashekar:
> I'll have to second other maintainers as well, failing on configuration absence is the behaviour we keep for CRS at the moment too.
> 
> ```
> [kube-state-metrics] go run . --custom-resource-state-config-file="x.yaml"                                                                  main
> E0717 23:11:29.242328    1152 wrapper.go:84] "Error reading Custom resource configuration file" err="open x.yaml: no such file or directory" file="x.yaml"
> exit status 1
> ```

I think there is some confusion. The comments in the earlier section is to fail in case of invalid config. I don’t think there are concerns about continuing to run in case of absent config. We could update the behavior for both CRs and config file if needed in case of absent config files and make them backward compatible with a flag. @rexagod - pls let me know what you think.

rashmichandrashekar:
Ping on this PR. Maintainers @richabanker @rexagod @mrueg @CatherineF-dev  - pls et me know the next steps and changes needed to proceed.

internal/wrapper.go
```diff
@@ -65,13 +71,15 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
-		}
+		if err == nil {
+			// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
+			configFile, err := os.ReadFile(filepath.Clean(file))
+			if err != nil {
+				klog.ErrorS(err, "failed to read options configuration file", "file", file)
+			}
 
-		yaml.Unmarshal(configFile, opts)
+			yaml.Unmarshal(configFile, opts)
```
richabanker, line 83: would this preserve the fields already set in opts that are not specified in configFile ?

richabanker:
@mrueg @rexagod  any other outstanding concerns here?

internal/wrapper.go
richabanker: oh this is previous behavior too, nvm.

richabanker:
I guess my last thoughts on this are that with this change, ksm will continue to run even if the config file is missing, which is a substantial shift in user experience (users who rely on the application's exit code to validate their deployments may find their checks no longer work as expected) and should be called out in the release notes if we do decide to go forward with this.

rashmichandrashekar:
> I guess my last thoughts on this are that with this change, ksm will continue to run even if the config file is missing, which is a substantial shift in user experience (users who rely on the application's exit code to validate their deployments may find their checks no longer work as expected) and should be called out in the release notes if we do decide to go forward with this.

Thanks @richabanker. That makes sense. If it helps with backward compatibility I could also make it configurable with a cli parameter and the new behavior of continuing to run, could be set with this so that there is no breaking change. Please let me know your thoughts. Also tagging @rexagod.

pkg/app/server.go
mrueg: Ah I remember, why we implemented this to allow soft-reloading a config file. yes this should be fine then.

rashmichandrashekar:
> > I guess my last thoughts on this are that with this change, ksm will continue to run even if the config file is missing, which is a substantial shift in user experience (users who rely on the application's exit code to validate their deployments may find their checks no longer work as expected) and should be called out in the release notes if we do decide to go forward with this.
> 
> Thanks @richabanker. That makes sense. If it helps with backward compatibility I could also make it configurable with a cli parameter and the new behavior of continuing to run, could be set with this so that there is no breaking change. Please let me know your thoughts. Also tagging @rexagod.

Ping on this. @richabanker - pls let me know if we want to make this backward compatible with a parameter or just documenting with the release is fine. Thanks!

richabanker:
> > > I guess my last thoughts on this are that with this change, ksm will continue to run even if the config file is missing, which is a substantial shift in user experience (users who rely on the application's exit code to validate their deployments may find their checks no longer work as expected) and should be called out in the release notes if we do decide to go forward with this.
> > 
> > 
> > Thanks @richabanker. That makes sense. If it helps with backward compatibility I could also make it configurable with a cli parameter and the new behavior of continuing to run, could be set with this so that there is no breaking change. Please let me know your thoughts. Also tagging @rexagod.
> 
> Ping on this. @richabanker - pls let me know if we want to make this backward compatible with a parameter or just documenting with the release is fine. Thanks!

Ideally expecting a response from the project leads @rexagod or @mrueg , do you have an opinion here? IS this change as proposed breaking?

rexagod:
> If it helps with backward compatibility I could also make it configurable with a cli parameter and the new behavior of continuing to run, could be set with this so that there is no breaking change.

I'm sorry, but why do we need this exactly? Introducing a breaking feature and putting that behind a flag, defaulting it to off, can set an unwanted precedent for the project.

I don't mean to block you here, but I'm not sure we can introduce this unless there's a general consensus that the addition goes beyond downstream applications and helps the larger community.

rashmichandrashekar:
> > If it helps with backward compatibility I could also make it configurable with a cli parameter and the new behavior of continuing to run, could be set with this so that there is no breaking change.
> 
> I'm sorry, but why do we need this exactly? Introducing a breaking feature and putting that behind a flag, defaulting it to off, can set an unwanted precedent for the project. I don't mean to block you here, but we cannot introduce something like this unless there's a general consensus that the addition goes beyond downstream applications and helps the larger community.

The scenario would be that the config file could be provided at any time and not required during startup, for example having users configure this as a configmap mounted to the container after the container has started running and it can pick up changes dynamically which for the most part works today except that it expects it at startup. In a managed component scenario, this provides the flexibility for users to configure it after it is deployed.

rexagod:
Given how we have live-reloading for both KSM and CRS configurations, I can see the two complementing each other. However, wouldn't you need to do the same for CRS configuration as well (to deliver similar UX)? Would it make sense to have flags for both of these with the flag descriptions also highlighting the use-case for their inception?

rashmichandrashekar:
> Given how we have live-reloading for both KSM and CRS configurations, I can see the two complementing each other. However, wouldn't you need to do the same for CRS configuration as well (to deliver similar UX)? Would it make sense to have flags for both of these with the flag descriptions also highlighting the use-case for their inception?

Yes. Currently we dont support CR configuration. So, I didnt explore that. But I can make the change consistent across the two. I can add comments about the use cases for the config. So, just double checking before I implement - suggestion is to add a config flag that enables this behavior to provide for backward compatibility and make it consistent across KSM and CR configuration with comments.  @rexagod - could you pls confirm if this sounds good?

rexagod:
Sounds good.

To reiterate a bit, it'd be nice to have a condensed version of the scenario you went over in your previous comment, in that flag(s) definition ([here](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/options/options.go#L143-L179)) as well, so other users know exactly when to use it (and thus deduce any similar use cases).

rashmichandrashekar:
> Sounds good.
> 
> To reiterate a bit, it'd be nice to have a condensed version of the scenario you went over in your previous comment, in that flag(s) definition ([here](https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/options/options.go#L143-L179)) as well, so other users know exactly when to use it (and thus deduce any similar use cases).

@rexagod - Addressed to add support for both and added description, pls check. Thanks!

internal/wrapper.go
```diff
@@ -43,17 +43,25 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
 		cfgViper.SetConfigFile(file)
-		if err := cfgViper.ReadInConfig(); err != nil {
+		var err error
+		if err = cfgViper.ReadInConfig(); err != nil {
 			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
-				klog.ErrorS(err, "Options configuration file not found", "file", file)
+				klog.InfoS("Options configuration file not found at startup", "file", file)
+			} else if _, err = os.Stat(filepath.Clean(file)); os.IsNotExist(err) {
+				// TODO: Remove this check once viper.ConfigFileNotFoundError is working as expected, see this issue -
+				// https://github.com/spf13/viper/issues/1783
```
rexagod: Nice catch, thank you!

internal/wrapper.go
```diff
@@ -65,25 +73,33 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 		})
 		cfgViper.WatchConfig()
 
-		// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
-		configFile, err := os.ReadFile(filepath.Clean(file))
-		if err != nil {
-			klog.ErrorS(err, "failed to read options configuration file", "file", file)
-		}
+		if err == nil {
```
rexagod: Can we use a different variable name here, something more specific such as `cfgViperReadInConfigErr`, so this doesn't end up shadowing any other `err` values introduced before this line and after `cfgViper.ReadInConfig`?

pkg/app/server.go
```diff
@@ -539,11 +546,17 @@ func resolveCustomResourceConfig(opts *options.Options) (customresourcestate.Con
 		return yaml.NewDecoder(strings.NewReader(s)), nil
 	}
 	if file := opts.CustomResourceConfigFile; file != "" {
-		f, err := os.Open(filepath.Clean(file))
-		if err != nil {
-			return nil, fmt.Errorf("unable to open Custom Resource State Metrics file: %v", err)
+		if opts.ContinueWithoutCustomResourceConfigFile {
+			if _, err := os.Stat(filepath.Clean(file)); os.IsNotExist(err) {
+				klog.InfoS("custom resource config file does not exist,ignoring", "file", file)
```
rexagod:
```suggestion
				klog.InfoS("custom resource config file does not exist, ignoring", "file", file)
```

pkg/app/server.go
```diff
@@ -539,11 +546,17 @@ func resolveCustomResourceConfig(opts *options.Options) (customresourcestate.Con
 		return yaml.NewDecoder(strings.NewReader(s)), nil
 	}
 	if file := opts.CustomResourceConfigFile; file != "" {
-		f, err := os.Open(filepath.Clean(file))
-		if err != nil {
-			return nil, fmt.Errorf("unable to open Custom Resource State Metrics file: %v", err)
+		if opts.ContinueWithoutCustomResourceConfigFile {
+			if _, err := os.Stat(filepath.Clean(file)); os.IsNotExist(err) {
```
rexagod: We'll want to surface any non-`IsNotExist` errors too, for visibility.

pkg/options/options.go
```diff
@@ -156,13 +158,15 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
 	o.cmd.Flags().Float64Var(&o.AutoGoMemlimitRatio, "auto-gomemlimit-ratio", float64(0.9), "The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. (experimental)")
 	o.cmd.Flags().StringVar(&o.CustomResourceConfig, "custom-resource-state-config", "", "Inline Custom Resource State Metrics config YAML (experimental)")
 	o.cmd.Flags().StringVar(&o.CustomResourceConfigFile, "custom-resource-state-config-file", "", "Path to a Custom Resource State Metrics config file (experimental)")
+	o.cmd.Flags().BoolVar(&o.ContinueWithoutCustomResourceConfigFile, "continue-without-cr-config-file", false, "If true, kube-state-metrics continues to run even if the config file specified by --custom-resource-state-config-file is not present. This is useful for scenarios where config file is not provided at startup but is provided later ex - via configmap. Kube-state-metrics will not exit with an error if the custom-resource-state-config file is not found, instead watches and reloads when it is created")
```
rexagod:
```suggestion
	o.cmd.Flags().BoolVar(&o.ContinueWithoutCustomResourceConfigFile, "continue-without-custom-resource-state-config-file", false, "If true, Kube-state-metrics continues to run even if the config file specified by --custom-resource-state-config-file is not present. This is useful for scenarios where config file is not provided at startup but is provided later, for e.g., via configmap. Kube-state-metrics will not exit with an error if the custom-resource-state-config file is not found, instead watches and reloads when it is created.")
```

pkg/options/options.go
```diff
@@ -47,19 +47,21 @@ type Options struct {
...
-	Node                     NodeType `yaml:"node"`
-	Pod                      string   `yaml:"pod"`
-	TLSConfig                string   `yaml:"tls_config"`
-	TelemetryHost            string   `yaml:"telemetry_host"`
-
-	Config string
+	cmd                                     *cobra.Command
+	Apiserver                               string   `yaml:"apiserver"`
+	CustomResourceConfig                    string   `yaml:"custom_resource_config"`
+	CustomResourceConfigFile                string   `yaml:"custom_resource_config_file"`
```
rexagod:
```suggestion
	CustomResourceConfigFile                string   `yaml:"custom_resource_state_config_file"`
```

pkg/options/options.go
```diff
@@ -47,19 +47,21 @@ type Options struct {
...
+	ContinueWithoutCustomResourceConfigFile bool     `yaml:"continue_without_custom_resource_config_file"`
```
rexagod:
```suggestion
	ContinueWithoutCustomResourceConfigFile bool     `yaml:"continue_without_custom_resource_state_config_file"`
```

pkg/options/options.go
```diff
@@ -156,13 +158,15 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
...
 	o.cmd.Flags().StringVar(&o.Kubeconfig, "kubeconfig", "", "Absolute path to the kubeconfig file")
 	o.cmd.Flags().StringVar(&o.Namespace, "pod-namespace", "", "Name of the namespace of the pod specified by --pod. "+autoshardingNotice)
 	o.cmd.Flags().StringVar(&o.Pod, "pod", "", "Name of the pod that contains the kube-state-metrics container. "+autoshardingNotice)
 	o.cmd.Flags().StringVar(&o.TLSConfig, "tls-config", "", "Path to the TLS configuration file")
 	o.cmd.Flags().StringVar(&o.TelemetryHost, "telemetry-host", "::", `Host to expose kube-state-metrics self metrics on.`)
 	o.cmd.Flags().StringVar(&o.Config, "config", "", "Path to the kube-state-metrics options config file")
+	o.cmd.Flags().BoolVar(&o.ContinueWithoutConfig, "continue-without-config", false, "If true, kube-state-metrics continues to run even if the config file specified by --config is not present. This is useful for scenarios where config file is not provided at startup but is provided later ex - via configmap. Kube-state-metrics will not exit with an error if the config file is not found, instead watches and reloads when it is created")
```
rexagod:
```suggestion
	o.cmd.Flags().BoolVar(&o.ContinueWithoutConfig, "continue-without-config", false, "If true, kube-state-metrics continues to run even if the config file specified by --config is not present. This is useful for scenarios where config file is not provided at startup but is provided later, for e.g., via configmap. Kube-state-metrics will not exit with an error if the config file is not found, instead watches and reloads when it is created.")
```

rashmichandrashekar:
Thanks @mrueg for triggering the test. I am running into issues with the cli-arguments. Is there a way to auto generate this file from the options file?

rashmichandrashekar:
@rexagod - Thanks for all the feedback. Addressed all the comments, pls take a look.

rashmichandrashekar:
> Could you please squash the commits here as well? Thanks!

Done.

rexagod:
I believe https://github.com/kubernetes/kube-state-metrics/pull/2703/files#r2251349420 is left to address.

Also, you'll need to rebase, since the other PR is in now.

pkg/app/server.go
rashmichandrashekar: @rexagod - I have updated this to log the error. pls let me know if you wanted me to do it another way. Since we want to continue to run, we can log the message and the error.

rashmichandrashekar:
> I believe https://github.com/kubernetes/kube-state-metrics/pull/2703/files#r2251349420 is left to address.
> 
> Also, you'll need to rebase, since the other PR is in now.

@rexagod - I think all of this should be addressed now, pls take a look. I did add logs to surface all errors (since I had done it only for one earlier).

pkg/app/server.go
```diff
@@ -596,11 +604,17 @@ func resolveCustomResourceConfig(opts *options.Options) (customresourcestate.Con
...
+				klog.ErrorS(err, "failed to open Custom Resource State Metrics file, ignoring", "file", file)
+			}
+		} else {
```
rexagod:
```suggestion
		if opts.ContinueWithoutCustomResourceConfigFile {
			if _, err := os.Stat(filepath.Clean(file)); err != nil {
				klog.ErrorS(err, "failed to open Custom Resource State Metrics file, ignoring", "file", file)
			}
		} else {
```

I meant surfacing errors when anything except the file not being found, otherwise, IIUC, that information will never get propagated to the user?

rashmichandrashekar: This is how it shows up with the current implementation (but happy to update this to the suggestion, if needed) - 
image

rashmichandrashekar: Commited the suggestion since that will capture all errors

pkg/app/server.go
```diff
@@ -178,15 +183,18 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 	}
 
 	if opts.CustomResourceConfigFile != "" {
-		crcFile, err := os.ReadFile(filepath.Clean(opts.CustomResourceConfigFile))
-		if err != nil {
-			return fmt.Errorf("failed to read custom resource config file: %v", err)
+		if _, err := os.Stat(filepath.Clean(opts.CustomResourceConfigFile)); err != nil {
+			klog.ErrorS(err, "config file does not exist, ignoring", "file", opts.CustomResourceConfigFile)
```
rexagod:
```suggestion
		if _, err := os.Stat(filepath.Clean(opts.CustomResourceConfigFile)); err != nil {
			klog.ErrorS(err, "encountered error while processing the file", "file", opts.CustomResourceConfigFile)
```

pkg/app/server.go
```diff
@@ -125,31 +125,36 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
...
-			// DO NOT end the process.
-			// We want to allow the user to still be able to fix the misconfigured config (redeploy or edit the configmaps) and reload KSM automatically once that's done.
-			klog.ErrorS(err, "failed to unmarshal opts config file")
-			// Wait for the next reload.
-			klog.InfoS("misconfigured config detected, KSM will automatically reload on next write to the config")
-			klog.InfoS("waiting for config to be fixed")
-			configSuccess.WithLabelValues("config", filepath.Clean(got)).Set(0)
-			<-ctx.Done()
+		if _, err := os.Stat(filepath.Clean(got)); err != nil {
+			klog.ErrorS(err, "config file does not exist, ignoring", "file", got)
```
rexagod:
```suggestion
			klog.ErrorS(err, "encountered error while processing the file", "file", got)
```

rexagod (commented):
One final nit regarding the error messages to reflect your last changes in the `if-else` conditionals. I believe we should be good to go after this.

rashmichandrashekar:
> One final nit regarding the error messages to reflect your last changes in the `if-else` conditionals. I believe we should be good to go after this.

@rexagod - Sure, made the changes, please check. Thanks!

rashmichandrashekar:
> > One final nit regarding the error messages to reflect your last changes in the `if-else` conditionals. I believe we should be good to go after this.
> 
> @rexagod - Sure, made the changes, please check. Thanks!

@rexagod - Gentle ping, please let me know if this is good to merge or if there is my other feedback. Thanks!

rashmichandrashekar:
> One final nit regarding the error messages to reflect your last changes in the `if-else` conditionals. I believe we should be good to go after this.

@rexagod - Gentle ping, pls take a look and let me know if everything looks good, looking forward to merging this. Thanks for all the help so far!

internal/wrapper.go
```diff
@@ -43,17 +43,25 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
 	if file := options.GetConfigFile(*opts); file != "" {
 		cfgViper := viper.New()
 		cfgViper.SetConfigType("yaml")
 		cfgViper.SetConfigFile(file)
-		if err := cfgViper.ReadInConfig(); err != nil {
-			if errors.Is(err, viper.ConfigFileNotFoundError{}) {
-				klog.ErrorS(err, "Options configuration file not found", "file", file)
+		var cfgViperReadInConfigErr error
+		if cfgViperReadInConfigErr = cfgViper.ReadInConfig(); cfgViperReadInConfigErr != nil {
+			if errors.Is(cfgViperReadInConfigErr, viper.ConfigFileNotFoundError{}) {
```
mrueg, line 53: Have you tested if this works?
Can you take a look at https://github.com/spf13/viper/issues/1491#issuecomment-2981286182 ?

rashmichandrashekar: @mrueg - This doesn't work as expected because of this bug - https://github.com/spf13/viper/issues/1783
The else if detects file doesn't exist reliably.

rexagod:
Apologies for the delay, I'll take a look by tomorrow at the latest.

internal/wrapper.go
mrueg: What I wanted to point at is the following comment:

```
errors.Is is intended for comparing against sentinel error variables (like os.ErrNotExist), not error types or struct values.
viper.ConfigFileNotFoundError{} is a struct, not a sentinel variable
```

rashmichandrashekar: @mrueg - Thanks for pointing that out. The first if condition that you are pointing to always returns false currently because of the bug mentioned in the comment. We can fix that to also address your comment, when the original issue is fixed and remove the else if completely (please note this was something that existed before). The being said, I did test this out locally to make sure the flow works as expected.

rexagod:
/lgtm
/hold

Holding since @mrueg's reviews are ongoing, feel free to unhold once they are addressed.

internal/wrapper.go
rashmichandrashekar: @mrueg - please let me know if this looks good or if you have additional concerns, thank you! It’s waiting on your review to merge

pkg/app/server.go
```diff
@@ -596,11 +604,17 @@ func resolveCustomResourceConfig(opts *options.Options) (customresourcestate.Con
 		return yaml.NewDecoder(strings.NewReader(s)), nil
 	}
 	if file := opts.CustomResourceConfigFile; file != "" {
-		f, err := os.Open(filepath.Clean(file))
-		if err != nil {
-			return nil, fmt.Errorf("unable to open Custom Resource State Metrics file: %v", err)
+		if opts.ContinueWithoutCustomResourceConfigFile {
+			if _, err := os.Stat(filepath.Clean(file)); err != nil {
+				klog.ErrorS(err, "failed to open Custom Resource State Metrics file, ignoring", "file", file)
```
mrueg: Should this be demoted to a Warning? Or is my assumption that Errors are fatal most of the time wrong?

rashmichandrashekar: Thanks @mrueg. Updated this, pls take a look

rashmichandrashekar: @mrueg - Gentle ping, pls let me know if this looks good. Thanks!

mrueg:
/lgtm

/hold cancel

Thanks for your contribution!

ringerc:
This breaks the config file option `custom_resource_config_file`:  https://github.com/kubernetes/kube-state-metrics/issues/2803

internal/wrapper.go
```diff
@@ -65,25 +73,33 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
...
-		}
+		if cfgViperReadInConfigErr == nil {
+			// Merge configFile values with opts so we get the CustomResourceConfigFile from config as well
+			configFile, err := os.ReadFile(filepath.Clean(file))
+			if err != nil {
+				klog.ErrorS(err, "failed to read options configuration file", "file", file)
+			}
 
-		yaml.Unmarshal(configFile, opts)
+			yaml.Unmarshal(configFile, opts)
```
ringerc, line 83: Somehow this has broken in this patch. See https://github.com/kubernetes/kube-state-metrics/issues/2803#issuecomment-3464511667 . In 2.17.0 a config file containing

```
custom_resource_config_file: custom-resource-config.yaml
```

results in an `opts.CustomResourceConfigFile` that's the empty string.

[In 2.16.0 it instead gets the correct value.](https://github.com/kubernetes/kube-state-metrics/issues/2803#issuecomment-3464541147)

pkg/options/options.go
```diff
@@ -47,19 +47,21 @@ type Options struct {
...
-	Node                     NodeType `yaml:"node"`
-	Pod                      string   `yaml:"pod"`
-	TLSConfig                string   `yaml:"tls_config"`
-	TelemetryHost            string   `yaml:"telemetry_host"`
-
-	Config string
+	cmd                                     *cobra.Command
+	Apiserver                               string   `yaml:"apiserver"`
+	CustomResourceConfig                    string   `yaml:"custom_resource_config"`
+	CustomResourceConfigFile                string   `yaml:"custom_resource_state_config_file"`
```
ringerc, line 53: This is the cause of https://github.com/kubernetes/kube-state-metrics/issues/2803

```
-	CustomResourceConfigFile string   `yaml:"custom_resource_config_file"`
+	CustomResourceConfigFile                string   `yaml:"custom_resource_state_config_file"`
```

The config option was renamed silently, with no documentation or warning, and no check to deprecate the old config option name.

Since it wasn't really properly documented in the first place it's arguably not a breakage of a proper user-facing interface, but still frustrating.

PR #2705, opened 2025-07-09
fix: Config file overrides apply to some fields but not other

### LINKED ISSUE #2704: Config file overrides apply to some fields but not other
**What happened**:
--config option used to provide an optional config file overrides some options but not all, leading to inconsistencies.
When using k-s-m with cli args and config file some fields get overriden with what is present in the config file. For ex- shards, kubeconfig etc (any field that is not of type slice of structs), but for types such as ResourceSet and MetricSet which are of type slice of structs the values in the file get appended to the existing set of values which is the default behavior of yaml.Unmarshal which is used [here](https://github.com/kubernetes/kube-state-metrics/blob/56d3b561e6954e0055ea0d6f2d7034f6d898b6c6/pkg/app/server.go#L136).

This causes the following problems - 
1. Inconsistencies between some fields overridden while not others.
2. When config file is updated, the old values are persisted even though they are not present in the cli args or in the config file.

Intersecting options are overriden by the config file for all fields

Create a config file with -
metric_allowlist: 
  "kube_job_spec_completions": {}

later update this to only contain - 
metric_allowlist: 
  "kube_job_status_succeeded": {}

The old values are retained even though they do not exist

* kube-state-metrics version: v2.16.0

rashmichandrashekar:
Fixes bug where some option values are overriden with config file but not others.
Provides optional behavior for backward compatiblity

Does not change cardinality

Fixes #2704


rashmichandrashekar:
Hi @rexagod - You have been assigned as a reviewer. Could you please review this PR? Thanks!

rashmichandrashekar:
@dgrisonnet @CatherineF-dev - Gentle ping, pls take a look at the pr and provide your review, thanks!

rashmichandrashekar:
@CatherineF-dev - Added tests as per your feedback, pls take a look. Thanks!

pkg/app/server.go
```diff
@@ -378,6 +379,45 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 	return nil
 }
 
+func configureResourcesAndMetrics(opts *options.Options, configFile []byte) *options.Options {
+	// If the config file is set, we will overwrite the opts with the config file.
+	// This is only needed for types that are a slice of structs because the default behaviour of yaml.Unmarshal is to append instead of overwrite
```
rexagod:
```suggestion
func configureResourcesAndMetrics(opts *options.Options, configFile []byte) *options.Options {
	// If the config file is set, we will overwrite the opts with the config file.
	// This is only needed for maps because the default behaviour of yaml.Unmarshal is to append keys (and overwrite any conflicting ones).
```

I'd suggest adding a comment about the config file taking precedence over options in the flag description.

Also, I believe we need to do this for `AnnotationsAllowList` and `LabelsAllowList` as well?

rashmichandrashekar: Good catch. Added this too. @rexagod - please take a look

richabanker:
@rashmichandrashekar this seems to be running into some CI failures.

rashmichandrashekar:
> @rashmichandrashekar this seems to be running into some CI failures.

Thanks @richabanker. Could you pls retrigger the runs?

mrueg:
Triggered the CI run

/ok-to-test
/lgtm

for others to review:
/hold

rashmichandrashekar:
Pinging other reviewers - @dgrisonnet @CatherineF-dev - This PR is waiting on additional reviewers. Could you pls take a look? Thanks!

rexagod:
> I'd suggest adding a comment about the config file taking precedence over options in the flag description.

I believe this is remaining here?

rashmichandrashekar:
> > I'd suggest adding a comment about the config file taking precedence over options in the flag description.
> 
> I believe this is remaining here?

Thanks! added

CatherineF-dev:
> In apiserver, command-line arguments override any settings provided in a configuration file. 

>  https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#start-a-kubelet-process-configured-via-the-config-file

@richabanker and I find standard practice is using command-line arguments to override configuration file.
@dgrisonnet 

If so, is your project blocked by this? Could you find other ways?

rashmichandrashekar:
> > In apiserver, command-line arguments override any settings provided in a configuration file.
> 
> > https://kubernetes.io/docs/tasks/administer-cluster/kubelet-config-file/#start-a-kubelet-process-configured-via-the-config-file
> 
> @richabanker and I find standard practice is using command-line arguments to override configuration file. @dgrisonnet
> 
> If so, is your project blocked by this? Could you find other ways?

@CatherineF-dev  - The current behavior is that the config file overrides the cli args. This PR is making it consistent for all params. Yes, if the change is to break this behavior, it is blocking.

richabanker:
/lgtm 

Re-iterating from the sig meeting, since this fix makes the behavior of overriding CLI flags with config file consistent for slice params, makes sense to me to have this fix merged now.

While this is opposite to the pattern we have for the main K8s components, but this was a choice made early on, which will require a separate dedicated discussion/enhancement if we want to adopt that for KSM due to the breaking nature of the change.

richabanker:
/hold 
for @CatherineF-dev and @rexagod to express any outstanding concerns

docs/developer/cli-arguments.md
```diff
@@ -44,7 +44,7 @@ Flags:
       --auth-filter                                If true, requires authentication and authorization through Kubernetes API to access metrics endpoints
       --auto-gomemlimit                            Automatically set GOMEMLIMIT to match container or system memory limit. (experimental)
       --auto-gomemlimit-ratio float                The ratio of reserved GOMEMLIMIT memory to the detected maximum container or system memory. (experimental) (default 0.9)
-      --config string                              Path to the kube-state-metrics options config file
+      --config string                              Path to the kube-state-metrics options config file, which is a YAML file that contains the options to be used by kube-state-metrics. If this flag is set, the values in the yaml file override the command line flags for options.
```
rexagod:
```suggestion
      --config string                              Path to the kube-state-metrics options config YAML file. If this flag is set, the flags defined in the file override the command line flags.
```

Brevity check.

rashmichandrashekar:
> Could you please squash the commits here as well?

Done

rashmichandrashekar:
@rexagod - Addressed all comments. pls take a look

PR #2708, opened 2025-07-21
feat: add kube_deployment_status_replicas_terminating and kube_replicaset_status_terminating_replicas metrics

atiratree:

This is needed to support the https://github.com/kubernetes/enhancements/issues/3973 observability and to promote DeploymentReplicaSetTerminatingReplicas to Beta https://github.com/kubernetes/kubernetes/pull/133087

**How does this change affect the cardinality of KSM**: *increases*

Fixes #


CatherineF-dev:
Error: ../../../go/pkg/mod/k8s.io/apiserver@v0.32.1/pkg/cel/environment/base.go:53:30: undefined: featuregate.DefaultComponentGlobalsRegistry
Error: ../../../go/pkg/mod/k8s.io/apiserver@v0.32.1/pkg/cel/environment/base.go:53:94: undefined: featuregate.DefaultKubeComponent
Error: ../../../go/pkg/mod/k8s.io/apiserver@v0.32.1/pkg/cel/environment/base.go:55:30: undefined: utilversion.DefaultKubeEffectiveVersion

Could you upgrade k8s version in a separate pull request?

atiratree:
I removed the go.mod bump requirement and rebased the PR.

Should be ready for a review @CatherineF-dev @richabanker @rexagod 

@soltysh for SIG Apps

atiratree:
@CatherineF-dev @dgrisonnet @mrueg @rexagod  Can you please take a look?

internal/store/deployment.go
```diff
@@ -160,6 +160,26 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 				}
 			}),
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_deployment_status_replicas_terminating",
```
dgrisonnet: nit: I think terminating_replicas would be a slightly better name following the Prometheus best practices: https://prometheus.io/docs/practices/naming/#metric-names

More precisely this part:
```
...SHOULD have a suffix describing the unit, in plural form.
```

It would also match perfectly the field on the object

atiratree: This was originally done to use the same format as the other deployment metrics. But, I agree that the suggested format is better. Updated.

PR #2715, opened 2025-08-03
docs: High Availability Setup documentation

### LINKED ISSUE #2081: High Availability setup documentation request

I would love to ask for the recommendation on High Availability setup for kube-state-metrics.

**Why is this needed**:

We use some of kube-state-metrics metrics to calculate our service SLI metric. Unfortunately during host rebalancing or host upgrades we do get gaps in SLI metrics because kube-state-metrics gets restarted/rescheduled (becomes unavailable).

It would be great to have a paragraph in the README describing HA setup, similar to [scaling recommendations](https://github.com/kubernetes/kube-state-metrics#scaling-kube-state-metrics).

**Additional context**

We're using kube-state-metrics bundled by [kube-prometheus project](https://github.com/prometheus-operator/kube-prometheus/blob/main/manifests/kubeStateMetrics-deployment.yaml#L12) (which dues run a single instance of kube-state-metrics).

Rishab87:
Adds documentation for high availability setups

does not change cardinality

Fixes #2081

### DIFF
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ are deleted they are no longer visible on the `/metrics` endpoint.
   * [Horizontal sharding](#horizontal-sharding)
     * [Automated sharding](#automated-sharding)
   * [Daemonset sharding for pod metrics](#daemonset-sharding-for-pod-metrics)
+* [High Availability](#high-availability)
 * [Setup](#setup)
   * [Building the Docker container](#building-the-docker-container)
 * [Usage](#usage)
@@ -303,6 +304,50 @@ spec:
 ```
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
+
+### High Availability
+
+Kube-state-metrics is a stateless service that reads from the Kubernetes API server. Be aware that multiple replicas increase the load on the Kubernetes API. Therefore, in most cases a single replica is also an option since most users scrape with a 30s interval. If you have the need for a higher scrape frequency or you have other constraints that require multiple replica, you can increase the availablity of kube-state-metrics in the following way:
+
+For high availability, run multiple kube-state-metrics replicas to prevent a single point of failure. A common setup uses at least 2 replicas, pod anti-affinity rules to ensure they run on different nodes, and a PodDisruptionBudget (PDB) with `minAvailable: 1` to protect against voluntary disruptions.
+
+```yaml
+apiVersion: apps/v1
+kind: Deployment
+metadata:
+  name: kube-state-metrics
+spec:
+  replicas: 2
+  template:
+    spec:
+      affinity:
+        podAntiAffinity:
+          preferredDuringSchedulingIgnoredDuringExecution:
+          - weight: 100
+            podAffinityTerm:
+              labelSelector:
+                matchLabels:
+                  app: kube-state-metrics
+              topologyKey: kubernetes.io/hostname
+      containers:
+      - name: kube-state-metrics
+        image: registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.10.0
+        ports:
+        - containerPort: 8080
+          name: http-metrics
+---
+apiVersion: policy/v1
+kind: PodDisruptionBudget
+metadata:
+  name: kube-state-metrics-pdb
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app: kube-state-metrics
+```
+
+If metrics get scraped at the service level via a ServiceMonitor (Prometheus-Operator) or similar, metrics won't need to be deduplicated in a HA setup.
 
 ### Setup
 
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -304,6 +305,12 @@ spec:
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
 
+### High Availability
+
+For high availability, run multiple kube-state-metrics replicas with anti-affinity rules to prevent single points of failure. Configure 2 replicas, anti-affinity rules on hostname, rolling update strategy with `maxUnavailable: 1`, and a PodDisruptionBudget with `minAvailable: 1`.
```
mrueg: What is the value-add specifically for ksm here? I believe this would apply to most Kubernetes deployments if you want to run them highly available.

README.md
```diff
@@ -304,6 +305,12 @@ spec:
...
+
+When using multiple replicas, Prometheus will scrape all instances resulting in duplicate metrics with different instance labels. Handle deduplication in queries using `avg without(instance) (metric_name)`. Brief inconsistencies may occur during state transitions but resolve quickly as replicas sync with the API server.
```
mrueg: This depends on how you scrape the metrics with prometheus, on a pod level or a service level.

Rishab87: yeah right, I'll update the section to clarify the differences.

README.md
Rishab87: I'll update the documentation to explain that users must perform query-time deduplication

Rishab87:
@mrueg I've addressed reviews can you have a look again?

README.md
```diff
@@ -304,6 +305,12 @@ spec:
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
 
+### High Availability
+
+For high availability, run multiple kube-state-metrics replicas to prevent a single point of failure. A standard setup uses at least 2 replicas, pod anti-affinity rules to ensure they run on different nodes, and a PodDisruptionBudget (PDB) with `minAvailable: 1` to protect against voluntary disruptions.
```
mrueg: I would suggest to add an introductory paragraph:
* It should mention that multiple replica increase the load on the Kubernetes' API as a trade-off.
* It should mention that most likely you don't need HA if you scrape every 30s and you can tolerate a few missing scrapes (which usually is the case) 

* Does a "standard" setup exist?

README.md
```diff
@@ -304,6 +305,12 @@ spec:
...
+
+When scraping the individual pods directly in an HA setup, Prometheus will ingest duplicate metrics distinguished only by the instance label. This requires you to deduplicate the data in your queries, for example, by using `max without(instance) (your_metric)`. The correct aggregation function (max, sum, avg, etc.) is important and depends on the metric type, as using the wrong one can produce incorrect values for timestamps or during brief state transitions.
```
mrueg: Is Pod Scraping that common?  I would assume most folks will scrape at the service level via a ServiceMonitor / Prometheus-Operator or similar.

Rishab87: yeah right

Rishab87:
@mrueg made the changes can you please re-review?

Rishab87:
@mrueg any updates on this one?

README.md
```diff
@@ -304,6 +305,50 @@ spec:
 
 Other metrics can be sharded via [Horizontal sharding](#horizontal-sharding).
 
+### High Availability
+
+Multiple replicas increase the load on the Kubernetes API as a trade-off. Most likely you don't need HA if you scrape every 30s and you can tolerate a few missing scrapes (which usually is the case).
```
mrueg:
```suggestion
Kube-state-metrics is a stateless service that reads from the Kubernetes API server. Be aware that multiple replicas increase the load on the Kubernetes API. Therefore, in most cases a single replica is also an option since most users scrape with a 30s interval. If you have the need for a higher scrape frequency or you have other constraints that require multiple replica, you can increase the availablity of kube-state-metrics in the following way:
```

Rishab87:
@mrueg thanks for the review, made the changes

README.md
```diff
@@ -304,6 +305,50 @@ spec:
...
+metadata:
+  name: kube-state-metrics-pdb
+spec:
+  minAvailable: 1
+  selector:
+    matchLabels:
+      app: kube-state-metrics
+```
+
+Most users will scrape at the service level via a ServiceMonitor / Prometheus-Operator or similar. When scraping the individual pods directly in an HA setup, Prometheus will ingest duplicate metrics distinguished only by the instance label. This requires you to deduplicate the data in your queries, for example, by using `max without(instance) (your_metric)`. The correct aggregation function (max, sum, avg, etc.) is important and depends on the metric type, as using the wrong one can produce incorrect values for timestamps or during brief state transitions.
```
mrueg:
```suggestion
If metrics get scraped at the service level via a ServiceMonitor (Prometheus-Operator) or similar, metrics won't need to be deduplicated in a HA setup.
When scraping the individual pods directly in an HA setup, Prometheus will ingest duplicate metrics distinguished only by the instance label. This means the data in your queries will need to be deduplicated, for example, by using `max without(instance) (your_metric)`. The correct aggregation function (max, sum, avg, etc.) is important and depends on the metric type, as using the wrong one can produce incorrect values for timestamps or during brief state transitions.
```

Rishab87:
@mrueg thanks for the review, made the suggested change

PR #2719, opened 2025-08-04
feat: include reason label to `kube_deployment_status_condition`

### LINKED ISSUE #1991: Include `reason` label to the `kube_deployment_status_condition` metric

Rishab87:
Building upon the work of #2146, I've addressed the reviews and created this.

This PR adds reason label to `kube_deployment_status_condition` metrics.

It's necessary to distinguish between [Progressing Deployment](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#progressing-deployment) and [Complete Deployment ](https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#complete-deployment).

no changes in cardinality

Fixes #1991

### DIFF
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -54,6 +54,7 @@ var (
 		"DeploymentResumed":          {},
 		"MinimumReplicasAvailable":   {},
 		"MinimumReplicasUnavailable": {},
+		"":                           {},
 	}
 )
 
--- a/internal/store/deployment_test.go
+++ b/internal/store/deployment_test.go
@@ -201,6 +201,7 @@ func TestDeploymentStore(t *testing.T) {
 				Status: v1.DeploymentStatus{
 					Conditions: []v1.DeploymentCondition{
 						{Type: v1.DeploymentAvailable, Status: corev1.ConditionFalse, Reason: "ThisReasonIsNotAllowed"},
+						{Type: v1.DeploymentProgressing, Status: corev1.ConditionTrue},
 					},
 				},
 				Spec: v1.DeploymentSpec{
@@ -220,6 +221,9 @@ func TestDeploymentStore(t *testing.T) {
         kube_deployment_status_replicas_ready{deployment="depl3",namespace="ns3"} 0
         kube_deployment_status_replicas_unavailable{deployment="depl3",namespace="ns3"} 0
         kube_deployment_status_replicas_updated{deployment="depl3",namespace="ns3"} 0
+	    kube_deployment_status_condition{condition="Progressing",deployment="depl3",namespace="ns3",reason="",status="false"} 0
+        kube_deployment_status_condition{condition="Progressing",deployment="depl3",namespace="ns3",reason="",status="true"} 1
+        kube_deployment_status_condition{condition="Progressing",deployment="depl3",namespace="ns3",reason="",status="unknown"} 0
 `,
 		},
 	}


internal/store/deployment.go
```diff
@@ -174,8 +182,13 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 					for j, m := range conditionMetrics {
 						metric := m
 
-						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						reason := c.Reason
+						if _, ok := allowedDeploymentReasons[reason]; !ok {
+							reason = ""
```
mrueg: should this be empty? or should this be "unknown" 

CC: @rexagod

Rishab87: I went with an empty string for cleaner queries, but I'm open to changing it if you want.

mrueg: it depends, is the reason ever empty? then we could use unknown to differ between an empty reason and something we don't recognize.

if the reason field is never empty, this approach is fine.

Rishab87: yeah right, seems like reason field can be empty, I've changed it to unknown, thanks for pointing out!

Rishab87:
@mrueg I've made the changes, can you have a look again?

internal/store/deployment.go
```diff
@@ -174,8 +182,13 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 					for j, m := range conditionMetrics {
 						metric := m
 
-						metric.LabelKeys = []string{"condition", "status"}
-						metric.LabelValues = append([]string{string(c.Type)}, metric.LabelValues...)
+						reason := c.Reason
+						if _, ok := allowedDeploymentReasons[reason]; !ok {
```
mrueg, line 195: if it's empty, will it still be unknown?
Can you create a test with the reason field being empty as well as with one test with a non allowed deployment reason?

Rishab87: yeah that was the case earlier, fixed it if its empty then it will remain empty and will show unknown if reason is not in the allowed list, also added the tests for both cases.

Rishab87:
@mrueg made changes, can you have another look?

internal/store/deployment.go
```diff
@@ -183,8 +183,10 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 						metric := m
 
 						reason := c.Reason
-						if _, ok := allowedDeploymentReasons[reason]; !ok {
-							reason = "unknown"
+						if reason != "" {
```
mrueg: Can we add "" to the allowedDeploymentReasons instead?

Rishab87: yeah, added it

Rishab87:
@mrueg I've addressed the reviews, can you look again?

internal/store/deployment.go
```diff
@@ -41,6 +41,15 @@ var (
 	descDeploymentLabelsDefaultLabels = []string{"namespace", "deployment"}
 )
 
+var (
+	allowedDeploymentReasons = map[string]struct{}{
+		"MinimumReplicasAvailable": {},
```
mrueg: Looking at these strings again
https://github.com/kubernetes/kubernetes/blob/master/pkg/controller/deployment/util/deployment_util.go#L67

Can we refer to these here instead and create a complete list based on what's available in deployment_util.go?
```suggestion
import deploymentutil "k8s.io/kubernetes/pkg/controller/deployment/util"
var (
	allowedDeploymentReasons = map[string]struct{}{
		deploymentutil.MinimumReplicasAvailable: {},
```

Rishab87: should I also remove the empty string too then?

Rishab87: @mrueg, When I'm trying to:
```bash
go get k8s.io/kubernetes/pkg/controller/deployment/util
```

I'm getting this:
```bash
k8s.io/controller-manager/pkg/features: reading k8s.io/controller-manager/go.mod at revision v0.0.0: unknown revision v0.0.0
```
Is this a good idea to use internal k8s packages directly? should I directly use strings?

mrueg: Good point, you're right!

Can you copy all other reasons over as a string and add a comment that they are taken from deployment_utils.go?

Rishab87: @mrueg thanks for the review, made the changes

CatherineF-dev:
/lgtm
/hold
@mrueg for the final review

internal/store/deployment.go
```diff
@@ -41,6 +41,22 @@ var (
...
+		"ReplicaSetUpdated":          {},
+		"ReplicaSetCreateError":      {},
+		"NewReplicaSetCreated":       {},
+		"FoundNewReplicaSet":         {},
+		"NewReplicaSetAvailable":     {},
+		"ProgressDeadlineExceeded":   {},
+		"DeploymentPaused":           {},
+		"DeploymentResumed":          {},
+		"MinimumReplicasAvailable":   {},
+		"MinimumReplicasUnavailable": {},
```
mrueg, line 56: This is missing the "" if it's unavailable.

Rishab87: @mrueg added it.

mrueg:
/hold cancel

/lgtm

Thanks for your work on this @Rishab87 !

PR #2724, opened 2025-08-09
refactor: Replace ListerWatcher with ListerWatcherWithContext 

### LINKED ISSUE #2721: Migrate away from deprecated ListWatch to the new ListerWithContext/WatcherWithContext
We currently use the deprecated Lister interface: https://github.com/kubernetes/client-go/blob/v0.33.3/tools/cache/listwatch.go#L36

**Why is this needed**:
Drop a linting warning
Use the newer versions that are context-aware.

naseyro:
`ListerWatcher` is deprecated and we need to move KSM code to use instead `ListerWatcherWithContext`
N/A
**Which issue(s) this PR fixes:
Fixes #2721

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -516,12 +516,13 @@ func (b *Builder) buildIngressClassStores() []cache.Store {
 func (b *Builder) buildStores(
 	metricFamilies []generator.FamilyGenerator,
 	expectedType interface{},
-	listWatchFunc func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher,
+	listWatchWithContextFunc func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcherWithContext,
 	useAPIServerCache bool, objectLimit int64,
 ) []cache.Store {
 	metricFamilies = generator.FilterFamilyGenerators(b.familyGeneratorFilter, metricFamilies)
 	composedMetricGenFuncs := generator.ComposeMetricGenFuncs(metricFamilies)
 	familyHeaders := generator.ExtractMetricFamilyHeaders(metricFamilies)
+	var listerWatcher func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher
 
 	if b.namespaces.IsAllNamespaces() {
 		store := metricsstore.NewMetricsStore(
@@ -531,7 +532,7 @@ func (b *Builder) buildStores(
 		if b.fieldSelectorFilter != "" {
 			klog.InfoS("FieldSelector is used", "fieldSelector", b.fieldSelectorFilter)
 		}
-		listWatcher := listWatchFunc(b.kubeClient, v1.NamespaceAll, b.fieldSelectorFilter)
+		listWatcher := listerWatcher(b.kubeClient, v1.NamespaceAll, b.fieldSelectorFilter)
 		b.startReflector(expectedType, store, listWatcher, useAPIServerCache, objectLimit)
 		return []cache.Store{store}
 	}
@@ -545,7 +546,7 @@ func (b *Builder) buildStores(
 		if b.fieldSelectorFilter != "" {
 			klog.InfoS("FieldSelector is used", "fieldSelector", b.fieldSelectorFilter)
 		}
-		listWatcher := listWatchFunc(b.kubeClient, ns, b.fieldSelectorFilter)
+		listWatcher := listerWatcher(b.kubeClient, ns, b.fieldSelectorFilter)
 		b.startReflector(expectedType, store, listWatcher, useAPIServerCache, objectLimit)
 		stores = append(stores, store)
 	}
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -522,7 +522,6 @@ func (b *Builder) buildStores(
 	metricFamilies = generator.FilterFamilyGenerators(b.familyGeneratorFilter, metricFamilies)
 	composedMetricGenFuncs := generator.ComposeMetricGenFuncs(metricFamilies)
 	familyHeaders := generator.ExtractMetricFamilyHeaders(metricFamilies)
-	var listerWatcher func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher
 
 	if b.namespaces.IsAllNamespaces() {
 		store := metricsstore.NewMetricsStore(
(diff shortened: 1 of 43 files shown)


internal/store/builder.go
```diff
@@ -516,12 +516,13 @@ func (b *Builder) buildIngressClassStores() []cache.Store {
...
 	metricFamilies []generator.FamilyGenerator,
 	expectedType interface{},
-	listWatchFunc func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher,
+	listWatchWithContextFunc func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcherWithContext,
 	useAPIServerCache bool, objectLimit int64,
 ) []cache.Store {
 	metricFamilies = generator.FilterFamilyGenerators(b.familyGeneratorFilter, metricFamilies)
 	composedMetricGenFuncs := generator.ComposeMetricGenFuncs(metricFamilies)
 	familyHeaders := generator.ExtractMetricFamilyHeaders(metricFamilies)
+	var listerWatcher func(kubeClient clientset.Interface, ns string, fieldSelector string) cache.ListerWatcher
```
naseyro: That `listerWatcher` func will be removed (I prevent `startReflector` from raising errors)
should be replaced with `listWatchWithContextFunc`. further information I will explain at `startReflector` function below

naseyro:
I haven't fixed/worked on tests yet, and still doing some changes so I'd prefer to keep this as a draft until I finish my changes and I'd mention you for a review, so you can ignore this for couple of days. I'd revisit this tomorrow.

internal/store/builder.go
```diff
@@ -617,12 +614,12 @@ func (b *Builder) buildCustomResourceStores(resourceName string,
...
 	store cache.Store,
-	listWatcher cache.ListerWatcher,
+	listWatcherWithContext cache.ListerWatcherWithContext,
 	useAPIServerCache bool,
 	objectLimit int64,
 ) {
-	instrumentedListWatch := watch.NewInstrumentedListerWatcher(listWatcher, b.listWatchMetrics, reflect.TypeOf(expectedType).String(), useAPIServerCache, objectLimit)
-	reflector := cache.NewReflectorWithOptions(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, cache.ReflectorOptions{ResyncPeriod: 0})
+	instrumentedListWatchWithContext := watch.NewInstrumentedListerWatcher(listWatcherWithContext, b.listWatchMetrics, reflect.TypeOf(expectedType).String(), useAPIServerCache, objectLimit)
+	reflector := cache.NewReflectorWithOptions(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatchWithContext), expectedType, store, cache.ReflectorOptions{ResyncPeriod: 0})
```
naseyro, line 622: **issue**:
`newReflectorWithOptions` which is used from `client-go` is taking a `ListerWatcher` as an argument and inside its logic converts it to `ListerWatcherWithContext`
but we need to send `ListerWatcherWithContext` from the beginning without the need of a cast.
https://github.com/kubernetes/client-go/blob/v0.33.3/tools/cache/reflector.go#L259

naseyro:
@dgrisonnet @rexagod Hi 👋🏻 I need help here, I guess I messed things up 🤔

naseyro:
Hi @mrueg 👋🏻 can you help me in your free time?
this is an initial solution, if you have a better solution please let me know and I can start working on it.

internal/store/builder.go
mrueg: This is a good catch. Do you want to work with client-go folks to update the implementation to support ListerWatcherWithContext as well?

naseyro: Absolutely Yes! but my question is should we update the parameter list to receive `ListerWatcherWithContext` but wouldn't this make a dependency conflict? Or we should support another version of the function with different parameter list as we would do an override.
or Let's move this question to a new issue at client-go repo? WDYT?

mrueg: Since lister watcher without context is deprecated, I think we need a new NewReflector function in client-go. Yeah please open up an issue with client-go

aojea: is it deprecated? 

https://github.com/kubernetes/kubernetes/blob/17d6c9c551f917b63797cb1cfe9bbc4b01ea6a43/staging/src/k8s.io/client-go/tools/cache/listwatch.go#L105-L111

naseyro: Not sure, I guess yes?
@mrueg

pohly: No, `ListerWatcher` is not deprecated. There's no realistic path towards ever removing it. Code which passes around a `ListerWatcher` is fine. Is that perhaps an approach that you can use instead of converting APIs to `ListerWatcherWithContext`?

Ideally implementers should also provide the context-aware methods, but they should not get deprecation warnings when they don't.

The most important part is that callers should use the context-aware methods. They should get a deprecation warning to alert them of that. Such code can use `ToListerWatcherWithContext` to cast a `ListerWatcher`.

naseyro: Hi @pohly Excuse my little knowledge, Aren't we able to create a new API that supports `ListerWatcherWithContext` as an argument?
and keep the current API that takes `ListerWatcher` and converts its to with context using the `ToListerWatcherWithContext`?

naseyro: If this is not possible, can I know why? thanks 😄

pohly: There were too many existing APIs and structs which just pass through a `ListerWatcher`, without actually doing anything with it. Changing all of those to have variants with `ListWatcherWithContext` would have been a much bigger change, without some actual benefits.

naseyro: @pohly As Manuel Ruger seems kinda busy, can we move forward with [this solution you proposed?](https://github.com/kubernetes/kubernetes/issues/133567#issuecomment-3208011498) What's your opinion?
If we can proceed with this, I'll wait for the approval from manuel with yours too.

assu-2000:
> @dgrisonnet @rexagod Hi 👋🏻 I need help here, I guess I messed things up 🤔

Hi @onasser1, you definitely haven’t messed anything up. I’d recommend taking some time to get familiar with the codebase, that will help you build confidence and make it easier to contribute. I also suggest joining the bi-weekly [meeting](https://docs.google.com/document/d/1FE4AQ8B49fYbKhfg4Tx0cui1V0eI4o3PxoqQPUwNEiU/edit?tab=t.0#heading=h.qpfxt91hdl2x), where you can ask any questions you might have and get valuable insights into the code. That way, you’ll avoid falling into the trap. You’re already doing amazing work ,keep it up!  

Btw, maintainers and reviewers are at bandwidth, so please understand if they can’t always respond within your expected timeline.

naseyro:
@assu-2000 thank you for kind words 😃 I plan to do this actually.

internal/store/builder.go
pohly: I'd prefer to revisit the underlying assumption of this PR ("ListerWatcher is deprecated", which is not true) and instead address whatever linter warnings the current code gets without changing APIs.

naseyro: Thanks Patrick, I've reached out to Manuel, but he is busy with the new release of KSM as it's the current priority, I think we will revisit this once he has capacity to review.

dgrisonnet:
@onasser1 could you please address the CI failures?

PR #2730, opened 2025-08-11
fix: Fix logic for plain text fallback format

### LINKED ISSUE #2587: Metrics handler not falling back to gauge on text/plain negotiations
**What happened**: 

When a prometheus request is sent with the protobuf format, the KSM [metrics handler](https://github.com/kubernetes/kube-state-metrics/blob/d3f0c1853f11b387d7dc2c89bd5e52760f65079c/pkg/metricshandler/metrics_handler.go#L204) should negotiate the content type to be text as protobuf is not supported. However, the metrics handler also responds with type `info` which is not accepted by prometheus for this content type: `invalid metric type "info"`.
 
**What you expected to happen**: Similar to issue [#2270](https://github.com/kubernetes/kube-state-metrics/pull/2270), one solution is to fallback to `gauge` type when negotiating `text/plain` as well as protobuf. 

Running prometheus with `--enable-feature=native-histograms` with a scrape job that defines a fallback `PrometheusText0,0,4` formatm and run KSM with any `info` metrics. 

this issue seems to have been referenced [here](https://github.com/kubernetes/kube-state-metrics/issues/2248#issuecomment-1824568927) and [here](https://github.com/kubernetes/kube-state-metrics/pull/2270#issuecomment-2553137275).

Manually tested if the `invalid metric type: info` error popped up for various prometheus request + fallback formats:

| Prometheus requested format(s) | Fallback format          | Result |
| --------------------------- | ------------------- | ------ |
| OpenMetrics0.0.1, OpenMetrics1.0.0            | PrometheusText0.0.4              | OK     |
| OpenMetrics0.0.1, OpenMetrics1.0.0            | N/A                 | OK     |
| PrometheusProto             | PrometheusText0.0.4              | Fail   |
| PrometheusProto             | OpenMetrics1.0.0             | Fail   |
| PrometheusProto             | OpenMetrics0.0.1             | Fail   |
| PrometheusText1.0.0| PrometheusText0.0.4| Fail   |

There is another issue here (#2248) where v2.12 was supposed to have the fix for the issue I am seeing, but the `invalid type` error persists. 

**Environment**:

* kube-state-metrics version: 2.14
* Kubernetes version (use `kubectl version`):client v1.28.0 + server v1.31.1-gke.2105000
* Other info: Prometheus version 3.0

timonegk:

This pull requests fixes a logic error in metrics_writer.go where metrics headers are replaced when a protobuf format is requested. However, the existing logic is never used because the content type negotiation is already done in a previous step (in metrics_handler.go). There, the content type for proto-based formats is changed to text/plain before passing the argument to SanitizeHeaders.

The pull request changes the condition in SanitizeHeaders to check for the plain-text format instead of protobuf. I changed the signature of SanitizeHeaders to accept expfmt.Format instead of string. This makes checking the content type a bit cleaner. If this is considered a breaking change, we can also change it to a string prefix comparison.

I encountered the error when I tried to use native histogram parsing in prometheus and found errors while parsing kube-state-metrics' metrics. The issue is already described in #2587.

Does not change cardinality.

Fixes #2587

### DIFF
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -92,7 +92,7 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 }
 
 // SanitizeHeaders sanitizes the headers of the given MetricsWriterList.
-func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWriterList {
+func SanitizeHeaders(contentType expfmt.Format, writers MetricsWriterList) MetricsWriterList {
 	var lastHeader string
 	for _, writer := range writers {
 		if len(writer.stores) > 0 {
@@ -104,8 +104,9 @@ func SanitizeHeaders(contentType string, writers MetricsWriterList) MetricsWrite
 				// Skip this step if we encounter a repeated header, as it will be removed.
 				if header != lastHeader && strings.HasPrefix(header, "# HELP") {
 
-					// If the requested content type was proto-based (such as FmtProtoDelim, FmtProtoText, or FmtProtoCompact), replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' protobuf machinery.
-					if strings.HasPrefix(contentType, expfmt.ProtoType) {
+					// If the requested content type is text/plain, replace "info" and "statesets" with "gauge", as they are not recognized by Prometheus' plain text machinery.
+					// When Prometheus requests proto-based formats, this branch is also used because any requested format that is not OpenMetrics falls back to text/plain in metrics_handler.go
+					if contentType.FormatType() == expfmt.TypeTextPlain {
 						infoTypeString := string(metric.Info)
 						stateSetTypeString := string(metric.StateSet)
 						if strings.HasSuffix(header, infoTypeString) {
--- a/pkg/metrics_store/metrics_writer_test.go
+++ b/pkg/metrics_store/metrics_writer_test.go
@@ -276,6 +276,46 @@ func TestSanitizeHeaders(t *testing.T) {
 		expectedHeaders []string
 	}{
 		{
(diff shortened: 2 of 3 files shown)


timonegk:
I tested the PR locally using the following custom resource state config file:
```
spec:
  resources:
  - groupVersionKind:
      group: autoscaling.k8s.io
      kind: VerticalPodAutoscaler
      version: v1
    labelsFromPath:
      namespace:
      - metadata
      - namespace
      target_api_version:
      - spec
      - targetRef
      - apiVersion
      target_kind:
      - spec
      - targetRef
      - kind
      target_name:
      - spec
      - targetRef
      - name
      verticalpodautoscaler:
      - metadata
      - name
    metrics:
    - each:
        info:
          labelsFromPath:
            name:
            - metadata
            - name
        type: Info
      help: VPA container recommendations. Kubernetes labels converted to Prometheus
        labels
      name: verticalpodautoscaler_labels
```
and checked the type of the `kube_customresource_verticalpodautoscaler_labels` metric which is now `gauge` instead of `info` for protobuf or plain text, but is still `info` for OpenMetricsText.

naseyro:
Hi @timonegk , to let the Lint pull request title check pass, you need to change the PR title to follow:
fix: Logic for plain text fallback format
You can copy this and I suppose this to work.

timonegk:
I amended the commit to fix the test. I think that the behavior in the test now reflects the correct behavior: text/plain [does not support info](https://prometheus.io/docs/instrumenting/exposition_formats/#comments-help-text-and-type-information), OpenMetricsText [does](https://github.com/prometheus/OpenMetrics/blob/v1.0.0/specification/OpenMetrics.md#type).

mrueg:
/lgtm

Thanks for your contribution!

PR #2733, opened 2025-08-12
feat: Add pod affinity/anti-affinity metrics for deployments

### LINKED ISSUE #2701: Add schedule spec and status for workload
Currently, kubernetes scheduler supports various scheduling primitives, such as pod affinity/anti-affinity, pod disruption budget, pod topology spread constraints and so on. 

**Why is this needed**:
some scheduling primitives are provided by `prefered` logical instead of `required` which may in the lifecycle of the workload, some break variation may happen because pod priority preemption or node pressure eviction.

I think we need to collect the spec and status of the workload pod distribution or the dummy scheduler attribute for every workload and do some alarms when needed.
NA. I have not consider a solution about this. Will give it a try in the later weekends.
**Additional context**
NA

SoumyaRaikwar:
Adds explicit rule-based pod affinity and anti-affinity metrics for deployments to provide granular visibility into Kubernetes scheduling constraints, addressing issue #2701. 

**Refactored from count-based to explicit rule-based approach** following maintainer feedback for enhanced operational value.

## Which issue(s) this PR fixes
Fixes #2701

## Metrics Added
- `kube_deployment_spec_affinity` - Pod affinity and anti-affinity rules with granular labels

**Labels provided:**
- `affinity` - `podaffinity` | `podantiaffinity`  
- `type` - `requiredDuringSchedulingIgnoredDuringExecution` | `preferredDuringSchedulingIgnoredDuringExecution`
- `topology_key` - The topology key for the rule
- `label_selector` - The formatted label selector string


mrueg:
How would you use this metric for alerting and/or showing information about the deployment?

SoumyaRaikwar:
> How would you use this metric for alerting and/or showing information about the deployment?

These metrics enable critical alerting on scheduling constraint violations. For example: (kube_deployment_spec_pod_anti_affinity_preferred_rules > 0) and (kube_deployment_spec_pod_anti_affinity_required_rules == 0) alerts when deployments rely only on soft anti-affinity rules that can be ignored during node pressure, creating single points of failure.

They also help monitor missing protection: (kube_deployment_spec_pod_anti_affinity_required_rules == 0) and (kube_deployment_spec_pod_anti_affinity_preferred_rules == 0) identifies deployments without any anti-affinity rules.

For dashboards, you can visualize cluster-wide scheduling health with count(kube_deployment_spec_pod_anti_affinity_required_rules > 0) to show how many deployments have proper distribution protection.

During incidents, these metrics help correlate why workloads ended up co-located or why pods failed to schedule due to overly complex constraints.

This addresses #2701's core need: visibility into "preferred vs required" scheduling logic to maintain reliable workload distribution during cluster events. Thanks @mrueg for the question - these use cases demonstrate the operational value of these scheduling constraint metrics!

mrueg:
i think the metric should be explicit, something like:
```
kube_deployment_affinity{affinity="podaffinity", type="requiredDuringSchedulingIgnoredDuringExecution",topologyKey="foo",labelSelector="matchExpression foo in bar,baz"}  1
```

then you can count over these and get the desired result, as well as gather exactly that information about the specific affinity setting.

I'm not sure about the labelSelector at this point, if this should be split into subtypes as well or just calling https://github.com/kubernetes/apimachinery/blob/master/pkg/apis/meta/v1/helpers.go#L171 is enough.

SoumyaRaikwar:
Thanks @mrueg for the feedback! I understand you're looking for more explicit metrics that expose individual affinity rule details rather than just counts.

You're absolutely right that explicit metrics would provide much more granular visibility. Instead of simple count metrics

SoumyaRaikwar:
Hi @mrueg, 

I've successfully refactored the implementation to use explicit rule-based metrics as you requested.

**Key Changes:**

- Replaced 4 count-based metrics with single `kube_deployment_spec_affinity` metric  
- Added granular labels for individual rule visibility and flexible querying  
- Used `metav1.FormatLabelSelector()` for consistent labelSelector formatting  
- Updated comprehensive tests and documentation  

The new approach provides exponentially more operational value while maintaining low cardinality and following the individual object-level data principle from the best practices document.

sadathastuto:
would it be more better to add that metric under pod_metrics ?

SoumyaRaikwar:
Hi @mrueg,
When you have a moment, could you please review my recent PR?

SoumyaRaikwar:
Hi @mrueg,
When you have a moment, could you please review my PR?

SoumyaRaikwar:
Hi @CatherineF-dev, @logicalhan, and @rexagod — could you please review this PR when you have a chance?

internal/store/deployment_test.go
```diff
@@ -141,94 +137,74 @@ func TestDeploymentStore(t *testing.T) {
...
+        kube_deployment_spec_replicas{deployment="depl-with-affinity",namespace="ns1"} 3
+        kube_deployment_status_observed_generation{deployment="depl-with-affinity",namespace="ns1"} 1
+        kube_deployment_status_replicas_available{deployment="depl-with-affinity",namespace="ns1"} 3
+        kube_deployment_status_replicas_unavailable{deployment="depl-with-affinity",namespace="ns1"} 0
+        kube_deployment_status_replicas_updated{deployment="depl-with-affinity",namespace="ns1"} 3
+        kube_deployment_status_replicas{deployment="depl-with-affinity",namespace="ns1"} 3
+        kube_deployment_status_replicas_ready{deployment="depl-with-affinity",namespace="ns1"} 3
 `,
 		},
+		// Continue with other test cases...
```
mrueg: Why this comment here?

internal/store/deployment.go
```diff
@@ -398,3 +407,67 @@ func createDeploymentListWatch(kubeClient clientset.Interface, ns string, fieldS
...
+
+	return &metric.Family{Metrics: metrics}
+}
+
+// formatLabelSelector converts a LabelSelector to a string representation
+func formatLabelSelector(selector *metav1.LabelSelector) string {
+	if selector == nil {
+		return ""
+	}
+	// Use Kubernetes helper function as suggested by @mrueg
```
mrueg: No need to put this comment here :)

internal/store/deployment.go
```diff
@@ -398,3 +407,67 @@ func createDeploymentListWatch(kubeClient clientset.Interface, ns string, fieldS
...
+		}
+	}
+
+	return &metric.Family{Metrics: metrics}
+}
+
+// formatLabelSelector converts a LabelSelector to a string representation
+func formatLabelSelector(selector *metav1.LabelSelector) string {
+	if selector == nil {
+		return ""
```
mrueg, line 535: This would produce a metric with an empty label selector, can the label selector be empty?

internal/store/deployment_test.go
```diff
@@ -1,12 +1,9 @@
 /*
 Copyright 2016 The Kubernetes Authors All rights reserved.
-
```
mrueg, line 3: Keep the empty lines here please

internal/store/deployment.go
```diff
@@ -398,3 +407,67 @@ func createDeploymentListWatch(kubeClient clientset.Interface, ns string, fieldS
...
 }
+func generateDeploymentAffinityMetrics(d *v1.Deployment) *metric.Family {
+	var metrics []*metric.Metric
+
+	if d.Spec.Template.Spec.Affinity == nil {
+		return &metric.Family{Metrics: metrics}
+	}
+
+	// Handle pod affinity rules
+	if d.Spec.Template.Spec.Affinity.PodAffinity != nil {
```
mrueg, line 476: Do we need to handle namespaceSelector and namespaces list as well?

SoumyaRaikwar: Thanks for catching that! I’ve updated the metric to include both namespace_selector and namespaces for PodAffinity and PodAntiAffinity (required and preferred). namespace_selector is rendered with metav1.FormatLabelSelector, and namespaces is a comma-separated list. Tests and docs have been updated accordingly.

internal/store/deployment.go
SoumyaRaikwar: Yes, an empty LabelSelector is valid in Kubernetes (it matches all). I now use metav1.FormatLabelSelector for non-nil selectors and keep the metric even when the selector is empty. If the selector is nil, the label value is an empty string, which represents the “match all” case.

internal/store/deployment_test.go
SoumyaRaikwar: Removed the stray comment.

internal/store/deployment.go
SoumyaRaikwar: Done. formatLabelSelector calls metav1.FormatLabelSelector. Also removed the inline comment.

internal/store/deployment_test.go
SoumyaRaikwar: •  Done. Restored the header spacing in deployment_test.go to match repository style.

SoumyaRaikwar:
Hi @mrueg, can you review it please sir.

examples/standard/kustomization.yaml
```diff
@@ -1,10 +0,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
```
mrueg, line 1: It seems this file got deleted by accident.

examples/autosharding/kustomization.yaml
```diff
@@ -1,13 +0,0 @@
-apiVersion: kustomize.config.k8s.io/v1beta1
```
mrueg, line 1: It seems this file got deleted by accident.

internal/store/deployment_test.go
```diff
@@ -1,12 +1,9 @@
...
 Licensed under the Apache License, Version 2.0 (the "License");
 you may not use this file except in compliance with the License.
 You may obtain a copy of the License at
-
```
mrueg, line 7: still removes empty lines here.

SoumyaRaikwar: reverted it

examples/standard/kustomization.yaml
SoumyaRaikwar: Thanks for spotting that — the kustomization.yaml files in examples/autosharding and examples/standard were accidentally dropped during rebase. I’ve restored them from upstream/main and pushed the fix.

examples/autosharding/kustomization.yaml
SoumyaRaikwar: Thanks for spotting that — the kustomization.yaml files in examples/autosharding and examples/standard were accidentally dropped during rebase. I’ve restored them from upstream/main and pushed the fix.

SoumyaRaikwar:
Hi @mrueg — I’ve restored the deleted kustomization.yaml files in examples/autosharding and examples/standard.  
Reverted the whitespace change in internal/store/deployment_test.go; could you please take another look?

SoumyaRaikwar:
@mrueg @CatherineF-dev , could you please review my pr.

SoumyaRaikwar:
@rexagod  could you please review my pr

SoumyaRaikwar:
@mrueg  could you please review my pr when you have a chance

SoumyaRaikwar:
@mrueg could you review my pr when you have a moment

SoumyaRaikwar:
@mrueg please review my pr.

SoumyaRaikwar:
@mrueg  i have resolved all the merge conflict, could please review my pr ?

SoumyaRaikwar:
@mrueg  could you please review sir?

PR #2739, opened 2025-08-19
Add info metric for client-go version

fpetkovski:

KSM supports a range of Kubernetes versions which makes it important to keep up with new versions as clusters upgrade.

In order to allow for alerting on unsupported cluster versions, this commit adds an info metric for the client-go version used by KSM. This way users can set up alerts if the cluster in which KSM is running is outside of the supported range.

Cardinality is unchanged

Fixes #


pkg/app/server.go
```diff
@@ -86,6 +90,16 @@ func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options) erro
 func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 	ksmMetricsRegistry := prometheus.NewRegistry()
 	ksmMetricsRegistry.MustRegister(versionCollector.NewCollector("kube_state_metrics"))
+
+	clientGoVersion := promauto.With(ksmMetricsRegistry).NewGaugeVec(
+		prometheus.GaugeOpts{
+			Name: "kube_state_metrics_client_go_info",
```
mrueg: I wonder if we can make this part of the kube_state_metrics_build_info metrics, but then on the other hand we don't have that many metrics in this metric path.

pkg/app/server.go
```diff
@@ -86,6 +90,16 @@ func RunKubeStateMetricsWrapper(ctx context.Context, opts *options.Options) erro
...
+			Help: "An info metric for the client-go version used by kube-state-metrics",
```
mrueg: Can we mark it as experimental?

pkg/app/server.go
fpetkovski: Adding the label to `kube_state_metrics_build_info` makes sense. Does this metric exist though? I wasn't able to find it on port 8081.

mrueg: https://github.com/prometheus/client_golang/pull/1860 PR is here to add support.

mrueg:
Dropping this from the v2.17 milestone, I would prefer to have it in the build_info metric via the client_golang patch.

mrueg:
@fpetkovski https://github.com/prometheus/client_golang/pull/1860 got merged. Do you want to update the client_golang dep to point at a commit on main and make the change here?

pkg/app/server.go
fpetkovski: @mrueg sorry for the late reply, had to step away from this for a while. I've added the label as part of `kube_state_metrics_build_info` now. I had to update client-golang to the latest main commit since you're change is not present on the latest release.

pkg/app/server.go
fpetkovski: I assume this suggestion is now outdated. Do I need to update information on the `kube_state_metrics_build_info`  metric?

pkg/app/server.go
mrueg: thanks!

PR #2742, opened 2025-08-22
feat: Add ResourceMetricsConfig for configurable core resource metrics

### LINKED ISSUE #2165: Resource Metrics as Config

@dgrisonnet described this idea in a private chat a while ago and I want to avoid that we forget it, so I'm tracking it here.

The goal would be to replace existing core metrics with a configuration based approach similar to existing customresourcestate metrics. That way users can alter, remove and update metrics via a configuration to ensure they contain what the users need.
KSM would still supply a standard config.

**Why is this needed**:
Some users would like to create metrics that are not desired in standard workloads (e.g. due to high cardinality). With metrics as config, we allow these users to define their own metric without causing issues or additional complexity (allowlists/denylists) for the rest of the users.

Similar to customresourcestate metrics but with a more intuitive configuration syntax.

**Additional context**
-

Rishab87:
Implements ResourceMetricsConfig to allow users to define custom metrics for core Kubernetes resources (Pods, Deployments, etc.) via YAML configuration. This enables users to create only the metrics they need, reducing cardinality and eliminating the complexity of allowlists/denylists. The feature reuses the proven CustomResourceState metric generation system for familiar configuration syntax.

Decreases cardinality by allowing users to define only specific metrics they need instead of all default metrics for a resource type.

Fixes #2165

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -215,10 +215,29 @@ func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.Registry
 		} else {
 			gvrString = f.Name()
 		}
-		if _, ok := availableStores[gvrString]; ok {
+
+		key := gvrString
+		if _, ok := availableStores[f.Name()]; ok {
+			key = f.Name()
+			klog.InfoS("Overriding core resource store with custom factory", "resource", key)
+		} else if _, ok := availableStores[gvrString]; ok {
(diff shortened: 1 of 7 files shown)


mrueg:
Just before you continue on this, it would be good to sync with @rexagod and myself as there's https://github.com/kubernetes/enhancements/pull/4811 https://github.com/rexagod/resource-state-metrics out there as well :)

PR #2751, opened 2025-08-28
chore: Reduce code duplication for metadata metric families

mrueg:
There are common metrics shared among all resources for creation, deletion, labels and annotations. We can deduplicate code this way and currently I implemented it for pods (I will apply it to other resources as well).

None
Fixes #


internal/store/metadata.go
```diff
@@ -0,0 +1,155 @@
...
+
+	"k8s.io/kube-state-metrics/v2/pkg/metric"
+
+	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
+)
+
+var (
+	descAnnotationsHelp = "Kubernetes annotations converted to Prometheus labels."
+	descLabelsHelp      = "Kubernetes labels converted to Prometheus labels."
+	descCreationHelp    = "Unix creation timestamp"
```
CatherineF-dev, line 33: qq: why does it have 4 new metrics while the above only deletes 2 metrics (annotation and labels)?

Seems tests are passing, just want to see what happened.

mrueg: I moved the _created _labels _annotations _deleted metrics into the metadata file. This is still WIP, hoping to reduce code duplication. I ran into some issues with the csr resource which relies on info outside of the metadata scope.

Issue #2756, opened 2025-09-01
[LFX] Design Proposal for Automating the Release Process


Rishab87:
**What would you like to be added:**
A pipeline to automate the release process.

**Why is this needed:**
To save some effort on repetitive release task.

**Describe the solution you'd like:**
- [ ] Create bash script for updating `data.yaml`.
- [ ]  Create a pre-release github action which takes in a version number as input and creates the prep PR after running all the tests.
- [ ] Release notes template taking main changes from `changelog.md` + full changelog + welcoming new contributors, so release can be easily created from the command line (or automation). 
- [ ] Create a post-release github action for sending a PR to k8s.io for image promotion. It will also create PR to merge changes of release back to main. 

**Few things to decide on:**
- [ ] Do we need to keep a changelog.md or are we fine with the changelog in github's release?
- [ ] Explore and decide on tooling like [goreleaser.com](http://goreleaser.com/) [ko.build](https://app.slack.com/client/T09NY5SBT/ko.build/)

**Additional context:**

PR #2660, opened 2025-04-23
chore: Replace gojsontoyaml with gojq

mrueg:
**What this PR does / why we need it**:
https://github.com/brancz/gojsontoyaml hasn't had a commit since 4 years +, replace it with a maintained alternative.

**How does this change affect the cardinality of KSM**: *(increases, decreases or does not change cardinality)*
None
**Which issue(s) this PR fixes** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*:
Fixes #

PR #2769, opened 2025-09-07
feat: Added bash script for updating `data.yaml`

Rishab87:
This PR automates the process of updating the `data.yaml` file. To run specify the new version like this:
```bash
./update-data-yaml.sh v2.18.0
```

For each new version, it fetches the latest Kubernetes releases (we keep the k8s version one behind the latest release), adds the new compatibility mapping, and prunes the oldest entry. This ensures our compatibility data is kept current automatically with each new release.

does not change

Related to #2756

### DIFF
--- a/scripts/update-data-yaml.sh
+++ b/scripts/update-data-yaml.sh
@@ -1,0 +1,150 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( cd -- "${SCRIPT_DIR}/.." &> /dev/null && pwd )
+DATA_FILE="${REPO_ROOT}/data.yaml"
+NUM_RELEASES_TO_KEEP=5
+
+GOJQ="go tool github.com/itchyny/gojq/cmd/gojq"
+
+check_command() {
+    if ! command -v "$1" &> /dev/null; then
+        echo "Error: Required command '$1' not found. Please install it to continue." >&2
+        exit 1
+    fi
+}
+
+check_command "go"
+check_command "sort"
+
+if [ -z "$1" ]; then
+    echo "Usage: $0 <new_version>"
+    echo "Example: $0 v2.17.0"
+    exit 1
+fi
+
+if [ ! -f "${DATA_FILE}" ]; then
+    echo "Error: Data file not found at ${DATA_FILE}" >&2
+    exit 1
+fi
+
+NEW_VERSION_WITH_V=$1
+NEW_VERSION_WITHOUT_V=${NEW_VERSION_WITH_V#v}
+
+echo "Starting update process for version ${NEW_VERSION_WITH_V}..."
+
+echo "Checking k8s.io/client-go version from go.mod..."
+GO_MOD="${REPO_ROOT}/go.mod"
+
+if [ ! -f "${GO_MOD}" ]; then
+    echo "Error: go.mod file not found at ${GO_MOD}" >&2
+    exit 1
+fi
+
+CLIENT_GO_VERSION=$(go list -m -f '{{.Version}}' k8s.io/client-go)
+
+if [ -z "$CLIENT_GO_VERSION" ]; then
+    echo "Error: Could not find k8s.io/client-go version in go.mod." >&2
+    exit 1
+fi
+
+K8S_MINOR=$(echo "${CLIENT_GO_VERSION}" | cut -d. -f2)
+
+K8S_VERSION_FOR_NEW_RELEASE="1.${K8S_MINOR}"
+echo "New release ${NEW_VERSION_WITH_V} will be mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+
+# Convert YAML data file to JSON format
+DATA_FILE_JSON_DATA=$(${GOJQ} -r --yaml-input '.' "${DATA_FILE}")
+
+EXISTING_EXACT_MATCH=$(\
+echo "${DATA_FILE_JSON_DATA}" |\
+# Query for existing entry with same version and k8s mapping
+${GOJQ} -r --arg version "${NEW_VERSION_WITH_V}" --arg k8s "${K8S_VERSION_FOR_NEW_RELEASE}" ".compat[] | select(.version == \$version and .kubernetes == \$k8s) | .version // empty"\
+)
+
+if [ -n "${EXISTING_EXACT_MATCH}" ]; then
+    echo "Entry for ${NEW_VERSION_WITH_V} with Kubernetes ${K8S_VERSION_FOR_NEW_RELEASE} already exists. No changes needed."
+    exit 0
+fi
+
+EXISTING_KSM_VERSION_ENTRY=$(\
+echo "${DATA_FILE_JSON_DATA}" |\
+# Check if KSM version already exists (with different k8s mapping)
+${GOJQ} -r --arg version "${NEW_VERSION_WITH_V}" ".compat[] | select(.version == \$version) | .version // empty"\
+)
+
+if [ -n "${EXISTING_KSM_VERSION_ENTRY}" ]; then
+    echo "Version ${NEW_VERSION_WITH_V} found with a different K8s mapping. Updating..."
+    cat > "${DATA_FILE}" << EOF
+# This configuration tracks the last five releases, and acts as the source of truth for those entries in the README.
+#
+# This marks the latest release.
+version: "${NEW_VERSION_WITHOUT_V}"
+
+# List KSM-to-K8s version mapping for the last five releases here, and the default branch.
+compat:
+EOF
+    
+    
+    {
+        echo "${DATA_FILE_JSON_DATA}" |\
+        # Filter out the version being updated and format as YAML
+        ${GOJQ} -r --arg version "${NEW_VERSION_WITH_V}" --arg k8s "${K8S_VERSION_FOR_NEW_RELEASE}" \
+            ".compat[] | select(.version != \$version) | \"  - kubernetes: \\\"\" + .kubernetes + \"\\\"\\n    version: \" + .version"
+        
+        echo "  - kubernetes: \"${K8S_VERSION_FOR_NEW_RELEASE}\""
+        echo "    version: ${NEW_VERSION_WITH_V}"
+    } >> "${DATA_FILE}"
+    
+    echo "Successfully updated existing entry for ${NEW_VERSION_WITH_V}."
+    echo "--- Final ${DATA_FILE} content ---"
+    cat "${DATA_FILE}"
+    exit 0
+fi
+
+echo "Adding new version ${NEW_VERSION_WITH_V} and pruning old releases..."
+
+VERSIONS_LIST_WITHOUT_MAIN=$(\
+echo "${DATA_FILE_JSON_DATA}" |\
+# Extract version|kubernetes pairs, excluding main branch
+${GOJQ} -r ".compat[] | select(.version != \"main\") | \"\(.version)|\(.kubernetes)\"" 2>/dev/null || true\
+)
+
+FULL_VERSIONS_LIST=$(printf "%s\n%s|%s" "${VERSIONS_LIST_WITHOUT_MAIN}" "${NEW_VERSION_WITH_V}" "${K8S_VERSION_FOR_NEW_RELEASE}")
+
+SORTED_VERSIONS=$(\
+echo "${FULL_VERSIONS_LIST}" |\
+# Remove empty lines
+grep -v '^$' |\
+# Sort by version (ascending) using version sort
+sort -t'|' -k1,1 -V |\
+# Keep only the most recent N releases
+tail -n "${NUM_RELEASES_TO_KEEP}"\
+)
+
+cat > "${DATA_FILE}" << EOF
+# This configuration tracks the last five releases, and acts as the source of truth for those entries in the README.
+#
+# This marks the latest release.
+version: "${NEW_VERSION_WITHOUT_V}"
+
+# List KSM-to-K8s version mapping for the last five releases here, and the default branch.
+compat:
+EOF
+
+while IFS='|' read -r version k8s_ver; do
+    if [ -n "${version}" ]; then
+        echo "  - version: \"${version}\"" >> "${DATA_FILE}"
+        echo "    kubernetes: \"${k8s_ver}\"" >> "${DATA_FILE}"
+    fi
+done <<< "${SORTED_VERSIONS}"
+
+echo "  - version: \"main\"" >> "${DATA_FILE}"
+echo "    kubernetes: \"${K8S_VERSION_FOR_NEW_RELEASE}\"" >> "${DATA_FILE}"
+
+echo "Successfully updated and pruned ${DATA_FILE}."
+echo "New release (${NEW_VERSION_WITH_V}) is mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+echo "Main branch is mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+echo "--- Final ${DATA_FILE} content ---"
+cat "${DATA_FILE}"
--- a/go.mod
+++ b/go.mod
@@ -172,6 +172,7 @@ require (
 	github.com/mailru/easyjson v0.7.7 // indirect
 	github.com/mattn/go-colorable v0.1.14 // indirect
 	github.com/mattn/go-isatty v0.0.20 // indirect
+	github.com/mattn/go-runewidth v0.0.15 // indirect
 	github.com/mdlayher/socket v0.4.1 // indirect
 	github.com/mdlayher/vsock v1.2.1 // indirect
 	github.com/mitchellh/go-homedir v1.1.0 // indirect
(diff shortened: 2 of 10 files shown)


scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,111 @@
...
+fi
+
+LATEST_K8S_VERSION=$(echo "${LATEST_K8S_FULL_VERSION}" | sed 's/^v//' | cut -d. -f1,2)
+echo "Latest stable Kubernetes version (N): ${LATEST_K8S_VERSION}"
+
+K8S_MAJOR=$(echo "${LATEST_K8S_VERSION}" | cut -d. -f1)
+K8S_MINOR=$(echo "${LATEST_K8S_VERSION}" | cut -d. -f2)
+PREV_K8S_MINOR=$((K8S_MINOR - 1))
+K8S_VERSION_FOR_NEW_RELEASE="${K8S_MAJOR}.${PREV_K8S_MINOR}"
+echo "New release ${NEW_VERSION_WITH_V} will be mapped to Kubernetes (N-1): ${K8S_VERSION_FOR_NEW_RELEASE}"
```
mrueg: I think it should check k8s.io/client-go's version in go.mod when bumping, if I understand it correctly. :)

Rishab87: yup right, changed it

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,114 @@
...
+check_command() {
+    if ! command -v "$1" &> /dev/null; then
+        echo "Error: Required command '$1' not found. Please install it to continue." >&2
+        exit 1
+    fi
+}
+
+check_command "yq"
+check_command "jq"
+check_command "curl"
```
mrueg: Looks like we don't need curl anymore. :)

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,114 @@
...
+
+check_command() {
+    if ! command -v "$1" &> /dev/null; then
+        echo "Error: Required command '$1' not found. Please install it to continue." >&2
+        exit 1
+    fi
+}
+
+check_command "yq"
+check_command "jq"
```
mrueg: Can we use github.com/itchyny/gojq/cmd/gojq instead?

Once https://github.com/kubernetes/kube-state-metrics/pull/2660 is in, it will be used anyway as part of the project. It supports yaml via --yaml-input --yaml-output args.

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,114 @@
...
+
+echo "Checking k8s.io/client-go version from go.mod..."
+GO_MOD_FILE="${REPO_ROOT}/go.mod"
+
+if [ ! -f "${GO_MOD_FILE}" ]; then
+    echo "Error: go.mod file not found at ${GO_MOD_FILE}" >&2
+    exit 1
+fi
+
+CLIENT_GO_FULL_VERSION=$(grep 'k8s.io/client-go' "${GO_MOD_FILE}" | awk '{print $2}')
```
mrueg: You can get the version like this: https://github.com/kubernetes/kube-state-metrics/pull/2739/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R23

scripts/update-data-yaml.sh
Rishab87: yup

scripts/update-data-yaml.sh
Rishab87: added this, havent pushed `go.mod` and `go.sum` since this is to be merged after that PR

Rishab87:
@mrueg I've addressed all the reviews, can you please re-review?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+
+if [ -z "$CLIENT_GO_FULL_VERSION" ]; then
+    echo "Error: Could not find k8s.io/client-go version in go.mod." >&2
+    exit 1
+fi
+
+K8S_MINOR=$(echo "${CLIENT_GO_FULL_VERSION}" | cut -d. -f2)
+
+K8S_VERSION_FOR_NEW_RELEASE="1.${K8S_MINOR}"
+echo "New release ${NEW_VERSION_WITH_V} will be mapped to Kubernetes (N-1): ${K8S_VERSION_FOR_NEW_RELEASE}"
```
mrueg: What do you mean with the n-1 here?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+
+
+JSON_DATA=$(cat "${DATA_FILE}" | gojq -r --yaml-input '.')
```
mrueg: Do you need to use cat or can gojq also read from a file?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+compat:
+EOF
+    
+    
+    echo "${JSON_DATA}" | gojq -r --arg version "${NEW_VERSION_WITH_V}" --arg k8s "${K8S_VERSION_FOR_NEW_RELEASE}" '
+        .compat[] | select(.version != $version) | "  - kubernetes: \"" + .kubernetes + "\"\n    version: " + .version
+    ' >> "${DATA_FILE}"
+    
+    echo "  - kubernetes: \"${K8S_VERSION_FOR_NEW_RELEASE}\"" >> "${DATA_FILE}"
+    echo "    version: ${NEW_VERSION_WITH_V}" >> "${DATA_FILE}"
```
rexagod: I maybe off but wouldn't append `main` to the list as well?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+
+if [ -n "${EXISTING_KSM_VERSION_ENTRY}" ]; then
+    echo "Version ${NEW_VERSION_WITH_V} found with a different K8s mapping. Updating..."
+    cat > "${DATA_FILE}" << EOF
+# The purpose of this config is to keep all versions in a single place.
+#
+# Marks the latest release
+version: "${CLEAN_NEW_VERSION}"
+
+# List at max 5 releases here + the main branch
```
rexagod:
```suggestion
# This configuration tracks the last five releases, and acts as the source of truth for those entries in the README.
#
# This marks the latest release.
version: "${CLEAN_NEW_VERSION}"

# List KSM-to-K8s version mapping for the last five releases here, and the default branch.
```

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+
+SORTED_VERSIONS=$(echo "${FULL_VERSIONS_LIST}" | grep -v '^$' | sort -t'|' -k1,1 -Vr | head -n "${NUM_RELEASES_TO_KEEP}" | sort -t'|' -k1,1 -V)
+
+cat > "${DATA_FILE}" << EOF
+# The purpose of this config is to keep all versions in a single place.
+
+# Marks the latest release
+version: "${CLEAN_NEW_VERSION}"
+
+# List at max 5 releases here + the main branch
```
rexagod: Same here.

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+    
+    echo "Successfully updated existing entry for ${NEW_VERSION_WITH_V}."
+    echo "--- Final ${DATA_FILE} content ---"
+    cat "${DATA_FILE}"
+    exit 0
+fi
+
+echo "Adding new version ${NEW_VERSION_WITH_V} and pruning old releases..."
+
+VERSIONS_LIST=$(echo "${JSON_DATA}" | gojq -r '.compat[] | select(.version != "main") | "\(.version)|\(.kubernetes)"' 2>/dev/null || true)
```
rexagod:
```suggestion
VERSIONS_LIST_WITHOUT_MAIN=$(echo "${JSON_DATA}" | gojq -r '.compat[] | select(.version != "main") | "\(.version)|\(.kubernetes)"' 2>/dev/null || true)
```

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+FULL_VERSIONS_LIST=$(printf "%s\n%s|%s" "${VERSIONS_LIST}" "${NEW_VERSION_WITH_V}" "${K8S_VERSION_FOR_NEW_RELEASE}")
+
+SORTED_VERSIONS=$(echo "${FULL_VERSIONS_LIST}" | grep -v '^$' | sort -t'|' -k1,1 -Vr | head -n "${NUM_RELEASES_TO_KEEP}" | sort -t'|' -k1,1 -V)
```
rexagod: Again, I maybe off here since I've not run this, but wouldn't this take into account the current 5 releases, instead of dropping the oldest one, and appending the given release and then `main`?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,125 @@
...
+done <<< "${SORTED_VERSIONS}"
+
+echo "  - version: \"main\"" >> "${DATA_FILE}"
+echo "    kubernetes: \"${K8S_VERSION_FOR_NEW_RELEASE}\"" >> "${DATA_FILE}"
+
+echo "Successfully updated and pruned ${DATA_FILE}."
+echo "New release (${NEW_VERSION_WITH_V}) is mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+echo "Main branch is mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+echo "--- Final ${DATA_FILE} content ---"
+cat "${DATA_FILE}"
```
rexagod:
```suggestion
cat "${DATA_FILE}"

```

rexagod (commented):
Also, generally I'd like to see the more complex and lengthy pipe-based commands here broken down and commented at each step (concisely) so they're easier to maintain.

```diff
-FOO=$(a|b|c)
+FOO=$(\
+# what "a" does
+ a|\
+# what "b" does
+ b|\
+# what "c" does
+ c\
+)
```

scripts/update-data-yaml.sh
Rishab87: yes it takes into account current 5 releases excluding the main, so it removes the oldest release, adds the new one and updates the main version too if needed

scripts/update-data-yaml.sh
Rishab87: yes it's being appended separately below

Rishab87:
@rexagod I've made the suggested changes, can you please re-review it? Also I've updated `go.sum` and `go.mod` to add gojq as a dependency in this PR, so this PR does not depend on #2660 to get merged

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,155 @@
...
+JSON_DATA=$(\
+# Convert YAML data file to JSON format
+# Parse YAML input and output as JSON
+gojq -r --yaml-input '.' < "${DATA_FILE}"\
+)
+
+EXISTING_EXACT_MATCH=$(\
+echo "${JSON_DATA}" |\
+# Query for existing entry with same version and k8s mapping
+gojq -r --arg version "${NEW_VERSION_WITH_V}" --arg k8s "${K8S_VERSION_FOR_NEW_RELEASE}" ".compat[] | select(.version == \$version and .kubernetes == \$k8s) | .version // empty"\
```
mrueg: You can call gojq via go tool similar to how we do it in the makefile

scripts/update-data-yaml.sh
Rishab87: added that by mistake I was thinking something while writing down this script, removing it.

scripts/update-data-yaml.sh
Rishab87: Yes, gojq can read directly from a file without using cat. I've already removed cat from here

Rishab87:
@mrueg I've addressed all the reviews, can you please re-review it?

Rishab87:
@mrueg @rexagod  just a follow up on this one

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,152 @@
+#!/bin/bash
```
mrueg, line 1: I believe this file isn't executable by default, can you run chmod +x and commit it again?

Rishab87: done

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,152 @@
...
+)
+
+FULL_VERSIONS_LIST=$(printf "%s\n%s|%s" "${VERSIONS_LIST_WITHOUT_MAIN}" "${NEW_VERSION_WITH_V}" "${K8S_VERSION_FOR_NEW_RELEASE}")
+
+SORTED_VERSIONS=$(\
+echo "${FULL_VERSIONS_LIST}" |\
+# Remove empty lines
+grep -v '^$' |\
+# Sort by version (descending) using version sort
+sort -t'|' -k1,1 -Vr |\
```
mrueg: Why do we need to sort twice here?  Could we use tail instead of head?

Rishab87: yup you're right, thanks for pointing it out made the changes

scripts/update-data-yaml.sh
rexagod: I think that's supposed to mean the latest-1 release, relative to the latest KSM release.

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,150 @@
...
+    exit 1
+fi
+
+if [ ! -f "${DATA_FILE}" ]; then
+    echo "Error: Data file not found at ${DATA_FILE}" >&2
+    exit 1
+fi
+
+NEW_VERSION_WITH_V=$1
+CLEAN_NEW_VERSION=${NEW_VERSION_WITH_V#v}
```
rexagod:
```suggestion
NEW_VERSION_WITHOUT_V=${NEW_VERSION_WITH_V#v}
```

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,150 @@
+#!/bin/bash
+set -euo pipefail
+
+SCRIPT_DIR=$( cd -- "$( dirname -- "${BASH_SOURCE[0]}" )" &> /dev/null && pwd )
+REPO_ROOT=$( cd -- "${SCRIPT_DIR}/.." &> /dev/null && pwd )
+DATA_FILE="${REPO_ROOT}/data.yaml"
```
rexagod, line 6: How about maintaining a couple of mock data files (inputs and outputs), and populating them with the various use-cases covered here, to ensure functionality doesn't break over time?

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,150 @@
...
+
+echo "Checking k8s.io/client-go version from go.mod..."
+GO_MOD_FILE="${REPO_ROOT}/go.mod"
+
+if [ ! -f "${GO_MOD_FILE}" ]; then
+    echo "Error: go.mod file not found at ${GO_MOD_FILE}" >&2
+    exit 1
+fi
+
+CLIENT_GO_FULL_VERSION=$(go list -m -f '{{.Version}}' k8s.io/client-go)
```
rexagod:
```suggestion
CLIENT_GO_VERSION=$(go list -m -f '{{.Version}}' k8s.io/client-go)
```

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,150 @@
...
+    exit 1
+fi
+
+NEW_VERSION_WITH_V=$1
+CLEAN_NEW_VERSION=${NEW_VERSION_WITH_V#v}
+
+echo "Starting update process for version ${NEW_VERSION_WITH_V}..."
+
+echo "Checking k8s.io/client-go version from go.mod..."
+GO_MOD_FILE="${REPO_ROOT}/go.mod"
```
rexagod:
```suggestion
GO_MOD="${REPO_ROOT}/go.mod"
```

scripts/update-data-yaml.sh
```diff
@@ -0,0 +1,150 @@
...
+    exit 1
+fi
+
+K8S_MINOR=$(echo "${CLIENT_GO_FULL_VERSION}" | cut -d. -f2)
+
+K8S_VERSION_FOR_NEW_RELEASE="1.${K8S_MINOR}"
+echo "New release ${NEW_VERSION_WITH_V} will be mapped to Kubernetes: ${K8S_VERSION_FOR_NEW_RELEASE}"
+
+# Convert YAML data file to JSON format
+JSON_DATA=$(${GOJQ} -r --yaml-input '.' "${DATA_FILE}")
```
rexagod:
```suggestion
DATA_FILE_JSON_DATA=$(${GOJQ} -r --yaml-input '.' "${DATA_FILE}")
```

scripts/update-data-yaml.sh
Rishab87: yeah I think so I added it earlier because I was subtracting one version behind the latest k8s version but later Manuel suggested to take the k8s version from `go.mod` instead of fetching from github api and subtracting it.

scripts/update-data-yaml.sh
Rishab87: yeah nice idea, added them

Rishab87:
@rexagod I've addressed all your reviews, can you please re-review it?

Rishab87:
@rexagod just a follow up on this one.

scripts/update-data-yaml.sh
rexagod: Very cool, thanks!

mrueg:
/lgtm

Thanks @Rishab87 !

Issue #2727, opened 2025-08-10
Support kube_job_status_ready metrics


hunshcn:
**What would you like to be added:**

`kube_job_status_ready` metric

**Why is this needed:**

like `kube_job_status_active` (both pending && running), but with ready pods only.
`kube_job_status_ready` can help observe the real running status of job.

**Describe the solution you'd like:**

I am willing to provide pr for this.

**Additional context:**

PR #2771, opened 2025-09-08
feat: Add kube_job_status_ready metric

nmn3m:
This PR adds a new metric `kube_job_status_ready`.  
It works similarly to `kube_job_status_active`, but reports only the number of **ready pods** in a Job.  
This provides better visibility into the real running status of a Job, distinguishing between pods that are simply scheduled/active and those that are actually ready to serve work.  

Does not increase cardinality — it introduces a new metric with the same label set as existing Job status metrics.  

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2727


hunshcn:
I think the data source should be the status of job but not pod lister.

I report this issue because I have found exception in updating job status.

Also it will be closer to the semantics of metric. ( kube_job_status_ready -> job.status.ready ), keep consistent to other metrics like kube_job_status_*

rexagod:
Makes sense, let's use [`Ready`](https://pkg.go.dev/k8s.io/api/batch/v1@v0.33.4#JobStatus.Ready) here.

nmn3m:
> I think the data source should be the status of job but not pod lister.
> 
> I report this issue because I have found exception in updating job status.
> 
> Also it will be closer to the semantics of metric. ( kube_job_status_ready -> job.status.ready ), keep consistent to other metrics like kube_job_status_*

Noticed, and fixed. 
Thanks for your Review @hunshcn

nmn3m:
> Makes sense, let's use [`Ready`](https://pkg.go.dev/k8s.io/api/batch/v1@v0.33.4#JobStatus.Ready) here.
Thanks, applied. 
@rexagod  can you take a look ?.

hunshcn (approved):

PR #2773, opened 2025-09-09
feat: add container status resource metrics from runtime state

mrIncompetent:

The new metrics track currently applied resource limits and requests as reported by the container runtime, which may differ from pod spec during in-place resource updates.

Adds:
- kube_pod_container_status_resource_limits
- kube_pod_container_status_resource_requests
- kube_pod_init_container_status_resource_limits
- kube_pod_init_container_status_resource_requests

Adds 4 new metric families with the same label dimensions as existing container resource metrics (`container`, `namespace`, `node`, `pod`, `resource`, `uid`, `unit`). The cardinality increase is proportional to the number of containers and init containers across all pods.


mrueg:
I believe this is a duplicate of https://github.com/kubernetes/kube-state-metrics/pull/2702 isn't it?

mrIncompetent:
Mostly it is, with slight differences:
- Refactoring of the ResourceList parsing (nice to have)
- This here covers init containers as well as sidecars can be resized too
- Different metric names: `kube_pod_container_status_resource_requests` vs `kube_pod_container_actual_resource_requests`

Edit:
Commented https://github.com/kubernetes/kube-state-metrics/pull/2702#issuecomment-3299633562

PR #2775, opened 2025-09-13
feat: Added bash script for generating release notes

Rishab87:
It adds a bash script which can generate release notes. It can save some effort on manually writing it on every release.

To run:
```bash
./generate-release-notes.sh --tag v2.17.0 --out ./release/test_release.md
```

does not change

Related to #2756

### DIFF
--- a/scripts/generate-release-notes.sh
+++ b/scripts/generate-release-notes.sh
@@ -1,0 +1,219 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# Minimal release notes generator (gh-only)
+# Usage:
+#   ./scripts/generate-release-notes-min.sh --tag v2.17.0 --out release/test_release.md
+#
+# Requires: gh (GitHub CLI) authenticated in the environment.
+
+OUT="release/release_notes.md"
+TAG=""
+
+while [[ $# -gt 0 ]]; do
+  case "$1" in
+    --tag) TAG="$2"; shift 2 ;;
+    --out) OUT="$2"; shift 2 ;;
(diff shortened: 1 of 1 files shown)


scripts/generate-release-notes.sh
```diff
@@ -0,0 +1,220 @@
...
+    login="${rest##*|}"
+    FULL_CHANGELOG_LINES+=("- ${title} by @${login} in #${prnum}")
+    if [[ -n "$login" && "$login" != "null" && "${AUTHORS_IN_ORDER[*]}" != *"$login"* ]]; then
+      AUTHORS_IN_ORDER+=("$login")
+    fi
+  done <<< "$PR_LINES"
+fi
+
+NEW_CONTRIBUTORS=()
+NEW_CONTRIBUTORS=()
```
rexagod: Redeclared?

rexagod:
Of course, if no existing tools match up to this functionality, the solution proposed here would be good as is.

scripts/generate-release-notes.sh
Rishab87: whoops fixed it

Rishab87:
@rexagod from what I’ve seen, we’re already generating SBOMs and other artifacts automatically with the standard process when a release is made. I think we can just add a release-note block in the PR description and a kind/ label on every PR, then use release-notes (the standard k8s tool) to generate release notes instead of maintaining this script. We could also add a GitHub Action to generate release notes and create a new release whenever we push tags.

Rishab87:
as discussed on slack we're moving on with goreleaser for generating release notes and pre-release creation, closing this one.

PR #2782, opened 2025-09-21
feat: add kube_deployment_owner metric

### LINKED ISSUE #2674: Add kube_deployment_owner

There are more and more Operators in the ecosystem. Some of them just start Deployments and it would be nice to get the owner of those deployments

**Why is this needed**:

Add kube_deployment_owner like the kube_pod_owner 

**Additional context**

nmn3m:
This PR adds a new metric `kube_deployment_owner` to track ownership information for Deployments. This is particularly useful for identifying Deployments managed by Operators or other controllers in the Kubernetes ecosystem. The metric includes information about the owner's kind, name, and whether it's a controller, making it easier to understand the ownership hierarchy in clusters.

**How does this change affect the cardinality of KSM:** *increases cardinality*
This change adds one new metric with additional labels (`owner_kind`, `owner_name`, `owner_is_controller`) for each Deployment in the cluster.

Fixes #2674

The new metric provides the following information:
- Owner kind (e.g., Application, CustomResource)
- Owner name


CatherineF-dev:
Could you provide some use cases for this metric?

nmn3m:
@CatherineF-dev 
- Tracking Ownership & Orphaned Deployments: Identifies which Deployments are owned by controllers, and helps quickly find orphaned resources.

- Debugging, Monitoring & Auditing: Aids in troubleshooting by linking issues to the controller, provides visibility for monitoring, and supports auditing/compliance.

internal/store/deployment.go
```diff
@@ -60,6 +61,51 @@ var (
...
 func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []generator.FamilyGenerator {
 	return []generator.FamilyGenerator{
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_deployment_owner",
+			"Information about the Deployment's owner.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapDeploymentFunc(func(d *v1.Deployment) *metric.Family {
+				labelKeys := []string{"owner_kind", "owner_name", "owner_is_controller"}
```
CatherineF-dev: What are other possible values other than controller?

nmn3m: Application or custom resources

CatherineF-dev: Should we use an enum instead of boolean here?

nmn3m: why we should do this?, I found that `owner_is_controller` is used in multiple workload like pods, jobs and replicaset, all of them is in boolean form

CatherineF-dev: Discussed with Richa,

What is the use case of owner_is_controller? Could we remove this field until we finalize whether we should use enum or boolean.

Could you create a bug around  whether`owner_is_controller` field should use enum or boolean?

nmn3m: Thanks for the feedback! I've removed the owner_is_controller field from kube_deployment_owner and created an [issue](https://github.com/kubernetes/kube-state-metrics/issues/2793)  to track the enum vs boolean question for all owner metrics.

PR #2786, opened 2025-09-29
test: fixing issue realted to test indentation

### LINKED ISSUE #2776: Test Files Pass CI Despite Improper Indentation
### Issue
PR #[[2749](https://github.com/kubernetes/kube-state-metrics/pull/2749)] passed CI checks despite containing improper indentation in test files. Current linting configuration doesn't catch indentation issues in test files.

### Example
```go
# From horizontalpodautoscaler_test.go
        # TYPE kube_horizontalpodautoscaler_spec_min_replicas gauge
                # TYPE kube_horizontalpodautoscaler_spec_target_metric gauge
    # TYPE kube_horizontalpodautoscaler_status_condition gauge
```
### Solution
- Update `.golangci.yml` to enable whitespace linting for test files  
- Add CI check for consistent test file indentation  
- Add style check to `make lint` command  

### Impact
- Maintains consistent codebase style  
- Catches formatting issues before merge

nmn3m:
This PR fixes an issue where test files contained inconsistent indentation (spaces and tabs) that were not caught by linting.  
We now normalize leading whitespace in test parsing by using:  

```go
trimmedLine := strings.TrimLeft(l, " \t")
```
Additionally, some failing tests were updated to align with the corrected logic.

How does this change affect the cardinality of KSM:
does not change cardinality

Which issue(s) this PR fixes:
Fixes #2776

### DIFF
--- a/internal/store/testutils.go
+++ b/internal/store/testutils.go
@@ -149,7 +149,7 @@ func removeUnusedWhitespace(s string) string {
 	)
 
 	for _, l := range lines {
-		trimmedLine = strings.TrimSpace(l)
+		trimmedLine = strings.TrimLeft(l, " \t")
 
 		if len(trimmedLine) > 0 {
 			trimmedLines = append(trimmedLines, trimmedLine)
--- a/internal/store/endpointslice_test.go
+++ b/internal/store/endpointslice_test.go
@@ -157,10 +157,10 @@ func TestEndpointSliceStore(t *testing.T) {
 					# HELP kube_endpointslice_endpoints Endpoints attached to the endpointslice.
 					# HELP kube_endpointslice_endpoints_hints Topology routing hints attached to endpoints
 					# TYPE kube_endpointslice_endpoints gauge
-        			        # TYPE kube_endpointslice_endpoints_hints gauge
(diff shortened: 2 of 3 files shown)


internal/store/testutils.go
```diff
@@ -149,7 +149,7 @@ func removeUnusedWhitespace(s string) string {
 	)
 
 	for _, l := range lines {
-		trimmedLine = strings.TrimSpace(l)
+		trimmedLine = strings.TrimLeft(l, " \t")
```
mrueg, line 152: why do we need to check for tabs as well? Could we only do spaces?

nmn3m: I went with `strings.TrimLeft(l, "  \t")` instead of just `" "` because when I only trimmed spaces, a lot of tests failed. Using both spaces and tabs handled the mixed indentation and fixed the test failures.

mrueg: Can we fix the indentation instead to only use spaces or won't that work?

nmn3m: I tried to fix the indentation using the white spaces, and the tests failed.
I think using the combination between white spaces and tab will be a good choice in that case.

mrueg:
/lgtm 

Thanks!

PR #2787, opened 2025-09-30
feat: Added a workflow for automating pre-release PR

Rishab87:
**What this PR does / why we need it:**
This PR adds a workflow for automating the pre-release PR creation process. It runs all the commands mentioned in `RELEASE.md` for pre release as well as all the tests.

Test runs: https://github.com/Rishab87/kube-state-metrics/pull/8

**How does this change affect the cardinality of KSM:** *(increases, decreases or does not change cardinality)*
does not changes

**Which issue(s) this PR fixes:** *(optional, in `fixes #<issue number>(, fixes #<issue_number>, ...)` format, will close the issue(s) when PR gets merged)*
Related to #2756
To be merged after #2769

PR #2788, opened 2025-09-30
feat: Added goreleaser config for generating release notes and pre release creation

Rishab87:
In this PR I've added goreleaser config and a workflow which on tag push helps to generate release notes and create a pre release on github.

Test release: https://github.com/Rishab87/kube-state-metrics/releases/tag/v2.18.0 

does not changes

Related to #2756

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,0 +1,46 @@
+name: Pre-release
+
+on:
+  push:
+    tags:
+      - 'v*'
+
+permissions:
+  contents: write
+  discussions: write
+
+jobs:
+  build-and-create-prerelease:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout (full history)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Extract versions
+        id: versions
+        run: |
+          # Extract Go version from Dockerfile
+          GO_VERSION=$(grep '^ARG GOVERSION=' Dockerfile | cut -d'=' -f2)
+          echo "go-version=${GO_VERSION}" >> $GITHUB_OUTPUT
+          
+          # Extract k8s.io/client-go version from go.mod
+          CLIENT_GO_VERSION=$(go list -m -f '{{.Version}}' k8s.io/client-go)
+          echo "client-go-version=${CLIENT_GO_VERSION}" >> $GITHUB_OUTPUT
+          
+          echo "📦 Go version: ${GO_VERSION}"
+          echo "📦 k8s.io/client-go version: ${CLIENT_GO_VERSION}"  
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ steps.versions.outputs.go-version }}
+        id: go
+      - name: Make pre-release
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: latest
+          args: release --clean
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          GO_VERSION: ${{ steps.versions.outputs.go-version }}
+          K8S_CLIENT_VERSION: ${{ steps.versions.outputs.client-go-version }}
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -1,0 +1,77 @@
+version: 2
+
+archives:
+  - formats: [tar.gz]
+    # this name template makes the OS and Arch compatible with the results of `uname`.
+    name_template: >-
+      {{ .ProjectName }}_
+      {{- title .Os }}_
+      {{- if eq .Arch "amd64" }}x86_64
+      {{- else if eq .Arch "386" }}i386
+      {{- else }}{{ .Arch }}{{ end }}
+      {{- if .Arm }}v{{ .Arm }}{{ end }}
+
+changelog:
+  use: github
+  sort: asc
+  groups:
+    - title: "Features"
+      regexp: "(?i)(feat|feature|\\[feature\\]|\\[feat\\]|\\bFEATURE\\b)"
+      order: 1
+    - title: "Bug Fixes"  
+      regexp: "(?i)(fix|bugfix|\\[bugfix\\]|\\[BUGFIX\\]|\\bBUGFIX\\b)"
+      order: 2
+    - title: "Documentation"
+      regexp: "(?i)(docs?|documentation|\\[docs?\\])"
+      order: 3
+    - title: "Dependencies"
+      regexp: "(?i)(build\\(deps\\)|bump|dependencies)"
+      order: 4
+    - title: "Maintenance"
+      regexp: "(?i)(chore|refactor|\\[chore\\])"
+      order: 5
+    - title: "Other Changes"
+      regexp: ".*"
+      order: 99
+
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
+      - 'merge conflict'
+      - 'Merge branch'
+      - 'Merge pull request'
+
+release:
+  github:
+    owner: kubernetes     
+    name: kube-state-metrics      
+
+  prerelease: "true"
+
+  header: |
+    ## {{ .Tag }} / {{ .Now.Format "2006-01-02" }}
+    
+    {{- if index .Env "GO_VERSION" }}
+    ## Note
+    - This release builds with Go `{{ index .Env "GO_VERSION" }}`
+    {{- end }}
+
+    {{- if index .Env "K8S_CLIENT_VERSION" }}
+    - This release builds with `k8s.io/client-go`: `{{ index .Env "K8S_CLIENT_VERSION" }}`
+    {{- end }}
+
+    {{- if index .Env "GITHUB_ACTOR" }}
+    - Release coordinator: @{{ index .Env "GITHUB_ACTOR" }}
+    {{- end }}
+
+  footer: |
+    **Full Changelog**: {{ .GitURL }}/compare/{{ .PreviousTag }}...{{ .Tag }}
+
+  mode: replace
+  
+  # If set, will create a release discussion in the category specified.
+  # Warning: do not use categories in the 'Announcement' format.
+  # Check https://github.com/goreleaser/goreleaser/issues/2304 for more info.
+  # Default: ''.
+  discussion_category_name: Releases


.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - '^docs:'
+      - '^test:'
+      - 'merge conflict'
+      - 'Merge branch'
+      - 'Merge pull request'
+
+release:
+  github:
+    owner: Rishab87     
+    name: kube-state-metrics
```
rexagod, line 48: Is this supposed to be the release author (the one that the release will be attributed to)? If so, this should ideally infer the person who pushed the tag, as they will be responsible for reviewing the overall changeset.

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+      - 'merge conflict'
+      - 'Merge branch'
+      - 'Merge pull request'
+
+release:
+  github:
+    owner: Rishab87     
+    name: kube-state-metrics      
+
+  prerelease: "true"
```
rexagod, line 50: > First a release candidate (e.g. v1.2.0-rc.0) should be cut. If after a period of 7 days no bugs or issues were reported after publishing the release candidate, a stable release (e.g. v1.2.0) can be cut.

Do we plan on running a `prerelease: false` workflow after 7 days?

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+  skip_upload: true
+
+  header: |
+    ## {{ .Tag }} / {{ .Now.Format "2006-01-02" }}
+    
+    ## Note
+    - This release builds with Go `v1.24.6`
+    - This release builds with `k8s.io/client-go`: `v0.33.4`
+    - This release is the last release that has endpoint metrics enabled by default. In the next release we will enable endpointslices metrics and disable endpoint metrics by default, since the endpoint resources are deprecated. You can still alter this behaviour in kube-state-metrics' config.
```
rexagod: I believe this should be dynamic?

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+    exclude:
+      - '^docs:'
+      - '^test:'
+      - 'merge conflict'
+      - 'Merge branch'
+      - 'Merge pull request'
+
+release:
+  github:
+    owner: Rishab87
```
rexagod:
```suggestion
    owner: kubernetes    
```

(once you're done testing on your fork)

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+  filters:
+    exclude:
+      - '^docs:'
+      - '^test:'
+      - 'merge conflict'
+      - 'Merge branch'
+      - 'Merge pull request'
+
+release:
```
rexagod, line 45: Discussions should allow user to report concerns between the prerelease-to-release period.

```
  # If set, will create a release discussion in the category specified.
  #
  # Warning: do not use categories in the 'Announcement' format.
  #  Check https://github.com/goreleaser/goreleaser/issues/2304 for more info.
  #
  # Default: ''.
  discussion_category_name: Releases
```

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+    
+    ## Note
+    - This release builds with Go `v1.24.6`
+    - This release builds with `k8s.io/client-go`: `v0.33.4`
+    - This release is the last release that has endpoint metrics enabled by default. In the next release we will enable endpointslices metrics and disable endpoint metrics by default, since the endpoint resources are deprecated. You can still alter this behaviour in kube-state-metrics' config.
+
+  footer: |
+    **Full Changelog**: {{ .GitURL }}/compare/{{ .PreviousTag }}...{{ .Tag }}
+
+  mode: replace
```
rexagod, line 71:
```
  # You can add extra pre-existing files to the release.
  # The filename on the release will be the last part of the path (base).
  # If another file with the same name exists, the last one found will be used.
  #
  # Templates: allowed.
  extra_files:
    - glob: ./path/to/file.txt
    - glob: ./glob/**/to/**/file/**/*
    - glob: ./glob/foo/to/bar/file/foobar/override_from_previous
    - glob: ./single_file.txt
      name_template: file.txt # note that this only works if glob matches 1 file only
```

I believe we'll need to use this to export the openvex and SBOMs? Or is that done without this already?

.goreleaser.yaml
rexagod: I wonder if you can use

```
  # Additional templated extra files to add to the release.
  # Those files will have their contents pass through the template engine,
  # and its results will be added to the release.
  #
  # This feature is only available in GoReleaser Pro.
  # Templates: allowed.
  templated_extra_files:
    - src: LICENSE.tpl
      dst: LICENSE.txt
```

to achieve this.

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
+version: 2
+
+builds:
+  - skip: true
```
rexagod: KSM isn't shipped [as a library at the moment](https://github.com/goreleaser/goreleaser/blob/5d21342dcecc3deb92ce86d548f79ed2f2713ffd/www/docs/cookbooks/release-a-library.md?plain=1#L8).

.goreleaser.yaml
```diff
@@ -0,0 +1,58 @@
...
+
+archives: []
```
rexagod: Since we do release tarballs:

```
archives:
  - formats: [tar.gz]
    # this name template makes the OS and Arch compatible with the results of `uname`.
    name_template: >-
      {{ .ProjectName }}_
      {{- title .Os }}_
      {{- if eq .Arch "amd64" }}x86_64
      {{- else if eq .Arch "386" }}i386
      {{- else }}{{ .Arch }}{{ end }}
```

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,29 @@
+name: Create prerelease on tag
```
rexagod:
```suggestion
name: Pre-release
```

Also, can we change the file-name to `pre-release.yml` too?

.github/workflows/release-notes.yml
```diff
@@ -0,0 +1,29 @@
...
+    steps:
+      - name: Checkout (full history)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: stable
+      - name: Run GoReleaser
```
rexagod:
```suggestion
      - name: Make pre-release
```
(nit)

rexagod: I see https://github.com/kubernetes/kube-state-metrics/pull/2787/files#diff-8f0c4d810b263ab478547020c847ee0297cc98c11c67fc2209ce604e52719607 is already using that name. Let's rename that to `pre-release-pull-request` then.

.goreleaser.yaml
Rishab87: no this is supposed to be the org, I'll change this

.goreleaser.yaml
Rishab87: I was thinking of creating a seperate post release workflow (for merging back to main and running kpromo first) after which we can mark the release as the main release manually (we need to do mark it as main manually because it needs to be done after the merging of the kpromo PR & the 7 day wait period).

.goreleaser.yaml
Rishab87: they're built currently on creating a main release automatically, I don't think we should add them in pre-release here.

.goreleaser.yaml
Rishab87: I've made the versions dyanmic by getting them from `go.mod` in wokflow and passing them in env

Rishab87:
@rexagod thanks for the review, I've addressed them, can you have a look again?
Test release after making changes: https://github.com/Rishab87/kube-state-metrics/releases/tag/v2.18.0

The full changelog compares to v2.16.0, I think that's because while releasing v2.17.0 & merging back to main, instead of `release-2.17` the `prep-v2.17.0` branch was merged back to main.

.goreleaser.yaml
Rishab87: due to skip build set to true , no artifacts are being uploaded as an asset

.goreleaser.yaml
Rishab87: I should build it then?

.goreleaser.yaml
Rishab87: for now I've removed build skip and I'm able to build these tarballs

.goreleaser.yaml
rexagod: I was wondering if we can close the discussions once that period is over?

Rishab87: nope we need to do that manually

rexagod:
I'd push for keeping the release artifacts same as what we have currently:

Screenshot 2025-10-22 at 5 57 30 PM

and not generate per-arch artifacts since we do not have arch-specific build tags in KSM, it runs the same everywhere:

Screenshot 2025-10-22 at 5 58 44 PM

.goreleaser.yaml
rexagod: You should, as long as it invokes the in-house `build` target.

.goreleaser.yaml
rexagod: Hmm, not sure if we should skip them for pre-releases, I'd rather keep the pre-release and release workflow as coherent as possible, so the [pre-release artifacts](https://github.com/Rishab87/kube-state-metrics/releases/tag/v2.18.0) can be inspected as well. This pattern will allow us to have a nearly identical release copy to verify.

.goreleaser.yaml
rexagod: I'd rather see that automated after a 7-day cron, mostly because it's easy to lose track of. Also, I believe the `kpromo` and "merging back into main" effort needs to be done for the pre-release too, ideally, all [release steps](https://github.com/kubernetes/kube-state-metrics/blob/main/RELEASE.md#prepare-your-release) need to be done for both pre-release and final release, so they may be sourced from a single script.

.goreleaser.yaml
rexagod: I'd prefer if this was still attributed to the person running the release, to maintain ownership records. So instead of a per-org. release like so:

Screenshot 2025-10-22 at 5 36 09 PM

I'd suggest a per-author release such as:

Screenshot 2025-10-22 at 5 37 33 PM

This is all assuming that the name shown in the image and attributed to the release is indeed derived from here.

.goreleaser.yaml
Rishab87: so for that I can create a seperate PR changing the existing workflow which exports openvex and sboms to run on both pre release and release?

.goreleaser.yaml
Rishab87: cool, I'll create a seperate PR with cron to check for pre release and mark it as main

.goreleaser.yaml
Rishab87: it's not possible to show the tag pusher name here, either we need a PAT of someone or we can show on the changelog itself `released by: username`.

.goreleaser.yaml
Rishab87: I just checked the krpomo and merging back to main does not needs to be done for pre-release, it needs to be done before marking the release as main, this is what is mentioned in `RELEASE.md`, please do correct if I'm wrong, so do we still need a cron? Since we have to manually run post release after prow build is finished and then merge the PRs after 7 days.

mrueg:
> I'd push for keeping the release artifacts same as what we have currently:
> Screenshot 2025-10-22 at 5 57 30 PM
> 
> and not generate per-arch artifacts since we do not have arch-specific build tags in KSM, it runs the same everywhere:
> Screenshot 2025-10-22 at 5 58 44 PM

These artifacts are binaries, so we need to have them per arch unless I'm missing anything here.

.goreleaser.yaml
rexagod: Ah, I see. Nevermind, then!

rexagod: I'm not quite sure what you mean by setting this to the "org"? I'd reckon it'd be a bot alias (since that doesn't require a PAT)?

.goreleaser.yaml
rexagod: Right, I guess it made sense to do a kpromo promotion for pre-release as well, so users can test out the image. But since we don't do that at the moment, and no one's asked us to do that anyway, let's keep it as you mentioned.

.goreleaser.yaml
Rishab87: From what I've seen there is no direct way of changing bot alias in a release in goreleaser.
> I'm not quite sure what you mean by setting this to the "org"?

In the snippet here I've changed the owner to k8s since it was supposed to be the org name here, it is not at all related to the author's name on release which we are discussing here.

Rishab87:
@rexagod @mrueg from the discussions above, I think this is fine as it is?

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,47 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+        id: go
+      - name: Extract versions from go.mod
+        id: versions
+        run: |
+          # Extract Go version from go.mod
+          GO_VERSION=$(grep '^go ' go.mod | awk '{print $2}')
```
mrueg: This isn't necessarily the go version it is built with. You would probably need to get this from the Dockerfile base image.

Rishab87: thanks, I've changed it.

.goreleaser.yaml
rexagod: Setting the owner to `kubernetes` seems odd since technically `sig-instrumentation` owns this project, which is a SIG under `kubernetes`, so `sig-instrumentation-bit` would be a more accurate alias.

.goreleaser.yaml
rexagod: Sure, however you see fit!

rexagod:
> These artifacts are binaries, so we need to have them per arch unless I'm missing anything here.

We should be able to ship a single binary if KSM is built using Go, since we do not leverage arch-specific logic. But per-arch is fine too.

.goreleaser.yaml
Rishab87: But this is not an alias, this is the org on github under which this project comes, which is k8s in this case. It simply specifies the location of the repository where we want to create the release:
image

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,47 @@
...
+    steps:
+      - name: Checkout (full history)
+        uses: actions/checkout@v4
+        with:
+          fetch-depth: 0
+      
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
```
mrueg: This means, it's building with the version from go.mod, but not with the latest version of golang (do we want to control it here?)

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,47 @@
...
+        id: go
+      - name: Extract versions from go.mod
```
mrueg: This is not extracting from go.mod anymore

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,47 @@
...
+        id: versions
+        run: |
+          # Extract Go version from go.mod
```
mrueg: Same here :)

mrueg:
A few nitpicks, overall this looks good to me! Thanks for your work on this @Rishab87

.github/workflows/pre-release.yml
Rishab87: whoops, fixed it now

.github/workflows/pre-release.yml
Rishab87: yeah, right

Rishab87:
@mrueg thanks for the review, I've made the changes

mrueg:
Changes look good to me.
/approve

/hold 
for @rexagod to have another look

mrueg:
@Rishab87 one additional request here (or probably better in another PR): Please update the documentation in the RELEASE.md with the new steps on how to create a new release.

Rishab87:
@mrueg sure, I'll update in another PR once the whole workflow is in.

.goreleaser.yaml
rexagod: Ah, so it's different from the alias used to link the release to? That's fine then.

Rishab87: yeah

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,46 @@
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: ${{ steps.versions.outputs.go-version }}
+        id: go
+      - name: Make pre-release
+        uses: goreleaser/goreleaser-action@v6
+        with:
+          version: latest
+          args: release --clean
```
rexagod, line 42: Is this building as we discussed in https://github.com/kubernetes/kube-state-metrics/pull/2788#discussion_r2451921420?

Rishab87: It's not using the exact makefile command, let me update and test it to use the existing in house build command

Rishab87: nvm  to use pre built binaries we need goreleaser's pro version.  I don't think we explicitly need to run the command used inside the makefile, since go releaser's default builder is also ultimately running the exact same command, i.e `go build` (with exact same args).
@rexagod @mrueg

.github/workflows/pre-release.yml
```diff
@@ -0,0 +1,46 @@
...
+
+permissions:
+  contents: write
+  discussions: write
+
+jobs:
+  build-and-create-prerelease:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout (full history)
```
mrueg, line 16: fetch-depth 0 means it's checking it out without full history?

.github/workflows/pre-release.yml
mrueg: This is fine, I would like to move everything into goreleaser for build related stuff (also the Dockerfile) in a follow-up PR.

.github/workflows/pre-release.yml
Rishab87: no it means it checks out with full history. `fetch-depth: 1` is for shallow history

mrueg: Ah thanks!

.goreleaser.yaml
mrueg: Could we use `GITHUB_ACTOR` and mention it in the release note instead?

"Release coordinator: $GITHUB_ACTOR"

Rishab87: done you can see how it looks here: https://github.com/Rishab87/kube-state-metrics/releases/tag/v2.18.0

mrueg:
/lgtm

Thanks for all your work on this @Rishab87 ! I appreciate you!

mrueg:
/hold cancel

Let's get this in and refine later.

PR #2789, opened 2025-10-03
feat: Added a post-release workflow

Rishab87:
This PR adds a post release workflow, which helps in creating a PR for merging back to main & promoting images with `kpromo`. This needs to be run manually after pre release is created and images have been built on prow. (For making this workflow run automatically, one approach we can take is create a PR in test-infra repo which curls a github API when image build is completed after which this workflow is triggered triggered, open to suggestions and other possible approaches).

Test PR: https://github.com/Rishab87/kube-state-metrics/pull/16

does not changes

Related to #2756

### DIFF
--- a/.github/workflows/post-release.yml
+++ b/.github/workflows/post-release.yml
@@ -1,0 +1,92 @@
+name: Post Release
+on:
+  workflow_dispatch:
+
+permissions:
+  contents: write
+  pull-requests: write
+
+jobs:
+  promote-image:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v4
+        
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version-file: 'go.mod'
+          
+      - name: Install kpromo
+        run: go install sigs.k8s.io/promo-tools/v4/cmd/kpromo@latest
+        
+      - name: Get latest pre-release tag
+        id: extract_tag
+        run: |
+          TAG=$(gh release list --limit 1 --json tagName,isPrerelease --jq '.[] | select(.isPrerelease == true) | .tagName')
+          if [ -z "$TAG" ]; then
+            echo "Error: No pre-release found"
+            exit 1
+          fi
+          echo "tag=${TAG}" >> $GITHUB_OUTPUT
+          echo "Found pre-release tag: ${TAG}"
+        env:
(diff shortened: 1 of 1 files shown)


.github/workflows/post-release.yml
```diff
@@ -0,0 +1,92 @@
...
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        
+      - name: Create image promotion PR
+        run: |
+          kpromo pr --fork=${{ github.actor }} -i \
+            --project=kube-state-metrics \
+            --tag=${{ steps.extract_tag.outputs.tag }}
+        env:
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
rexagod, line 50: Hmm, I only see the second part of this workflow in the [test run](https://github.com/Rishab87/kube-state-metrics/actions/runs/18229006152/job/51907490769). Were you able to test this out on your local?

Rishab87: yes I tested this exact command when I was releasing `2.17.0` locally. I think running it again would create a PR in k8s.io. Should I still test it again? Just wanted to confirm here if it's not a problem if I create a PR in k8s.io for testing this out again.

rexagod: Did it create a PR earlier as well (when you were testing it)? The PR can be closed, so it's fine.

Rishab87: yes it did create a PR earlier, I was running it for promoting images for v2.17.0.

This command can't be run for v2.18.0 since its image is not built yet obviously.
```bash
FATA growing manifest with image filter [] and tag [v2.18.0]: no images survived filtering; double-check your --filter_* flag(s) for typos 
```

I tried running it for prev version (v2.17.0) here:
https://github.com/Rishab87/kube-state-metrics/actions/runs/19295972264/job/55177889064
and got:
```
level=info msg="No changes detected in the promoter images list, exiting without changes"
```

mrueg: Does the scope of the GITHUB_TOKEN work outside of its own repository for PRs?

Rishab87: No, I just checked it's repo scoped, I think then to create a PR in k8s.io we need PAT of that repo. Should I remove this step or should we include the PAT of k8s.io

PR #2807, opened 2025-11-10
perf: Reduce allocations when creating metric families

fpetkovski:

The metric family generates a series by using a `strings.Builder` to write data into, and then converts the string into a byte slice. This causes extra allocations which are not needed since we can directly write into a byte slice.

Benchmarks in CI show an improvement: https://github.com/kubernetes/kube-state-metrics/actions/runs/19236033135?pr=2807.

Has no effect.

Fixes #

### DIFF
--- a/pkg/metric/family.go
+++ b/pkg/metric/family.go
@@ -17,7 +17,7 @@ limitations under the License.
 package metric
 
 import (
-	"strings"
+	"bytes"
 )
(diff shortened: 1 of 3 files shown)


mrueg:
/hold
for further review from other maintainers.

/lgtm

Thanks for the contribution!

Issue #2682, opened 2025-06-09
Kube-state-metrics pod is getting restarted without any logs


nirmal20081028:
Kube-state-metrics pod is getting restarted without any events/logs.
Below checks are validated:
-- pod events 
-- pod logs
-- CPU and memory history.

Could you please help  what could the reasons.

helm.sh/chart = kube-state-metrics-5.27.0
pod output:
prod-kube-state-metrics-78b66b5600-gm9mv            1/1     Running   **13 (3d8h ago)**   27d

richabanker:
Hi, could you share more details about your cluster? How many instances / pods do you have or any other logs would help

richabanker:
ping @nirmal20081028

andreas-p:
I have the same issue on a 1.33 cluster, with 2.16 and 2.15. There are only endpoint deprecation and startup messages in the logs, the container exits with error code 2. So far, I haven't seen any correlation to events on the cluster (some 50 pods).

alexifsp:
I observed also some issues with the kube-state-metrics pod getting OOMKilled:

```
helm.sh/chart=kube-state-metrics-6.1.0
```

In my case I updated resources:

```
    Limits:
      cpu:                250m
      ephemeral-storage:  100Mi
      memory:             1Gi
    Requests:
      cpu:                100m
      ephemeral-storage:  50Mi
      memory:             256Mi
    Liveness:             http-get http://:8080/livez delay=5s timeout=5s period=10s #success=1 #failure=3
    Readiness:            http-get http://:8081/readyz delay=5s timeout=5s period=10s #success=1 #failure=3
    Environment:          <none>
```

@nirmal20081028 Could you describe your pod too? Just to eliminate this from possible causes?

andreas-p:
My issue is definitively not OOM, which would be error code 137 and show up in the host kern.log as well.

mrueg:
Please provide a full log output and information on your specific configuration (e.g. if you use customresourcestate), otherwise it's very difficult to help.

Also, were the /livez /healthz /metrics endpoints available before it crashed?

andreas-p:
The pods crash after hours or days of uptime, with no logging at all. The pods are generated from Helm Chart 27.28.0, with resources defined as cpu req=10m, limit=30m,  mem req=40Mi, limit=80Mi. I've never observed resource problems (currently top pod: cpu=1m, mem=53Mi)

assu-2000:
[...]

PR #2808, opened 2025-11-13
feat: warn on probe failing

mickael-carl:

We have a fairly common occurence of KSM getting restarted due to its liveness probe failing. This is seemingly caused by rolling deploys of the backing API server, but we don't know exactly why. This should make it slightly more visible what kind of response the API server when a probe fails, which in turn could help either tweak probes to weather such situations where the API server may be unavailable or overwhelmed (it's entirely possible it's returning KSM a bunch of 429s) or if we have a larger problem at hand.

Does not affect cardinality.

Potentially related to #2682.

### DIFF
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -515,6 +515,9 @@ func handleClusterDelegationForProber(client kubernetes.Interface, probeType str
 	return func(w http.ResponseWriter, _ *http.Request) {
 		got := client.CoreV1().RESTClient().Get().AbsPath(probeType).Do(context.Background())
 		if got.Error() != nil {
+			var statusCode int
+			got.StatusCode(&statusCode)
+			klog.Warningf("Failed to contact API server for %s: got %d", probeType, statusCode)
 			w.WriteHeader(http.StatusServiceUnavailable)
 			w.Write([]byte(http.StatusText(http.StatusServiceUnavailable)))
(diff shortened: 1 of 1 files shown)


pkg/app/server.go
```diff
@@ -513,8 +513,10 @@ func buildTelemetryServer(registry prometheus.Gatherer, authFilter bool, kubeCon
 
 func handleClusterDelegationForProber(client kubernetes.Interface, probeType string) http.HandlerFunc {
 	return func(w http.ResponseWriter, _ *http.Request) {
-		got := client.CoreV1().RESTClient().Get().AbsPath(probeType).Do(context.Background())
+		var statusCode int
+		got := client.CoreV1().RESTClient().Get().AbsPath(probeType).Do(context.Background()).StatusCode(&statusCode)
 		if got.Error() != nil {
+			klog.Warningf("Failed to contact API server for %s: got %d", probeType, statusCode)
```
mrueg, line 520: Can we do this instead? We would only care about it in case of an error.
```suggestion
            var statusCode int
            got.StatusCode(&statusCode)
			klog.Warningf("Failed to contact API server for %s: got %d", probeType, statusCode)
```

mickael-carl: Yeah makes complete sense, thanks! Done!

PR #2810, opened 2025-11-17
feat:  #2501 Count volumes attached to a node

### LINKED ISSUE #2501: Count volumes attached to a node

LiuQhahah:

Count volumes attached to a node

increaser two metrics: kube_node_volumes_attached_count and kube_node_volumes_in_use_count

Fixes #2501


CatherineF-dev:
Would you mind sharing the use case of this?

thommeo:
> Would you mind sharing the use case of this?

TLDR: The number of volumes that can be attached to a Node is hard-limited for different hosting providers and thus should be monitored and alerted when approaching the limit.

https://github.com/kubernetes/kube-state-metrics/issues/2501

internal/store/node.go
```diff
@@ -506,6 +508,44 @@ func createNodeStatusConditionFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+func createNodeVolumeCountGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGeneratorWithStability(
+		"kube_node_volumes_attached",
+		"Number of volumes attached to the node",
+		metric.Gauge,
+		basemetrics.STABLE,
```
skl, line 516: New metrics should be marked as experimental/alpha first, before graduated to stable at a later date. This applies to both new metrics:

```suggestion
		basemetrics.ALPHA,
```

Reference:

- https://github.com/kubernetes/kube-state-metrics/blob/main/docs/developer/guide.md#add-new-metrics

internal/store/node.go
```diff
@@ -506,6 +508,44 @@ func createNodeStatusConditionFamilyGenerator() generator.FamilyGenerator {
 	)
 }
 
+func createNodeVolumeCountGenerator() generator.FamilyGenerator {
```
skl, line 511: Should stay in line with the metric name in both cases:

```suggestion
func createNodeVolumesAttachedGenerator() generator.FamilyGenerator {
```

internal/store/node.go
```diff
@@ -506,6 +508,44 @@ func createNodeStatusConditionFamilyGenerator() generator.FamilyGenerator {
...
+					{
+						Value: float64(len(n.Status.VolumesAttached)),
+					},
+				},
+			}
+		}),
+	)
+}
+
+func createNodeVolumeInUseGenerator() generator.FamilyGenerator {
```
skl, line 530:
```suggestion
func createNodeVolumesInUseGenerator() generator.FamilyGenerator {
```

docs/metrics/cluster/node-metrics.md
```diff
@@ -14,3 +14,5 @@
...
+| kube_node_volumes_in_use | Gauge       | Number of volumes in use on the node                                                                                                |                                                                                                                                                                                   | `node`=&lt;node-address&gt;                                                                                                                                                                                                                                                                                                                                                                                                                               | STABLE |
```
skl, line 17: nit: suggest `|` alignment, they're floating in random places atm

skl (commented):
Seems like it would be useful to have this in KSM 👍

PR #2817, opened 2025-11-26
feat(nestedArray): parsing * on array of array

### LINKED ISSUE #2368: Parse Nested Arrays does not work
Have a custom CRD returning `conditions` under a nested structure like 

```yaml
status:
  parents:
  - conditions:
    - lastTransitionTime: "2024-04-05T16:27:15Z"
      message: References resolved
      observedGeneration: 1
      reason: ResolvedRefs
      status: "True"
      type: ResolvedRefs
    - lastTransitionTime: "2024-04-05T16:27:15Z"
      message: Accepted HTTPRoute
      observedGeneration: 1
      reason: Accepted
      status: "True"
      type: Accepted
    controllerName: projectcontour.io/gateway-controller
    parentRef:
      group: gateway.networking.k8s.io
      kind: Gateway
      name: contour
      namespace: projectcontour
      sectionName: https
```

Using `StateSet` to parse with path `['status', 'parents', 'conditions']` but getting an error like

```
registry_factory.go:685] "gatewayapi_grpcroute_status" err="[status,parents,conditions]: expected value for path to be string, got <nil>" 
```

Is this supported?

raf262:
**What this PR does / why we need it:** Enable parsing nested arrays.

We can now write this kind of path: `"status", "parents", "[*]", "conditions"`. 

Where `[*]`means listing all the parents object inside the list.

I updated the PR on this day (Jan 16, 2026) to allow the use of `[*]` at the `labelsFromPath` & `valueFrom`. This has been tested on our side and enable us to recover the `parentRef` name and the `condition` reason for each `condition` on `HTTPRoute->Status->Parents`

Fixes #2368

### DIFF
--- a/pkg/customresourcestate/example_config.yaml
+++ b/pkg/customresourcestate/example_config.yaml
@@ -47,6 +47,24 @@ spec:
               path: [status, other]
           errorLogV: 5
 
+        - name: "other_count_array_deep"
+          each:
+            type: Gauge
+            gauge:
+              path: [ status, other, '[*]', '[detail]' ]
+          errorLogV: 5
+
+        - name: "other_count_array_deep_multiple"
+          each:
+            type: Gauge
+            gauge:
+              path: [ status, other, '[*]' ]
+              labelsFromPath:
+                bar: [ detail ]
+                foo: [ foo, bar ]
+              valueFrom: [ count ]
+          errorLogV: 5
+
         - name: "info"
           each:
             type: Info
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -279,6 +279,14 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
 		}
 	case []interface{}:
 		for i, it := range iter {
+			if c.ValueFrom.hasWildcard() {
+				evs, fanErrs := c.fanout(it)
+				for _, e := range fanErrs {
+					onError(fmt.Errorf("[%d]: %w", i, e))
+				}
+				result = append(result, evs...)
+				continue
+			}
 			value, err := c.value(it)
 			if err != nil {
 				onError(fmt.Errorf("[%d]: %w", i, err))
@@ -444,6 +452,48 @@ func less(a, b map[string]string) bool {
 	return len(aKeys) < len(bKeys)
 }
 
+// fanout expands a single iteration context `it` into N eachValues by walking
+// `ValueFrom` and any `[*]`-containing label paths in parallel. Label paths
+// without `[*]` are broadcast as constants across all rows. Driven by the
+// length of `ValueFrom.GetAll(it)`.
+func (c *compiledGauge) fanout(it interface{}) ([]eachValue, []error) {
+	valueVals := c.ValueFrom.GetAll(it)
+	labelVals := make(map[string][]interface{}, len(c.LabelFromPath()))
+	for k, lp := range c.LabelFromPath() {
+		labelVals[k] = lp.GetAll(it)
+	}
+	var (
+		res  []eachValue
+		errs []error
+	)
+	for i, raw := range valueVals {
+		if raw == nil && !c.NilIsZero {
+			continue
+		}
+		f, err := toFloat64(raw, c.NilIsZero)
+		if err != nil {
+			errs = append(errs, fmt.Errorf("%s: %w", c.ValueFrom, err))
+			continue
+		}
+		labels := map[string]string{}
+		for k, lvs := range labelVals {
+			var v interface{}
+			switch {
+			case i < len(lvs):
+				v = lvs[i]
+			case len(lvs) > 0:
+				v = lvs[0]
+			}
+			if v == nil {
+				continue
+			}
+			labels[store.SanitizeLabelName(k)] = fmt.Sprintf("%v", v)
+		}
+		res = append(res, eachValue{Value: f, Labels: labels})
+	}
+	return res, errs
+}
+
 func (c compiledGauge) value(it interface{}) (*eachValue, error) {
 	labels := make(map[string]string)
 	got := c.ValueFrom.Get(it)
@@ -566,6 +616,45 @@ func (p valuePath) Get(obj interface{}) interface{} {
 	return obj
 }
 
+// hasWildcard reports whether the path contains a `[*]` fan-out marker.
+func (p valuePath) hasWildcard() bool {
+	for _, op := range p {
+		if op.part == "[*]" {
+			return true
+		}
+	}
+	return false
+}
+
+// GetAll walks the path like Get, but fans the walk out across `[*]` markers
+// when the current node is a []interface{}. For paths without `[*]`, it
+// returns a single-element slice containing what Get would return, so callers
+// can treat fan-out and single-value paths uniformly.
+func (p valuePath) GetAll(obj interface{}) []interface{} {
+	return getAll(p, obj)
+}
+
+func getAll(p valuePath, obj interface{}) []interface{} {
+	for i, op := range p {
+		if obj == nil {
+			return []interface{}{nil}
+		}
+		if op.part == "[*]" {
+			arr, ok := obj.([]interface{})
+			if !ok {
+				return []interface{}{nil}
+			}
+			var out []interface{}
+			for _, el := range arr {
+				out = append(out, getAll(p[i+1:], el)...)
+			}
+			return out
+		}
+		obj = op.op(obj)
+	}
+	return []interface{}{obj}
+}
+
 func (p valuePath) String() string {
 	var b strings.Builder
 	b.WriteRune('[')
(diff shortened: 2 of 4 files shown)


pkg/customresourcestate/registry_factory.go
```diff
@@ -279,16 +279,34 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+						onError(fmt.Errorf("[%d]: %w", j, err))
+						continue
+					}
+					if value == nil {
+						continue
+					}
+					addPathLabels(nestedIt, c.LabelFromPath(), value.Labels)
+					result = append(result, *value)
+				}
+			} else {
```
yannchabed: [nitpick](https://conventionalcomments.org/#labels): adopt early return (avoid else statements)

yannchabed (approved):

raf262:
Hi everyone, I don't know if this PR is missing something, please tell me. 
If it needs reword I'm available !

@mrueg maybe you have some times to review it ? 

Thanks guys

dademo (approved):

raf262:
Hi @rexagod, how could I help you (if I could) on this PR to be merged ?

pkg/customresourcestate/registry_factory.go
```diff
@@ -641,21 +713,32 @@ func compilePath(path []string) (out valuePath, _ error) {
...
+							}
+							if i < 0 || i >= len(s) {
+								return err2
+							}
+							return s[i]
 						}
-						if i < 0 || i >= len(s) {
-							return fmt.Errorf("list index out of range: %s", part)
+
+						// case we have a list and the part is an index
```
bhope: This seems like a potential breaking change. Previously applying a plain string key to a []interface{} returned nil (metric skipped). With your logic, it iterates the array and collects matching fields, so an existing path like that was a no-op before would silently start producing metrics. Any way we could avoid these unintended regressions?

pkg/customresourcestate/registry_factory.go
```diff
@@ -279,6 +279,42 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
 	case []interface{}:
 		for i, it := range iter {
+			// Check if it is a [][]interface{} if yes loop over it
+			// Else process normally
+			if nestedIter, ok := it.([]interface{}); ok {
+				for j, nestedIt := range nestedIter {
+					value, err := c.value(nestedIt)
+
+					if err != nil {
+						onError(fmt.Errorf("[%d]: %w", j, err))
```
bhope: It will be useful to emit both - i,j. Nested loops are tricky as is to debug.

bhope:
Thanks for the PR! Added a couple inlines. Also customresourcestate-metrics.md needs to be updated mentioning that [*] is a valid path segment after this change.

pkg/customresourcestate/registry_factory.go
raf262: Yes, I added bracket to avoid the breaking change, thanks !

pkg/customresourcestate/registry_factory.go
raf262: Done

bhope:
This looks in a good shape now. Thanks! 
@raf262  you may wanna fix the lint CI failures

/lgtm

raf262:
Ok seems good now, could we integrate this in the next release maybe ?

bhope:
/lgtm

@raf262 Can you do the quick fix for the example edit as suggested [here](https://github.com/kubernetes/kube-state-metrics/pull/2817#discussion_r3088071052). Also, please squash the commits.

pkg/customresourcestate/example_config.yaml
```diff
@@ -47,6 +47,25 @@ spec:
...
+
+        - name: "other_count_array_deep_multiple"
+          each:
+            type: Gauge
+            gauge:
+              path: [ status ]
+              labelsFromPath:
+                bar: [ other, '[*]', '[detail]' ]
+                foo: [ foo, bar ]
+              valueFrom: [ other, '[*]', '[count]' ]
```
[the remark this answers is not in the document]
mrueg: @raf262 what do you think about this suggestion?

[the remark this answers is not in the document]
raf262: Seems okay while in my mind we had `foo` in `status->foo`. That's another example, lets commit that change and people will play with their config

raf262:
Just fixes what copilot was suggesting and squashed the commits ✅

raf262:
👋  Hey guys, seems quite good now. Do you think we can merge this ? Thanks !

pkg/customresourcestate/registry_factory.go
```diff
@@ -515,43 +586,194 @@ func (f compiledFamily) BaseLabels(obj map[string]interface{}) map[string]string
...
+		if value == nil {
+			continue
+		}
+		result[store.SanitizeLabelName(e.key)] = fmt.Sprintf("%v", value)
+	}
+}
+
+// addPathLabels writes labels resolved from obj into result.
+func addPathLabels(obj interface{}, labels map[string]valuePath, result map[string]string) {
+	addPathLabelsByIndex(obj, labels, result, -1)
```
bhope: `addPathLabelsByIndex` is never called with a positive sliceIndex. Could we collapse these two back into one function and drop the parameter?

pkg/customresourcestate/registry_factory.go
```diff
@@ -515,43 +586,194 @@ func (f compiledFamily) BaseLabels(obj map[string]interface{}) map[string]string
...
+func isStarLabel(k string) bool {
+	return strings.HasPrefix(k, "*") || strings.HasSuffix(k, "*")
+}
+
+// resolveLabelValue resolves p against obj. When sliceIndex >= 0 and the
+// result is an array, it selects the element at that index — the caller is
+// iterating a parallel collection and wants values aligned by index. Returns
+// (value, true) when a value is available, or (_, false) when the caller
+// should skip this label (slice shorter than sliceIndex).
+func resolveLabelValue(p valuePath, obj interface{}, sliceIndex int) (interface{}, bool) {
```
bhope: Is `resolveLabelValue` a leftover from before resolveLabelsOnce/applyAt took over?  Seems like its slice indexing logic is never reached since `addPathLabelsByIndex` is never called for positive sliceIndex (as I called out on ln 712).

bhope:
Thanks for your patience working through the review feedback. A few small things to address. Left a couple inlines and following tests need to be added:
1. tests to cover `compileBracketOp` with an empty bracket []
2. test for [*] as a no-op on a flat array
3. test for applyAt's sliceIndex >= len(e.slice) condition

pkg/customresourcestate/registry_factory.go
```diff
@@ -278,18 +278,73 @@ func (c *compiledGauge) Values(v interface{}) (result []eachValue, errs []error)
...
+			// on each element. Emit one metric per index; labels are drawn
+			// from any parallel collections aligned by that same index.
+			if valueFromExpands {
+				if collected, ok := c.ValueFrom.Get(it).([]interface{}); ok {
+					// Resolve each label path against `it` once; per-idx work
+					// becomes a cheap index into the cached slice (or reuse of
+					// the cached scalar), avoiding redundant path traversals.
+					rl := resolveLabelsOnce(it, c.LabelFromPath())
+					for idx, v := range collected {
+						val, emit, err := c.resolveScalarValue(v)
```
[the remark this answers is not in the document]
bhope: This one is already handled. The compile-time guard in newCompiledMetric rejects this case.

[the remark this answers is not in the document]
raf262: Yeah I got lost trying to resolve all the comments from copilot

bhope:
Hi @raf262 - Just checking in if you're still planning to work on this one?

raf262:
> Hi @raf262 - Just checking in if you're still planning to work on this one?

Yes, it's still required for our company. But it seems that peoples here aren't very interested in this feature 😢

rexagod:
/assign @bhope

@raf262 You can look at https://github.com/kubernetes-sigs/resource-state-metrics which is only bounded by the functionality DSLs offer themselves. This is currently in BETA and inching towards GA soon, upon which CRS will be dropped from KSM entirely, in its favor.

pkg/customresourcestate/registry_factory.go
```diff
@@ -445,6 +452,48 @@ func less(a, b map[string]string) bool {
...
+			continue
+		}
+		labels := map[string]string{}
+		for k, lvs := range labelVals {
+			var v interface{}
+			switch {
+			case i < len(lvs):
+				v = lvs[i]
+			case len(lvs) > 0:
+				v = lvs[0]
```
bhope, line 485: fanout falls back to lvs[0] instead of erroring. Can we please make this error and not a fall back?

bhope (changes requested):
Hi @raf262 sorry for the gap, the implementation changed significantly since my last review. 

I left an inline comment. 

Can we also quantify the cardinality with this change and add the details in the cardinality section of the PR? It will be important for users to notify about this this feature can multiply cardinality across nested arrays.

Also, please check the CI failures.

bhope:
Hi @raf262 - we just merged #3050 which formally marks custom-resource-state as feature-frozen ahead of deprecation in favor of resource-state-metrics. Since this PR adds new parsing capability to the CRS DSL specifically, it falls into what's now frozen, so we won't be able to take it as-is.

Closing this one out - if you're still interested in this functionality, [resource-state-metrics](https://github.com/kubernetes-sigs/resource-state-metrics) is where new work in this space should go. Sorry about this - we really appreciate all the efforts you have put in and thanks again for surfacing the underlying gap (issue #2368).

FYI @mrueg @rexagod

PR #2823, opened 2025-12-02
feat: Add access mode metric for PersistentVolumes

viragvoros:
This PR adds a `kube_persistentvolume_access_mode` metric that exposes the access modes of `PersistentVolume` objects, in a way that mirrors the existing PVC access mode metric.

At the moment, if we want to filter or aggregate PersistentVolumes by access mode (e.g. `ReadWriteOnce`, `ReadWriteMany`, `ReadOnlyMany`), we have to rely on `kube_persistentvolumeclaim_access_mode` and join through
PVC metrics. That forces us to scrape the PVC access mode metrics even when we only care about PVs, and increases metric cardinality.

By exposing a dedicated PV access mode metric with a small label set, we can:
- filter and aggregate PVs directly by access mode
- avoid scraping PVC access mode metrics in setups where they are not
  otherwise needed,
while still keeping metric cardinality low.

**Details of the change**
- Add `kube_persistentvolume_access_mode` metric in the PV store:
   - creates one sample for each access mode a PersistentVolume lists
   - `access_mode` label uses the standard Kubernetes access
    modes (`ReadWriteOnce`, `ReadOnlyMany`, `ReadWriteMany`, etc.)
   - metric follows a similar style as the existing PVC access mode metric
-  Update `internal/store/persistentvolume_test.go`
   - add test coverage for the new PV access mode metric
- Update PV metrics documentation

Previously, dashboards and alerts that needed PV access mode information had to join against PVC access mode metrics and therefore scrape `kube_persistentvolumeclaim_access_mode`, which may have higher cardinality in large clusters.
With `kube_persistentvolume_access_mode`, users can query PV access modes directly from the PV metrics.


mrueg:
I wonder if it's worth adding a new metric here or enhancing kube_persistentvolume_volume_mode instead?

viragvoros:
I considered merging `access_mode` into the existing `kube_persistentvolume_volume_mode` metric, but I think keeping them separate is the better choice.
`volumeMode` and `accessMode` represent different shapes in the Kubernetes API. A `PersistentVolume` has exactly one `volumeMode`, while it may have multiple `accessModes`. Combining them would require emitting one sample per access mode, which means the volume mode metric would no longer represent a single property of the PV.

There is also a cardinality consideration for metric consumers. If only a combined `{volumemode, access_mode}` metric existed, users who are interested only in access modes would still have to deal with the extra volume mode label. This increases the number of series they need to aggregate over and makes otherwise simple queries (e.g. counting PVs by access mode) more expensive than using a dedicated access mode metric.

Users who need both dimensions can still correlate them via PV labels.

docs/metrics/storage/persistentvolume-metrics.md
```diff
@@ -12,6 +12,7 @@
...
+| kube_persistentvolume_access_mode | Gauge | Access modes of the PersistentVolume. Only the four official access modes are emitted.           |                         | `persistentvolume`=&lt;persistentvolume-name&gt; <br>`access_mode`=<ReadWriteOnce \| ReadOnlyMany \| ReadWriteMany \| ReadWriteOncePod> | STABLE       |
```
mrueg: new metrics should be marked as experimental and not stable

viragvoros:
Hi @mrueg, thank you for the review. I updated the new metric to match the guidance for newly introduced metrics. Also fixed the docs formatting and removed the sentence that said only the four official access modes are emitted, as the implementation exposes the values in `spec.accessModes`.

mrueg:
/remove lifecycle-rotten

/lgtm

sorry, it took a while to get back to this. Thanks for your contribution @viragvoros !

PR #2836, opened 2025-12-28
fix(horizontalpodautoscaler): Added container label to ContainerResource metrics 

### LINKED ISSUE #2403: Duplicate kube_horizontalpodautoscaler_spec_target_metric

bxrne:

Fixes duplicate `kube_horizontalpodautoscaler_spec_target_metric` and `kube_horizontalpodautoscaler_status_target_metric` metrics when multiple containers use the same resource (e.g., CPU) in a HorizontalPodAutoscaler. I also added a unit test case to verify this fix.

*BREAKING CHANGE*: Adding a new label to the stable
kube_horizontalpodautoscaler_status_target_metric

Previously, when an HPA specified multiple ContainerResource metrics with the same resource name but different containers, kube-state-metrics would generate metrics with identical label sets. This caused Prometheus 2.52.0+ to log duplicate metric errors and drop samples:

```
msg="Error on ingesting samples with different value but same timestamp" num_dropped=12
```

This PR adds a `container` label to ContainerResource metrics, making them distinguishable:
- Before: `kube_horizontalpodautoscaler_spec_target_metric{metric_name="cpu",metric_target_type="utilization"}`
- After: `kube_horizontalpodautoscaler_spec_target_metric{metric_name="cpu",metric_target_type="utilization",container="main"}`

For example, an HPA with three containers (main, sidecar1, sidecar2) all scaling on CPU now produces three distinct metrics instead of duplicates.

**How does this change affect the cardinality of KSM:** increases

Cardinality increases slightly for HPAs using ContainerResource metrics. Each container adds one additional metric cardinality dimension. This is the intended behavior to distinguish per-container resource targets.

Fixes #2403


bxrne:
@CatherineF-dev  @mrueg should i mark this PR as stale or backlog, i understand its priority is low?

ben-bertrands-hs:
Hi @bxrne,
Is there a technical reason for closing this pr?

bxrne:
> Hi @bxrne,
> Is there a technical reason for closing this pr?
>  

No. Just hasn't had any activity. 
Happy to reopen.

ben-bertrands-hs:
> Happy to reopen.

That would be awesome. Thanks

internal/store/horizontalpodautoscaler.go
```diff
@@ -50,7 +50,8 @@ var (
 	descHorizontalPodAutoscalerLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descHorizontalPodAutoscalerLabelsDefaultLabels = []string{"namespace", "horizontalpodautoscaler"}
 
-	targetMetricLabels = []string{"metric_name", "metric_target_type"}
+	targetMetricLabels          = []string{"metric_name", "metric_target_type"}
+	containerMetricTargetLabels = []string{"metric_name", "metric_target_type", "container"}
```
bhope: Having a separate label set for `ContainerResource` vs all other types means time series in the same metric family will have inconsistent label schemas --  some with container, some without. This makes it really hard to query across types because the label schemas don't match up.

For non-`ContainerResource` types, lets consider emitting `container=""` instead of omitting the label entirely. That way every metric in the family has the same set of labels.

bhope (changes requested):

internal/store/horizontalpodautoscaler.go
```diff
@@ -194,25 +195,37 @@ func createHPASpecTargetMetric() generator.FamilyGenerator {
...
 				case autoscaling.ResourceMetricSourceType:
 					metricName = string(m.Resource.Name)
 					metricTarget = m.Resource.Target
+					labelKeys = targetMetricLabels
+					labelValues = []string{metricName, ""}
 				case autoscaling.ContainerResourceMetricSourceType:
 					metricName = string(m.ContainerResource.Name)
 					metricTarget = m.ContainerResource.Target
+					labelKeys = containerMetricTargetLabels
+					labelValues = []string{metricName, "", m.ContainerResource.Container}
```
bhope: `kube_horizontalpodautoscaler_status_target_metric` is stable so adding a new label is a breaking change and should be called out as such in the PR description with a changelog. The old behavior was already broken so the fix is needed -- but users will still have to update their dashboards and/or alerts.

internal/store/horizontalpodautoscaler.go
```diff
@@ -253,25 +271,37 @@ func createHPAStatusTargetMetric() generator.FamilyGenerator {
...
 				case autoscaling.ResourceMetricSourceType:
 					metricName = string(m.Resource.Name)
 					currentMetric = m.Resource.Current
+					labelKeys = targetMetricLabels
+					labelValues = []string{metricName, ""}
 				case autoscaling.ContainerResourceMetricSourceType:
 					metricName = string(m.ContainerResource.Name)
 					currentMetric = m.ContainerResource.Current
+					labelKeys = containerMetricTargetLabels
+					labelValues = []string{metricName, "", m.ContainerResource.Container}
```
bhope: Same comment as R223

bhope (changes requested):

bhope:
Thanks for the PR! The fix makes sense and the motivation is solid. Prometheus 2.52+ silently drops these samples -- so users with ContainerResource HPAs are losing this metric entirely. 

I left a few inline comments. We need to be cautious as `kube_horizontalpodautoscaler_status_target_metric` is a stable metric now.

bhope:
Also, please update docs/metrics/workload/horizontalpodautoscaler-metrics.md to reflect container as new label in the existing label list.

bhope (approved):

PR #2844, opened 2026-01-12
feat: Resource query via parameter

mrueg:
Allows to add parameters to the metrics endpoint, to select specific resources.

Fixes #

### DIFF
--- a/README.md.tpl
+++ b/README.md.tpl
@@ -44,15 +44,15 @@ are deleted they are no longer visible on the `/metrics` endpoint.
   * [ECMAScript regular expression support for allow and deny lists](#ecmascript-regular-expression-support-for-allow-and-deny-lists)
   * [Conflict resolution in label names](#conflict-resolution-in-label-names)
 * [Kube-state-metrics self metrics](#kube-state-metrics-self-metrics)
-* [Resource recommendation](#resource-recommendation)
-* [Latency](#latency)
-* [A note on costing](#a-note-on-costing)
 * [kube-state-metrics vs. metrics-server](#kube-state-metrics-vs-metrics-server)
 * [Scaling kube-state-metrics](#scaling-kube-state-metrics)
   * [Resource recommendation](#resource-recommendation)
+  * [Latency](#latency)
+  * [A note on costing](#a-note-on-costing)
   * [Horizontal sharding](#horizontal-sharding)
     * [Automated sharding](#automated-sharding)
   * [Daemonset sharding for pod metrics](#daemonset-sharding-for-pod-metrics)
+  * [Resource filtering](#resource-filtering)
 * [Setup](#setup)
   * [Building the Docker container](#building-the-docker-container)
 * [Usage](#usage)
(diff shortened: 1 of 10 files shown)


README.md.tpl
```diff
@@ -305,6 +306,22 @@ spec:
...
+`curl 'http://localhost:8080/metrics?resources=pods,secrets'`
+
+Multiple resources can be specified as a comma-separated list, or by providing the `resources` parameter multiple times.
+
+You can also exclude specific resources using the `exclude_resources` query parameter. This is useful if you want to scrape all metrics except for a few specific ones.
+
+Example:
+`curl 'http://localhost:8080/metrics?exclude_resources=pods'`
+
+If both `resources` and `exclude_resources` are provided, the `resources` parameter acts as an allowlist, and `exclude_resources` acts as a denylist, filtering out any resources specified in the `exclude_resources` parameter from the allowed resources.
```
rexagod, line 322: May be worth specifying that the behavior is similar to the existing allow-deny list flags, and the fact that the deny list takes precedence.

rexagod: Also the fact that this builds on top of the (already) filtered resources based on the allow-deny list flags, if any, and doesn't override them in any manner.

pkg/metricshandler/metrics_handler.go
```diff
@@ -221,7 +221,40 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	}
 
 	m.metricsWriters = metricsstore.SanitizeHeaders(contentType, m.metricsWriters)
+
+	resources := r.URL.Query()["resources"]
```
rexagod: (nit) Could we use `Get("resources")` here and below (for "exclude_resources"), ~and add a testcase for irrelevant params like `foo` above, just as a sanity test~ (I see there's already a case for non-existent resources)?

pkg/metricshandler/metrics_handler.go
```diff
@@ -221,7 +221,40 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+		excludedResources = make(map[string]struct{})
+		for _, r := range excludedResourcesList {
+			for _, res := range strings.Split(r, ",") {
+				excludedResources[res] = struct{}{}
+			}
+		}
+	}
+
 	for _, w := range m.metricsWriters {
+		if requestedResources != nil {
```
rexagod, line 229:
```suggestion
        // If the resource parameter is not empty, any missing resource that was originally permitted will be filtered out.
		if requestedResources != nil {
```

pkg/app/server_filter_test.go
```diff
@@ -0,0 +1,178 @@
...
+	))
+	builder.WithAllowLabels(map[string][]string{})
+
+	handler := metricshandler.New(&options.Options{}, kubeClient, builder, false)
+	handler.ConfigureSharding(ctx, 0, 1)
+
+	// Wait for caches to fill
+	time.Sleep(time.Second)
+
+	tests := []struct {
```
rexagod, line 89: Can we add cases for `?resources=,,,`, `?resources=pods,,,services`, etc.? Additionally, can we add all cases covered in `?resources` for `?excluded_resources` too (comma separated, multiple params, nonexistent, multiple commas)?

README.md.tpl
mrueg: Good call, I've extend the readme a bit.

pkg/metricshandler/metrics_handler.go
mrueg: How would you use it here? Isn't Get() only supporting single value?

pkg/app/server_filter_test.go
mrueg: I've added more tests

pkg/metricshandler/metrics_handler.go
rexagod: ~I could be seeing it wrong but I suppose `r.URL.Query().Get("resources")`? Since we'll only ever supply either `resources` or `exclude_resources` here? But at this point this is more of a nitpick, so LMK if that's unnecessary and I'll merge this PR, as it looks good to go.~

rexagod: oof, dunno what I was thinking, sorry

PR #2846, opened 2026-01-15
feat: add metric about tls cert expiration in secrets

sylr:

Add a `kube_secret_tls_cert_not_after_seconds` to get certificate expiration dates.

Increase cardinality.


Serializator:
**The below is my opinion and a description of my concerns from a non-maintainer's perspective. There is no ill-intent and I respect the time and effort it takes to contribute. Please read this with a few 😄 in between!**

---

KSM may not be the right place to monitor details of a TLS certificate. I'd argue the **data** inside of a secret is not directly pertinent to the state of a Kubernetes cluster itself.

The inability to report problems with decoding the certificate may also add ambiguity when troubleshooting problems where the metric does not appear for a given secret containing a TLS certificate.

Kubernetes offers sugarcoating by means of the TLS secret type to more easily manage TLS certificates as secrets. However, the Kubernetes API itself does not directly interpret or subsequently expose  any details about the TLS certificate.

> The TLS Secret type is provided only for convenience. You can create an Opaque type for credentials used for TLS authentication. However, using the defined and public Secret type (kubernetes.io/tls) helps ensure the consistency of Secret format in your project. The API server verifies if the required keys are set for a Secret of this type.
> 
> _**Source -** https://kubernetes.io/docs/concepts/configuration/secret/#tls-secrets_

Decoding of (semi-)unknown data should not be KSM's responsibility.

mrueg:
I would agree with @Serializator that this an interesting idea but most likely not the right place in ksm and I would also from a ksm perspective avoid reading the content of a secret. 

Maybe https://github.com/kubernetes-sigs/resource-state-metrics/ could be a good option to implement such a metric?

sylr:
Closing because https://github.com/enix/x509-certificate-exporter does this already.

PR #2849, opened 2026-01-18
feat: add new kube_service_spec_ports metric

### LINKED ISSUE #2819: Missing port information in service metrics
It would be great to have a service metric that exposes port, similar to "kube_endpoint_ports" or "kube_endpointslice_ports" metrics.

**Why is this needed:**
Current service metrics are not accurately reflecting one of the most important aspects of service.

Add additional metric like "kube_service_ports".

**Additional context:**
n/a

Serializator:
Add new metric `kube_service_spec_ports` to represent ports on the service resource.

The `kube_service_spec_`  prefix matches with the naming convention for the rest of the metrics.

To stay aligned with other metrics, `.spec.ports.appProtocol` is not included as a label. The same is true for the `kube_endpoint_address` metric.

`.spec.ports.targetPort` is not included as a label due to its inconsistent value. It can be a named port or port number.

Increases cardinality with a new time series.

Fixes #2819


Serializator:
Hey @dgrisonnet & @mrueg! Would one of you have time to approve the workflow so tests can run? Thank you in advance!

internal/store/service.go
```diff
@@ -203,6 +204,42 @@ func serviceMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				}
 			}),
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_service_spec_ports",
+			"Service ports. One series for each port",
+			metric.Gauge,
+			basemetrics.STABLE,
```
bhope: I believe this should be `basemetrics.ALPHA`. We can only promote a metric to alpha->beta->stable after it has gone through multiple releases without any issues. Happy to be corrected if I am missing any context.

internal/store/service.go
```diff
@@ -203,6 +204,42 @@ func serviceMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
+					protocol := v1.ProtocolTCP
+					if port.Protocol != "" {
+						protocol = port.Protocol
+					}
+
+					keys := []string{"port_name", "port_protocol", "port_number"}
+					values := []string{port.Name, string(protocol), strconv.FormatInt(int64(port.Port), 10)}
+
+					// 0 means no node port is set. It is also a reserved port (RFC-6335) so there shouldn't be a problem.
+					if port.NodePort > 0 {
```
bhope, line 225: `node_port_number` should always be present per the project's best practices. Use "" when NodePort == 0 instead of conditionally adding the label.
Check the best practices here: https://github.com/kubernetes/kube-state-metrics/blob/main/docs/design/metrics-best-practices.md#optional-properties

bhope (changes requested):
Thanks for the PR. A few comments:
1. Stability: new metrics should start at basemetrics.ALPHA, not basemetrics.STABLE (see inline)
2. `node_port_number`: should always be present as a label, using "" when no NodePort is set (see inline)

internal/store/service.go
Serializator: Thank you for the review! I updated it to be `basemetrics.ALPHA`

internal/store/service.go
Serializator: Thank you! I've updated it to be an empty string when the node port is not set.

internal/store/service.go
```diff
@@ -203,6 +204,41 @@ func serviceMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
+					if port.Protocol != "" {
+						protocol = port.Protocol
+					}
+
+					nodePortNumber := ""	
+					if port.NodePort > 0 {
+						nodePortNumber = strconv.FormatInt(int64(port.NodePort), 10)
+					}
+
+					keys := []string{"port_name", "port_protocol", "port_number", "node_port_number"}
```
bhope: nit: probably should sit outside for loop

docs/metrics/service/service-metrics.md
```diff
@@ -8,5 +8,6 @@
...
 | kube_service_spec_external_ip             | Gauge       | Service external ips. One series for each ip                                                                                            |                         | `service`=&lt;service-name&gt; <br> `namespace`=&lt;service-namespace&gt; <br> `uid`=&lt;service-uid&gt; <br> `external_ip`=&lt;external-ip&gt;                                                                                                                                                                                          | STABLE       |
+| kube_service_spec_ports                   | Gauge       | Service port. One series for each port                                                                                                  |                         | `service`=&lt;service-name&gt; <br> `namespace`=&lt;service-namespace&gt; <br> `uid`=&lt;service-uid&gt; <br> `port_number`=&lt;port-number&gt; <br> `node_port_number`=&lt;node-port-number&gt;                                                                                                                                         | EXPERIMENTAL |
```
bhope: Can you verify this? Your code emits port_name and port_protocol too

internal/store/service.go
Serializator: Good point 😉

docs/metrics/service/service-metrics.md
Serializator: You're correct. Good catch! I added `port_name` and `port_protocol` to the documentation.

bhope:
Overall its in a good shape now. Can you squash the commits?

Serializator:
@bhope squashed the commits and also resolved a linting problem.

bhope:
Thanks 

/lgtm

mrueg:
Can we quantify the number of timeseries it generates per port, service, etc.?

bhope:
Hi @Serializator - Can you address @mrueg 's comment?
I ran a quick check on my end and it seems with this change the metric count would be about 2 times the per service metric count we have today (the code generates 1 TS per port per service).
Can you confirm on your end and post your evals here?

bhope:
Ran the numbers to quantify this. Built this branch and deployed against a local cluster with 20 services and realistic multi-port mix):

| Metric | Series | Notes |
|---|---:|---|
| kube_service_info | 20 | baseline, 1/service |
| kube_service_spec_ports (this PR) | **30** | 1.5x baseline |
| kube_endpoint_ports (existing STABLE) | 30 | identical, same per-port pattern |

Max ports on any single service was 3 (kube-dns, kubelet). `kube_service_spec_ports` matches the cardinality of the already stable `kube_endpoint_ports` exactly. User can also opt out via `--metric-denylist=kube_service_spec_ports`.

bhope (commented):
Hi @Serializator after discussing this with other maintainers, we decided to wait on this to receive more signal on the usage and need of this metric since it's increasing the cardinality by 1.5x

Feel free to share your use-case here.

PR #2850, opened 2026-01-18
chore: Merge `release-2.18` back into `main`

rexagod:
Update `main` manifests with v2.18 release chores.

### DIFF
--- a/.github/workflows/pre-release.yml
+++ b/.github/workflows/pre-release.yml
@@ -1,9 +1,9 @@
 name: Pre-release
 
-on:
-  push:
-    tags:
-      - 'v*'
+# on:
+#   push:
+#     tags:
+#       - 'v*'
 
 permissions:
   contents: write


rexagod:
IIUC pre-release workflows shouldn't be triggered on merging back into main. @Rishab87 could you take another look at the workflow?

Rishab87:
yup pre release workflow won't trigger on merging back to main, it will trigger when you push a tag

rexagod:
I see, so it shouldn't have run on this PR? In that same light, PCMIIW but I believe pre-release should be a non-PR-associated workflow, when a tag is pushed before a release is created by it (as it's running for the tag and not the PR)?

rexagod:
I just noticed the workflow replaced the v2.18 changelog in the release, which is fine, however, the Go version was 1.24, which should've been 1.25.6. However since we had a couple of notes to explicitly mention this release, I'll edit this back. Just wanted to point this out.

PR #2852, opened 2026-01-19
fix: avoid mutating metric families during write

### LINKED ISSUE #2805: Unpredictable format of info metrics
**What happened**:
Hi. Our scraper which collects metric from KSM use such header - Accept: application/openmetrics-text;version=1.0.0;charset=utf-8,application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3
and it receives info metric in such format - # TYPE resource_info info 

but if i make one request to ksm with such headers(like from browser) - Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7
it returns metric in such format - # TYPE resource_info gauge

and after this request our scraper receives this metric on all requests as -  # TYPE resource_info gauge

I expect that ksm should return metric in same format to all requests with specific headers

```bash
curl -H 'Accept: application/openmetrics-text;version=1.0.0;charset=utf-8,application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3' http://ip-appress-here:8080/metrics | grep 'TYPE resource_info'
# TYPE resource_info info

curl -H 'Accept: text/html,application/xhtml+xml,application/xml;q=0.9,image/avif,image/webp,image/apng,*/*;q=0.8,application/signed-exchange;v=b3;q=0.7' http://ip-appress-here:8080/metrics | grep 'TYPE resource_info'
# TYPE resource_info gauge

# now all requests with specific headers receive gauge, which previously received info
curl -H 'Accept: application/openmetrics-text;version=1.0.0;charset=utf-8,application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.7,text/plain;version=0.0.4;q=0.3' http://ip-appress-here:8080/metrics | grep 'TYPE resource_info'
# TYPE resource_info gauge
```

**Environment**:

* kube-state-metrics version: 2.17.0
* Kubernetes version (use `kubectl version`): v1.31.8
* Other info:

bhope:

Fixes a bug where metric family types could be mutated across requests when serving different content types.

When serving Prometheus text output, `info` and `stateset` metrics were rewritten to `gauge` by mutating shared metric families. This caused subsequent OpenMetrics requests to return incorrect types. This change ensures per-request sanitization does not affect stored metrics.

Does not change cardinality.

Fixes #2805

### DIFF
--- a/pkg/metrics_store/metrics_store.go
+++ b/pkg/metrics_store/metrics_store.go
@@ -28,11 +28,10 @@ import (
 // interface. Instead of storing entire Kubernetes objects, it stores metrics
 // generated based on those objects.
 type MetricsStore struct {
-	// metrics is a map indexed by Kubernetes object id, containing a slice of
-	// metric families, containing a slice of metrics. We need to keep metrics
-	// grouped by metric families in order to zip families with their help text in
-	// MetricsStore.WriteAll().
-	metrics sync.Map
+	// metrics points to a sync.Map indexed by Kubernetes object id, containing a slice of
+	// metric families, containing a slice of metrics. It's a pointer so cloned stores can
+	// safely share the same backing storage without copying or mutating it.
+	metrics *sync.Map
 
 	// generateMetricsFunc generates metrics based on a given Kubernetes object
 	// and returns them grouped by metric family.
@@ -48,7 +47,7 @@ func NewMetricsStore(headers []string, generateFunc func(interface{}) []metric.F
 	return &MetricsStore{
 		generateMetricsFunc: generateFunc,
 		headers:             headers,
-		metrics:             sync.Map{},
+		metrics:             &sync.Map{},
 	}
 }
 
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -26,6 +26,21 @@ import (
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 )
 
+const (
+	helpPrefix = "# HELP "
+	typePrefix = "# TYPE "
+)
+
+var (
+	infoTypeString     = string(metric.Info)
+	stateSetTypeString = string(metric.StateSet)
+	gaugeTypeString    = string(metric.Gauge)
+	// gaugeNewline is the pre-built suffix used when rewriting info/stateset TYPE lines.
+	// Pre-computing it reduces rewriteTypeLine to a 2-string concat (concatstring2),
+	// avoiding the overhead of a 3-string concat (concatstring3) on every rewrite.
+	gaugeNewline = gaugeTypeString + "\n"
+)
+
 // MetricsWriterList represent a list of MetricsWriter
 type MetricsWriterList []*MetricsWriter
 
@@ -59,20 +74,28 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 	}
 
 	for i, help := range m.stores[0].headers {
-		if help != "" && help != "\n" {
-			help += "\n"
-		}
-
 		var err error
-		m.stores[0].metrics.Range(func(_ interface{}, _ interface{}) bool {
-			_, err = w.Write([]byte(help))
+
+		// SanitizeHeaders blanks duplicate headers to preserve header/family
+		// index alignment. An empty header means suppress the header text but
+		// still emit the metric family bytes at this index.
+		if help != "" {
+			// Avoid allocating a new string if the header lacks a trailing newline:
+			// check once and emit "\n" as a second write if needed.
+			needsNewline := help[len(help)-1] != '\n'
+			m.stores[0].metrics.Range(func(_ interface{}, _ interface{}) bool {
+				_, err = io.WriteString(w, help)
+				if err != nil {
+					return false
+				}
+				if needsNewline {
+					_, err = io.WriteString(w, "\n")
+				}
+				return false
+			})
 			if err != nil {
-				err = fmt.Errorf("failed to write help text: %v", err)
+				return fmt.Errorf("failed to write help text: %v", err)
 			}
-			return false
-		})
-		if err != nil {
-			return err
 		}
 
 		for _, s := range m.stores {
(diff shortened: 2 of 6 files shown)


bhope:
@mrueg @rexagod this PR is waiting on workflow approval (CI hasn’t started yet). When you have a moment, could you approve the workflow so checks can run? Happy to get a review once CI is green. Thanks!

bhope:
All CI checks have passed now. Thanks.

pkg/metrics_store/metrics_store.go
mrueg: can you update the comment here to say that it's now a pointer to a map?

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
 				}
 
-				// Nullify duplicate headers after the sanitization to not miss out on any new candidates.
-				if header == lastHeader {
-					writer.stores[0].headers = append(writer.stores[0].headers[:i], writer.stores[0].headers[i+1:]...)
-
-					// Do not increment the index, as the next header is now at the current index.
-					continue
+				if shouldRemove {
+					writer.stores[0].headers[i] = ""
```
mrueg: Is this better than appending the slice before and after?

bhope: @mrueg Good question. I’m intentionally not splicing the slice to keep the header slice length stable, thus avoiding index shifts during iteration. Blanking the entry keeps the slice stable while safely treating the header as a no-op downstream. Happy to adjust if there’s a stronger preference for physical removal, but this felt like the safer tradeoff given how headers are consumed.

pkg/metrics_store/metrics_store.go
bhope: @mrueg Good catch, updated the comment to reflect that metrics is a pointer to sync.Map. Thanks!

bhope:
Thanks for the review! I’ve addressed the inline comments and updated the PR accordingly. Ready for another look when you have a chance.

bhope:
@mrueg gentle ping in case this got buried - addressed all comments and CI is green. Would appreciate another look when you have time. Thanks!

bhope:
I ran local tests and a small benchmark on the sanitize path. All unit tests pass (including immutability and duplicate handling).

For a worst-case scenario with 100k duplicate headers, `SanitizeHeaders` is ~19ms/op with ~11MB/op and ~200k allocs/op on my machine. I ran `go test ./pkg/metrics_store -bench BenchmarkSanitizeHeaders -benchmem`

The changes in this PR keep the header slice length stable and marks duplicates as "" instead of deleting them. The intention is to preserve index alignment with `metricFamilies[i]` in the non-consecutive duplicate case where deletion risks shifting indices.

pkg/metrics_store/metrics_writer.go
```diff
@@ -59,7 +59,12 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
 
 	for i, help := range m.stores[0].headers {
-		if help != "" && help != "\n" {
+		// Skip empty headers (set by SanitizeHeaders for duplicates)
+		if help == "" {
+			continue
+		}
+
+		if help != "\n" {
 			help += "\n"
```
mrueg: Could we implement this as part of the SanitizeHeaders func instead, so we do not need to touch the string here again?

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+					case strings.HasPrefix(line, "# HELP "):
+						fields := strings.Fields(line)
+						if len(fields) >= 3 {
+							metricName := fields[2]
+							if _, seen := seenHELP[metricName]; seen {
+								shouldRemove = true
+								break
+							}
+							seenHELP[metricName] = struct{}{}
+							modifiedLines = append(modifiedLines, line)
```
mrueg: Move this line out and you can remove the else path

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
-							header = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
-							writer.stores[0].headers[i] = header
+				lines := strings.Split(header, "\n")
+				shouldRemove := false
+				modifiedLines := make([]string, 0, len(lines))
+
+				for _, line := range lines {
+					switch {
+					case strings.HasPrefix(line, "# HELP "):
+						fields := strings.Fields(line)
```
mrueg: We can work on the line string directly here, something like this should work to extract the metric name?
```
helpPrefix := "# HELP "
# Get the Index of the second space
spaceIdx := strings.IndexByte(line[len(helpPrefix):], ' ')
if spaceIdx != -1 {
metricName := line[len(helpPrefix):len(helpPrefix) + spaceIdx]
...
}
```

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
 						}
-						if strings.HasSuffix(header, stateSetTypeString) {
-							header = header[:len(header)-len(stateSetTypeString)] + string(metric.Gauge)
-							writer.stores[0].headers[i] = header
+						fields := strings.Fields(line)
+						if len(fields) >= 3 {
+							metricName := fields[2]
+							modifiedLine := line
+							if contentType.FormatType() == expfmt.TypeTextPlain {
+								infoTypeString := string(metric.Info)
```
mrueg: Should these be moved out and simply be const?

pkg/metrics_store/metrics_writer.go
mrueg: Similar applies for the # TYPE below

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+							break
 						}
-						if strings.HasSuffix(header, stateSetTypeString) {
-							header = header[:len(header)-len(stateSetTypeString)] + string(metric.Gauge)
-							writer.stores[0].headers[i] = header
+						fields := strings.Fields(line)
+						if len(fields) >= 3 {
+							metricName := fields[2]
+							modifiedLine := line
+							if contentType.FormatType() == expfmt.TypeTextPlain {
```
mrueg: We only modify for TypeTextPlain, can we check that before we even touch anything?

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,49 +100,87 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+			}
+			// Share the metrics backing storage by sharing the pointer.
+			clonedStore.metrics = store.metrics
+			clonedStores = append(clonedStores, clonedStore)
+		}
+		clonedWriters = append(clonedWriters, &MetricsWriter{stores: clonedStores, ResourceName: writer.ResourceName})
+	}
+
+	// Deduplicate by metric name across all writers to handle non-consecutive duplicates during CRS reload.
+	seenHELP := make(map[string]struct{})
```
mrueg: Can we preallocate this using len(writer.stores[0].headers)?

pkg/metrics_store/metrics_writer.go
bhope: Good call - moved newline handling into SanitizeHeaders. WriteAll now writes headers as-is

pkg/metrics_store/metrics_writer.go
bhope: fixed.

pkg/metrics_store/metrics_writer.go
bhope: Agreed, switched to index-based extraction to avoid allocations.

pkg/metrics_store/metrics_writer.go
bhope: Yep, moved to file scope.

pkg/metrics_store/metrics_writer.go
bhope: Good suggestion, hoisted into a single isTextPlain check.

pkg/metrics_store/metrics_writer.go
bhope: Makes sense, added a capacity hint based on total header count.

bhope:
@mrueg Thanks for the thorough review.

Updates in the latest push:
- WriteAll() no longer mutates headers (now a dumb writer)
- newline normalization moved to SanitizeHeaders()
- preallocated HELP/TYPE dedupe maps
- replaced strings.Fields with slice/index parsing
- hoisted TypeTextPlain guard
- moved repeated strings to file scope

Please let me know if you'd like anything adjusted further.

bhope:
Hi @mrueg - wanted to check if you had a chance to review the latest changes. Happy to discuss or make further adjustments. Thanks for your time!

pkg/metrics_store/metrics_writer.go
mrueg:
```suggestion
			_, err = io.WriteString(w, help)
```

bhope: Thanks for pointing this out. Fixed now.

bhope:
Hi @mrueg resolved your latest inline comment - ready for another look whenever you have time!

bhope:
`ci-e2e-tests` is currently failing before tests execute due to image tag detection in the e2e script (it was parsing Docker’s human-readable output and treating the image ID as a tag).

I’ve opened a separate fix here: #2875 

This failure is unrelated to the changes in this PR. Once that merges and CI is green. This PR should be merge-ready as-is.

bhope:
@mrueg Thanks for merging #2875 . CI is green now and this PR should be ready. Would appreciate a final look when you have a moment!

bhope:
Rebased and squashed.

pkg/metrics_store/metrics_writer.go
```diff
@@ -59,13 +90,14 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 	}
 
 	for i, help := range m.stores[0].headers {
-		if help != "" && help != "\n" {
-			help += "\n"
+		// Skip empty headers (set by SanitizeHeaders for duplicates)
+		if help == "" {
+			continue
 		}
```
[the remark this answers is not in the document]
bhope: addressed - added logic to not skip metric families with empty headers in WriteAll

pkg/metrics_store/metrics_writer.go
```diff
@@ -95,34 +124,94 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
-				// When Prometheus requests proto-based formats, this branch is also used because any requested format that is not OpenMetrics falls back to text/plain in metrics_handler.go.
-				if strings.HasPrefix(header, "# HELP") && contentType.FormatType() == expfmt.TypeTextPlain {
-					infoTypeString := string(metric.Info)
-					stateSetTypeString := string(metric.StateSet)
-					if strings.HasSuffix(header, infoTypeString) {
-						header = header[:len(header)-len(infoTypeString)] + string(metric.Gauge)
-					}
-					if strings.HasSuffix(header, stateSetTypeString) {
-						header = header[:len(header)-len(stateSetTypeString)] + string(metric.Gauge)
+			for i := 0; i < len(writer.stores[0].headers); i++ {
```
mrueg: this looks like we're reverting the change from #2866

bhope: You are right, my bad -- will fix the rebase.

bhope: @mrueg -- it took me a bit to wrap my head around this. The actual fix in #2866 was blanking instead of removing headers from the slice. It's back in my PR but we are still blanking, -- so index stability is preserved. So technically this isn't reverting the fix from #2866

pkg/metrics_store/metrics_writer.go
```diff
@@ -59,15 +90,16 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
...
+			continue
 		}
 
 		var err error
 		m.stores[0].metrics.Range(func(_ interface{}, _ interface{}) bool {
-			_, err = w.Write([]byte(help))
+			_, err = io.WriteString(w, help)
 			if err != nil {
-				err = fmt.Errorf("failed to write help text: %v", err)
+				return false
```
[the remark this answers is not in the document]
bhope: fixed

bhope:
Hi @mrueg - The CI is green and Copilot review is addressed too. Ready for your final review.

bhope:
Thanks @mrueg for collaborating with me on this PR. The changes look good! It looks in a good shape.

PR #2854, opened 2026-01-19
feat: implement discovery of apiservices

### LINKED ISSUE #2471: `CustomResourceState` does not produce metrics for the aggregation layer if there is no `CustomResourceDefinition` defined
**What happened**:

As per Kubernetes [documentation](https://kubernetes.io/docs/concepts/extend-kubernetes/api-extension/), there are two ways to extend the Kubernetes API: custom resource definitions and the aggregation layer. The custom resource state metrics feature provided by kube-state-metrics is capable of generating metrics for both kinds of custom resources. The bug appears however when there are no custom resource definitions in the Kubernetes API but only custom resources from the aggregation layer. In such case, no metrics for custom resources are generated.

We think the problem might be caused by the [CRDiscoverer](https://github.com/kubernetes/kube-state-metrics/blob/25a1d8da057cf761d614c59a52785335d34082d1/internal/discovery/discovery.go#L69), because it [polls](https://github.com/kubernetes/kube-state-metrics/blob/25a1d8da057cf761d614c59a52785335d34082d1/internal/discovery/discovery.go#L266-L275) only if a custom resource definition is discovered. A quick workaround is to initialize the `CRDiscoverer` with the `WasUpdated` property set to true, but this is probably not the right way to fix the bug:

```
diff --git a/pkg/app/server.go b/pkg/app/server.go
index 9d25fd0e..ff70aece 100644
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -308,6 +308,7 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
                        CRDsAddEventsCounter:    crdsAddEventsCounter,
                        CRDsDeleteEventsCounter: crdsDeleteEventsCounter,
                        CRDsCacheCountGauge:     crdsCacheCountGauge,
+                       WasUpdated: true,
                }
                // This starts a goroutine that will watch for any new GVKs to extract from CRDs.
                err = discovererInstance.StartDiscovery(ctx, kubeConfig)
```

As far as we understand, watching the `customresourcedefinitions` resource in the `apiextensions.k8s.io` API group is only required to support [wildcard matching](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md#wildcard-matching-of-version-and-kind-fields). However, this might not be taking custom resources from the aggregation layer into account. We tested wildcard matching for custom resources from the aggregation layer and it did not work for us. It seems kube-state-metrics is only able to generate metrics for these when the group, version and kind are provided in the custom resource state configuration. Note watching the `customresourcedefinitions` resource might not be necessary in this case because the configuration already contains all information needed to call the API.

We would like to challenge if kube-state-metrics could watch custom resources from the aggregation layer as well. We think this would not only be a better way to fix this bug, but it would also allow the custom resource state feature to work likewise for any custom resource (aggregation layer or custom resource definitions). The `/openapi/v2` endpoint could be seen or discussed as an alternative to watch which resources do exist in the Kubernetes API:

```
kubectl get --raw /openapi/v2 | jq -r '.paths | keys[]'
```

We would expect metrics for custom resources from the aggregation layer to be exposed as defined in the custom resource state configuration, independently of if there are custom resource definitions or not.

```bash
# Execute kube-state-metrics against a Kubernetes cluster without custom resource definitions, and
# use a custom resource state configuration that generates metrics from custom resources from the aggregation layer.

# This should show no resources
kubectl get crd

go run main.go --custom-resource-state-only --custom-resource-state-config-file <custom-resource-config-file.yaml> --kubeconfig $KUBECONFIG

# Check kube-state-metrics and validate the configured custom resource metrics are not generated
curl http://localhost:8080/metrics

# Now apply a dummy custom resource definition, e.g., from the custom resource definition from the Kubernetes documentation
# (in the command below, deleting the selectableFields entry might not be required depending on the Kubernetes version)
curl -s https://raw.githubusercontent.com/kubernetes/website/main/content/en/examples/customresourcedefinition/shirt-resource-definition.yaml | yq 'del(.spec.versions[0].selectableFields)' | kubectl --kubeconfig apply -f -

# Check kube-state-metrics again and validate the configured custom resource metrics appear after a few moments
curl http://localhost:8080/metrics
```

N/A

**Environment**:

* kube-state-metrics version: 2.13
* Kubernetes version (use `kubectl version`): 1.30.3
* Cloud provider or hardware configuration: N/A
* Other info: N/A

alexandernorth:
Enables discovery and metric collection for Custom Resources managed by aggregated API servers which do not have a local CRD. It does this by querying non-local apiservices for the resources they handle.

By default, no change 

Fixes #2471

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -17,10 +17,12 @@ package discovery
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
 	"k8s.io/apimachinery/pkg/runtime/schema"
+	k8sdiscovery "k8s.io/client-go/discovery"
 	"k8s.io/client-go/dynamic"
 	"k8s.io/client-go/dynamic/dynamicinformer"
 	"k8s.io/client-go/rest"
@@ -37,17 +39,84 @@ import (
 // Interval is the time interval between two cache sync checks.
 const Interval = 3 * time.Second
 
+type gvkExtractor func(obj interface{}) []groupVersionKindPlural
+
 // StartDiscovery starts the discovery process, fetching all the objects that can be listed from the apiserver, every `Interval` seconds.
 // resolveGVK needs to be called after StartDiscovery to generate factories.
 func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config) error {
+	err := r.startCRDDiscovery(ctx, config)
+	if err != nil {
+		return err
+	}
+	err = r.startAPIServiceDiscovery(ctx, config)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *CRDiscoverer) runInformer(ctx context.Context, informer cache.SharedIndexInformer, gvkExtractor gvkExtractor) error {
+	stopper := make(chan struct{})
+	_, err := informer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		AddFunc: func(obj interface{}) {
+			gvkps := gvkExtractor(obj)
+			r.SafeWrite(func() {
+				r.AppendToMap(gvkps...)
+				r.WasUpdated = true
+			})
+			r.SafeWrite(func() {
+				r.CRDsAddEventsCounter.Inc()
+				r.CRDsCacheCountGauge.Inc()
+			})
+		},
+		UpdateFunc: func(oldObj, newObj interface{}) {
+			oldGVKPs := gvkExtractor(oldObj)
+			newGVKPs := gvkExtractor(newObj)
+			r.SafeWrite(func() {
+				r.RemoveFromMap(oldGVKPs...)
+				r.AppendToMap(newGVKPs...)
+				r.WasUpdated = true
+			})
+			r.SafeWrite(func() {
+				r.CRDsUpdateEventsCounter.Inc()
+			})
+		},
+		DeleteFunc: func(obj interface{}) {
+			gvkps := gvkExtractor(obj)
+			r.SafeWrite(func() {
+				r.RemoveFromMap(gvkps...)
+				r.WasUpdated = true
+			})
+			r.SafeWrite(func() {
+				r.CRDsDeleteEventsCounter.Inc()
+				r.CRDsCacheCountGauge.Dec()
+			})
+		},
+	})
+	if err != nil {
+		return err
+	}
+
+	// Respect context cancellation.
+	go func() {
+		for range ctx.Done() {
+			klog.InfoS("context cancelled, stopping discovery")
+			close(stopper)
+			return
+		}
+	}()
+	go informer.Run(stopper)
+	return nil
+}
+
+func (r *CRDiscoverer) startCRDDiscovery(ctx context.Context, config *rest.Config) error {
 	client := dynamic.NewForConfigOrDie(config)
 	factory := dynamicinformer.NewFilteredDynamicInformer(client, schema.GroupVersionResource{
 		Group:    "apiextensions.k8s.io",
 		Version:  "v1",
 		Resource: "customresourcedefinitions",
 	}, "", 0, nil, nil)
-	informer := factory.Informer()
-	stopper := make(chan struct{})
+
 	extractGVKPs := func(obj interface{}) []groupVersionKindPlural {
 		objSpec := obj.(*unstructured.Unstructured).Object["spec"].(map[string]interface{})
 		var gvkps []groupVersionKindPlural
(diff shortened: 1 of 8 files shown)


internal/discovery/discovery.go
```diff
@@ -118,6 +109,84 @@ func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config)
...
+	processAPIService := func(obj interface{}) []groupVersionKindPlural {
+		serviceSpec := obj.(*unstructured.Unstructured).Object["spec"].(map[string]interface{})
+		if svc, ok := serviceSpec["service"]; !ok || svc == nil {
+			return nil
+		}
+
+		group := serviceSpec["group"].(string)
+		version := serviceSpec["version"].(string)
+
+		resourceList, err := discoveryClient.ServerResourcesForGroupVersion(fmt.Sprintf("%s/%s", group, version))
```
bhope: This call runs on every add/update event for an APIService and can make it quite chatty, especially since such updates can occur frequently. Do we want any dedup or backoff per group/version to avoid repeated discovery calls and potential API churn?

alexandernorth: Good point - I will look into this

bhope (commented):
Since `runInformer()` is now used for both CRDs and APIServices, the `CRDsAddEventsCounter` and  `CRDsCacheCountGauge` metrics will also count APIService events. That might be a bit confusing from a metrics perspective. Should we consider renaming them or splitting by source (CRD vs APIService)?

internal/discovery/discovery.go
alexandernorth: I have improved the logic so that the APIService must be `Available` before querying the API Server for resources, reducing calls which would return no data (and thus hopefully the amount of churn too).

I am under the impression we should always perform this query whenever there is an update as we need to track all available resources and a change to an APIService could result in new/removed `Kind`s, and this means that it becomes tricky with dedup/backoff as we might miss updates. As resources are queried by group+version which should be unique within the cluster, I don't see a case where we have multiple calls per update for the same group/version combinations - please correct me if I am missing something here though.

alexandernorth:
As APIServices are dynamic I realised I could not use exactly the same system as was present for CRDs, as if the APIService is no longer available then it cannot be queried for its Kinds (to remove them from the cache map). I refactored the cache map, which is now keyed by the source of the discovered resource - the benefit being that we can handle the case where an APIService becomes unavailable. It does mean that we no longer index by Group/Version, but this could be implemented if it is a requirement.

Regarding the generated metrics, I have consolidated and renamed them to apply for both CRDs and APIServices. I also removed the `add` metric as in the new implementation add and update are implemented the same way. I also wonder if the `delete` metric is necessary or if it makes sense to simply have that also increment the `update` counter. If I should not remove existing metrics I can add them back, and optionally add a consolidated counter with `add/update/delete`s.

I refactored the `GVK->chan` map, moving the stop channel to be part of the `DiscoveredResource`, and lifecycling it as part of the `Update/Delete` process.

The refactor also fixes a missing synchronisation where the cache map could be read outside of being locked (via the `ResolveGVKToGVKPs` function called in `pkg/customresourcestate/config.go#L191`)

internal/discovery/discovery.go
bhope: Thanks, that makes sense. The availability check sounds like a good improvement and should help reduce chattiness.
My original thought was mainly around repeated updates where the APIService object itself changes (like status churn) without a change in group/version, but I agree dedup/backoff gets tricky if we want to avoid missing updates.

bhope (commented):
Thanks for those refactors and detailed walkthrough. 

On the metrics side, since these are already released (hence probably consumed), I’d prefer we keep the delete metric rather than folding it into update. Having delete counted separately is still useful to understand churn (resources dropping vs being refreshed). Besides, removing it would be a breaking change for existing dashboards.

internal/discovery/discovery.go
alexandernorth: Yes that's true, I also considered this but I didn't find a "nice" solution which would ensure we received all updates - although in my (short term) observations around this, unless the aggregation service is very unstable there are not that many updates triggered.

alexandernorth:
That makes sense - I have added back the `add` metric - the metrics have the same name as before, but now are counting the 'merged' CRD/APIServer events (the naming was not CRD specific). I updated the help text to be more generic, but I think this is fine as it should not change anything for existing consumers unless they now specify aggregation layer resources

bhope (commented):
Thanks for the update and incorporating the feedback. Overall, looks good to me.

internal/discovery/discovery.go
bhope: Makes sense, thanks

internal/discovery/types.go
```diff
@@ -21,99 +21,205 @@ import (
...
 }
 
-// AppendToMap appends the given GVKs to the cache.
-func (r *CRDiscoverer) AppendToMap(gvkps ...groupVersionKindPlural) {
-	if r.Map == nil {
-		r.Map = map[string]map[string][]kindPlural{}
-	}
-	if r.GVKToReflectorStopChanMap == nil {
-		r.GVKToReflectorStopChanMap = map[string]chan struct{}{}
+// UpdateSource replaces all resources for a source with new resources.
```
dgrisonnet, line 83: I don't think this will work. The EventHandler will call UpdateSource with only one CRD or one Aggregated API, not the full list. So with the way you wrote this function, the latest updated resource will always override the other.

internal/discovery/types.go
```diff
@@ -21,99 +21,205 @@ import (
...
+// If resources is nil, this is a noop.
+// If resources is empty, all resources for the source are removed.
```
dgrisonnet, line 85: This shouldn't happen because the Updates and Creations will always pass a valid resource here. The only scenario where source are removed is on Delete

dgrisonnet (commented):
Nice work @alexandernorth :+1:

This is looking good, but would you mind splitting this in two PRs?
One to refactor the CRD discovery and another one adding the APIService discovery? This would make this work easier to review.

alexandernorth:
Thanks for the positive first feedback @dgrisonnet .

I have attempted to split the PR into this one and #2872, where this one implements the APIService discovery (due to the branch name making more sense) and then #2872 does the refactor of the discovery logic - I hope this is ok for you.

I was unable to change the target branch, so this also contains the same commits as the new PR but this has the APIService changes added back, hence the size of changes. For the APIService changes only, please refer to https://github.com/alexandernorth/kube-state-metrics/compare/feature/refactor-discovery...alexandernorth:kube-state-metrics:feature/discover-aggregation-layer-resources

internal/discovery/types.go
alexandernorth: I had a look into this, but I think it should work as expected.

The new implementation groups the resources by source (either CRD or APIService), if a CRD/APIService triggers an update, the discovered types under that source will be removed and replaced with the new data (from the CRD itself or the discovered types from the apiservice). The intention being that if we update a source, we need to 'rediscover' any resources owned by it, but other sources are not affected

internal/discovery/types.go
alexandernorth: This is needed mainly for APIServices. 
A noop is for the case where there was an interruption in the discovery (specifically for APIServices) - if `ServerResourcesForGroupVersion` fails, I thought it would be better to not clear the discovered types, as they might still be available on the cluster.
The removal of the source's resources is based on the fact that an APIService might change/remove the resources it offers, or become not ready without a deletion of the apiservice resource itself (so the informer would only see an update to the status field, for example, but that could mean that the offered types might be empty e.g. apiservice is not ready, or have new/fewer types e.g. update to the aggregation layer)

PR #2863, opened 2026-01-29
feat(ingress): add kube_ingress_status_load_balancer metric

### LINKED ISSUE #1366: Support for ingress status hostname

redbaron:
Captures load balancer created for an Ingress and reported in `status.loadBalancer.ingress[]` array.

Some times we need to know load balancer created for an ingress .

Name `kube_ingress_status_load_balancer` is chosen to follow pattern established by the `kube_service_status_load_balancer_ingress` metric.

It introduces new fairly low-cardinality metric `kube_ingress_status_load_balancer`. Only cardinality concern is IP, but on major cloud providers I am aware it is not reported. Plus if it was a problem in some setups I suppose IP label can be blacklisted.

Fixes #1366


internal/store/ingress.go
```diff
@@ -206,6 +206,34 @@ func ingressMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
...
 		),
+		*generator.NewOptInFamilyGenerator(
+			"kube_ingress_status_load_balancer",
+			"Ingress load balancer status.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapIngressFunc(func(i *networkingv1.Ingress) *metric.Family {
+				if len(i.Status.LoadBalancer.Ingress) == 0 {
+					return &metric.Family{
```
Serializator, line 217: Is there a need for the early return?

If `i.Status.LoadBalancer.Ingress` is empty then `ms` will be equally empty and thus `Metrics: ms`, will be equal to `Metrics: []*metric.Metric{}` in this early return.

internal/store/ingress_test.go
```diff
@@ -235,6 +235,82 @@ func TestIngressStore(t *testing.T) {
 				`,
 			MetricNames: []string{"kube_ingress_info", "kube_ingress_metadata_resource_version", "kube_ingress_created", "kube_ingress_labels", "kube_ingress_path", "kube_ingress_tls"},
 		},
+		{
```
Serializator, line 238: A test case to verify no metric family is returned when no load balancer ingress exists will verify there is no unwanted behavior either.

Serializator:
Thank you for contributing! 🎉 

I'm not a maintainer but I like to help out where I can. The comments I left are some low-hanging fruit you could take a look at early in the process so there is less waiting involved when a maintainer takes a look at your PR 😉

PR #2864, opened 2026-02-01
[WIP] feat: support dynamic namespace discovery

### LINKED ISSUE #2820: Dynamic namespace filtering

It could be great to have an option to dynamically filter the namespaced resources.

**Why is this needed:**

Having a static namespaces option is good, but it forces one to re-configure KSM each time metrics for a new namespace are needed.

I’d like to be able to configure a namespace label selector, so that KSM can dynamically discover namespaces based on labels and emit metrics only for objects in these namespaces.

**Additional context:**

Serializator:
Add support for dynamic namespace discovery. This allows the user to filter namespaces by label, providing more dynamic and granular control over what namespaces to watch.

This is a work in progress but I'd really appreciate early feedback.

The most impactful change is actually the refactoring of the metrics store to filter objects by their namespace at the event handler level instead of starting a dedicated reflector per namespace.

I'd love feedback on this approach. Performance-wise this reduces the number of events being handled concurrently at any point in time as there is one goroutine per resource instead of per resource and namespace.

The counter-part is a goroutine per resource and namespace. With dynamic namespace discovery instead of a static list, there will be rate limiting at the API level due to the design of one reflector (list & watch) per namespace.

No change.

Fixes #2820


Serializator:
@mrueg what do you think of these proposed changes?

Serializator:
Hey @mrueg! Just a quick mention in case it was overlooked. If you rather have me bother someone else with this, let me know!

bhope:
@Serializator Thanks for the PR. Do you mind rebasing it, we can do a review then.

Serializator:
@bhope no problem! I'll rebase this evening. It's 10:04 AM (UTC+2) for me right now. Just as an indicator 😉

Serializator:
@bhope I rebased. Unfortunately `internal/discovery/*_test.go` are running indefinitely for some reason. They were running fine before. I pushed so at least the code can be reviewed with the latest changes from `main` incorporated.

bhope (changes requested):
Thanks for the PR draft and chalking out the core concept. The `NamespaceDiscoverer` and single reflector approach for dynamic discovery seem reasonable to me and the test coverage is good too.

However, the refactoring that is done for the buildstores (I get the reasoning is to simplify for cluster wide scope) is creating a security regression for multi-tenant clusters. It would require ClusterRoleBinding for all users, breaking the existing pattern where KSM can be scoped to specific namespaces via RoleBindings.

I would suggest to keep the original per-namespace reflectors intact while adding your cluster_wide reflectors and the new code to support namespace-label-selector set separately. This keeps the RBAC footprint unchanged for existing users while enabling the new dynamic discovery feature for those who opt in.

Also, once you have made the changes, try building locally to fix any compile errors (I believe there are some at the moment).

Serializator:
Thank you for reviewing @bhope!

Apologies for the late reply. I haven't had the time to work on anything open source for the last few months.

I'll dive back into this pull request and get it to completion.

PR #2866, opened 2026-02-05
fix: order-dependent metric loss from header dedupe

jfremy-openai:
Custom resource metrics could disappear depending on metric declaration order when consecutive duplicate headers were sanitized. The old SanitizeHeaders implementation removed duplicate header entries from the header slice, but metric family byte slices remained indexed at their original positions. WriteAll zips headers and families by index, so deleting headers caused index drift and could skip later metric families entirely.

This was especially visible when one duplicate family emitted no samples (e.g. missing nested paths), but the root cause is header/family misalignment and can occur even without empty families.

Fix: preserve positional alignment by blanking duplicate/empty headers instead of removing them. WriteAll already skips empty headers naturally, so output remains deduplicated while family indices stay stable.

Tests: add two regressions in metrics_writer_test
- duplicate header + empty middle family still preserves later family output
- duplicate headers without empty families still preserve later family output
Also update TestSanitizeHeaders expectations for index-preserving blanking behavior.

**How does this change affect the cardinality of KSM:** *does not change cardinality*

### DIFF
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -98,43 +98,28 @@ func SanitizeHeaders(contentType expfmt.Format, writers MetricsWriterList) Metri
 	var lastHeader string
 	for _, writer := range writers {
 		if len(writer.stores) > 0 {
-			for i := 0; i < len(writer.stores[0].headers); {
-				header := writer.stores[0].headers[i]
-
-				// Removes duplicate headers from the given MetricsWriterList for the same family (generated through CRS).
-				// These are expected to be consecutive since G** resolution generates groups of similar metrics with same headers before moving onto the next G** spec in the CRS configuration.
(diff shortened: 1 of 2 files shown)


jfremy-openai:
@mrueg @CatherineF-dev let me know if you need more details. But overall, if you have multiple entries in a CRD metrics config that have the same metric name (in my case it was to report on a v1.ResourceList, I have to have one entry per resource type), KSM will skip some later metrics for that CRD

bhope:
Thanks for the PR - the index misalignment from removing headers is subtle. Blanking them instead is the right call. Regression tests look good too.
LGTM.

jfremy-openai:
@bhope there might be need for an lgtm label or is somebody else going to give it?

bhope:
/lgtm overall.

cc @mrueg for approval

mrueg:
/lgtm

Thanks for your contribution

PR #2869, opened 2026-02-07
fix(discovery): prevent duplicate GVKs and avoid stop channel leaks

mrueg:

In CRDiscoverer.AppendToMap, GVKs were unconditionally appended to the internal map and new stop channels were created on every call. This could lead to duplicate metric registration and leaked channels/goroutines when existing stop channels were overwritten.

This change:
- Only creates a new stop channel if one does not already exist for the GVK.
- Checks for existing Kinds in the discovery map before appending to prevent duplicates.

Fixes #

### DIFF
--- a/internal/discovery/types.go
+++ b/internal/discovery/types.go
@@ -84,8 +84,28 @@ func (r *CRDiscoverer) AppendToMap(gvkps ...groupVersionKindPlural) {
 		if _, ok := r.Map[gvkp.Group][gvkp.Version]; !ok {
 			r.Map[gvkp.Group][gvkp.Version] = []kindPlural{}
 		}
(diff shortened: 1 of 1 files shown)


alexandernorth:
Hi @mrueg , I wanted to quickly jump in as the work I did in #2854 should also address this issue (#2870 would still be needed, as the `stopCh` still exist although I think it might be more appropriate to use child `context`s instead of these floating channels). Happy to support or discuss my PR further, as it changes how discovery is handled.

bhope (commented):
@mrueg looks like #2920 fixed this. Can you confirm?

mrueg:
@bhope yes we should be good there.

PR #2870, opened 2026-02-07
fix(store): ensure CR reflectors stop on context cancellation

mrueg:

Previously, custom resource reflectors were only stopped when their specific stop channel in GVKToReflectorStopChanMap was closed (e.g., during discovery updates). If the main application context was cancelled, these reflectors would continue to run, potentially causing leaks.

This change wraps the GVK-specific stop channel with a select that also listens to the main application context cancellation, ensuring all CR reflectors are properly shut down when the application exits.

Fixes #


bhope (commented):
@mrueg this is covered in #2920. Can you confirm?

mrueg:
Yes we can close this one. Thanks for checking!

/close

PR #2873, opened 2026-02-12
feat: allow wildcards as part of annotations and labels in allowlists

skoef:
In my clusters, I've add some annotations to nodes, pods and PVCs. I want to use these annotations to filter on in my alertmanager queries. For instance: I want to alert if a pod has high CPU usage, except when an annotation on that pod is set like `monitoring.mycompany.org/ignore-cpu` or whatever. I have a whole bunch of these annotations.

Currently, if I want to add these annotations to KSMs allowlist, I have to either use `*` for all annotations (I'd rather not do this) or mention _all_ my annotations, for `nodes=`, `pods=`, and `persistentvolumeclaims=`. This could get rather annoying so I figured: why can't I use partial wildcards instead? Something like `nodes=[monitoring.mycompany.org/*],pods=[monitoring.mycompany.org/*],persistentvolumeclaims=[monitoring.mycompany.org/*]` or perhaps even `*=[monitoring.mycompany.org/*]`.

With this MR this would be possible, to the limit of 1 `*` in every match. I would like to get some feedback on this idea! 🙏 

## Summary by CodeRabbit

* **New Features**
  * Added support for resource wildcards and single-asterisk key patterns in metric annotation and label allowlists.
  * Added wildcard matching for Prometheus labels, including bounded pattern expansion.

* **Validation**
  * Allowlists now reject entries containing more than one wildcard.

* **Documentation**
  * Updated CLI help and developer documentation with wildcard syntax and usage details.


internal/store/utils_test.go
```diff
@@ -319,3 +319,181 @@ func TestMergeKeyValues(t *testing.T) {
...
+			},
+			allowList:    []string{"*key*"},
+			expectKeys:   []string{},
+			expectValues: []string{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			if tc.name != "multipleWildcards" {
```
bhope: This skips every case except multipleWildcards. Should be == not !=

```suggestion
			if tc.name == "multipleWildcards" {
```

internal/store/utils.go
```diff
@@ -178,24 +178,47 @@ func isPrefixedNativeResource(name v1.ResourceName) bool {
...
 
-		for _, l := range allowList {
-			v, found := allKubeData[l]
-			if found {
-				allowedKubeData[l] = v
+		for k, v := range allKubeData {
+			if found, _ := regexp.MatchString(expandWildcard(l, 1), k); found {
+				allowedKubeData[k] = v
 			}
 		}
```
bhope, line 214: regexp.MatchString recompiles the pattern on every call. With N allowlist entries and M annotation keys, it would be N×M compilations per reconciliation -- that's a lot. Compile once per allowlist entry instead.

internal/store/utils.go
```diff
@@ -178,24 +178,47 @@ func isPrefixedNativeResource(name v1.ResourceName) bool {
...
 	}
+
 	return kubeMapToPrometheusLabels(prefix, allowedKubeData)
 }
 
+// expandWildcard expands wildcards (*) to regular expressions, up to a limited number of wildcards
+func expandWildcard(pattern string, limit uint) string {
```
bhope, line 221: Should we just validate upfront and return an error when the pattern has more wildcards than the limit allows?

bhope:
Thanks for the PR -- this solves a real issue with namespaced annotations. Left a few inline comments that need attention.

internal/store/utils_test.go
skoef: Actually, the whole condition should be removed. I used this for easier debugging and it shouldn't be committed at all. I'll remove it!

internal/store/utils.go
skoef: Yes, makes sense. Changed this.

internal/store/utils.go
skoef: First of all, the limit of 1 wildcard is an arbitrary limit that fits my use-case but perhaps is too strict. I added a check for the amount of wildcards during parsing `LabelsAllowList`, but technically there is no reason to limit it to 1 occurrence of `*`. I left the `expandWildcard` function as is with a variable limit, but now there is a hardcoded `1` in two different packages (`internal/store` and `pkg/options`) that I'd rather replace with a constant then.

internal/store/utils.go
bhope: `regexp.MatchString` still recompiles the pattern on every call. Use `regexp.Compile` once outside the inner loop. Something like:

re, err := regexp.Compile(expandWildcard(l, 1))
if err != nil {
    continue
}
for k, v := range allKubeData {
    if re.MatchString(k) {
        allowedKubeData[k] = v
    }
}

internal/store/utils.go
bhope: Makes sense. Since you introduced both instances of this limit in this PR, could you define it as a shared constant and reference it in both utils.go and types.go? I was initially thinking of tracking this via a separate issue, but that's putting us in weird loop - can't reference this change in an issue when it's not yet merged. So we'd rather fix it here.

bhope (changes requested):

internal/store/utils.go
skoef: That's much better indeed.

internal/store/utils.go
skoef: I found the `pkg/constant` package a perfect fit for such a shared constant. It's implemented

docs/developer/cli-arguments.md
```diff
@@ -60,9 +60,9 @@ Flags:
...
+      --metric-annotations-allowlist string                  Comma-separated list of Kubernetes annotation keys that will be used in the resource' labels metric. By default the annotations metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes annotation keys you would like to allow for them (Example: '=namespaces=[kubernetes.io/team,...],pods=[kubernetes.io/team],...)'. A single '*' can be provided per resource instead to allow any annotations, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided for resources, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple annotations, i.e., `=pods=[something.examle.org/foo-*] will resolve to `=pods=[something.examle.org/foo-bar,something.examle.org/foo-baz,...]`.
```
bhope: nit
```suggestion
      --metric-annotations-allowlist string                  Comma-separated list of Kubernetes annotation keys that will be used in the resource' labels metric. By default the annotations metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes annotation keys you would like to allow for them (Example: '=namespaces=[kubernetes.io/team,...],pods=[kubernetes.io/team],...)'. A single '*' can be provided per resource instead to allow any annotations, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided for resources, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple annotations, i.e., `=pods=[something.example.org/foo-*] will resolve to `=pods=[something.example.org/foo-bar,something.example.org/foo-baz,...]`.
```

docs/developer/cli-arguments.md
```diff
@@ -60,9 +60,9 @@ Flags:
...
+      --metric-labels-allowlist string                       Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple labels, i.e., `=pods=[something.examle.org/foo-*] will resolve to `=pods=[something.examle.org/foo-bar,something.examle.org/foo-baz,...]`.
```
bhope: nit
```suggestion
      --metric-labels-allowlist string                       Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple labels, i.e., `=pods=[something.example.org/foo-*] will resolve to `=pods=[something.example.org/foo-bar,something.example.org/foo-baz,...]`.
```

bhope (commented):
Thanks for addressing the feedback. Left one inline nit, otherwise it's in pretty good shape.

However, the docs CI is failing -- cli-arguments.md is auto-generated from the cli flag usage strings. The wildcard documentation needs to be added to the flag descriptions in the code and then regenerate by running `make doccheck` locally.

docs/developer/cli-arguments.md
skoef: Yeah took me a while to notice the doc-job failed, it's fixed!

bhope (approved):
Looks good to me.
We will wait for the CI to be green and one of the maintainers, if they have any comments.

skoef:
@CatherineF-dev could you review this MR please?

pkg/constant/utils.go
```diff
@@ -0,0 +1,24 @@
...
+limitations under the License.
+*/
+
+package constant
+
+const (
+	// MaxPartialWildcardsPerLabel defines the amount of times a wildcard (*) may
+	// occur in a Label (or Annotation) passed with --metric-labels-allowlist or
+	// --metric-annotations-allowlist
+	MaxPartialWildcardsPerLabel = 1
```
rexagod: I'd suggest moving this to `types.go`, the convention has been to move something to a higher "util" package if its imported across packages and we want to avoid any dependency loops, and since `internal/store` already imports from `pkg/options`.

pkg/options/types_test.go
```diff
@@ -382,6 +382,21 @@ func TestLabelsAllowListSet(t *testing.T) {
...
+				"*": {
+					"somelabel/*",
+				},
+			}),
+		},
+		{
+			Desc:   "[invalid] with multiple partial wildcards per label",
+			Value:  "*=[somelabel/*-foo-*]",
+			Wanted: LabelsAllowList(map[string][]string{}),
+			err:    true,
```
rexagod: To make this a bit more robust, can we refactor this to actually see if `errLabelsAllowListMultipleWildcards` was thrown in this case?

internal/store/utils_test.go
```diff
@@ -319,3 +319,178 @@ func TestMergeKeyValues(t *testing.T) {
...
+			},
+			allowList:    []string{"*key*"},
+			expectKeys:   []string{},
+			expectValues: []string{},
+		},
+	}
+
+	for _, tc := range testCases {
+		t.Run(tc.name, func(t *testing.T) {
+			gotKeys, gotValues := createPrometheusLabelKeysValues(tc.prefix, tc.kubeData, tc.allowList)
```
rexagod:
```suggestion
			gotKeys, gotValues := createPrometheusLabelKeysValues("metric", tc.kubeData, tc.allowList)
```
(nit, to simplify the struct above a bit)

internal/store/utils.go
```diff
@@ -178,24 +179,52 @@ func isPrefixedNativeResource(name v1.ResourceName) bool {
...
 
-		for _, l := range allowList {
-			v, found := allKubeData[l]
-			if found {
-				allowedKubeData[l] = v
+		// prepare regular expression based on potential wildcards
+		re, err := regexp.Compile(expandWildcard(l, constant.MaxPartialWildcardsPerLabel))
+		if err != nil {
+			continue
+		}
```
rexagod, line 198: We'd want to surface these to the user.

rexagod:
Also PTAL at the Copilot reviews.

pkg/options/options.go
```diff
@@ -168,8 +168,8 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
...
+	o.cmd.Flags().Var(&o.LabelsAllowList, "metric-labels-allowlist", "Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple labels, i.e., '=pods=[something.example.org/foo-*]' will resolve to '=pods=[something.example.org/foo-bar,something.example.org/foo-baz,...]'")
```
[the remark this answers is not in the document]
skoef: done

docs/developer/cli-arguments.md
```diff
@@ -60,9 +60,9 @@ Flags:
...
+      --metric-labels-allowlist string                       Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be use to match multiple labels, i.e., '=pods=[something.example.org/foo-*]' will resolve to '=pods=[something.example.org/foo-bar,something.example.org/foo-baz,...]'
```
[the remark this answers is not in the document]
skoef: done

pkg/constant/utils.go
skoef: done

internal/store/utils_test.go
skoef: done

pkg/options/types_test.go
skoef: done

internal/store/utils.go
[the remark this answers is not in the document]
skoef: Since we use the label(-description) as the basis for the regex, we can't prepare the regex outside of this loop.

skoef: How do you suggest I'd implement that? The function `createPrometheusLabelKeysValues` currently doesn't return an error. I could go over the labels/annotations and throw this error in `LabelsAllowList.Set` instead during startup?

skoef:
@rexagod Hi, could you perhaps re-review my changes? Thanks!

skoef:
@rexagod could you re-review my changes please?

skoef:
Just a subtle re-poke @rexagod

bhope (changes requested):
Hi @skoef - Thank you for your patience on this one and appreciate the solid test coverage. If we could address. following points, we would be in a good spot to approve this.

1. Docs are out of sync - make doccheck is failing in CI. Could you run make generate and commit the actual output
2. `createPrometheusLabelKeysValues` runs for every object of every resource type on every add/update. Could we precompile the pattern once instead of on every call?
3. Circling back to @rexagod's earlier comment on internal/store/utils.go. It'd be good to at least log a warning here rather than fail silently.

Let me know if you need any help with above.

pkg/options/options.go
```diff
@@ -178,8 +178,8 @@ func (o *Options) AddFlags(cmd *cobra.Command) {
...
+	o.cmd.Flags().Var(&o.LabelsAllowList, "metric-labels-allowlist", "Comma-separated list of additional Kubernetes label keys that will be used in the resource' labels metric. By default the labels metrics are not exposed. To include them, provide a list of resource names in their plural form and Kubernetes label keys you would like to allow for them (Example: '=namespaces=[k8s-label-1,k8s-label-n,...],pods=[app],...)'. A single '*' can be provided per resource instead to allow any labels, but that has severe performance implications (Example: '=pods=[*]'). Additionally, an asterisk (*) can be provided as a key, which will resolve to all resources, i.e., assuming '--resources=deployments,pods', '=*=[*]' will resolve to '=deployments=[*],pods=[*]'. Wildcards can also be used to match multiple labels, but only a single '*' is supported per key pattern, i.e., '=pods=[something.example.org/foo-*]' will match labels such as 'something.example.org/foo-bar' and 'something.example.org/foo-baz'.")
```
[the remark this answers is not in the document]
skoef: addresses

skoef:
I think I addressed all concerns now! @rexagod @bhope @CatherineF-dev @mrueg could you please review this and hopefully approve it?

bhope (commented):
Hi @skoef the changes now look good to me. Can you address the CI failures?

bhope (approved):
Thanks @skoef for the patience and help with addressing the review comments. I did a small commit on your branch to fix the minor CI failures. The PR looks good now. 

/lgtm

skoef:
@bhope thank you very much!

PR #2875, opened 2026-02-18
fix: tests/e2e - derive image tag using docker --format

bhope:
The e2e script parsed `docker images -a` human-readable output and extracted the second column via awk, which can return the image ID depending on Docker version/format.

This caused `kind load docker-image` to attempt loading `<image>:<imageID>`, which is not a valid tag, leading to CI failures before tests ran.

Use `docker images --format` to reliably select the built tag for `kube-state-metrics` (excluding 'latest') and load that into kind.

Link to example failures in e2e that this PR fixes: https://github.com/kubernetes/kube-state-metrics/actions/runs/22114858830/job/63976232563?pr=2852

### DIFF
--- a/tests/e2e.sh
+++ b/tests/e2e.sh
@@ -189,8 +189,18 @@ kubectl version
 # query kube-state-metrics image tag
 REGISTRY="registry.k8s.io/kube-state-metrics" make container
 docker images -a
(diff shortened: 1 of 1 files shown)


mrueg:
/lgtm
Thanks for the fix

PR #2880, opened 2026-03-01
feat: add SchedulingGated to kube_pod_status_reason

### LINKED ISSUE #2848: Add `SchedulingGated` reason to `kube_pod_status_reason` metric

Currently, there is no metrics exposing if a pod is in `Pending` phase because it is `SchedulingGated`. Having `kube_pod_status_reason` exposing this would help to write better alerting rules to prevent sending alerts for pending pods if they are just in `SchedulingGated` phase.

**Why is this needed:**

Preventing alert rules to trigger on `Pending` pods that are in `SchedulingGated` state

Add `SchedulingGated` reason to the `kube_pod_status_reason` metrics

**Additional context:**

I can open a PR to fix this issue if that can help.

bhope:
`SchedulingGated` is a standard Kubernetes reason set on the `PodScheduled` condition when a pod is held by scheduling gates. Adding this as a valid reason state gives a way to distinguish intentionally gated pods from genuinely stuck pending pods. This would help filtering alerts between genuinely pending and scheduling gated state pods.

Cardinality impact is bounded as `SchedulingGated` is a fixed pod reason and not a dynamic value.

Fixes #2848


bhope:
Thanks @nmn3m for `lgtm`. Squash done.
cc @mrueg @dgrisonnet

mrueg:
/lgtm

Thanks for your contribution!

PR #2884, opened 2026-03-08
fix(customresourcestate): handle nil path gracefully in StateSet metrics

### LINKED ISSUE #2482: CustomResourceDefinitions status fields cause spam of errors that cannot be fixed
**What happened**: Spam of errors that look like this:

```
"kube_customresource_phase" err="[status,phase]: expected value for path to be string, got <nil>"
"kube_customresource_initialized" err="[status,initialized]: got nil while resolving path"
```

Status fields are not guaranteed to exist at resource creation. The behavior is not consistent with known types where a default value is taken.

Create a CRD with status and a valid object. Do not run a controller (this is one of possible scenarios).

```
kubectl apply -f https://raw.githubusercontent.com/kubernetes/sample-controller/master/artifacts/examples/example-foo.yaml
kubectl apply -f https://raw.githubusercontent.com/kubernetes/sample-controller/master/artifacts/examples/crd.yaml
...
E0823 13:09:18.205000   60178 registry_factory.go:685] "cr_replicas" err="[status,replicas]: got nil while resolving path"
```

The error repeats for every instance of a resource, and there can be thousands of such resources.

```
E0823 13:11:49.963518   60382 registry_factory.go:685] "cr_replicas" err="[status,replicas]: got nil while resolving path"
E0823 13:11:49.963601   60382 registry_factory.go:685] "cr_replicas" err="[status,replicas]: got nil while resolving path"
```

I believe is a general problem for all CRDs and all status fields.

**Environment**: kind or any other Kubernetes cluster

* kube-state-metrics version: commit f7304dc8d8337ad4028c6e9f02dd47ab2fb0aa52
* Kubernetes version (use `kubectl version`): Client Version: v1.30.2 (shouldn't matter)
* Cloud provider or hardware configuration: kind Server Version: v1.30.0
* Other info: n/a

Br1an67:
Fixes #2482

When CustomResourceDefinition status fields don't exist at resource creation time, StateSet metrics would previously log an error for each resource instance. This caused error spam like:

```
registry_factory.go:685] "cr_test" err="[status,phase]: expected value for path to be string, got <nil>"
```

Status fields are not guaranteed to exist at resource creation, so this behavior was inconsistent with known types where nil values are handled gracefully (e.g., Gauge with `NilIsZero`).

This PR modifies `compiledStateSet.values()` to return empty results instead of an error when the path resolves to nil, consistent with how Gauge handles nil values.

**How does this change affect the cardinality of KSM:** Does not change cardinality

Fixes #2482

## Summary by CodeRabbit

* **Bug Fixes**
  * State-set metrics now treat unresolved source or value paths as absent values without returning errors.
  * Improved error details for invalid non-string values by including the complete value path.

* **Documentation**
  * Clarified how unresolved paths and non-string values affect StateSet metrics.

* **Tests**
  * Added coverage for missing source paths, missing value paths, and invalid numeric values.

### DIFF
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -690,6 +690,10 @@ spec:
 Metrics of type `StateSet` will generate a metric for each value defined in `list` for each resource.
 The value will be 1, if the value matches the one in list.
 
+If `path` (or `valueFrom`, if set) does not resolve, for example because the status field is not populated yet,
+no metrics are produced for that resource and no error is logged. Once the field is set, the metrics appear.
+A value that resolves to a non-string type is reported as an error.
+
 Produces the metric:
 
 ```prometheus
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -403,6 +403,12 @@ func (c *compiledStateSet) values(v interface{}) (result []eachValue, errs []err
 	comparable := c.ValueFrom.Get(v)
 	value, ok := comparable.(string)
 	if !ok {
+		// If the path doesn't exist (nil), return empty results instead of an error.
+		// This is consistent with how Gauge handles nil values and is expected for
+		// status fields that don't exist at resource creation time.
+		if comparable == nil {
+			return []eachValue{}, nil
+		}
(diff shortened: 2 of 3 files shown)


mrueg:
@Br1an67 can you sign the CLA?

mrueg:
@Br1an67 can you address the comments from copilot? Thanks

PR #2891, opened 2026-03-10
feat: add support for label generation for ephemeral volumes in pod

### LINKED ISSUE #2490: Export pod ephemeral PVCs metrics
**What would you like to be added**: kube-state-metrics exposes metrics about PVC usage by pods through metrics like `kube_pod_spec_volumes_persistentvolumeclaims_info` and `kube_pod_spec_volumes_persistentvolumeclaims_readonly`. I'd like similar metrics to be available for [Ephemeral Volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) mounts since those are also backed by PVCs.

**Why is this needed**: We use prometheus metrics to determine whether a PVC is not mounted, giving us a reminder to drop it it was left behind for some reason. Our alerting rule lists PVCs in a namespace with `kube_persistentvolumeclaim_info` and excludes mounted ones with `kube_pod_spec_volumes_persistentvolumeclaims_info`. Ephemeral volumes generate a PVC which appears in `kube_persistentvolumeclaim_info` but not in `kube_pod_spec_volumes_persistentvolumeclaims_info` since the volume does not have `PersistentVolumeClaim.ClaimName` defined. Adding a metric exposing ephemeral PVCs would give us a way to avoid false alarms when a pod is using an ephemeral PVC.

**Describe the solution you'd like**: Exposing another metric `kube_pod_spec_volumes_ephemeral_persistentvolumeclaims_info` seems acceptable, or updating `kube_pod_spec_volumes_persistentvolumeclaims_info` to add a `ephemeral` label would work as well.

Implementation note: while the PodSpec does not have a field explicitly giving the PVC name, the docs clarify how it's derived from the pod and volume name:
> Naming of the automatically created PVCs is deterministic: the name is a combination of the Pod name and volume name, with a hyphen (-) in the middle.

Alternatively, exposing PVC ownership data (`ownerReferences` metadata) would also address my use case, although I think it would be hard to integrate to my alerting rule.

**Additional context**
We sometimes run temporary workloads that need to store large amounts of data. Since we don't need the data to persist across pod executions, we use [Ephemeral Volumes](https://kubernetes.io/docs/concepts/storage/ephemeral-volumes/) to ensure the PVC is removed when we drop the pod.

Here's a pod manifest example (we use these pods to perform operations on our databases by `exec`-ing into them, this avoids tunneling and guards against connection drops):
```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    run: tmp-workload
  name: tmp-workload
spec:
  terminationGracePeriodSeconds: 3
  containers:
  - args:
    - bash
    - -c
    - sleep infinity
    image: postgres
    name: tmp-workload
    volumeMounts:
    - name: workdir
      mountPath: /workdir
    resources:
      limits:
        memory: 1Gi
        cpu: "1"
  volumes:
  - name: workdir
    ephemeral:
      volumeClaimTemplate:
        spec:
          accessModes: ["ReadWriteOnce"]
          resources:
            requests:
              storage: 1Ti
```

eminaktas:
Add support for generating the persistentvolumeclaim label for ephemeral volumes in pod volume metrics.

Kubernetes creates PVCs for ephemeral volumes using the naming convention <pod-name>-<volume-name>. This change aligns kube-state-metrics with the upstream implementation to ensure the generated label matches the actual PVC name created by Kubernetes.

Upstream references:

Ephemeral controller PVC creation:
https://github.com/kubernetes/kubernetes/blob/21b427c29960dac4e362d6a9f92817d0b01d78ec/pkg/controller/volume/ephemeral/controller.go#L261

PVC naming helper:
https://github.com/kubernetes/component-helpers/blob/b20f683efd864cebf95ec4a62f4c271cc69b9571/storage/ephemeral/ephemeral.go#L41-L43

Fixes #2490

**What this PR does / why we need it:** Add support for exposing ephemeral PVCs in `kube_pod_spec_volumes_persistentvolumeclaims_info`.

**How does this change affect the cardinality of KSM:** increases

**Which issue(s) this PR fixes:** Fixes #2490

## Summary by CodeRabbit

* **New Features**
  * Extended pod volume metrics to include ephemeral volumes alongside persistentvolumeclaim-backed volumes.
  * Added an `ephemeral` label and standardized how `persistentvolumeclaim` is populated for ephemeral volumes (using the generated `<pod-name>-<volume-name>` format).
  * Ephemeral volumes report `read_only` as `0`.
* **Documentation**
  * Updated metric descriptions to explicitly document ephemeral volume behavior and the read-only reporting rule.


internal/store/pod.go
```diff
@@ -1273,6 +1273,12 @@ func createPodSpecVolumesPersistentVolumeClaimsInfoFamilyGenerator() generator.F
 						LabelValues: []string{v.Name, v.PersistentVolumeClaim.ClaimName},
 						Value:       1,
 					})
+				} else if v.Ephemeral != nil {
```
bhope, line 1422: `kube_pod_spec_volumes_persistentvolumeclaims_readonly` in `createPodSpecVolumesPersistentVolumeClaimsReadonlyFamilyGenerator()` uses the same volume loop and should also be updated for consistency. A user querying both metrics would get asymmetric results otherwise.

bhope (changes requested):

bhope:
Thanks for your contributions! Other than the inline comment, you'd also need to update the docs here: [docs/metrics/workload/pod-metrics.md ](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md?plain=1#L51)

internal/store/pod.go
eminaktas: I added the change for readonly metric. Ephemeral volume cannot be mounted as readonly, so it will be set to 0.

bhope:
Thinking about this a bit more, today these metrics are dropped for pods with ephemeral storage. So this change would mean those metrics will now start flowing in, so there would be a slight increase in metric volume (not necessarily cardinality). We should call that out in cardinality section of the PR description.

eminaktas:
Hi @nmn3m @mrueg @dgrisonnet — could you please take a look at this PR when you get a chance? Thanks!

internal/store/pod.go
```diff
@@ -1300,6 +1306,12 @@ func createPodSpecVolumesPersistentVolumeClaimsReadonlyFamilyGenerator() generat
 						LabelValues: []string{v.Name, v.PersistentVolumeClaim.ClaimName},
 						Value:       boolFloat64(v.PersistentVolumeClaim.ReadOnly),
 					})
+				} else if v.Ephemeral != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"volume", "persistentvolumeclaim"},
+						LabelValues: []string{v.Name, p.Name + "-" + v.Name},
+						Value:       0,
+					})
```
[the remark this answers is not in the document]
eminaktas: This is added as suggested [here](https://github.com/kubernetes/kube-state-metrics/pull/2891#discussion_r2919514622). To have consistency between metrics, we are adding here as well. This cannot be ready only so this reason I am adding static value.

Please let me know if we should remove it @mrueg.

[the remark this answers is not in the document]
bhope: @eminaktas We can keep the value 0 but we will have to document in the metric HELP string that for ephemeral volumes this reflects the absence of a spec-level readOnly flag. Without documentation, users will misinterpret it.

bhope (commented):
Overall looks good. Let's add the documentation in the metric help string for 0 value to avoid user misinterpretation.

eminaktas:
> Overall looks good. Let's add the documentation in the metric help string for 0 value to avoid user misinterpretation.

TY. Updated doc for the metric helper.

internal/store/pod.go
```diff
@@ -1273,6 +1274,12 @@ func createPodSpecVolumesPersistentVolumeClaimsInfoFamilyGenerator() generator.F
 						LabelValues: []string{v.Name, v.PersistentVolumeClaim.ClaimName},
 						Value:       1,
 					})
+				} else if v.Ephemeral != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"volume", "persistentvolumeclaim"},
+						LabelValues: []string{ephemeral.VolumeClaimName(p, &v)},
```
bhope, line 1280: Sorry, just realized, this may panic at metrics writer (before tests) as it would expect two labels.

```suggestion
						LabelValues: []string{v.Name, ephemeral.VolumeClaimName(p, &v)},
```

internal/store/pod.go
```diff
@@ -1300,6 +1307,12 @@ func createPodSpecVolumesPersistentVolumeClaimsReadonlyFamilyGenerator() generat
 						LabelValues: []string{v.Name, v.PersistentVolumeClaim.ClaimName},
 						Value:       boolFloat64(v.PersistentVolumeClaim.ReadOnly),
 					})
+				} else if v.Ephemeral != nil {
+					ms = append(ms, &metric.Metric{
+						LabelKeys:   []string{"volume", "persistentvolumeclaim"},
+						LabelValues: []string{ephemeral.VolumeClaimName(p, &v)},
```
bhope, line 1313: Same as above
```suggestion
						LabelValues: []string{v.Name, ephemeral.VolumeClaimName(p, &v)},
```

eminaktas:
Thanks for the review guys. I believe It should be good now.

eminaktas:
I am not sure if this is forgotten. Just reminder ping :)

@bhope @mrueg

bhope:
Thanks for the changes and patience. 
/lgtm
/hold for other approvers to review and for CI to pass

bhope (approved):

internal/store/pod.go
```diff
@@ -1300,6 +1307,12 @@ func createPodSpecVolumesPersistentVolumeClaimsReadonlyFamilyGenerator() generat
 						LabelValues: []string{v.Name, v.PersistentVolumeClaim.ClaimName},
 						Value:       boolFloat64(v.PersistentVolumeClaim.ReadOnly),
 					})
+				} else if v.Ephemeral != nil {
```
mrueg, line 1455: Should we expose that this is an ephemeral pvc volume in metrics as well?

eminaktas: I think it make sense to have it. Should we introduce a new metric or a new label?

bhope: @eminaktas - to keep it simple for promql queries, we can add a label to the already stable metrics (`kube_pod_spec_volumes_persistentvolumeclaims_info` and `kube_pod_spec_volumes_persistentvolumeclaims_readonly`)

eminaktas: I just updated the PR for this change.

PR #2892, opened 2026-03-11
feat: add PreemptionByScheduler and TerminationByKubelet to kube_pod_status_reason

bhope:
Adds two new reason values to the `kube_pod_status_reason` metric:

  - `PreemptionByScheduler` -- set when a pod is terminated because the scheduler preempted it to make room for a higher priority pod
  - `TerminationByKubelet` -- set when the kubelet terminates a pod

These complement the existing reasons -- Evicted, NodeAffinity, NodeLost, SchedulingGated, Shutdown, UnexpectedAdmissionError

This is a fixed-cardinality expansion bounded by the static `podStatusReasons` slice, so it's safe and predictable.

Follows the same pattern as #2848


docs/metrics/workload/pod-metrics.md
```diff
@@ -50,7 +50,7 @@
...
+| kube_pod_status_reason                                | Gauge       | The pod status reasons                                                                                                                                                              |                                                | `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `reason`=&lt;Evicted\|NodeAffinity\|NodeLost\|PreemptionByScheduler\|SchedulingGated\|Shutdown\|TerminationByKubelet\|UnexpectedAdmissionError&gt; <br> `uid`=&lt;pod-uid&gt;                                                                                                                                                                                | EXPERIMENTAL | -      |
```
CatherineF-dev: add EvictionByEvictionAPI?

Issue #2895, opened 2026-03-12
Remove viper ConfigFileNotFoundError workaround


bxrne:
**What happened**:

There is a workaround in `internal/wrapper.go` that handles the case when a config file is not found. This workaround uses `os.Stat` as a fallback because `viper.ConfigFileNotFoundError` was not working correctly (see [viper#1783](https://github.com/spf13/viper/issues/1783)).

**What you expected to happen**:

The viper issue has been resolved, so the workaround can be removed as the go dependency lock statement is a higher version than the release of the fix

**Proposed solution**:

Remove the `os.Stat` fallback checks since `viper.ConfigFileNotFoundError` now works correctly.
Solution: #2896 

**Additional context**:

This cleanup was identified when reviewing the TODO comments in the code that referenced the viper issue.

PR #2896, opened 2026-03-12
chore(wrapper): removed viper os fallback due to underlying update

bxrne:

This PR removes a workaround in `internal/wrapper.go` that was added to handle a bug in viper's `ConfigFileNotFoundError` (see [viper#1783](https://github.com/spf13/viper/issues/1783)). The workaround used `os.Stat` as a fallback to check if a config file exists when viper failed to detect it.

Since the viper issue has been resolved, this workaround is no longer needed and can be removed.

**Changes made:**
- Removed `os.Stat` fallback for main config file and for custom resource config file
- Removed associated TODO comments referencing the viper issue

Does not change cardinality - this is a code cleanup with no functional changes, but is in hot path.

#2895

### DIFF
--- a/internal/wrapper.go
+++ b/internal/wrapper.go
@@ -52,10 +52,6 @@ func RunKubeStateMetricsWrapper(opts *options.Options) {
 		if cfgViperReadInConfigErr = cfgViper.ReadInConfig(); cfgViperReadInConfigErr != nil {
 			if errors.Is(cfgViperReadInConfigErr, viper.ConfigFileNotFoundError{}) {
 				klog.ErrorS(cfgViperReadInConfigErr, "Options configuration file not found at startup", "file", file)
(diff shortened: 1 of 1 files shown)


bhope:
In KSM we pin to viper v1.21.0. That viper release was cut in Sept 2025 and does not have the fix you are referring to (https://github.com/spf13/viper/pull/2027 ). Unfortunately, we would have to wait until we pin to newer viper release that has that fix (which at the moment there isn't).

/do-not-merge
/hold

bhope:
In KSM we pin to viper v1.21.0. That viper release was cut in Sept 2025 and does not have the fix you are referring to (https://github.com/spf13/viper/pull/2027 ). Unfortunately, we would have to wait until we pin to newer viper release that has that fix (which at the moment there isn't).

/hold

Issue #2354, opened 2024-03-29
[regression] /metrics port down when not existing CRD are listed in config file


bmm-alc:
**What happened**:
we use https://github.com/fluxcd/flux2-monitoring-example/blob/bcae1ef44963da89caabe962a93280c0dfad3381/monitoring/controllers/kube-prometheus-stack/kube-state-metrics-config.yaml as part of our config for FluxCD, but we didn´t deploy  resources`imagepolicies` and `imageupdateautomations` from `apiGroup` `image.toolkit.fluxcd.io` (they are optional and we don't use them).
Before 2.11.0 release it would not cause any problem, except recurring warning messages.

since version 2.11.0, as long as theses resources are listed, the '/metrics'  is down from prometheus POV
removing them, brought back the endpoint up in prometheus.

**What you expected to happen**:
`/metrics` to be up and running

**How to reproduce it (as minimally and precisely as possible)**:
declaring not deployed resources should trigger the issue.

**Anything else we need to know?**:

**Environment**:

* kube-state-metrics version: 2.11.0
* Kubernetes version (use `kubectl version`): 1.29.1
* Cloud provider or hardware configuration: scaleway

CatherineF-dev:
Could you provide steps to reproduce this issue?

bmm-alc:
closing because I don't have time to investigate this

CatherineF-dev:
Okay, we can keep it open and wait others to provide steps to reproduce. Feel it might be an issue.

/reopen

sebastiangaiser:
Hey, I stumbled up on this when searching for the expected behavior for a not existing CRD. For us this works without any errors (v2.16.0) but I'm wondering about the logging. I would expect to see a `warning`/`info` at startup that a specified CRD is not installed in the cluster but it didn't showed up (even using `-v=10` added to deployment args). How should this work and is this somewhere in docs and I didn't found it 🤔

rexagod:
IIUC we'd want that throw logs, +1. Do you mind taking this up?

Marking this as a good first issue.

PR #2903, opened 2026-03-19
fix(customresourcestate): log when configured CRD is not installed in cluster

maksimp13:
## What happened

When a resource is listed in the custom resource state config but its CRD is not installed in the cluster, `ResolveGVKToGVKPs` returns an empty list with no error. Previously this was silently skipped with no indication to the user.

## What this PR does

Adds an `InfoS` log when a configured CRD is not found in the cluster, so users understand why metrics are not being collected for that resource.

## How to test

1. Create a CRS config referencing a CRD that is not installed in the cluster
2. Run KSM with `--custom-resource-state-config-file` pointing to that config
3. Observe the log: `CRD for resource is not installed in the cluster, skipping`

Ref #2354

## Summary by CodeRabbit

* **Bug Fixes**
  * Added an informational log when a configured resource has no matching custom resource definitions.
  * Resources without matching definitions continue to be skipped safely.

### DIFF
--- a/pkg/customresourcestate/config.go
+++ b/pkg/customresourcestate/config.go
@@ -191,6 +191,8 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 			resolvedSet /* GVKPs */, err := discovererInstance.ResolveGVKToGVKPs(schema.GroupVersionKind(resource.GroupVersionKind))
 			if err != nil {
 				klog.ErrorS(err, "failed to resolve GVK", "gvk", resource.GroupVersionKind)
+			} else if len(resolvedSet) == 0 {
+				klog.InfoS("no matching CRD found for configured GVK, skipping", "gvk", resource.GroupVersionKind)
 			}
 			for _, resolved /* GVKP */ := range resolvedSet {
 				// Set their G** attributes to various resolutions of the GVK.


pkg/customresourcestate/config.go
```diff
@@ -191,6 +191,8 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
 			resolvedSet /* GVKPs */, err := discovererInstance.ResolveGVKToGVKPs(schema.GroupVersionKind(resource.GroupVersionKind))
 			if err != nil {
 				klog.ErrorS(err, "failed to resolve GVK", "gvk", resource.GroupVersionKind)
+			} else if len(resolvedSet) == 0 {
```
nmn3m, line 194: The change looks good to me — adding visibility for silently skipped CRDs is a nice improvement.

**One note**: this PR adds a log line, but issue #2354 reports that /metrics goes down entirely when non-existent CRDs are listed. A log doesn't fix that behavior, so Fixes #2354 might be inaccurate. 
Consider changing it to Ref #2354 unless the endpoint issue has already been resolved separately.

/cc @rexagod

maksimp13: Hmm.. I wasn't able to reproduce the /metrics crash locally - it seems it was already fixed in a previous versions. The endpoint returns 200 OK even when a configured CRD is not installed in the cluster. 

Updated Fixes #2354 → Ref #2354.

pkg/customresourcestate/config.go
```diff
@@ -191,6 +191,8 @@ func FromConfig(decoder ConfigDecoder, discovererInstance *discovery.CRDiscovere
...
+				klog.InfoS("CRD for resource is not installed in the cluster, skipping", "gvk", resource.GroupVersionKind)
```
mrueg: How often does this log? Only at the start up of ksm or regularly?

maksimp13: The log fires once per CRD informer event (add/update/delete), not on every poll tick. `generateMetrics()` only runs when WasUpdated == true - set by the informer on any CRD change, reset after each run. So it runs once at startup, then only if the cluster CRD state changes.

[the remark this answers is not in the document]
mrueg: @maksimp13 can you address this comment?

bhope (commented):
/lgtm
/hold for others to review

mrueg:
/hold cancel 

Thanks for your contribution!

PR #2907, opened 2026-03-25
fix(cronjob): replace panic on timezone load failure with graceful error skip

### LINKED ISSUE #2898: When a CronJob has spec.timeZone set and the kube-state-metrics container does not have tzdata installed, kube-state-metrics panics and crashes, taking down all cluster metrics collection.
**What happened**:

When a CronJob has `spec.timeZone` set to a non-UTC timezone (e.g. `Asia/Singapore`) and the
kube-state-metrics container image does not include tzdata, kube-state-metrics panics and enters
`CrashLoopBackOff`, taking down all cluster metrics collection.

The panic originates in `internal/store/cronjob.go:254` where `getNextScheduledTime()` prefixes
the schedule with `CRON_TZ=<timezone>`, passes it to `cron.ParseStandard()`, which calls
`time.LoadLocation()` — failing without tzdata — and the error is passed directly to `panic(err)`
with no recovery.

The error should be logged and the `kube_cronjob_next_schedule_time` metric skipped for the
affected CronJob. The process should continue running and serving metrics for all other resources.

```bash
# Apply any CronJob with spec.timeZone set to a named timezone
kubectl apply -f - <<EOF
apiVersion: batch/v1
kind: CronJob
metadata:
  name: tz-test
spec:
  schedule: "0 4 * * 1-5"
  timeZone: "Asia/Singapore"
  jobTemplate:
    spec:
      template:
        spec:
          containers:
          - name: test
            image: busybox
            command: ["echo", "hello"]
          restartPolicy: OnFailure
EOF

# kube-state-metrics will panic immediately with:
# panic: Failed to parse cron job schedule 'CRON_TZ=Asia/Singapore 0 4 * * 1-5':
#        provided bad location Asia/Singapore: unknown time zone Asia/Singapore
```

`spec.timeZone` is a stable Kubernetes API field since v1.27. The `panic(err)` in `cronjob.go:254`
means any cluster using this field with kube-state-metrics deployed without tzdata will lose all
metrics. The fix is two-fold:

1. Replace `panic(err)` with a logged error that skips the metric for the affected CronJob
2. Include tzdata in the container image so named timezones resolve correctly

Suggested fix for `cronjob.go:254`:

```go
if err != nil {
    klog.Errorf("failed to compute next schedule time for cronjob %s/%s: %v",
        j.Namespace, j.Name, err)
    return &metric.Family{Metrics: ms}
}
```

**Environment**:

* kube-state-metrics version: *(please fill in)*
* Kubernetes version: `v1.32.12`
* Cloud provider or hardware configuration: AWS EKS
* Other info: kube-state-metrics container image does not include tzdata; `time.LoadLocation()` fails for any named timezone

aryanputta:

When a CronJob has `spec.timeZone` set to a named timezone (e.g. `Asia/Singapore`) and the kube-state-metrics container image does not include `tzdata`, `cron.ParseStandard` cannot resolve the location. The error was passed directly to `panic(err)` in the `kube_cronjob_next_schedule_time` generator, crashing the process and taking down all cluster metrics collection.

This change replaces `panic(err)` with `klog.Errorf` so the failure is logged at error level and the `kube_cronjob_next_schedule_time` metric is silently skipped for the affected CronJob. All other CronJobs and resource types continue serving metrics normally. Also adds `"k8s.io/klog/v2"` to the import block, which was previously absent from this file.

**How does this change affect the cardinality of KSM:** Does not change cardinality. The affected metric is omitted for CronJobs whose timezone cannot be resolved; it was previously not emittable at all because the process crashed.

**Which issue(s) this PR fixes:** Fixes #2898

**Special notes for your reviewer:** The fix is two lines in `internal/store/cronjob.go`. No new tests needed beyond running existing tests with a CronJob that has a named timezone -- the process no longer panics.

### DIFF
--- a/internal/store/cronjob.go
+++ b/internal/store/cronjob.go
@@ -31,6 +31,8 @@ import (
 	"k8s.io/client-go/tools/cache"
 	basemetrics "k8s.io/component-base/metrics"
 
+	"k8s.io/klog/v2"
+
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
 )
@@ -251,8 +253,10 @@ func cronJobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				// If the cron job is suspended, don't track the next scheduled time
 				nextScheduledTime, err := getNextScheduledTime(j.Spec.Schedule, j.Status.LastScheduleTime, j.CreationTimestamp, j.Spec.TimeZone)
 				if err != nil {
-					panic(err)
-				} else if j.Spec.Suspend == nil || !*j.Spec.Suspend {
+					klog.ErrorS(err, "Failed to compute next schedule time for cronjob", "namespace", j.Namespace, "cronjob", j.Name, "schedule", j.Spec.Schedule, "timezone", j.Spec.TimeZone)
+					return &metric.Family{Metrics: ms}
+				}
+				if j.Spec.Suspend == nil || !*j.Spec.Suspend {
 					ms = append(ms, &metric.Metric{
 						LabelKeys:   []string{},
 						LabelValues: []string{},
--- a/internal/store/cronjob_test.go
+++ b/internal/store/cronjob_test.go
@@ -533,13 +533,16 @@ func TestCronJobStore(t *testing.T) {
 func TestGetNextScheduledTime(t *testing.T) {
 
 	testCases := []struct {
+		Desc             string
 		schedule         string
 		lastScheduleTime metav1.Time
 		createdTime      metav1.Time
 		timeZone         string
 		expected         time.Time
+		expectErr        bool
 	}{
 		{
+			Desc:             "valid schedule with UTC timezone",
 			schedule:         "0 */6 * * *",
 			lastScheduleTime: metav1.Time{Time: ActiveRunningCronJob1LastScheduleTime},
 			createdTime:      metav1.Time{Time: ActiveRunningCronJob1LastScheduleTime},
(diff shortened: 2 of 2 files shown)


mrueg:
@aryanputta can you ensure the CLA is signed? Thanks!

aryanputta:
I think I did it, sorry for thr trouble.

aryanputta:
> /check-cla

i fixed it, I think the bot was glitched cuz of copilot. Sorry for the trouble..

bhope (changes requested):
@aryanputta Thank you for the PR! Could you add a test case covering the invalid timezone path? The existing getNextScheduledTime unit tests ignore the returned error entirely, so the new code path introduced by your change has no test coverage.

aryanputta:
Thanks for the review @bhope. Pushed a follow-up that:

- captures the previously-discarded `err` from `getNextScheduledTime`
- adds a case with a bad named timezone (`Not/A_Real_Zone`) that exercises the `cron.ParseStandard` failure path the PR's new logging branch handles
- adds a case with a malformed schedule string for symmetry on the error contract
- asserts a non-nil error and a zero-value `time.Time` on the error path, and asserts no unexpected error on success

Kept the existing table-driven style (`Desc` field, plain for-loop, no testify) to match the rest of `internal/store`.

aryanputta:
Rebased onto the current \ and force-pushed. The requested test coverage from the earlier review is already included on this branch, so this should be ready for another look. @bhope @dgrisonnet

aryanputta:
Rebased onto current main and force-pushed. The requested test coverage is already included on this branch, so this should be ready for another look. @bhope @dgrisonnet

bhope:
Hi @aryanputta - Thank you so much for your contribution to this issue. The fix for this was covered under broader efforts in #2978 which is now merged.

Closing this PR.

PR #2916, opened 2026-04-07
chore: modernize using slices, min builtin, and %w error wrapping

mrueg:

Mechanical modernization of code that predates newer standard library facilities. No behavior change.

- Replace `sort.StringSlice` + `.Sort()` with `[]string` + `slices.Sort()`
- Replace `sort.Slice` with `slices.SortFunc` (typed comparator)
- Replace `sort.Strings` with `slices.Sort` across multiple packages and tests
- Replace `math.Min(float64(len(...)))` with the `min()` builtin (Go 1.21)
- Replace `fmt.Errorf("%v", err)` with `%w` so error chains survive wrapping
- Replace `reflect.DeepEqual` with `slices.Equal` where the operands are slices
- Replace the hand-rolled `copyList` helper with `maps.Clone`
- Update tests to compare error messages rather than concrete error types

**How does this change affect the cardinality of KSM:** does not change cardinality.

**Which issue(s) this PR fixes:** none.

---

This PR was originally larger. It has been split so the mechanical change is reviewable on its own:

- #3059 — `fix(pod)`: nil `*metric.Metric` panic on unparseable pod IPs (was buried in the modernize commit here, and owned the `k8s.io/utils/net` → `net/netip` swap)
- #3060 — `perf`: header precomputation, `mapToPrometheusLabels` allocations, and single-buffer family rendering, with benchstat numbers

All three branch off `main` independently. #2916 and #3060 both touch `internal/store/utils.go`, `internal/store/utils_test.go` and `pkg/metrics_store/metrics_writer.go`, but they merge cleanly in either order (test-merged).

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved error reporting across configuration, startup, metric collection, and output operations while preserving underlying error details.
  - Ensured metric labels, values, statuses, and resource listings appear in a consistent order across outputs and environments.
  - Improved allow/deny list handling, including reliable behavior when no initial settings are provided.

- **Tests**
  - Updated automated coverage for deterministic results and more precise error handling.

### DIFF
--- a/pkg/allowdenylist/allowdenylist.go
+++ b/pkg/allowdenylist/allowdenylist.go
@@ -18,6 +18,7 @@ package allowdenylist
 
 import (
 	"errors"
+	"maps"
 	"strings"
 	"sync"
 	"time"
@@ -59,10 +60,10 @@ func New(allow, deny map[string]struct{}) (*AllowDenyList, error) {
 
 	// Default to denylisting
 	if len(allow) != 0 {
-		list = copyList(allow)
+		list = maps.Clone(allow)
 		isAllowList = true
 	} else {
-		list = copyList(deny)
+		list = maps.Clone(deny)
 		isAllowList = false
 	}
 
@@ -171,10 +172,4 @@ func (l *AllowDenyList) Test(generator generator.FamilyGenerator) bool {
 	return isIncluded
 }
 
-func copyList(l map[string]struct{}) map[string]struct{} {
-	newList := map[string]struct{}{}
-	for k, v := range l {
-		newList[k] = v
-	}
-	return newList
-}
+
--- a/pkg/allowdenylist/allowdenylist.go
+++ b/pkg/allowdenylist/allowdenylist.go
@@ -41,6 +41,7 @@ type AllowDenyList struct {
 	list        map[string]struct{}
 	rList       []*regexp.Regexp
 	isAllowList bool
+	cache       *sync.Map
 }
 
 // New constructs a new AllowDenyList based on a allow- and a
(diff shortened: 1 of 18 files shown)


mrueg:
Split this into three PRs so the mechanical change and the behavioral ones can be reviewed separately:

- #3059 — `fix(pod)`: the nil `*metric.Metric` panic on unparseable pod IPs, which was buried in the modernize commit here.
- #3060 — `perf`: header precomputation, `mapToPrometheusLabels`, and single-buffer family rendering, each with benchstat numbers.

This PR now contains only the no-behavior-change modernization (13 files, +78/-84).

pkg/allowdenylist/allowdenylist.go
```diff
@@ -59,10 +60,10 @@ func New(allow, deny map[string]struct{}) (*AllowDenyList, error) {
...
 	// Default to denylisting
 	if len(allow) != 0 {
-		list = copyList(allow)
+		list = maps.Clone(allow)
 		isAllowList = true
 	} else {
-		list = copyList(deny)
+		list = maps.Clone(deny)
 		isAllowList = false
 	}
```
bhope, line 65:
```suggestion
    if len(allow) != 0 {
            list = maps.Clone(allow)
            isAllowList = true
    } else {
            list = maps.Clone(deny)
            isAllowList = false
    }
    if list == nil {
            list = map[string]struct{}{}
    }
```
This is ensure code doesn't panic later on nil list as Clone returns nil on empty input

mrueg: Good catch, fixed it! Thanks

bhope:
/lgtm

Thanks

PR #2925, opened 2026-04-14
fix(deps): bump google.golang.org/grpc to v1.79.3

sturman:
Fixes CVE-2026-33186: gRPC-Go authorization bypass via missing leading
slash in :path pseudo-header (CVSS 9.1 Critical).

- Bumps `google.golang.org/grpc` from v1.75.1 to v1.79.3
- Fixes **CVE-2026-33186**: gRPC-Go authorization bypass via missing leading slash in `:path` pseudo-header (CVSS 9.1 Critical)
- Transitive dependency updates required by the grpc upgrade (cel.dev/expr, grpc-gateway, go-jose, go-spiffe, etc.)

## References
- https://nvd.nist.gov/vuln/detail/CVE-2026-33186
- https://github.com/grpc/grpc-go/releases/tag/v1.79.3

Does not change cardinality. This is a dependency-only update with no changes to metrics or labels.

N/A — no existing issue for this CVE.


mrueg:
Thanks for the heads up, I don't think this is in the active code path. Govulncheck didn't report it.
Anything I'm missing here?

sturman:
> Thanks for the heads up, I don't think this is in the active code path. Govulncheck didn't report it. Anything I'm missing here?

You're right that govulncheck doesn't report it at the symbol level — the vulnerable functions aren't directly called. Running `govulncheck -show verbose` on `main` does show it at the **package level** though:
[GO-2026-4762](https://pkg.go.dev/vuln/GO-2026-4762?q)
 ```
❯ govulncheck -show verbose ./...
...
...
=== Package Results ===

Vulnerability #1: GO-2026-4762
    Authorization bypass in gRPC-Go via missing leading slash in :path in
    google.golang.org/grpc
  More info: https://pkg.go.dev/vuln/GO-2026-4762
  Module: google.golang.org/grpc
    Found in: google.golang.org/grpc@v1.75.1
    Fixed in: google.golang.org/grpc@v1.79.3
...
...
```

So the vulnerable package is imported (via otelgrpc/otlptracegrpc) even though the specific authorization bypass symbols aren't invoked. The bump is low-risk — all tests pass and it's just transitive dependency updates.

mrueg:
Thanks for checking again.

We can bump it to reduce noise.

/lgtm

Issue #2679, opened 2025-06-03
KSM as re-packaged metrics-exporter


bavarianbidi:
**What would you like to be added**:

KSM default configuration generation should be moved into a more generic solution.
Currently it's indirectly done defining and parsing the cli flags in https://github.com/kubernetes/kube-state-metrics/blob/main/pkg/options/options.go#L102

**Why is this needed**:

Assume you are a provider of an own controller with own CRDs.
With that you want to ship a very opinionated monitoring-binary which is in fact a re-packaged `ksm`.

example code:

```golang
package main

import (
	"context"

	ksm "k8s.io/kube-state-metrics/v2/pkg/app"
	"k8s.io/kube-state-metrics/v2/pkg/options"
)

var (
	//go:embed resource_config.yaml
	resourceConfig string
)

func main() {
	ctx := context.Background()

	opts := options.NewOptions()
	opts.Kubeconfig = "/tmp/my-kubeconfig"
	opts.CustomResourcesOnly = true
	opts.CustomResourceConfig = resourceConfig

	opts.Port = 8080
	// opts.ServerReadTimeout = 60 * time.Second
	// opts.ServerWriteTimeout = 60 * time.Second
	// opts.ServerIdleTimeout = 5 * time.Minute
	// opts.ServerReadHeaderTimeout = 5 * time.Second
	opts.TotalShards = 1

	ksm.RunKubeStateMetricsWrapper(ctx, opts)
}
```

You have to know all the values from the `Options` struct, otherwise your self-build KSM won't start (e.g. `opts.TotalShards` is an `int` and without setting it explicitly to `1`, the default value will be `0` ... and with that you will never get any metrics if you try to scrape them.

**Describe the solution you'd like**

`options.NewOptions()` is doing the default-value generation.

**Additional context**

This is some kind of a theoretical problem. I've stumbled over that once I've supported a User in KSM slack
 with his very custom deployment method. (xref: [slack thread in `kube-state-metrics`](https://kubernetes.slack.com/archives/CJJ529RUY/p1748928466003409))

If this seems to be a valid use-case, I'm happy to take over the necessary implementation steps.

mrueg:
It's an interesting approach for sure, my main concern is that the project lacks already maintainer time for the stand alone. We currently do not recommend to use ksm as a library, even if others are doing it as we don't want to/can't guarantee stability of an API.

bavarianbidi:
[...]

PR #2928, opened 2026-04-15
fix: make `discovery` package public

L3n41c:

Move `internal/discovery/` to `pkg/discovery/` so that external consumers of the `kube-state-metrics` library can use the CRD discovery mechanism.

I work on a project that leverages `kube-state-metrics` to collect data about the Kubernetes objects of a cluster.
In order to support huge clusters, we do not spawn a `kube-state-metrics` process from which we scrape OpenMetrics data that have to be parsed. Instead, we use `kube-state-metrics` as a library to get the data directly thanks to a custom store.
This works extremely well for all metrics but CRDs because the discoverer is currently inaccessible.

The `customresourcestate.FromConfig()` function requires a `*discovery.CRDiscoverer` parameter. Since `discovery` is an `internal` package, external consumers cannot construct or pass this type. Moving it to `pkg/` makes the API usable by projects that embed `kube-state-metrics` as a library.

**How does this change affect the cardinality of KSM:** It doesn’t.


mrueg:
The intention of this project still is to be run standalone, even if third parties include pieces of it in their tooling. We still don't have the capacity to support a stable library for ksm. 

https://github.com/kubernetes/kube-state-metrics/issues/2679#issuecomment-2939022916

L3n41c:
Hello @mrueg,

Thank you for your reply!
We are fully aware that `kube-state-metrics` aims at being used as a standalone binary and that, as such, internal APIs are not stable and can change at any time.
That’s a trade-off that we’ve accepted so far and that explains why we’ve implemented our needs ourselves in the past.
Although I understand that the change proposed in this PR doesn’t bring anything for the supported use case of standalone binary, I’m wondering if it would be considered as too intrusive to be accepted.
In the end `internal` or not packages don’t make any difference for standalone binaries.

CatherineF-dev:
Agree with @mrueg

@L3n41c Could you list the problems you have without this change. Let's see whether we can have other alternative solutions.

PR #2929, opened 2026-04-16
feat: Add public interface to use the CRD discoverer

lavigne958:
Add a plublic go API to use the module in other go programs.

Add a small to ensure the API is persisted and does not change.


lavigne958:
reopen PR properly 🙃

PR #2937, opened 2026-04-27
fix: Skip landing page registration on NewLandingPage error

carterpewpew:

When `web.NewLandingPage()` returns an error, the handler value is nil and the API is telling us not to use it. The previous code logged the error but still called `mux.Handle("/", landingPage)`, registering a nil `*web.LandingPageHandler`. Because a nil pointer stored in an `http.Handler` interface is non-nil in Go, the default mux accepts it silently. The first `GET /` then panics when `ServeHTTP` dereferences the nil receiver (e.g. reading `routePrefix`).

This change:
- Adds a `newLandingPage` function parameter to `buildTelemetryServer()` and `buildMetricsServer()` so the landing page factory is injected per call. Production code passes `web.NewLandingPage`; tests can inject a failing factory without mutable global state or data races.
- Returns the mux after logging, without registering the `/` handler, when the factory returns an error. The rest of the server (metrics, health, pprof, etc.) continues to work and the process does not crash on the root path if landing page construction fails.
- Adds focused tests for both the happy path (landing page registered, `/` returns 200) and the error path (landing page skipped, `/` returns 404, other routes like `/metrics` and `/healthz` still work). All tests are safe to run with `t.Parallel()` and pass `go test -race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved server startup handling when landing pages cannot be created.
  * Prevented invalid landing-page handlers from being registered.
  * Ensured telemetry and metrics servers serve valid landing pages correctly.

### DIFF
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -346,7 +346,7 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 		)
 	}
 
-	telemetryMux := buildTelemetryServer(ksmMetricsRegistry, opts.AuthFilter, kubeConfig)
+	telemetryMux := buildTelemetryServer(ksmMetricsRegistry, opts.AuthFilter, kubeConfig, web.NewLandingPage)
 	telemetryListenAddress := net.JoinHostPort(opts.TelemetryHost, strconv.Itoa(opts.TelemetryPort))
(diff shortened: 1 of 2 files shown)


mrueg:
@carterpewpew if you have time, can you fix the CI issues?

PR #2940, opened 2026-04-28
fix: Ignore non-served CRD versions in discovery

boomer41:

When an API version is still available, but not served by the API server, any usage of that will result in errors.
Thus, ignore the CRD versions that are not served.

This happened with ExternalSecrets.
It still adds v1beta1 to the CRD definition, but sets it as non-stored and non-served.
If you have the version for Kube State Metrics set to `*`, it will repeatedly fail with the following message:

```
"Failed to watch" err="failed to list external-secrets.io/v1beta1, Kind=ExternalSecret: the server could not find the requested resource" logger="UnhandledError" reflector="pkg/mod/k8s.io/client-go@v0.34.3/tools/cache/reflector.go:290" type="external-secrets.io/v1beta1, Kind=ExternalSecret"
```

It removes CRD versions from the discovery which are not served by the API server (and any request will fail), but still defined in the CRD.

Fixes #

## Summary by CodeRabbit

- **Bug Fixes**
  - Custom-resource metrics now process only API-served versions.
  - Metrics are skipped for unserved or unavailable versions and logged for visibility.
  - Improved handling of deleted resources and unsupported objects during discovery.

- **Documentation**
  - Documented version-serving requirements and the resulting metric behavior.
  - Clarified that metrics remain disabled until a matching served version is available.


mrueg:
change makes sense to me. wdyt @rexagod

boomer41:
> change makes sense to me. wdyt @rexagod

Hi @mrueg,
before I start looking that the AI comments, is the general idea now acceptable?

Thing is, Copilot talks about missing Unit Tests, where no Unit Tests exist at all for the code referenced.
Also, I just followed the style of code around that area.

boomer41:
push

mrueg:
@boomer41 it makes sense to me if they are not served. WDYT @rexagod ?

bhope (changes requested):
The approach looks sound to me, I like that it's targeting the root cause (KSM not checking served) rather than the symptom (the resulting errors).

@boomer41 could you address the Copilot's comments, especially the one regarding extracting extractGVKPs into a package level function and add a test with mixed served true/false versions?

PR #2941, opened 2026-04-28
fix: bump go-jose/v4 to v4.1.4 for CVE-2026-34986

marioferh:

Bumps the indirect dependency github.com/go-jose/go-jose/v4 from v4.1.3 to v4.1.4. That release fixes a denial-of-service where decrypting a malformed JWE (key-wrapping alg with an empty encrypted_key) could panic. kube-state-metrics does not use JWE directly; the library is pulled in through the module graph (e.g. tooling / transitive deps). Updating removes the vulnerable version from the resolved graph.

No effect. This is a dependency-only change; it does not add, remove, or relabel metrics.

Internal security / dependency hygiene (go-jose JWE DoS, fixed in ≥ v4.1.4). Replace with your tracker ID or N/A if you do not link public issues.


rexagod:
/approve

Thank you for this PR, will merge once tests pass.

PR #2956, opened 2026-05-12
build: replace Docker-based image build with goreleaser/ko

mrueg:
Replace the manual multi-arch Docker build process with GoReleaser's `ko` integration, which builds multi-arch container images directly from Go source without requiring a Dockerfile.

Changes:
- `.goreleaser.yaml`: add `builds` section with ldflags matching Makefile behavior, and a `kos` section for multi-arch container builds (`amd64`, `arm/v6`, `arm/v7`, `arm64`, `ppc64le`, `s390x`) using `gcr.io/distroless/static-debian13:nonroot` as the base image.
- `Makefile`: replace Docker-based container/push targets with GoReleaser equivalents; `container` builds a local snapshot; add `all-container` alias; `push` runs GoReleaser release; update `clean` target to remove `dist/`.
- `.github/workflows/ci.yml`: update build job to use `goreleaser-action` with `fetch-depth: 0` on checkout.
- `.github/workflows/pre-release.yml`: remove stale Dockerfile reference.
- `scripts/generate-help-text.sh`: accept `GOOS` and `GOARCH` from environment with fallbacks to `go env`.
- `Dockerfile`: remove; `ko` builds directly from source (default ports `8080`/`8081` are set in options).
- `SECURITY-INSIGHTS.yml`: update Dockerfile reference to `.goreleaser.yaml`.
- `tests/e2e.sh`: remove architecture suffix from image name to match `ko` output.
- `.gitignore`: add `/dist` (GoReleaser build output).

None

Fixes #


Makefile
```diff
@@ -114,31 +102,13 @@ test-benchmark-compare-release:
 
 all: all-container
```
bhope, line 115: After testing this locally,  all-container has no recipe in the current version, so `make all` is essentially a no-op currently. Suggested a fix below that correctly tries to run goreleaser

Makefile
```diff
@@ -114,31 +102,13 @@ test-benchmark-compare-release:
...
-# Container push, push is the target to push for multiple architectures as defined in ALL_ARCH
-
-push: $(addprefix sub-push-,$(ALL_ARCH)) push-multi-arch;
-
-sub-push-%: container-% do-push-% ;
+# Container build and push using goreleaser / ko
 
-do-push-%:
-	${DOCKER_CLI} push $(IMAGE)-$*:$(TAG)
+container:
```
bhope, line 121:
```suggestion
all-container: container

container:
```
Adding this alias ensures that it attempts to build all containers. Let me know if this makes sense.

bhope (commented):
Left an inline for `all: all-container`. Otherwise the change looks good 👍

bhope (commented):
Thanks for this fix. It looks to be in a good shape now. `make build` is succeeding locally.
/lgtm

Feel free to unhold!

bhope:
Thanks for refactoring and getting this to a good state :) 

/lgtm
/hold

Feel free to unhold!

PR #2960, opened 2026-05-19
fix: Wildcard label/annotation clobbering

### LINKED ISSUE #2488: labels_allow_list and annotations_allow_list wildcards clobber resource specific configuration
**What happened**:

When configuring kube-state-metrics to emit `kube_{resourcekindplural}_labels` info-metrics, I found that a `*` entry in the label allow map *clobbers all configuration for all other resource types*.

Expressed in kube-state-metrics yaml config format rather than CLI format for readability, for example, adding the wildcard in the below:

```yaml
    labels_allow_list:
      "*":
        - some_custom_label
      nodes:
        # AWS EKS node pool
        - eks.amazonaws.com/nodegroup
        # Azure AKS node pool
        - agentpool
        # google GKE node pool
        - cloud.google.com/gke-nodepool
```

will cause `kube_node_labels` to have _only_ `label_some_custom_label`.

The addition of the wildcard _removes_ the previously present `label_eks_amazonaws_com_nodegroup`, `label_agentpool` and `label_cloud_google_com_gke_nodepool`.

So the net result of this configuration is equivalent to:

```yaml
    labels_allow_list:
      nodes:
        - some_custom_label
```

I expected the resource-specific configuration to either _append to_ or _override_ the wildcard configuration, so the net effective configuration would be either:

```yaml
    labels_allow_list:
      nodes:
        # labels from "*"
        - some_custom_label
        # labels from "nodes"
        - eks.amazonaws.com/nodegroup
        - agentpool
        - cloud.google.com/gke-nodepool
      otherresource:
        # labels from "*"
        - some_custom_label
      # ...
```

or (if node-specific overrides `*` rather than appending:

```yaml
    labels_allow_list:
      nodes:
        # labels from "nodes"
        - eks.amazonaws.com/nodegroup
        - agentpool
        - cloud.google.com/gke-nodepool
      otherresource:
        # labels from "*"
        - some_custom_label
      # ...
```

Run `kube-state-metrics` with CLI arguments for wildcard added:

```
"--metric-labels-allowlist=nodes=[kubernetes.io/arch],*=[somenonexistentlabel]"
```

and query its metrics endpoint over port-forward (or use prometheus) e.g.

```
curl -sSLf1 http://127.0.0.1:8080/metrics |grep ^kube_node_labels
```

you will note that the node labels do _not_ contain `label_kubernetes_io_arch`.

Now re-launch `kube-state-metrics`, but this time with CLI arguments omitting the wildcard:

```
"--metric-labels-allowlist=nodes=[kubernetes.io/arch]"
```

If you query the metrics endpoint, the `label_kubernetes_io_arch` label will appear on the metrics.

This looks like it was intentional, per https://github.com/kubernetes/kube-state-metrics/blame/c864c93606db61e1c424b9313da03522f9f11adb/internal/store/builder.go#L235-L239

It was added in https://github.com/kubernetes/kube-state-metrics/commit/0b76e7d4f484e8eb7f71f06c66b679ab9e119d11#diff-a1639ee623bffb002ce1b1d3d18893f1d3ca6460a15d030cd272281f3126a7be

It just doesn't make sense though, there's no purpose having support for both a wildcard and non-wildcard when the wildcard unconditionally clobbers the non-wildcards.

It should either:

* use the resource-kind-specific config if found, and fall back to the wildcard if not (recommended); or
* append the wildcard config to the resource-specific config and do a unique sort

The former option is preferred because it allows the config author to say "add these labels to all resource kinds, except for this specific kind where I want to leave some of them out".

**Environment**:

* kube-state-metrics version: `registry.k8s.io/kube-state-metrics/kube-state-metrics:v2.13.0`
* Kubernetes version (use `kubectl version`): server v1.29.2
* Cloud provider or hardware configuration: Repro'd on `kind`, but seen on CSP-managed k8s too
* Other info: 

```[tasklist]
### Tasks
```

ringerc:
Wildcard entries in `labels_allow_list` and `annotations_allow_list` clobbered resource-specific entries per https://github.com/kubernetes/kube-state-metrics/issues/2488. So it was not possible to express the configuration "label A for all resource types, and additionally label B for this specific resource type".

Fix by merging the wildcard and resource-specific discovered labels.

No effect on cardinality _except_ possibly if an existing configuration has masked config due to this bug; if any  annotations (or labels, where on mutable resources) have high rates of churn on underlying resources, the newly-unclobbered labels might increase cardinality. This is really a user misconfiguration, but was previously masked by this bug.

Fixes #2488

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -232,7 +232,7 @@ func (b *Builder) WithCustomResourceStoreFactories(fs ...customresource.Registry
 }
 
 // allowList validates the given map and checks if the resources exists.
-// If there is a '*' as key, return new map with all enabled resources.
+// If there is a '*' as key, merge wildcard labels with resource-specific labels.
 func (b *Builder) allowList(list map[string][]string) (map[string][]string, error) {
 	if len(list) == 0 {
 		return nil, nil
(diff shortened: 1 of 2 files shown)


bhope:
Thanks for the PR to fix this. 

With this change, users who had both * and resource-specific entries configured may see additional labels appearing after this fix since the old behavior was silently dropping the resource-specific ones. Worth calling this out.

PR #2961, opened 2026-05-20
fix: Remove deprecated Endpoints store

rexagod:
The v1 Endpoints API is deprecated in Kubernetes 1.33+ in favor of discovery.k8s.io/v1 EndpointSlice. When the endpoints resource is enabled, every List/Watch call triggers a deprecation warning that floods logs with ~10K+ repeated messages, obscuring genuinely important entries.

The endpointslices store has been the default since v2.14.0 and provides a superset of functionality. This removes the endpoints store entirely to eliminate the deprecation warning flood.


mrueg:
@rexagod we deprecated it in v2.17.0, should we remove it in v2.20.0 or mention the removal there and remove it in v2.21.0?

rexagod:
Ah, good catch. I'll open a PR to explicitly log a warning about the removal in the next-to-next release, when someone uses the Endpoints store.

Closing this for now, will revisit once we start preparing for that 2.21.

mrueg:
feel free to keep it open and mark it as /hold :)

PR #2969, opened 2026-05-26
feat: Add OutOfcpu and OutOfmemory to kube_pod_status_reason

### LINKED ISSUE #2968: Add OutOfcpu and OutOfmemory to kube_pod_status_reason metric

When pods fail due to CPU or memory resource exhaustion, Kubernetes sets the pod status reason to OutOfcpu or OutOfmemory. However, kube_pod_status_reason currently only tracks: Evicted, NodeAffinity, NodeLost, PreemptionByScheduler, SchedulingGated, Shutdown, TerminationByKubelet, UnexpectedAdmissionError.

This means there is no way to detect or alert on resource-exhaustion pod failures using kube-state-metrics, creating a significant gap in cluster observability for capacity planning and application health monitoring.

Extend the podStatusReasons slice with the two missing reasons, following the same pattern as #2880 (SchedulingGated addition):

podStatusReasons = []string{"Evicted", "NodeAffinity", "NodeLost", "OutOfcpu", "OutOfmemory", "PreemptionByScheduler", "SchedulingGated", "Shutdown", "TerminationByKubelet", "UnexpectedAdmissionError"}
I'm happy to submit a PR for this.

**Additional context:**

Minimal change — same approach as #2880
No new metrics, no API changes, no cardinality risk (just 2 additional fixed label values)
These are standard Kubernetes reasons set by the kubelet when pod resource limits are exceeded

krish0005:
Extends the hardcoded podStatusReasons list to include OutOfcpu and
OutOfmemory, enabling observability of resource-exhaustion pod failures.

This follows the same pattern as #2880 (SchedulingGated).

Fixes #2968

## Summary by CodeRabbit

* **New Features**
  * Pod metrics now report dedicated status reasons for pods that run out of CPU or memory, instead of grouping them under “Other.”

* **Documentation**
  * Updated metric documentation to include the new CPU and memory exhaustion status reasons.

* **Tests**
  * Added coverage verifying both status reasons are emitted correctly.


krish0005:
Hi @CatherineF-dev and @nmn3m do you have any estimate to perform review on this PR ? 
Thanks

internal/store/pod_test.go
```diff
@@ -1304,6 +1304,8 @@ func TestPodStore(t *testing.T) {
 				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="Evicted",uid="uid4"} 0
 				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="NodeAffinity",uid="uid4"} 0
 				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="NodeLost",uid="uid4"} 1
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="OutOfcpu",uid="uid4"} 0
```
CatherineF-dev, line 1307: Add test case to make it non-zero?

krish0005: Added test cases for both OutOfcpu and OutOfmemory with value=1. Thanks for the feedback! @CatherineF-dev

mrueg:
Okay, I think at this point we should reconsider the approach here to reduce cardinality. Maybe we should not use a stateset here @dgrisonnet @bhope @CatherineF-dev @rexagod wdyt?

bhope:
> Okay, I think at this point we should reconsider the approach here to reduce cardinality. Maybe we should not use a stateset here @dgrisonnet @bhope @CatherineF-dev @rexagod wdyt?

I Agree, this is probably the right moment to fix the pattern. #2880 (mine) used the same argument of cardinality being bounded but we're likely to keep getting fair requests for more reasons over time and each one compounds the same problem.

I propose that we move to a new standard where we only emit when the pod actually has a reason set (instead of emitting one per known reason). This would decouple cardinality from the number of reasons we support.

Opened #3004  to track this separately. Let's hold this PR until that's resolved, unless others feel differently.

dgrisonnet:
:+1: for moving away from the stateset approach

internal/store/pod.go
```diff
@@ -37,7 +37,7 @@ import (
 
 var (
 	descPodLabelsDefaultLabels = []string{"namespace", "pod", "uid"}
-	podStatusReasons           = []string{"Evicted", "NodeAffinity", "NodeLost", "PreemptionByScheduler", "SchedulingGated", "Shutdown", "TerminationByKubelet", "UnexpectedAdmissionError"}
+	podStatusReasons           = []string{"Evicted", "NodeAffinity", "NodeLost", "OutOfcpu", "OutOfmemory", "PreemptionByScheduler", "SchedulingGated", "Shutdown", "TerminationByKubelet", "UnexpectedAdmissionError"}
 )
```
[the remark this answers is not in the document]
krish0005: Hi! thanks! i updated pod-metrics.md now. see here: https://github.com/kubernetes/kube-state-metrics/pull/2969/changes/dd2021c59ff273240fde888fae045b9de55082ea

krish0005:
@CatherineF-dev  @nmn3m  Hi! what is missing here to get lgtm and approved labels ? thanks

bhope:
Hi @krish0005  - we're rethinking the stateset approach to keep cardinality down (tracked in #3004), so we'll have to hold this until that's settled.

/hold

CatherineF-dev:
Agree on reducing cardinliaty. RAM is more and more expensive..

CatherineF-dev:
I am thinking adding a feature gate --migrated-to-stateset-less-metrics instead of creating a new metric.

The risk seems low because all conditions here are bad condition "Evicted\|NodeAffinity\|NodeLost\|OutOfcpu\|OutOfmemory\|PreemptionByScheduler\|SchedulingGated\|Shutdown\|TerminationByKubelet\|UnexpectedAdmissionError", so consumers will not compute a ratio.

CatherineF-dev:
@krish0005 I have added this topic in next Friday's sig-instrumentation meeting agenda.

internal/store/pod_test.go
```diff
@@ -1545,11 +1561,71 @@ func TestPodStore(t *testing.T) {
...
+				},
+			},
+			Want: `
+				# HELP kube_pod_status_reason The pod status reasons
+				# TYPE kube_pod_status_reason gauge
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="Evicted",uid="uid4"} 0
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="NodeAffinity",uid="uid4"} 0
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="NodeLost",uid="uid4"} 0
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="OutOfcpu",uid="uid4"} 1
+				kube_pod_status_reason{namespace="ns4",pod="pod4",reason="OutOfmemory",uid="uid4"} 0
```
rexagod, line 1594: Can you link to the source for both of these upstream?

CatherineF-dev:
/lgtm 

/hold for [rexagod](https://github.com/rexagod)'s comment.

A future improvement is https://github.com/kubernetes/kube-state-metrics/issues/3004

internal/store/pod_test.go
krish0005: @rexagod  Both are defined as constants in the kubelet's predicate admission handler, 
predicate.go :

```
// These reasons are used to denote that the pod has reject admission
// to the node because there's not enough resources to run the pod.
OutOfCPU              = "OutOfcpu"
OutOfMemory           = "OutOfmemory"
OutOfEphemeralStorage = "OutOfephemeral-storage"
OutOfPods             = "OutOfpods"
They are assigned to pod.Status.Reason in the Admit method when an InsufficientResourceError occurs for the corresponding resource (CPU / memory).
```

krish0005:
@rexagod both OutOfcpu and OutOfmemory are defined as constants in the kubelet's predicate admission handler (
predicate.go
):

```
// These reasons are used to denote that the pod has reject admission
// to the node because there's not enough resources to run the pod.
OutOfCPU              = "OutOfcpu"
OutOfMemory           = "OutOfmemory"
OutOfEphemeralStorage = "OutOfephemeral-storage"
OutOfPods             = "OutOfpods"
```
They're assigned to pod.Status.Reason in the Admit method when an InsufficientResourceError occurs for the corresponding resource. Scoped this PR to CPU/memory as those were the requested reasons in #2968.

krish0005:
Rebased onto latest main and adapted the change to the new stateless kube_pod_status_reason implementation from #3004. OutOfcpu and OutOfmemory are now added to podStatusReasons so they're emitted with their real label instead of falling into Other. Docs and tests updated to match the new format.

@rexagod  I've answered your question above re: the upstream source — could you take another look so the /hold can be lifted? @CatherineF-dev this will also need a fresh /lgtm since the rebase removed the previous one. Thanks both!

bhope:
Thanks for rebasing from #3089 . Verified the source checks out.
/lgtm

krish0005:
@nmn3m can this be merged ?

Issue #2967, opened 2026-05-26
Crashing on "exotic" CronJobs


tun0:
**What happened**:

After upgrading to v2.19.0 it panics when there's a CronJob present with an exotic schedule like `0 1 */32,1-7 * 3`. 

**What you expected to happen**:

Not panic.

**How to reproduce it (as minimally and precisely as possible)**:

Just a create a CronJob resource containing `schedule: 0 1 */32,1-7 * 3`

**Anything else we need to know?**:

After additional investigation, it turns out v2.18.0 is slightly less affected. v2.18.0 only panics when such a cronjob is present at startup. Whereas v2.19.0 also panics when such a CronJob appears while KSM is already running.

**Environment**:

* kube-state-metrics version: v2.19.0
* Kubernetes version (use `kubectl version`): v1.34.6-gke.1307000
* Cloud provider or hardware configuration: GKE
* Other info: -

MarcWort:
Got the same since 2.19.0: 
```
 panic: failed to parse cron job schedule '*/120 4-22 * * *': step (120) must be less than range size (60): "*/120" [recovered, repanicked]
```

mrueg:
We switched from https://github.com/robfig/cron to https://github.com/netresearch/go-cron in v2.19.0 https://github.com/kubernetes/kube-state-metrics/pull/2874 maybe the parsing there is more strict? It might be worth to file an issue with upstream if you can reproduce it with the new library.

tun0:
> We switched from https://github.com/robfig/cron to https://github.com/netresearch/go-cron in v2.19.0 [#2874](https://github.com/kubernetes/kube-state-metrics/pull/2874) maybe the parsing there is more strict? 

That was my first assumption too. But, given the issue is present in both 2.18.0 and 2.19.0, albeit slightly different, I don't think upstream is to blame here? As in: it's effectively already reproduced with the new library.

Both 2.18.0 and 2.19.0 have issues with this cronjob schedule. Those versions do differ in when the panic manifests:

2.18.0: only at startup
2.19.0: at both startup and also whenever such a schedule pops up while already running

Basically, there's 2 issues with possibly a single fix:

* KSM not supporting "exotic" cronjob schedules
* KSM 2.19.0 becoming even more sensitive to such schedules.

mxey:
I'm surprised that kube-state-metrics is not reusing the very same parser that Kubernetes itself uses.

bhope:
[...]

PR #2970, opened 2026-05-26
fix(cronjob): guard nil suspend in next schedule metric

immanuwell:

`kube_cronjob_next_schedule_time` blows up if a CronJob omits `spec.suspend`.
That field is optional in the Kubernetes API, so this can happen for real.
This patch treats `nil` like `false` and adds a regression test, nice and small.

does not change cardinality

Related to #2967

**How to reproduce:**

1. Use a CronJob object with no `spec.suspend` field.
2. Before this patch, run `go test ./internal/store -run TestCronJobStore -count=1`.
3. It panics in `internal/store/cronjob.go` on a nil pointer deref.

**Verification:**

- `go test ./internal/store -run TestCronJobStore -count=1`
- `make test-unit`


mrueg:
/lgtm 

Thanks for your contribution!

PR #2974, opened 2026-06-01
fix: watch-list bookmarks with sharding

### LINKED ISSUE #2973: Missing metrics following 2.19.0 upgrade
**What happened**:
Following KSM upgrade from 2.18.0 to 2.19.0, metric `kube_namespace_status_phase` is missing for some namespaces. In my cluster, with 20 namespaces currently, it lists only 8 distinct namespaces after update. With 2.18.0, all 20 were listed as expected. Same issue with `kube_namespace_created`.
Logs show a lot of warnings that were not present in 2.18.0 (multiple occurences repeating every 10 seconds):
```
I0601 14:33:38.774781       1 reflector.go:1159] "Warning: event bookmark expired" err="pkg/mod/k8s.io/client-go@v0.35.4/tools/cache/reflector.go:289: awaiting required bookmark event for initial events stream, no events received for 10.083216873s"
I0601 14:33:48.730618       1 reflector.go:1159] "Warning: event bookmark expired" err="pkg/mod/k8s.io/client-go@v0.35.4/tools/cache/reflector.go:289: hasn't received required bookmark event marking the end of initial events stream, received last event 16.874743671s ago"
```

`kube_namespace_status_phase` and `kube_namespace_created` should exist for all namespaces, not just a subset.

Not sure if this is environment specific. I simply deployed kube-prometheus-stack 86.1.0 (which comes with kube-state-metrics 2.19.0), with the values specified below, and observed the issue.
```yaml
kube-state-metrics:
  replicas: 2
  autosharding:
    enabled: true
  metricLabelsAllowlist:
    - pods=[*]
    - persistentvolumeclaims=[*]
```
Then I query `count(count by (namespace) (kube_namespace_status_phase))` or check some default prometheus stack dashboards that rely on these metrics.

If I rollback to kube-state-metrics 2.18.0, using the following values, the issue is resolved and all namespaces are listed as expected.
```yaml
kube-state-metrics:
  image:
    tag: v2.18.0
  replicas: 2
  autosharding:
    enabled: true
  metricLabelsAllowlist:
    - pods=[*]
    - persistentvolumeclaims=[*]
```

**Environment**:

* kube-state-metrics version: 2.19.0
* Kubernetes version (use `kubectl version`): 1.35.3
* Cloud provider or hardware configuration: Scaleway Kapsule
* Other info:

Image

jfremy-openai:
KSM 2.19.0 switched to client-go 0.35.4 which enables the `WatchListClient` feature by default.
The sharding code does not handle it properly. 
Bookmark events are processed by the shard filter while they should be propagated to every shard (so they know the initial sync is complete).
As a result when using KSM 2.19.0 with sharding, only one shard can initialize successfully (the one which receives the bookmark event)
This change special case bookmark events and make sure every shard will see them

Fixes #2973

### DIFF
--- a/pkg/sharding/listwatch.go
+++ b/pkg/sharding/listwatch.go
@@ -81,6 +81,12 @@ func (s *shardedListWatch) Watch(options metav1.ListOptions) (watch.Interface, e
 	}
 
 	return watch.Filter(w, func(in watch.Event) (out watch.Event, keep bool) {
+		// Bookmarks are stream control events. They carry a resource version but
+		// no UID, so filtering them would route every bookmark to a single shard.
+		if in.Type == watch.Bookmark {
+			return in, true
+		}
+
 		a, err := meta.Accessor(in.Object)
 		if err != nil {
 			// TODO(brancz): needs logging
--- a/pkg/sharding/listwatch_test.go
+++ b/pkg/sharding/listwatch_test.go
@@ -70,23 +70,31 @@ func TestShardedListWatchPassesInitialEventsEndBookmarkToEveryShard(t *testing.T
 		},
 	}
 
+	source := watch.NewBroadcaster(1, watch.WaitIfChannelFull)
+	t.Cleanup(source.Shutdown)
+
+	var shardedWatches []watch.Interface
 	for shard := int32(0); shard < 4; shard++ {
-		t.Run(strconv.Itoa(int(shard)), func(t *testing.T) {
-			source := watch.NewRaceFreeFake()
-			lw := &cache.ListWatch{
-				WatchFunc: func(metav1.ListOptions) (watch.Interface, error) {
-					return source, nil
-				},
-			}
+		lw := &cache.ListWatch{
+			WatchFunc: func(metav1.ListOptions) (watch.Interface, error) {
+				return source.Watch()
+			},
+		}
 
-			shardedWatch, err := cache.ToWatcherWithContext(NewShardedListWatch(shard, 4, lw)).WatchWithContext(context.Background(), metav1.ListOptions{})
-			if err != nil {
-				t.Fatalf("failed to create sharded watch: %v", err)
-			}
-			defer shardedWatch.Stop()
+		shardedWatch, err := cache.ToWatcherWithContext(NewShardedListWatch(shard, 4, lw)).WatchWithContext(context.Background(), metav1.ListOptions{})
+		if err != nil {
+			t.Fatalf("failed to create sharded watch: %v", err)
+		}
+		t.Cleanup(shardedWatch.Stop)
+		shardedWatches = append(shardedWatches, shardedWatch)
+	}
 
-			source.Action(watch.Bookmark, bookmark)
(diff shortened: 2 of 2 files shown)


jfremy-openai:
> New changes are detected. LGTM label has been removed.

Fixed lint checks

odinuge:
/lgtm

Thanks! We also started hitting this yesterday, and the patch looks good.

pkg/sharding/listwatch.go
```diff
@@ -81,6 +81,12 @@ func (s *shardedListWatch) Watch(options metav1.ListOptions) (watch.Interface, e
 	}
 
 	return watch.Filter(w, func(in watch.Event) (out watch.Event, keep bool) {
+		// Bookmarks are stream control events. They carry a resource version but
+		// no UID, so filtering them would route every bookmark to a single shard.
+		if in.Type == watch.Bookmark {
```
odinuge, line 86: I partially wonder if we should also keep `watch.Error` as well and only filter out `Added`, `Modified` and `Deleted`. But I think this is also fine as-is.

jfremy-openai: I'd keep it as is for now, but I can look into a follow up PR
Let me know if there is anything else needed

jfremy-openai:
@mrueg do you know what else we need to get it merged?

pkg/sharding/listwatch_test.go
```diff
@@ -54,3 +59,42 @@ func TestSharding(t *testing.T) {
...
+				},
+			}
+
+			shardedWatch, err := cache.ToWatcherWithContext(NewShardedListWatch(shard, 4, lw)).WatchWithContext(context.Background(), metav1.ListOptions{})
+			if err != nil {
+				t.Fatalf("failed to create sharded watch: %v", err)
+			}
+			defer shardedWatch.Stop()
+
+			source.Action(watch.Bookmark, bookmark)
```
rexagod: Ideally, we should have the source initialization happen once, with a single stream of events, which are then fanned out to each shard, and verified.

rexagod (commented):
/hold

Thank you for the patch. PTAL my comment below. Everything else LGTM.

@mrueg IMO This warrants for an immediate patch release, since sharding is broken in 2.19.

jfremy-openai:
> /hold
> 
> Thank you for the patch. PTAL my comment below. Everything else LGTM.
> 
> @mrueg IMO This warrants for an immediate patch release, since sharding is broken in 2.19.

I've updated the test to use a single source (Broadcaster) and then it fans out to all watches. Does that work for you?

pkg/sharding/listwatch_test.go
jfremy-openai: updated

jfremy-openai:
@mrueg @rexagod are we good to land? Any additional changes needed?

CatherineF-dev:
I am thinking enable streaming feature in v2.20.0, since v2.19.0 has some security fixes https://github.com/kubernetes/kube-state-metrics/releases/tag/v2.19.0

bhope (approved):
/lgtm

PR #2977, opened 2026-06-03
fix(discovery): Stop the discoverer when the context is canceled

L3n41c:

Fix a goroutine and memory leak when the context passed to `CRDDiscoverer.StartDiscovery` expires.

`StartDiscovery` spawns an informer goroutine when invoked.
It takes a `Context` as parameter and should stop the goroutine when that `Context` expires.

The currently used pattern is: https://github.com/kubernetes/kube-state-metrics/blob/e77145f237f7038fc5c6c4a25c3cb9617d9abba6/internal/discovery/discovery.go#L117-L124

A `for range ch` on a channel waits for values to be fed in the channel. But when the channel is closed, the `for` loop exits without executing its body so that the `stopper` channel is never closed and the informer goroutine is leaked.

I replaced it with:
```go
<-ch
```
which returns when either a value has been enqueued or the channel is closed.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -116,11 +116,9 @@ func (r *CRDiscoverer) StartDiscovery(ctx context.Context, config *rest.Config) 
 	}
 	// Respect context cancellation.
 	go func() {
-		for range ctx.Done() {
-			klog.InfoS("context cancelled, stopping discovery")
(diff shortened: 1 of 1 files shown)


mrueg:
/lgtm 

/hold

@rexagod @bhope this might be a good one for a patch release as well.

PR #2978, opened 2026-06-04
fix: don't panic on CronJobs with unparseable schedules

### LINKED ISSUE #2967: Crashing on "exotic" CronJobs

atosatto:

KSM panics and crashes when a CronJob has a schedule the cron parser rejects, e.g. `*/120 4-22 * * *` (the step `120` exceeds the minute range size of `60`). 

The Kubernetes API/KCM accept and run such schedules, but the `kube_cronjob_next_schedule_time` generator calls `panic(err)` on any parse error, taking down the whole exporter.

This became worse in v2.19.0: with [`robfig/cron`](https://github.com/robfig/cron) (≤ v2.18.0) the crash only happened at startup, but since the switch to [`netresearch/go-cron`](https://github.com/netresearch/go-cron) (#2874) the panic also fires at runtime whenever such a CronJob appears while KSM is already running — so a single user-created CronJob can crash a running exporter.

Instead of panicking on an un-parseable schedule, the next-schedule-time generator now logs the CronJob and skips only that metric, continuing to generate every other metric for the object (and all other CronJobs). This mirrors the `kubernetes/kubernetes` behaviour, which tolerates these CronJobs rather than failing hard.

No changes to cardinality.

Fixes https://github.com/kubernetes/kube-state-metrics/issues/2967


internal/store/cronjob.go
```diff
@@ -251,7 +252,11 @@ func cronJobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				// If the cron job is suspended, don't track the next scheduled time
 				nextScheduledTime, err := getNextScheduledTime(j.Spec.Schedule, j.Status.LastScheduleTime, j.CreationTimestamp, j.Spec.TimeZone)
 				if err != nil {
-					panic(err)
+					// A single CronJob with an unparseable schedule must not crash the
+					// whole exporter. Skip its next-schedule metric and keep going, the
+					// same way the Kubernetes CronJob controller tolerates such schedules.
+					klog.InfoS("Skipping cronjob with unparseable schedule",
```
bhope:
```suggestion
					klog.ErrorS("Skipping cronjob with unparseable schedule",
```

Metric being dropped needs to be more visible.

internal/store/cronjob_test.go
```diff
@@ -563,3 +563,75 @@ func TestGetNextScheduledTime(t *testing.T) {
 	}
 
 }
+
+// TestCronJobStoreScheduleParsing ensures kube_cronjob_next_schedule_time is emitted for
```
bhope: Please remove this and other redundant comments

bhope (changes requested):

internal/store/cronjob.go
atosatto: Changed in [d388ea0](https://github.com/kubernetes/kube-state-metrics/pull/2978/commits/d388ea03d19c62315849d4fc8e592355354e0019)

internal/store/cronjob_test.go
atosatto: I have removed this comment, kept the other ones inline with the tests cause explain the specific test-case

bhope:
@atosatto Can you add a test case for invalid timezone?

atosatto:
> @atosatto Can you add a test case for invalid timezone?

Added to b3048f39f916d202c212f31c3255d9acb86ac7f7

bhope:
Thanks for your contribution and addressing the feedback! Looks good to me now. 
/triage accepted
/lgtm

FYI @mrueg @rexagod -- this is a patch candidate.

bhope (approved):

internal/store/cronjob.go
```diff
@@ -251,7 +252,11 @@ func cronJobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				// If the cron job is suspended, don't track the next scheduled time
 				nextScheduledTime, err := getNextScheduledTime(j.Spec.Schedule, j.Status.LastScheduleTime, j.CreationTimestamp, j.Spec.TimeZone)
 				if err != nil {
-					panic(err)
+					// A single CronJob with an unparseable schedule must not crash the
+					// whole exporter. Skip its next-schedule metric and keep going, the
+					// same way the Kubernetes CronJob controller tolerates such schedules.
+					klog.ErrorS(err, "Skipping cronjob with unparseable schedule",
```
mrueg: Should we expose a metric given that most likely nobody looks at logs?

atosatto: What would it be an acceptable cardinality of the metric? Something per-cronjob or a total count per namespace?

bhope: In this case, users can leverage prometheus `absent()` primitive to detect absence of metric. So a new metric would be redundant. WDYT?

internal/store/cronjob.go
```diff
@@ -251,7 +252,11 @@ func cronJobMetricFamilies(allowAnnotationsList, allowLabelsList []string) []gen
 				// If the cron job is suspended, don't track the next scheduled time
 				nextScheduledTime, err := getNextScheduledTime(j.Spec.Schedule, j.Status.LastScheduleTime, j.CreationTimestamp, j.Spec.TimeZone)
 				if err != nil {
-					panic(err)
+					// A single CronJob with an unparseable schedule must not crash the
+					// whole exporter. Skip its next-schedule metric and keep going, the
```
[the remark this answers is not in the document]
mrueg: this would not be ideal, and thus I think a metric would be better.

internal/store/cronjob.go
mrueg: how would you use that? Maybe we should pivot towards a more flexible parsing mechanism https://github.com/kubernetes/kubernetes/blob/42d9ed9c47706cb04457410e387e07889203b888/pkg/util/parsers/parsers.go#L59

atosatto: Kubernetes is using a `github.com/robfig/cron/v3` which is the very same library we were replacing in https://github.com/kubernetes/kube-state-metrics/pull/2874. So I am not sure this would a viable option.

`absent()` requires you to know what is the cronjob that is not being emitted, while a metric will allow you to easily spot the misconfigured ones.

atosatto: I have given this a try in 6bf78e8929b1cb2c9bfc656c2228d5aed46d568e adding a new metric `kube_cronjob_spec_schedule_invalid` to capture this. I have also changed back the message to from Error to Warning.

@mrueg  / @bhope  hopes this consolidates both of your feedback

internal/store/cronjob.go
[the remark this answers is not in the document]
atosatto: Removed the log line and replaced with metrics in 476e0848

internal/store/cronjob.go
bhope: @atosatto - Usually the `_spec` metric should mirror a certain spec. Since that's not the case here, can we drop `spec` from the name of this new metric and keep it simple like `kube_cronjob_schedule_invalid`

bhope (changes requested):

internal/store/cronjob.go
atosatto: @bhope changed the metric name to `kube_cronjob_schedule_invalid ` and squashed all of the commits so that we can see a clear diff since we went through quite a few iteration of the PR.

bhope (approved):

atosatto:
> /lgtm

Thanks @bhope - shall we assign this to @rexagod (as per Prow) in order to get this merged or will yours and @mrueg's review be sufficient?

bhope:
/assign @mrueg 

Since you were involved in the review.

PR #2991, opened 2026-06-10
fix(hpa): handle omitted minReplicas in HPA metrics

immanuwell:

Fixes a nil deref in `kube_horizontalpodautoscaler_spec_min_replicas`.
`spec.minReplicas` is optional in the HPA API and defaults to `1`, so omitting it is a valid object shape. Old code could just blow up on `*a.Spec.MinReplicas`. This patch falls back to `1`.

How does this change affect the cardinality of KSM:

does not change cardinality

Which issue(s) this PR fixes:

none

Repro:

1. Apply an HPA without `spec.minReplicas`, for example:
```yaml
apiVersion: autoscaling/v2
kind: HorizontalPodAutoscaler
metadata:
  name: repro
spec:
  scaleTargetRef:
    apiVersion: apps/v1
    kind: Deployment
    name: repro
  maxReplicas: 3
```
2. Scrape the HPA metrics.
3. Before this patch, the `kube_horizontalpodautoscaler_spec_min_replicas` path could panic.
4. After this patch, it exports `1`.

Tests:

`go test ./internal/store -run TestHPAStore -count=1`
`go test $(go list ./... | grep -v '/tests/e2e')`

### DIFF
--- a/internal/store/horizontalpodautoscaler_test.go
+++ b/internal/store/horizontalpodautoscaler_test.go
@@ -260,6 +260,30 @@ func TestHPAStore(t *testing.T) {
 			},
 		},
 		{
+			// Verify omitted minReplicas falls back to the API default.
+			Obj: &autoscaling.HorizontalPodAutoscaler{
+				ObjectMeta: metav1.ObjectMeta{
+					Name:      "hpa-default-min-replicas",
+					Namespace: "ns1",
+				},
+				Spec: autoscaling.HorizontalPodAutoscalerSpec{
+					MaxReplicas: 3,
+					ScaleTargetRef: autoscaling.CrossVersionObjectReference{
+						Kind: "Deployment",
+						Name: "deployment-default-min",
+					},
+				},
+			},
+			Want: `
+				# HELP kube_horizontalpodautoscaler_spec_min_replicas [STABLE] Lower limit for the number of pods that can be set by the autoscaler, default 1.
+				# TYPE kube_horizontalpodautoscaler_spec_min_replicas gauge
+				kube_horizontalpodautoscaler_spec_min_replicas{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 1
+			`,
+			MetricNames: []string{
+				"kube_horizontalpodautoscaler_spec_min_replicas",
+			},
+		},
+		{
 			// Verify populating base metric.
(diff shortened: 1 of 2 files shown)


internal/store/horizontalpodautoscaler_test.go
```diff
@@ -259,6 +259,50 @@ func TestHPAStore(t *testing.T) {
...
+				# TYPE kube_horizontalpodautoscaler_spec_max_replicas gauge
+				# TYPE kube_horizontalpodautoscaler_spec_min_replicas gauge
+				# TYPE kube_horizontalpodautoscaler_status_current_replicas gauge
+				# TYPE kube_horizontalpodautoscaler_status_desired_replicas gauge
+				kube_horizontalpodautoscaler_info{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1",scaletargetref_kind="Deployment",scaletargetref_name="deployment-default-min"} 1
+				kube_horizontalpodautoscaler_metadata_generation{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 0
+				kube_horizontalpodautoscaler_spec_max_replicas{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 3
+				kube_horizontalpodautoscaler_spec_min_replicas{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 1
+				kube_horizontalpodautoscaler_status_current_replicas{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 0
+				kube_horizontalpodautoscaler_status_desired_replicas{horizontalpodautoscaler="hpa-default-min-replicas",namespace="ns1"} 0
```
CatherineF-dev: qq: why is desired_replicas 0 when the minReplicas is 1?

immanuwell: because `kube_horizontalpodautoscaler_status_desired_replicas` is sourced from `status.desiredReplicas`, not from the `spec.minReplicas` default

in this test fixture I left `status` empty because the case was only meant to exercise the `spec.minReplicas == nil` path, so the zero-value status metric rendered as `0`
I've updated the test to narrow it to the behavior under test and only assert `kube_horizontalpodautoscaler_spec_min_replicas`

PR #2996, opened 2026-06-14
fix(discovery): log when a configured custom resource CRD is absent

### LINKED ISSUE #2354: [regression] /metrics port down when not existing CRD are listed in config file

aryanputta:
## What this does

When a fully specified custom resource GVK in the CustomResourceState config is not present in the discovery cache (its CRD is not installed in the cluster), `ResolveGVKToGVKPs` resolved it to an empty set and returned with no log. kube-state-metrics then silently generated no metrics for that resource, with nothing surfaced even at high verbosity.

This adds a `klog.InfoS` in `ResolveGVKToGVKPs` for the fully-specified-GVK path when there is no matching discovery cache entry, naming the GVK so a missing or misconfigured CRD is visible at default verbosity.

Resolution behavior is unchanged (the path still returns an empty set). The existing `fixed version and kind, no matching cache entry` test case in `discovery_test.go` already covers the empty-return contract of this branch and now also exercises the new log.

## Which issue(s) this fixes

Fixes #2354

In the issue discussion, @sebastiangaiser reported that a CRD specified in config but not installed produced no warning/info even at `-v=10`, and @rexagod confirmed logging this is the desired behavior and marked it a good first issue.

## How to verify

```
go test ./internal/discovery/ -run TestGVKMapsResolveGVK -v
```

The `fixed version and kind, no matching cache entry` case now emits:

```
"Configured custom resource was not found in the cluster, no metrics will be generated for it until its CRD is installed" gvk="testgroup/v1, Kind=TestObject1"
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved custom resource discovery so resolved resources no longer keep stale “missing” warnings.
  * Reduced repeated warnings when a resource is unavailable, while still logging once when it can’t be found.
  * Discovery now clears an old missing-resource warning if the resource becomes available again, so future missing events can be reported properly.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -152,6 +152,12 @@ func (r *CRDiscoverer) ResolveGVKToGVKPs(gvk schema.GroupVersionKind) (resolvedG
 				}, nil
 			}
 		}
+		// The GVK was fully specified but is absent from the discovery cache,
+		// which usually means the CRD is not installed in the cluster. Surface
+		// this so a missing or misconfigured CRD is visible instead of silently
+		// generating no metrics for it.
+		klog.InfoS("Configured custom resource was not found in the cluster, no metrics will be generated for it until its CRD is installed", "gvk", gvk)
+		return nil, nil
 	}
 	if hasVersion && !hasKind {
 		kinds := r.Map[g][v]
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -140,6 +140,9 @@ func (r *CRDiscoverer) ResolveGVKToGVKPs(gvk schema.GroupVersionKind) (resolvedG
 	if hasVersion && hasKind {
 		for _, el := range r.Map[g][v] {
 			if el.Kind == k {
+				// The CRD is present now; clear any prior "missing" warning so a
+				// future disappearance is logged again.
+				r.clearMissingGVKWarning(gvk)
 				return []groupVersionKindPlural{
 					{
 						GroupVersionKind: schema.GroupVersionKind{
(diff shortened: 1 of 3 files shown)


bhope:
Thanks for the PR! The issue with this approach is that `ResolveGVKToGVKPs` is called every 3 seconds so this will log continuously for any missing CRD. I'd recommend tracking warned GVKs in `CRDiscoverer` (the existing m sync.RWMutex covers it) and logging only on first occurrence, clearing when the GVK is successfully resolved. That way you get one log per missing episode at default verbosity without the noise.

internal/discovery/discovery.go
```diff
@@ -152,6 +152,12 @@ func (r *CRDiscoverer) ResolveGVKToGVKPs(gvk schema.GroupVersionKind) (resolvedG
 				}, nil
 			}
 		}
+		// The GVK was fully specified but is absent from the discovery cache,
```
bhope: Please remove redundant comments

bhope (changes requested):

aryanputta:
Thanks @bhope, that's the right call. Updated in 434c99b:

- Added a `warnedMissingGVKs` set on `CRDiscoverer`, guarded by the existing `m sync.RWMutex` via small `markMissingGVKWarned` / `clearMissingGVKWarning` helpers.
- `ResolveGVKToGVKPs` now logs a missing GVK only on the first miss. When the GVK is later resolved successfully the entry is cleared, so a subsequent disappearance is logged again (one log per missing episode, not per tick).
- Added `TestResolveGVKToGVKPsMissingWarnDedup` covering first-miss, repeat-miss suppression, clear-on-resolve, and re-warn-after-disappear.

`go build`, `go vet`, and `go test ./internal/discovery/...` all pass.

aryanputta:
@bhope removed the redundant comments. The large explanatory block on the missing-CRD branch and the inline comment on `clearMissingGVKWarning` are both gone; the helper names now carry the intent. Ready for another look when you have a moment.

bhope:
Thanks for the PR and for addressing the review comments. 

/lgtm

PR #2997, opened 2026-06-14
feat(pod): add init container state_started and last_terminated_timestamp metrics

emmayusufu:
## What this does

Adds two init-container metrics that mirror their existing regular-container counterparts:

- `kube_pod_init_container_state_started` (ALPHA) — mirrors `kube_pod_container_state_started`
- `kube_pod_init_container_status_last_terminated_timestamp` (ALPHA) — mirrors `kube_pod_container_status_last_terminated_timestamp`

## Why

The existing `state_started` / `last_terminated_timestamp` metrics only cover standard containers, so there's no way to monitor init-container start latency. That's the gap reported in #2627.

## Notes

This revives work that was already triage-accepted but stalled when the authors went inactive: #2904 (which had a complete implementation, credited via Co-authored-by), and earlier #2750 (triage-accepted) and #2638 ("Overall LGTM"). I'd like to carry it across the line and will stay responsive through review.

## Testing

- New unit cases in `internal/store/pod_test.go` for running, terminated, and last-terminated init containers
- `TestFullScrapeCycle` updated and passing
- `BenchmarkPodStore` family-count guard updated (55 -> 57)
- `make build`, `gofmt`, and the above tests pass locally

Closes #2627

PR #3000, opened 2026-06-15
feat(pod): add kube_pod_init_container_status_last_terminated_exitcode

emmayusufu:
**What this does**

Adds `kube_pod_init_container_status_last_terminated_exitcode`, the exit code of the last terminated init container, mirroring the existing `kube_pod_container_status_last_terminated_exitcode` for regular containers.

After #2997 added `state_started` and `last_terminated_timestamp` for init containers, the exit code was the only last-terminated field regular containers exposed that init containers did not. This completes container/init parity on last-terminated info, which is useful for alerting on init containers that died badly (for example OOMKilled with exit code 137).

The new generator is registered alongside the other init-container generators, the existing init-container test case is extended to assert the metric, and the metric is documented. Labels: `container`, plus the standard `pod`, `namespace`, `uid`. Stability: ALPHA, matching the container equivalent.


mrueg:
@emmayusufu can you address the comments here?

emmayusufu:
Thanks @mrueg, addressed both Copilot points: reordered the exit-code generator to sit before the timestamp one so the function definition order matches the registration order, and reformatted the new docs table row to match the column padding of the surrounding rows. Pushed.

mrueg:
/lgtm

Thanks!

PR #3006, opened 2026-06-16
fix(deployment): emit default rolling update metrics

immanuwell:

`kube_deployment_spec_strategy_rollingupdate_max_unavailable` and `kube_deployment_spec_strategy_rollingupdate_max_surge` are skipped when a Deployment leaves `spec.strategy.rollingUpdate` unset.

That case is pretty common. In `apps/v1`, Kubernetes defaults both fields to `25%`, so KSM was under-reporting these metrics for normal Deployments. This patch emits the defaulted values, and still skips `Recreate`.

Repro:
1. Apply a plain Deployment with `replicas: 10` and no `spec.strategy` / `spec.strategy.rollingUpdate`.
2. Scrape KSM before this patch.
3. `kube_deployment_spec_strategy_rollingupdate_max_unavailable{deployment="demo",namespace="default"}` and `kube_deployment_spec_strategy_rollingupdate_max_surge{deployment="demo",namespace="default"}` are missing, which is kinda wrong.
4. After this patch they show up as `2` and `3`.

### DIFF
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -39,6 +39,10 @@ var (
 	descDeploymentLabelsName          = "kube_deployment_labels"
 	descDeploymentLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descDeploymentLabelsDefaultLabels = []string{"namespace", "deployment"}
+	defaultDeploymentRollingUpdate    = v1.RollingUpdateDeployment{
+		MaxUnavailable: &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+		MaxSurge:       &intstr.IntOrString{Type: intstr.String, StrVal: "25%"},
+	}
 )
 
 // Reasons copied from kubernetes/pkg/controller/deployment/deployment_utils.go.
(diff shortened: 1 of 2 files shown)


internal/store/deployment.go
```diff
@@ -39,6 +39,10 @@ var (
 	descDeploymentLabelsName          = "kube_deployment_labels"
 	descDeploymentLabelsHelp          = "Kubernetes labels converted to Prometheus labels."
 	descDeploymentLabelsDefaultLabels = []string{"namespace", "deployment"}
+	defaultDeploymentRollingUpdate    = v1.RollingUpdateDeployment{
```
mrueg, line 42: where is the default being set?  is there a way not to hardcode percentages here but infer them from k8s codebase?

immanuwell: https://github.com/kubernetes/kubernetes/blob/master/pkg/apis/apps/v1/defaults.go#L54-L62

The defaults are defined in `kubernetes/kubernetes pkg/apis/apps/v1/defaults.go` (lines 54–62), where both MaxUnavailable and MaxSurge are explicitly set to "25%" via intstr.FromString("25%") when RollingUpdate is nil

k8s.io/kubernetes is not importable in downstream projects due to its module structure.
So the values are intentionally hardcoded here to match that upstream definition.
I can add a comment in the code pointing to that source if you want

immanuwell:
@mrueg WDYT?

immanuwell:
closing after re-checking the premise
Kubernetes API server defaults the deployment strategy, rollingUpdate, maxUnavailable, and maxSurge before returning the object, so kube-state-metrics should not receive the nil state handled here

PR #3011, opened 2026-06-18
feat(node): add kube_node_spec_pod_cidrs metric

### LINKED ISSUE #2932: Add kube_node_spec_pod_cidr metric for node podCIDRs

A new metric exposing the podCIDRs field from the Node spec (node.spec.podCIDRs), which contains the list of CIDR ranges assigned to the node for Pod IP allocation.

**Why is this needed:**

podCIDRs is the authoritative multi-value successor to podCIDR and is required for dual-stack clusters, where a node will have both an IPv4 and IPv6 CIDR assigned. There is currently no way to observe or alert on this field via kube-state-metrics. This would enable the capability to alert on misconfigurations. For example, you might spin up a 6 node cluster and grant it a /108 IPv6 CIDR range and assume 1,048,576 IPs would be enough for your cluster's Pods. But in fact the way kubernetes splits that CIDR up, you'll end up with 2 nodes without IPv6 ranges.

A presence-style metric with one series per CIDR entry:

```
kube_node_spec_pod_cidr{node="node1", cidr="10.0.1.0/24", ip_family="IPv4"} 1
kube_node_spec_pod_cidr{node="node1", cidr="fd00::/120", ip_family="IPv6"} 1
```

This follows the existing pattern used by kube_node_spec_taint and similar multi-valued fields. The ip_family label can be derived from the CIDR itself (presence of : indicates IPv6).

This metric wouldn't be disruptive, you should end up with a cardinality of 1-2 timeseries per node depending on if your cluster is single or dual stack.

**Additional context:**

This metric enables alerting on dual-stack completeness, for example:

```
promql# Alert on nodes missing dual-stack CIDR assignment
count by (node) (kube_node_spec_pod_cidr) < 2
```

The existing kube_node_info metric exposes the singular pod_cidr field, but this does not capture the full podCIDRs array and is insufficient for dual-stack environments.

emmayusufu:
`kube_node_info` exposes only the primary `Spec.PodCIDR`. On dual-stack nodes a second pod CIDR (e.g. the IPv6 range) lives in `Spec.PodCIDRs` and isn't surfaced anywhere. This adds `kube_node_spec_pod_cidr`, emitting one series per entry in `Spec.PodCIDRs`, the same list shape `kube_node_spec_taint` already uses.

New ALPHA metric, with a test case and a docs row.

Closes #2932


emmayusufu:
Thanks @mrueg. This is lgtm'd and approved, on hold for a second maintainer to review. It's a small node spec metric that mirrors `kube_node_spec_taint` for the dual-stack pod CIDRs that `kube_node_info` doesn't surface. Ready whenever someone has a moment for the second look.

docs/metrics/cluster/node-metrics.md
mrueg, line 13: we do have a plural metric here, should the metric be cidrs instead of cidr? This will also ensure that we avoid confusion with the existing field

emmayusufu: Good call, renamed it to `kube_node_spec_pod_cidrs` (plural) and updated the HELP text. Since it's one series per entry in `Spec.PodCIDRs`, the plural metric name reads right and keeps it clearly distinct from `kube_node_info`'s `pod_cidr`. Kept the per-series label as `pod_cidr` (each series is a single CIDR), matching how `kube_node_spec_taint` labels each taint. Pushed.

PR #3014, opened 2026-06-19
feat: Implement ValidationAdmissionPolicy (VAP) metrics

mrueg:

Implements initial metrics for ValidationAdmissionPolicy(Bindings)

 
Adds two metrics per VAP and three per VAPB

kube_validatingadmissionpolicy_info
kube_validatingadmissionpolicy_created

kube_validatingadmissionpolicybinding_info
kube_validatingadmissionpolicybinding_created
kube_validatingadmissionpolicybinding_validation_action

In general this currently is only helpful to verify which VAP/VAPB are installed. It would be nice to get more status fields, but that has the potential for high cardinality.


CatherineF-dev:
/hold 

feel free to unhold

PR #3015, opened 2026-06-19
feat(hpa): add scale up/down behavior tolerance metrics

### LINKED ISSUE #2951: Expose HPA `spec.behavior.scaleUp/scaleDown.tolerance` as metrics

Two new metrics exposing the `tolerance` field from `spec.behavior.scaleUp` and `spec.behavior.scaleDown`:

```
kube_horizontalpodautoscaler_spec_behavior_scale_up_tolerance{namespace="...", horizontalpodautoscaler="..."} 0.6
kube_horizontalpodautoscaler_spec_behavior_scale_down_tolerance{namespace="...", horizontalpodautoscaler="..."} 0.6
```

**Why is this needed:**

The `tolerance` field directly determines the dead band within which no scaling occurs. With a tolerance of `600m` (60%) and a CPU target of 50%, scaling only triggers outside the [30%, 80%] range — yet this threshold is invisible in Prometheus today.

This creates a concrete observability gap: it is impossible to build a dashboard or alert that explains *why* an HPA is not scaling without reading the raw Kubernetes spec. The effective thresholds (`target × (1 ± tolerance)`) cannot be derived from any existing metric.

Scrape `spec.behavior.scaleUp.tolerance` and `spec.behavior.scaleDown.tolerance` from the `HorizontalPodAutoscaler` object (already available at scrape time, no additional API call needed) and expose them as gauge metrics.

When the field is absent, expose the cluster-wide default value.

```promql
# Effective scale-up CPU threshold
kube_horizontalpodautoscaler_spec_target_metric{metric_name="cpu", metric_target_type="utilization"}
  * (1 + kube_horizontalpodautoscaler_spec_behavior_scale_up_tolerance)
```

**Additional context:**

https://kubernetes.io/docs/concepts/workloads/autoscaling/horizontal-pod-autoscale/#tolerance
https://kubernetes.io/blog/2025/04/28/kubernetes-v1-33-hpa-configurable-tolerance/

emmayusufu:
Exposes the `tolerance` field from `spec.behavior.scaleUp` and `spec.behavior.scaleDown` on HorizontalPodAutoscaler as two new gauges:

- `kube_horizontalpodautoscaler_spec_behavior_scale_up_tolerance`
- `kube_horizontalpodautoscaler_spec_behavior_scale_down_tolerance`

The tolerance sets the dead band within which no scaling happens, so the effective thresholds (`target * (1 ± tolerance)`) can't be derived from any existing metric today.

One deviation from the issue: it suggested falling back to the cluster-wide default when the field is unset. kube-state-metrics has no way to know that default (it's a kube-controller-manager flag), so the metrics are emitted only when the field is set, which is the usual convention here for optional spec fields.

Fixes #2951


bhope:
Thanks for the PR, this nice observability addition for HPA tolerance

/lgtm
/hold cancel

PR #3016, opened 2026-06-19
feat(pod): add container status resource request and limit metrics

### LINKED ISSUE #2980: Feature Request: Add metrics for actual allocated container resources (status.containerStatuses[*].resources) to support In-Place Pod Resize
I would like to add new metrics:
- kube_pod_container_status_resource_requests (Gauge) - The number of actual requested request resources currently allocated to a container (status.containerStatuses[*].resources.requests).
- kube_pod_container_status_resource_limits (Gauge) - The number of actual limit resources currently allocated to a container (status.containerStatuses[*].resources.limits).

**Why is this needed:**
Starting from Kubernetes v1.33 (Beta) and graduating to Stable in newer versions, the In-Place Pod Resize feature (InPlacePodVerticalScaling) allows users to change a container's CPU and memory resource requests/limits without restarting the Pod.
With this feature, the Pod architecture splits resource tracking into two parts:
1. spec.containers[*].resources - represents the desired resource state.
2. status.containerStatuses[*].resources - represents the actual/allocated resource state currently configured on the node's runtime.
Currently, kube-state-metrics exposes metrics like kube_pod_container_resource_requests and kube_pod_container_resource_limits based strictly on the spec section. However, when an in-place resize is triggered, there can be a significant mismatch or delay between spec and status (for example, if the node lacks resources and the resize is blocked or set to PodResizePending / Infeasible)

Introduce new gauge metrics (or add status-based alternatives) that track the actual allocated resources from the Pod status:
- kube_pod_container_status_resource_requests (Gauge) - The number of actual requested request resources currently allocated to a container (status.containerStatuses[*].resources.requests).
- kube_pod_container_status_resource_limits (Gauge) - The number of actual limit resources currently allocated to a container (status.containerStatuses[*].resources.limits).

**Additional context:**
Without these metrics, cluster administrators cannot easily detect via Prometheus when a Pod's desired resources (spec) deviate from its actual execution state (status) due to scheduling/node resource constraints.

**Questions / Contribution**
- Is there any ongoing work or existing discussion regarding supporting In-Place Pod Resize metrics?
- If not, our team is highly interested in implementing this feature. Would the maintainers accept a Pull Request (PR) for this? We can take over the implementation if guided on the preferred design/naming conventions.

emmayusufu:
Adds two metrics from `status.containerStatuses[*].resources`:

- `kube_pod_container_status_resource_requests`
- `kube_pod_container_status_resource_limits`

These expose the resources the kubelet has actually allocated to a running container, which can differ from the spec under In-Place Pod Resize. The existing `kube_pod_container_resource_requests`/`kube_pod_container_resource_limits` only read from the spec, so the resized values aren't visible anywhere today.

The new metrics mirror that shape (same `container`/`node`/`resource`/`unit` labels) but read from the container status, and emit nothing when the kubelet hasn't populated the field.

Fixes #2980

## Summary by CodeRabbit

* **New Features**
  * Added experimental metrics for pod container resource requests and limits.
  * Metrics include CPU and memory values with container, pod, namespace, node, resource, unit, and UID labels.
  * Resource values are reported consistently across supported units.

* **Documentation**
  * Documented the new pod container resource metrics.

* **Tests**
  * Added coverage for CPU and memory request and limit metrics.
  * Updated scrape validation and benchmark expectations.


internal/store/pod.go
```diff
@@ -296,6 +298,136 @@ func createPodContainerResourceRequestsFamilyGenerator() generator.FamilyGenerat
...
+				Metrics: ms,
+			}
+		}),
+	)
+}
+
+func createPodContainerStatusResourceRequestsFamilyGenerator() generator.FamilyGenerator {
+	return *generator.NewFamilyGeneratorWithStability(
+		"kube_pod_container_status_resource_requests",
+		"The request resources currently allocated to a container as reported by the kubelet (status.containerStatuses[*].resources.requests).",
```
mrueg:
```suggestion
		"The request resources currently allocated to a container",
```

emmayusufu: Done, applied the shorter wording to both the request and limit metrics so they stay parallel, and updated the docs and golden tests to match. Thanks.

CatherineF-dev:
Hi, these metrics should already exist in scheduler. https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/workload/pod-metrics.md

Should we close this?

Note: For pod resource requests and limits, kube-scheduler exposes kube_pod_resource_request and kube_pod_resource_limit on its /metrics/resources endpoint. These are computed from the pod's effective resource requirements and already account for pod-level resources (pod.spec.resources, beta and enabled by default since Kubernetes 1.34), so they are recommended over the container-level metrics below. kube-state-metrics does not expose a separate metric for pod-level resources. These metrics are defined in [KEP-1748](https://github.com/kubernetes/enhancements/tree/master/keps/sig-instrumentation/1748-pod-resource-metrics).

emmayusufu:
The scheduler's kube_pod_resource_request and kube_pod_resource_limit come from the pod's effective spec, and so do KSM's existing kube_pod_container_resource_requests and limits. This PR reads status.containerStatuses[].resources instead, which is what the kubelet actually allocated, so the two only diverge when in-place pod resize changes a running container. That divergence window is the reason I opened it, without a status-side series you cannot see whether a requested resize was actually applied. If that use case feels too narrow for now I can park it, but it is not a duplicate of the scheduler metrics.

mrueg:
One thought I had looking at this PR: Can we assume that scheduler metrics are available all the time for ksm users? I wonder if there are some cloud systems that do not expose them to the user.

emmayusufu:
That assumption does not hold on managed control planes. On EKS, GKE, and AKS the scheduler is not directly scrapable, so its metrics only come through each provider's own opt-in pipeline, CloudWatch on EKS, the control plane metrics config on GKE, managed Prometheus on AKS, meaning a plain Prometheus setup never sees kube_pod_resource_request or limit there. KSM reads the pod objects from the API server, so these metrics behave the same on any cluster. Also, even where the scheduler metrics are exposed, they report the effective spec, while this PR reads status.containerStatuses[].resources, so with in-place resize they answer different questions.

PR #2797, opened 2025-10-14
feat: add valueType field for explicit duration parsing in custom resources

SoumyaRaikwar:
## What this PR does 

This PR adds a `valueType` field to custom resource state metrics gauge configuration, enabling explicit parsing of Go duration strings to seconds. This is particularly useful for resources like cert-manager Certificates that store durations as strings (e.g., "2160h", "720h").

Currently, kube-state-metrics automatically detects value types but fails to parse Go duration strings, resulting in errors like:
```
strconv.ParseFloat: parsing "2160h": invalid syntax
```

## Which issue(s) this PR fixes:

Fixes #2625

### Code Changes
- Added `ValueType` field to `MetricGauge` struct with options: `duration`, `quantity`, and default (empty string)
- Implemented `parseDurationValue()` function using Go's `time.ParseDuration`
- Updated gauge value extraction to respect `valueType` in both direct path and optimization code paths
- Maintains full backward compatibility

### Documentation
- Added "Value Type Specification" section to `customresourcestate-metrics.md`
- Documented all available value types and usage scenarios
- Included complete cert-manager Certificate monitoring example
- Added duration format reference (Go `time.ParseDuration` formats)

## Example Usage

```yaml
kind: CustomResourceStateMetrics
spec:
  resources:
    - groupVersionKind:
        group: cert-manager.io
        version: v1
        kind: Certificate
      metrics:
        - name: "certificate_duration_seconds"
          help: "Certificate validity duration in seconds"
          each:
            type: Gauge
            gauge:
              path: [spec, duration]
              valueType: duration  # Parses "2160h" → 7776000.0 seconds

        - name: "certificate_renew_before_seconds"
          help: "Time before expiration when certificate should be renewed"
          each:
            type: Gauge
            gauge:
              path: [spec, renewBefore]
              valueType: duration  # Parses "720h" → 2592000.0 seconds
```

**Resulting metrics:**
```prometheus
kube_customresource_certificate_duration_seconds{customresource_group="cert-manager.io",customresource_kind="Certificate",customresource_version="v1",name="example-cert",namespace="default"} 7776000
kube_customresource_certificate_renew_before_seconds{customresource_group="cert-manager.io",customresource_kind="Certificate",customresource_version="v1",name="example-cert",namespace="default"} 2592000
```

## Supported Duration Formats

The `duration` valueType supports Go's `time.ParseDuration` format:
- Hours: `"1h"`, `"24h"`, `"2160h"` (90 days)
[...]

PR #3017, opened 2026-06-22
feat(customresourcestate): add valueType field for explicit duration parsing

### LINKED ISSUE #2625: Add support for parsing duration (`time.ParseDuration`) on custom resource state metrics

I would like custom state metrics to be able to parse resource duration fields into metric value. For example, they should be able to parse `duration: 1h` into `3600`.

**Why is this needed**:

Some resources (like cert-manager certificates) store valuable information, such as how long a certificate is valid for, as a duration string.

I would like KSM to support parsing these values, similarly to other values, as described [here](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/metrics/extend/customresourcestate-metrics.md#type-conversion-and-special-handling).

**Additional context**

This is not as trivial as adding a three line `if this can be parsed; set the metric to the time.ParseDuration` value. Custom resources already support parsing resource quantities such as `250m` (i.e. 1/4 of a CPU), which is indistinguishable from the duration `250m` (i.e. four hours and ten minutes). The config likely would need to support explicitly stating how a metric should be converted (which is probably desirable anyway).

emmayusufu:
Revives #2797 (closed as rotten when the original author went inactive) to implement the duration-parsing feature requested in #2625, which is `triage/accepted`. @mrueg suggested picking this up.

Adds a `valueType` field to the custom resource state gauge configuration. `valueType: duration` parses Go duration strings (for example `"2160h"`) with `time.ParseDuration` and converts them to seconds. This helps with resources such as cert-manager Certificates that store durations as strings and would otherwise fail float parsing.

Co-authored with the original author @SoumyaRaikwar.

Fixes #2625

## Summary by CodeRabbit

* **New Features**
  * Added a `valueType` option for gauge metrics to control `valueFrom` parsing: default/auto behavior or `duration` (converted to seconds).
* **Documentation**
  * Refined Custom Resource State Metrics examples, expanded gauge `valueType` guidance (including duration semantics), and clarified `status.conditions` key=value matching.
* **Bug Fixes**
  * Enforced validation for unsupported `valueType` values with clear errors and ensured consistent gauge value parsing (including nil handling via `NilIsZero`).
* **Tests**
  * Added coverage for duration parsing, nil/empty/error cases, backward compatibility, and unknown `valueType` rejection.

### DIFF
--- a/docs/metrics/extend/customresourcestate-metrics.md
+++ b/docs/metrics/extend/customresourcestate-metrics.md
@@ -634,7 +634,6 @@ By default, kube-state-metrics automatically detects and converts values using t
 The `valueType` field in gauge configuration accepts the following values:
 
 * `duration` - Explicitly parse Go duration strings (e.g., "1h", "30m", "1h30m45s") and convert them to seconds
-* `quantity` - Explicitly parse Kubernetes resource quantities (e.g., "250m", "5Gi")
 * (empty/omitted) - Use automatic type detection (default behavior)
 
 ###### Duration Value Type
@@ -703,10 +702,7 @@ Use `valueType: duration` when:
 * Auto-detection fails because the string isn't recognized as a number
 * You want explicit, predictable parsing behavior
 
-Use `valueType: quantity` when:
-
-* You want to ensure Kubernetes quantity parsing (even if auto-detection would work)
-* You want to make the parsing behavior explicit in configuration
+Kubernetes quantities like "250m" or "5Gi" need no explicit valueType; automatic detection already parses them.
 
 ###### valueType with valueFrom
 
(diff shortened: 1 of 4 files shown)


emmayusufu:
Hey @mrueg  so I went through all the comments again, including the ones hidden in the collapsed CodeRabbit sections, and pushed fixes for everything.

The nilIsZero regression was already fixed in 8024405 along with the missing test and field comment. In 7d52c6a I also pulled the parsing switch into one helper used by both paths so they can't drift, made an unknown valueType fail when the config compiles instead of erroring on every scrape, and gave nil durations a proper error message.

The last one, b199454, drops valueType: quantity. It turned out to do the same thing as leaving valueType out, since auto-detection already parses quantities, so keeping it was misleading. The issue only asks for duration anyway. If we want real strict quantity parsing later it can be its own PR. Docs updated to match, and the shadowed err got renamed too.

PR #3018, opened 2026-06-22
feat: add kube_pod_status_disruption_reason metric

### LINKED ISSUE #2774: [Feature Request] Support detailed disruption reason detection via new metric kube_pod_status_disruption_reason
Introduce a new metric `kube_pod_status_disruption_reason` that captures Pod disruption causes by reading the reason from pod.status.conditions (specifically where type == "DisruptionTarget"). This metric should support these possible reasons:
```go
podDisruptionConditionReasons = [
  "PreemptionByScheduler",
  "DeletionByTaintManager",
  "EvictionByEvictionAPI",
  "DeletionByPodGC",
  "TerminationByKubelet"
]
``` 
These are the complete set of disruption reasons are defined in the [official Kubernetes documentation on pod disruption conditions.](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/?utm_source=chatgpt.com#pod-disruption-conditions)

**Why is this needed:**
The existing `kube_pod_status_reason` metric is fundamentally limited and often returns zero for all reason labels—even in cases where pods are clearly being evicted. The root cause lies in its implementation: it only matches `pod.status.reason == "Evicted"`, which is only set when eviction is initiated directly by the kubelet (e.g., due to resource pressure or shutdown). This logic completely overlooks evictions triggered via the Eviction API (e.g., using kubectl drain), which instead set `pod.status.condition.reason = "EvictionByEvictionAPI"` but leave pod.status.reason unchanged, resulting in a constant metric value of zero￼.

By extending to all known disruption reasons from Kubernetes (as listed above), the proposed metric enables users to accurately identify and monitor diverse Pod disruption events—whether they’re triggered by the Kubelet, scheduler preemption, taint manager, garbage collection, or Eviction API.

- **Metric definition:** `kube_pod_status_disruption_reason` (Gauge) that captures disruption cause via pod.status.conditions where type equals “DisruptionTarget.”
- **Supported reason values**:
  - PreemptionByScheduler
  - DeletionByTaintManager
  - EvictionByEvictionAPI
  - DeletionByPodGC
  - TerminationByKubelet
- **Backward compatibility**: Retain the legacy `kube_pod_status_reason `for existing dashboards and alerts, while clarifying in docs when to prefer the new metric.
- **Implementation guidance**:
  - In KSM implementation, inspect pod.status.conditions for DisruptionTarget conditions and emit the corresponding reason label.
  - Add unit/integration tests to simulate each disruption scenario.
  - Update metrics.md documentation with clear guidance, usage examples, and migration notes.

**Additional context:**
**related issues:**
- https://github.com/kubernetes/kube-state-metrics/issues/2612
- https://stackoverflow.com/questions/76653767/prometheus-metric-kube-pod-status-reason-returns-hundreds-of-results-but-all-wi

pierluigilenoci:
## What type of PR is this?

/kind feature

Adds a new ALPHA metric `kube_pod_status_disruption_reason` that reports the disruption reason for pods carrying a `DisruptionTarget` condition.

The existing `kube_pod_status_reason` metric reports `0` for all reasons when pods are evicted via the Eviction API (#2612), because that code path does not set `pod.status.reason` — it only sets a `DisruptionTarget` condition. This metric closes that gap by reading the condition directly.

Supported `reason` label values (per the [Kubernetes pod disruption conditions spec](https://kubernetes.io/docs/concepts/workloads/pods/disruptions/#pod-disruption-conditions)):

| Reason | Set by |
|---|---|
| `DeletionByPodGC` | Pod GC controller |
| `DeletionByTaintManager` | Taint manager |
| `EvictionByEvictionAPI` | Eviction API |
| `PreemptionByScheduler` | kube-scheduler |
| `TerminationByKubelet` | kubelet |

## How to verify it

```bash
# Evict a pod via the Eviction API, then query:
kube_pod_status_disruption_reason{reason="EvictionByEvictionAPI"} 1
```

## Checklist
- [x] Added new metric `kube_pod_status_disruption_reason` (ALPHA)
- [x] Added unit tests in `internal/store/pod_test.go`
- [x] Build passes: `go build ./internal/store/...`
- [x] Tests pass: `go test ./internal/store/... -run TestGetDisruptionConditionReasonValue`

Fixes #2774
Ref: #2612

## Summary by CodeRabbit

* **New Features**
  * Added a new pod metrics gauge for disruption reasons, with one series per supported reason.
  * Included the metric in the full scrape output and added documentation for it.

* **Tests**
  * Updated scrape-cycle expectations to include the new metric.
  * Added unit tests to verify disruption-reason condition matching and correct default behavior when no match is found.


pierluigilenoci:
Hi @mrueg — I've fixed the benchmark test count (58 → 59, since the new metric adds one entry per pod). All other checks were already passing. Would you be willing to take a look at this PR when you get a chance? Thank you!

bhope (commented):
Closing in favor of #3088. Thanks again for the PR! Also this version loop over all 5 known reasons and emits a 0/1 row per pod - same cardinality issue #3089 just fixed on kube_pod_status_reason.

/close

PR #3026, opened 2026-06-29
feat(pod): add pod-level resource metrics

### LINKED ISSUE #2945: Support Pod-Level Resources

Since Kubernetes 1.34 it's possible to assign Pod-Level Resources: https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/

There should be such metric exposed.

**Why is this needed:**

When one decides to specify resources on Pod-level, instead container, kube-state-metrics does not provide metrics for it.

Analogously to `kube_pod_container_resource_requests` and `kube_pod_container_resource_limits`, there should exist:

* `kube_pod_resource_requests`
* `kube_pod_resource_limits`

**Additional context:**

* https://kubernetes.io/docs/tasks/configure-pod-container/assign-pod-level-resources/
* https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#pod-level-resource-specification

emmayusufu:
Kubernetes 1.34 stabilized pod-level resources (`pod.spec.resources`), letting you set requests/limits at the pod instead of per container. kube-state-metrics exposes `kube_pod_container_resource_requests`/`limits` but nothing for the pod-level field, so pods that use it are invisible to capacity planning.

This adds two ALPHA metrics read from `pod.spec.resources`, labeled by `node`, `resource`, and `unit` (mirroring the container metrics minus the `container` label):

- `kube_pod_spec_resource_requests`
- `kube_pod_spec_resource_limits`

On naming: I used `kube_pod_spec_resource_*` rather than `kube_pod_resource_*`, since KSM's own HELP text points to `kube_pod_resource_*` as the kube-scheduler metrics. I can rename if you'd prefer a different convention.

Includes a store test, the server_test golden HELP/TYPE, the BenchmarkPodStore family count, and docs. The issue isn't triage-accepted yet, so I'm glad to adjust scope or naming based on your steer.

Fixes #2945


CatherineF-dev:
cadvisor has these metrics

emmayusufu:
Thanks for taking a look.

The usual framing is that cAdvisor exposes the effect of requests and limits on the kernel (the enforced cgroup values), while kube-state-metrics exposes the logical requests and limits themselves from the API. That's why KSM already ships `kube_pod_container_resource_requests`/`limits` even though cAdvisor has `container_spec_cpu_quota` and `container_spec_memory_limit_bytes`.

This PR is the pod-level counterpart for `pod.spec.resources` (new in 1.34), which KSM doesn't cover yet. cAdvisor can show the pod cgroup's enforced limit, but not the declared per-resource requests and limits the way KSM does for containers today.

If you'd still rather not add it given the overlap, I'm fine closing. Issue #2945 asked for it and it lines up with the existing container metrics, so it seemed worth doing.

CatherineF-dev:
cc @dgrisonnet for inputs

docs/metrics/workload/pod-metrics.md
```diff
@@ -31,6 +31,8 @@
...
 | kube_pod_container_resource_requests                  | Gauge       | The number of requested request resource by a container. It is recommended to use the `kube_pod_resource_request` metric exposed by kube-scheduler instead, as it is more precise. | `cpu`=&lt;core&gt; <br> `memory`=&lt;bytes&gt; | `resource`=&lt;resource-name&gt; <br> `unit`=&lt;resource-unit&gt; <br> `container`=&lt;container-name&gt; <br> `pod`=&lt;pod-name&gt; <br> `namespace`=&lt;pod-namespace&gt; <br> `node`=&lt; node-name&gt; <br> `uid`=&lt;pod-uid&gt;                                                                                                                           | STABLE | -      |
```
dgrisonnet, line 32: @ndixita when I reviewed https://github.com/kubernetes/enhancements/pull/6180/, I forgot that the scheduler introduced its own `kube_pod_resource_request` metric and that we recommend using it in favor of the kube-state-metrics one.

I this something we could perhaps also use for pod-level resources metrics?

emmayusufu:
Thanks Damien, fair point, and I went and checked. The scheduler's `kube_pod_resource_requests` and `kube_pod_resource_limits` are computed through `resourcehelper.PodRequests` and `PodLimits` in `pkg/scheduler/metrics/resources/resources.go`, and that helper already accounts for pod-level resources (it handles `pod.Spec.Resources` through the pod-level path). So the effective requests and limits the scheduler exposes already reflect pod-level resources today.

The only thing this KSM metric would add is the raw declared `pod.spec.resources` as its own series, versus the scheduler's effective value (the max of pod-level and the container sum). That's a narrow config-audit view, and since KSM already points people to the scheduler metric over its container-level ones, a pod-level KSM metric cuts against that direction.

So I think closing this in favor of the scheduler metric is the right call. If a specific need for the raw declared pod-level field comes up later, that's a separate discussion. Thanks both for looking.

mrueg:
Could we maybe use this as an opportunity to document that these metrics are available from the scheduler in the README.md or within the pod.md?

emmayusufu:
Good idea. I've opened #3027 adding a note to pod-metrics.md pointing at the scheduler's `kube_pod_resource_requests` and `kube_pod_resource_limits`, including that they already cover pod-level resources. Closing this one in favor of that.

PR #3031, opened 2026-07-06
fix: negotiate OpenMetrics when protobuf is preferred in Accept header

### LINKED ISSUE #3010: Content negotiation ignores supported formats when highest-quality Accept entry is unsupported (protobuf)
**What happened**:
  When a client sends an Accept header listing application/vnd.google.protobuf with the highest q-value alongside supported formats (application/openmetrics-text, text/plain), KSM responds with text/plain; version=0.0.4 instead of the next best supported format.
                                                                                                                                                                                                                                                                             
  This is the standard Accept header sent by Prometheus scrapers:                                                                                                                                                                                                            
 ```                                                                                                                                                                                                                                                                            
  Accept: application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.6,                                                                                                                                                                  
          application/openmetrics-text;version=1.0.0;escaping=allow-utf-8;q=0.5,                                                                                                                                                                                             
          application/openmetrics-text;version=0.0.1;q=0.4,                                                                                                                                                                                                                  
          text/plain;version=1.0.0;escaping=allow-utf-8;q=0.3,                                                                                                                                                                                                               
          text/plain;version=0.0.4;q=0.2,                                                                                                                                                                                                                                    
          */*;q=0.1          
```                                                                                                                                                                                                                                                
                                                                                                                                                                                                                                                                             
  The response Content-Type is text/plain; version=0.0.4; charset=utf-8 — all OpenMetrics entries in the Accept header are silently ignored. 

KSM cannot serve protobuf, so it should skip it and select the next acceptable format it can serve. The response Content-Type should be:                                                                                                                                   
                  
  application/openmetrics-text; version=1.0.0; charset=utf-8                                                                                                                                                                                                                 
                                                                                                                                                                                                                                                                             
  This is the behavior mandated by https://datatracker.ietf.org/doc/html/rfc7231#section-5.3.2: the Accept header is a preference ordering, not a hard requirement. The server should select the highest-quality type it can produce and only fall back to a default if none 
  of the listed types are supported.
  # Start kube-state-metrics (any standard deployment)                                                                                                                                                                                                                       
  # Then scrape /metrics with the Prometheus default Accept header:                                                                                                                                                                                                          
  curl -v \                                                                                                                                                                                                                                                                  
    -H 'Accept: application/vnd.google.protobuf;proto=io.prometheus.client.MetricFamily;encoding=delimited;q=0.6,application/openmetrics-text;version=1.0.0;escaping=allow-utf-8;q=0.5,application/openmetrics-text;version=0.0.1;q=0.4,text/plain;version=1.0.0;escaping=all
  ow-utf-8;q=0.3,text/plain;version=0.0.4;q=0.2,*/*;q=0.1' \                                                                                                                                                                                                                 
    http://localhost:8080/metrics 2>&1 | grep "Content-Type"                                                                                                                                                                                                                 
  # Actual:   Content-Type: text/plain; version=0.0.4; charset=utf-8                                                                                                                                                                                                         
  # Expected: Content-Type: application/openmetrics-text; version=1.0.0; charset=utf-8                                                                                                                                                                                       
                                                                                          

 The root cause is in pkg/metricshandler/metrics_handler.go:                                                                                                                                                                                                                
```go           
                                                                                                                                                                                                                                                                             
  if contentType.FormatType() != expfmt.TypeOpenMetrics {                                                                                                                                                                                                                    
      contentType = expfmt.NewFormat(expfmt.TypeTextPlain)                                                                                                                                                                                                                   
  }                                                                                                                                                                                                                                                                          
  ```
  expfmt.NegotiateIncludingOpenMetrics correctly returns FmtProtoDelim (the highest-quality match from the Accept header). The guard then discards it because KSM doesn't support protobuf — but instead of re-negotiating against only the formats KSM does support, it     
  unconditionally falls back to text/plain; version=0.0.4.
                                                                                                                                                                                                                                                                             
  The practical impact is that any Prometheus scraper using the default Accept header will never receive OpenMetrics from KSM, even though both sides support it. This blocks OpenMetrics-only features from working correctly end-to-end.                                   
   
  Related: #2022                                            
**Environment**:

* kube-state-metrics version:
* Other info:

### PR DESCRIPTION
**Pull request by dpacgdm**

Fix content negotiation on `/metrics` so Prometheus scrapers receive OpenMetrics instead of falling back to legacy `text/plain; version=0.0.4`. When the Accept header lists protobuf with the highest q-value (Prometheus default), strip unsupported protobuf entries and re-negotiate against OpenMetrics and text/plain. Adds unit tests for the Prometheus default Accept header and protobuf-only fallback.

Fixes #3010

```release-note
Fixed `/metrics` content negotiation to prefer OpenMetrics when Prometheus sends an Accept header that lists protobuf first.
```

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metrics response `Content-Type` negotiation to more accurately follow client `Accept` headers.
  * When clients indicate protobuf preferences, those hints are ignored for the response, ensuring consistent output.
  * Responses now reliably fall back to OpenMetrics or plain text to avoid unexpected content-type selection.
* **Tests**
  * Added unit tests covering `Accept`-header negotiation and filtering of protobuf-related `Accept` values.

### DIFF
--- a/pkg/metricshandler/metrics_handler.go
+++ b/pkg/metricshandler/metrics_handler.go
@@ -197,13 +197,7 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 	resHeader := w.Header()
 	var writer io.Writer = w
 
-	contentType := expfmt.NegotiateIncludingOpenMetrics(r.Header)
-
-	// We do not support protobuf at the moment. Fall back to FmtText if the negotiated exposition format is not FmtOpenMetrics See: https://github.com/kubernetes/kube-state-metrics/issues/2022.
-
-	if contentType.FormatType() != expfmt.TypeOpenMetrics {
-		contentType = expfmt.NewFormat(expfmt.TypeTextPlain)
-	}
+	contentType := negotiateSupportedContentType(r.Header)
 	resHeader.Set("Content-Type", string(contentType))
 
 	if m.enableGZIPEncoding {
@@ -271,6 +265,49 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
 		}
 	}
 }
+
+// negotiateSupportedContentType selects the highest-quality exposition format
+// kube-state-metrics can serve. Protobuf is skipped so clients that prefer it
+// (e.g. Prometheus scrapers) still receive OpenMetrics or text/plain.
+func negotiateSupportedContentType(h http.Header) expfmt.Format {
+	contentType := expfmt.NegotiateIncludingOpenMetrics(h)
+	if isProtoFormat(contentType) {
+		filtered := http.Header{}
+		filtered.Set(hdrAccept, filterProtoFromAccept(h.Get(hdrAccept)))
+		contentType = expfmt.NegotiateIncludingOpenMetrics(filtered)
+	}
+	if contentType.FormatType() != expfmt.TypeOpenMetrics && contentType.FormatType() != expfmt.TypeTextPlain {
+		contentType = expfmt.NewFormat(expfmt.TypeTextPlain)
+	}
+	return contentType
+}
+
+func isProtoFormat(contentType expfmt.Format) bool {
+	switch contentType.FormatType() {
+	case expfmt.TypeProtoDelim, expfmt.TypeProtoText, expfmt.TypeProtoCompact:
+		return true
+	default:
+		return false
+	}
+}
+
+func filterProtoFromAccept(accept string) string {
+	if accept == "" {
+		return accept
+	}
+	parts := strings.Split(accept, ",")
+	filtered := make([]string, 0, len(parts))
+	for _, part := range parts {
+		trimmed := strings.TrimSpace(part)
+		if strings.HasPrefix(trimmed, expfmt.ProtoType) {
+			continue
+		}
+		filtered = append(filtered, trimmed)
+	}
+	return strings.Join(filtered, ",")
+}
+
+const hdrAccept = "Accept"
 
 func parseResources(params []string) map[string]struct{} {
 	if params == nil {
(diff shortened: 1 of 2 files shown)


pkg/metricshandler/metrics_handler.go
```diff
@@ -272,6 +266,49 @@ func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
...
+	default:
+		return false
+	}
+}
+
+func filterProtoFromAccept(accept string) string {
+	if accept == "" {
+		return accept
+	}
+	parts := strings.Split(accept, ",")
```
CatherineF-dev, line 298: It's a little hacky. Does it have standard library for this?

Does other metrics exporters (node-exporter eta) also use string split or not

dpacgdm:
@CatherineF-dev The actual format selection still goes through `expfmt.NegotiateIncludingOpenMetrics` (prometheus/common), which uses proper Accept parsing via `goautoneg`. There’s no API there for “negotiate, but skip protobuf,” which is why we filter protobuf offers out of `Accept` and call negotiate again.

The `strings.Split` is only that pre-filter. `mime.ParseMediaType` handles a single media type, not a full Accept list, so it doesn’t replace this step by itself. Exporters that serve protobuf (e.g. via `promhttp`) typically don’t need an equivalent; kube-state-metrics intentionally doesn’t support protobuf (#2022), so we have to demote those offers to land on OpenMetrics / text.

I can switch the filter to `goautoneg.ParseAccept` (same parser Negotiate uses) if you’d prefer that over the comma split — or reshape the approach if you have a preferred pattern.

adziauho:
@dpacgdm I think it is good idea trying to test recently added `NegotiateAccept` function https://github.com/prometheus/common/blob/5768e9c47379fd39b38190ca4d27fe3f72c330fa/expfmt/encode.go#L83

PR #3036, opened 2026-07-18
feat(pvc): add metrics for VolumeAttributesClass support

FRosner:
## Why

Kubernetes' `VolumeAttributesClass` API (GA, default-enabled since v1.34) lets a `PersistentVolumeClaim` reference a mutable class of storage attributes, and lets the CSI driver report progress as it applies changes. kube-state-metrics currently exposes no information about this on PVCs.

## What

Adds three new EXPERIMENTAL metrics to the PVC store:

* `kube_persistentvolumeclaim_volume_attributes_class` — the VolumeAttributesClass requested in the PVC spec.
* `kube_persistentvolumeclaim_status_current_volume_attributes_class` — the VolumeAttributesClass currently applied, as reported by the CSI driver. Kept separate from the spec metric so drift between desired and applied class can be detected.
* `kube_persistentvolumeclaim_status_modify_volume_status` — one-hot status (`Pending`/`InProgress`/`Infeasible`) of an in-progress `ModifyVolume` operation, with a `target_volume_attributes_class` label. Emits zero rows when no modification is in progress.

The two new `PersistentVolumeClaimConditionType` values (`ModifyVolumeError`, `ModifyingVolume`) introduced alongside this API are already covered for free by the existing generic `kube_persistentvolumeclaim_status_condition` metric.

This is scoped to PVC metrics only — no new standalone `VolumeAttributesClass` resource/store. I will add these metrics in a follow-up PR.

## Summary by CodeRabbit

- **New Features**
  - Added experimental Prometheus metrics for PersistentVolumeClaim volume attribute classes.
  - Added metrics for current volume attribute classes and modify-volume status, including target class details.

- **Bug Fixes**
  - Prevented volume attribute class metrics from being emitted with empty label values when the corresponding fields are unset.

- **Documentation**
  - Documented the new metrics, labels, gauge types, and experimental status.


FRosner:
<a href="https://easycla.lfx.linuxfoundation.org/#/?version=2">CLA Signed</a><br >The committers listed above are authorized under a signed CLA.<ul><li>:white_check_mark: login: FRosner / name: Frank Rosner (bdfe87a0d40cb1eb632fa7c6d3c6ad0b2eeb7055, d9330c6bd162d4c88aa843dc8f9cd1e29585de47)</li></ul>

CatherineF-dev:
@FRosner you can click the test failure link inside "[EasyCLA](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/18706487/58184885/3036/#/?version=2) — Missing CLA Authorization." presubmit, it will redirect you to sign.

FRosner:
> @FRosner you can click the test failure link inside "[EasyCLA](https://api.easycla.lfx.linuxfoundation.org/v2/repository-provider/github/sign/18706487/58184885/3036/#/?version=2) — Missing CLA Authorization." presubmit, it will redirect you to sign.

Yeah I'm doing it. I am trying to sign as an employee and I'm in contact with our CLA manager but it's a bit difficult for me. I went through this at least 4 times. Now a 5th time and it finally went through :)

FRosner:
@CatherineF-dev would you be the one to review the PR? What's the expected timeline to get a reviewer assigned? :)

CatherineF-dev:
1. What are the use cases of these metrics?
2. Is it possible to have fewer metrics to contain these data?

FRosner:
Thanks for picking this up @CatherineF-dev!

> What are the use cases of these metrics?

The motivating case: I have StatefulSets whose PVCs I want to migrate to a different `VolumeAttributesClass` at runtime (e.g. bumping IOPS/throughput tier), without recreating the PVC. Today kube-state-metrics gives no visibility into any of this. Concretely:

1. **Drift detection** — alert when the requested class and the applied class disagree for longer than expected:
   ```promql
   kube_persistentvolumeclaim_volume_attributes_class
     != on(namespace, persistentvolumeclaim)
   kube_persistentvolumeclaim_status_current_volume_attributes_class
   ```
2. **Stuck/failed modifications** — page when a `ModifyVolume` operation goes `Infeasible`, or has been `InProgress`/`Pending` too long:
   ```promql
   kube_persistentvolumeclaim_status_modify_volume_status{status="Infeasible"} == 1
   ```
3. **Rollout tracking** — during a fleet-wide VAC migration, count how many PVCs still have the old class vs. the new one, to track progress of the rollout across a StatefulSet.
4. **Auditing** — inventory which storage class/tier every PVC is actually running, for cost or compliance reporting.

None of this is derivable from any existing KSM metric today — `VolumeAttributesClassName` and `ModifyVolumeStatus` aren't surfaced anywhere else.

> Is it possible to have fewer metrics to contain these data?

I did consider collapsing these, but I think 3 is the right number given KSM's existing conventions:

- The **spec vs. status split** (`..._volume_attributes_class` vs. `..._status_current_volume_attributes_class`) follows the same pattern KSM already uses everywhere else for desired-vs-observed pairs (e.g. `kube_deployment_spec_replicas` / `kube_deployment_status_replicas`). Merging them into one metric with a `type="requested"|"current"` label would be a departure from that convention and would actually make the drift-detection query above *harder* to write, not easier.
- The **one-hot `modify_volume_status` metric** is intentionally implemented identically to `kube_persistentvolumeclaim_status_phase` (same `boolFloat64` one-hot-per-possible-value pattern, same file). It's not a new shape — it's reusing the existing idiom for "the status is one of a fixed small enum," which is what makes `== 1` alerting queries and `sum by (status)` aggregations work consistently across the codebase.

I could technically fold `modify_volume_status` into `status_current_volume_attributes_class` via extra labels, but that would conflate "what class is currently applied" (always present) with "is a modification in progress" (transient, and needs its own target/status labels) — two different lifecycles — and would break the existing convention that phase-like fields get their own one-hot family. Cardinality-wise the cost is also minimal: `modify_volume_status` emits **zero rows** when no modification is in progress, and the other two emit exactly one row per PVC either way, so there's no meaningful reduction available by merging.

Happy to adjust if you see a cleaner shape, though.

CatherineF-dev:
> The motivating case: I have StatefulSets whose PVCs I want to migrate to a different VolumeAttributesClass at runtime (e.g. bumping IOPS/throughput tier), 

Do you think it's a common use case for other users?

FRosner:
> Do you think it's a common use case for other users?

If they are using [VolumeAttributeClasses](https://kubernetes.io/docs/concepts/storage/volume-attributes-classes/), yes. I don't know how widely they are adopted, but they are stable since Kubernetes v1.36 so I think it makes sense to provide observability around them.

rndinfosecguy:
Don't see the point. These metrics are quite useful

CatherineF-dev:
I see.

VolumeAttributesClass is a new resource type.
Option1:
kube_persistentvolumeclaim_volumeattributesclass_info
kube_persistentvolumeclaim_volumeattributesclass_status

Option2: 
kube_volumeattributesclass_info
kube_volumeattributesclass_status

Option2 will increase query friction on joining pvc and vac metric

How do you think @dgrisonnet which option is preferred?

FRosner:
> VolumeAttributesClass is a new resource type.

Yes. Please note that I am going to submit another PR to add metrics about the VACs. This PR here is only limited to the VAC related fields in PVC, not the VAC resources themselves. So that means Option 2 is anyway going to happen and is not related to this PR.

> How about adding labels into kube_persistentvolumeclaim_info?

You asked that earlier. We could do it, it already carries the storage class and it would require one less join. Why I chose to add a new metric:

- *Stability policy*: `kube_persistentvolumeclaim_info` is `STABLE`. The developer guide is explicit: "Make metrics experimental first when introducing them" (`docs/developer/guide.md`, see #1910). Landing it as its own `ALPHA`/`EXPERIMENTAL` metric lets us iterate freely; landing it as a label on a `STABLE` metric locks us in immediately and makes any future correction a breaking change instead of a fast follow-up.
- *Mutability mismatch*: Every existing label on `_info` (`storageclass`, `volumename`, `volumemode`) is fixed at PVC creation and never changes for the life of the object. VolumeAttributesClassName is the opposite — it's designed to be mutated at runtime, which is the entire premise of this feature and my use case. Attaching a frequently-mutating value as a label on an otherwise-static info row means every VAC change kills one series and spawns another, which could be a meaningfully different churn profile than what `_info` consumers expect today.
- *Breaks the drift-detection use case*: The most valuable query here is comparing requested vs. currently-applied class. That needs two independent metrics so PromQL can do a native vector match (`!=/== on(namespace, persistentvolumeclaim)`). If both values were label keys on the same `_info` row, you can't diff two labels on one series without `label_replace()` gymnastics — it makes the primary use case harder, not easier.
- *`modify_volume_status` can't fit _info at all, regardless of the above*: it's a multi-row one-hot (`Pending`/`InProgress`/`Infeasible`) that emits zero rows when nothing is in progress. `_info` is always exactly one row per PVC. There's no clean way to represent that as additional labels on a singleton row.

I'm open to suggestions, just outlining my reasoning.

internal/store/persistentvolumeclaim.go
```diff
@@ -112,6 +112,94 @@ func persistentVolumeClaimMetricFamilies(allowAnnotationsList, allowLabelsList [
...
 		),
+		*generator.NewFamilyGeneratorWithStability(
+			"kube_persistentvolumeclaim_volume_attributes_class",
+			"The name of the VolumeAttributesClass requested by the persistent volume claim.",
+			metric.Gauge,
+			basemetrics.ALPHA,
+			"",
+			wrapPersistentVolumeClaimFunc(func(p *v1.PersistentVolumeClaim) *metric.Family {
+				volumeAttributesClassName := ""
+				if p.Spec.VolumeAttributesClassName != nil {
```
dgrisonnet: if it is nil, it would be better to skip the metric

dgrisonnet (commented):
> VolumeAttributesClass is a new resource type.
Option1:
kube_persistentvolumeclaim_volumeattributesclass_info
kube_persistentvolumeclaim_volumeattributesclass_status
> Option2:
kube_volumeattributesclass_info
kube_volumeattributesclass_status
> Option2 will increase query friction on joining pvc and vac metric
> How do you think @dgrisonnet which option is preferred?

For VAC metric, they should be named `kube_volumeattributesclass_` like your option 2, but for PVC metrics about VAC, it should be `kube_persistentvolumeclaim_volume_attributes_class` like it is in this PR

internal/store/persistentvolumeclaim.go
FRosner: Makes sense! https://github.com/kubernetes/kube-state-metrics/pull/3036/commits/bdfe87a0d40cb1eb632fa7c6d3c6ad0b2eeb7055

Issue #2988, opened 2026-06-09
kube-state-metrics panics (crashes whole process) on unparsable CronJob schedule


polisko:
KSM crashes (CrashLoopBackOff) when any CronJob in the cluster has a schedule that
the cron parser rejects but the Kubernetes CronJob controller accepts.

Version: v2.19.0 (panic still present on main as of f69b12ef)

Reproducer — apply this CronJob:

apiVersion: batch/v1
kind: CronJob
metadata:
name: ksm-panic-repro
spec:
schedule: "*/60 * * * *"
jobTemplate:
spec:
template:
spec:
restartPolicy: Never
containers:
- name: noop
image: busybox
command: ["true"]

Result — KSM panics and the whole process dies (not just this object's metric):

panic: failed to parse cron job schedule '*/60 * * * *':
 step (60) must be less than range size (60): "*/60"
at internal/store/cronjob.go:254(panic(err))
via metric generator for kube_cronjob_next_schedule_time
-> apimachinery HandleCrash re-panics -> process exit -> CrashLoopBackOff

Impact: one misconfigured CronJob takes down ALL kube-state-metrics output for the
entire cluster, not just kube_cronjob_next_schedule_time for that object.

Root cause: getNextScheduledTime() returns a parse error for schedules the stricter
cron parser (github.com/netresearch/go-cron) rejects; the metric generator has no
error return and calls panic(err) instead of logging and skipping the metric.

Suggested fix: log + skip the metric on parse error (graceful degradation), same as
the adjacent nil-Suspend case already handled in PR#2970.

emmayusufu:
I think this is already fixed on main. Commit 36cc81d4 from June 4, "fix: don't panic on CronJobs with unparseable schedules", swapped the panic in the kube_cronjob_next_schedule_time generator for log-and-skip, and a kube_cronjob_schedule_invalid metric now surfaces the bad schedule instead. There are no panic calls left in cronjob.go on main, and the reproducer doesn't crash a current build. The report lists v2.19.0, and f69b12ef looks like it sits just before that commit. Worth a re-check, and probably safe to close if that holds for you.

mpldr:
Duplicate of #2967

bhope:
Closing as fixed via #2978 - please reopen if you can still reproduce this on a build that includes v2.19.1 or later.

PR #3041, opened 2026-07-21
fix(cronjob): log and skip next schedule on unparseable schedules

dpacgdm:
## Summary
- Log and skip `kube_cronjob_next_schedule_time` when a CronJob schedule cannot be parsed (for example `*/60 * * * *`), instead of failing silently after the panic was removed in #2978.
- Add a unit-test case for the #2988 reproducer schedule so the log-and-skip path stays covered.

Panic crash itself was addressed by #2978 / #2967; this completes the log-and-skip behavior described in #2988.

## Test plan
- [x] `go test ./internal/store/ -run TestCronJobStoreScheduleParsing`
- [ ] Confirm a CronJob with an unparseable step schedule no longer risks exporter crash and emits `kube_cronjob_schedule_invalid`

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved handling of invalid CronJob schedules by recording an error and omitting only the affected schedule metric.
  * Preserved existing behavior for suspended CronJobs and valid schedules.

* **Tests**
  * Updated monitoring output expectations for invalid schedule scenarios.

### DIFF
--- a/internal/store/cronjob.go
+++ b/internal/store/cronjob.go
@@ -30,6 +30,7 @@ import (
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	basemetrics "k8s.io/component-base/metrics"
+	"k8s.io/klog/v2"
 
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
 	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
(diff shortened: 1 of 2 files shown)


dpacgdm:
Closing this — the crash in #2988 was already fixed by #2978 (skip + `kube_cronjob_schedule_invalid`). This PR only added logging on top of that path, and the log message was broader than schedule-parse failures. Not worth the review noise; happy to revisit a tighter log-only change later if maintainers want it.

PR #3055, opened 2026-08-03
fix(metricshandler): wait for store sync before swapping writers on rebuild

harishkashyap569:

When custom resource state (CRS) discovery rebuilds metrics writers, kube-state-metrics starts new reflectors with **empty** stores and previously could swap those writers in before the initial list completed. Scrapes during that window returned HTTP 200 with an incomplete metric set (notably missing large families such as `kube_pod_*`), which looks like intermittent metrics gaps to Prometheus.

This change waits for the new reflector generation to sync (configurable via `--store-sync-timeout`), swaps writers only after a successful sync, and keeps serving the previous writer generation until then. Each scrape also snapshots the active writers under the read lock so a long `/metrics` response is not blocked by a concurrent rebuild.

## How does this change affect the users?

- Reduces empty/incomplete scrapes during CRS-driven writer rebuilds.
- Adds `--store-sync-timeout` (default matches existing store sync behavior) so operators can tune how long a rebuild waits for the new generation to become ready.
- During the sync window, memory may briefly hold two generations of stores (old + new). That is intentional for correctness; operators with very large CRS sets should be aware of peak memory during frequent rebuilds.

## Related Issues / PRs

Relates to #3093 (discovery-side `ConfigureStore` migration). This PR does not fully close the builder mutation race; that lands in #3093.

## Checklist

- [x] I have read [CONTRIBUTING.md](https://github.com/kubernetes/kube-state-metrics/blob/main/CONTRIBUTING.md) and [Developer Guide](https://github.com/kubernetes/kube-state-metrics/blob/main/docs/developer/guide.md)
- [x] My change requires a change to the documentation
- [x] I have updated the documentation accordingly
- [x] I have updated CHANGELOG.md
- [x] I have added/updated unit tests

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved metric updates during resource state rebuilds by waiting for initial store synchronization.
  * Preserved existing metrics when synchronization fails or times out, with automatic retries.
  * Ensured consistent metric snapshots during concurrent updates.
  * Reduced unnecessary metric rebuilds when discovered resources have not changed.

* **New Features**
  * Added configurable store synchronization timeouts.
  * Added the `--store-sync-timeout` option, defaulting to 120 seconds.
  * Coalesced configuration changes during metric rebuilds.

* **Documentation**
  * Documented the store synchronization timeout option and behavior.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,9 @@
 # Changelog
 
+## main / unreleased
+
+* [BUGFIX] Wait for reflector initial list before swapping metrics writers on custom resource state rebuild
+
 ## v2.19.1 / 2026-06-10
 
 * This release builds with Go `v1.26.4`
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -40,6 +40,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -92,6 +93,11 @@ type Builder struct {
 	objectLimit        int64
 
 	GetGVKStopChan func(gvk string) chan struct{}
+
+	// reflectorsMu guards reflectors, which Build() replaces while
+	// WaitForStoresSync() reads it outside the caller's own lock.
+	reflectorsMu sync.Mutex
+	reflectors   []*cache.Reflector
 }
 
 // NewBuilder returns a new builder.
@@ -295,6 +301,10 @@ func (b *Builder) Build() metricsstore.MetricsWriterList {
 	if b.familyGeneratorFilter == nil {
 		panic("familyGeneratorFilter should not be nil")
 	}
+
+	b.reflectorsMu.Lock()
+	b.reflectors = nil
+	b.reflectorsMu.Unlock()
 
 	var metricsWriters metricsstore.MetricsWriterList
 	var activeStoreNames []string
@@ -723,12 +733,37 @@ func (b *Builder) startReflector(
 ) {
 	instrumentedListWatch := watch.NewInstrumentedListerWatcher(listWatcher, b.listWatchMetrics, reflect.TypeOf(expectedType).String(), useAPIServerCache, objectLimit, client)
 	reflector := cache.NewReflectorWithOptions(sharding.NewShardedListWatch(b.shard, b.totalShards, instrumentedListWatch), expectedType, store, cache.ReflectorOptions{ResyncPeriod: 0})
+	b.reflectorsMu.Lock()
+	b.reflectors = append(b.reflectors, reflector)
+	b.reflectorsMu.Unlock()
 	if cr, ok := expectedType.(*unstructured.Unstructured); ok {
 		gvkStopCh := b.GetGVKStopChan(cr.GroupVersionKind().String())
 		go reflector.Run(newCRReflectorStopCh(b.ctx, gvkStopCh))
 	} else {
 		go reflector.Run(b.ctx.Done())
 	}
+}
+
+// WaitForStoresSync blocks until every reflector started by the latest Build() has
+// completed its initial list, or until ctx/timeout elapses.
+func (b *Builder) WaitForStoresSync(ctx context.Context, timeout time.Duration) bool {
+	b.reflectorsMu.Lock()
+	reflectors := slices.Clone(b.reflectors)
+	b.reflectorsMu.Unlock()
+
+	if len(reflectors) == 0 {
+		return true
+	}
+
+	err := wait.PollUntilContextTimeout(ctx, ResourceDiscoveryInterval, timeout, true, func(context.Context) (bool, error) {
+		for _, reflector := range reflectors {
+			if reflector.LastSyncResourceVersion() == "" {
+				return false, nil
+			}
+		}
+		return true, nil
+	})
+	return err == nil
 }
 
 // cacheStoresToMetricStores converts []cache.Store into []*metricsstore.MetricsStore
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -39,6 +39,7 @@ import (
 	rbacv1 "k8s.io/api/rbac/v1"
 	storagev1 "k8s.io/api/storage/v1"
 	"k8s.io/apimachinery/pkg/apis/meta/v1/unstructured"
+	"k8s.io/apimachinery/pkg/util/wait"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/klog/v2"
@@ -87,6 +88,8 @@ type Builder struct {
 	objectLimit         int64
 
 	GetGVKStopChan func(gvk string) chan struct{}
+
+	reflectors []*cache.Reflector
 }
 
 // NewBuilder returns a new builder.
@@ -277,6 +280,8 @@ func (b *Builder) Build() metricsstore.MetricsWriterList {
 	if b.familyGeneratorFilter == nil {
 		panic("familyGeneratorFilter should not be nil")
 	}
+
+	b.reflectors = nil
 
 	var metricsWriters metricsstore.MetricsWriterList
 	var activeStoreNames []string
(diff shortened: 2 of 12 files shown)


internal/store/builder.go
```diff
@@ -675,6 +689,28 @@ func (b *Builder) startReflector(
...
+// WaitForStoresSync blocks until every reflector started by the latest Build() has
+// completed its initial list, or until ctx/timeout elapses.
+func (b *Builder) WaitForStoresSync(ctx context.Context, timeout time.Duration) bool {
+	b.reflectorsMu.Lock()
+	reflectors := slices.Clone(b.reflectors)
+	b.reflectorsMu.Unlock()
+
+	if len(reflectors) == 0 {
+		return true
+	}
```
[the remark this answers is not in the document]
harishkashyap569: this is pre-existing Build/CRS-client behavior; fixing it means teaching `Build()` to signal construction failure or omit failed CRS writers — out of scope for the sync-before-swap fix.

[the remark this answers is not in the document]
harishkashyap569: Thanks for the review — verifying against current code, I think this is **pre-existing Build/CRS-client behavior**, not something introduced by treating an empty reflector list as sync success.

- `WaitForStoresSync` returning `true` when there are **no reflectors** is intentional: there is nothing to wait for (same as “already synced”).
- A missing custom-resource client already returned `[]cache.Store{}` and `Build()` already created a `MetricsWriter` with zero stores **before** this PR. Previously those writers were swapped in immediately; this PR only waits for reflectors that were actually started.
- In a mixed build (core resources + CRS), the reflector list is **not** empty when core stores start successfully, so the “empty list ⇒ success” path is not what drives an incomplete CRS swap.

Hardening this properly means teaching `Build()` to signal construction failure (or omit failed CRS writers) so `doRebuild` can retain the previous generation — that’s a broader change than the sync-before-swap fix and I’d prefer to handle it in a follow-up if maintainers want it.

Happy to adjust if you’d rather fail closed on zero reflectors when enabled resources were non-empty.

pkg/metricshandler/metrics_handler.go
```diff
@@ -69,18 +83,129 @@ func New(opts *options.Options, kubeClient kubernetes.Interface, storeBuilder ks
...
+		return
+	}
+	m.rebuildRunning = true
+	m.rebuildMu.Unlock()
+
+	if syncFirst {
+		m.rebuildLoop(ctx)
+		return
+	}
+	go m.rebuildLoop(ctx)
```
bhope, line 103: Now that this runs async, could these writes to storeBuilder (https://github.com/kubernetes/kube-state-metrics/blob/main/internal/discovery/discovery.go#L244-L259) land while Build() is still reading it? Could this be a race condition?

harishkashyap569: @bhope Good catch — async rebuilds do widen the window where discovery’s `With*` calls on storeBuilder can overlap a rebuild in flight.

In practice I don’t think this explains scrape gaps or continuous metrics loss:

Discovery only mutates the builder when CRDs change (WasUpdated), not on every tick.
The long part of a rebuild is `WaitForStoresSync`, which only inspects the reflectors from the current `Build()`. Config writes during that wait mainly set `pendingRebuild`, and the coalesced follow-up rebuild picks up the latest config.

On sync failure we keep the previous writers, so a racy/partial rebuild shouldn’t blank the scrape endpoint.
Steady state (no CRD churn) doesn’t hit this path at all.

That said, it is a real concurrency hygiene issue: discovery still updates shared builder state without the same serialization `Build()` uses, and returning from `BuildWriters` immediately makes a later discovery pass more likely to land mid-rebuild. Properly fixing that means serializing discovery config updates with rebuild (or snapshotting builder config before `Build()`), which is adjacent to — but broader than — the sync-before-swap change here.

Happy to track that as a follow-up if you’d prefer it addressed separately; I can also tighten it in this PR if you want it gated with the fix. - For now, added it to the same PR.

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Changelog
 
+## main / unreleased
```
mrueg: no need to fill out the changelog, we do that on release.

mrueg:
If I understand this change correctly, on startup this would return an empty http/200 until the first sync is complete (before it would send an incomplete set of metrics). Can we use the `/readyz` probe to signal that the first store sync has happened.

mrueg:
Can you split https://github.com/kubernetes/kube-state-metrics/pull/3055/changes/a49586a754eb118db88f54bbf29ecf24ffa07fdd into its own PR? that would make two PRs more sizeable

harishkashyap569:
## Validation (sync-before-swap + `/readyz`)

Tested commit `470afd28` on an internal production-like cluster with CRS-enabled kube-state-metrics (multi-shard deployments + node-local daemonset).

**Results**
| Check | Result |
|-------|--------|
| Writer swap after store sync | Initial swap ~600–700ms; CRS rebuild swap ~1m16s on extended shards |
| Sync timeouts / rebuild retries | 0 across all shards |
| `/readyz` (telemetry :8081) | 200 OK |
| `/metrics` (:8080) | Populated (~14k `kube_*` series on shard-0) |
| Node-local daemonset | Swapped successfully (`writerCount=1`, ~101ms sync) |

Example log:
```
"Swapped metrics writers after store sync" writerCount=27 syncWaitDuration="700.768138ms"
"Swapped metrics writers after store sync" writerCount=32 syncWaitDuration="1m17.200093667s"
```

## Discovery split

Per @mrueg's feedback, discovery changes are now in a separate stacked PR on top of this one (no-op CRD rebuild handling + `ConfigureStore`).

harishkashyap569:
Discovery half of the split is now upstream as #3093 (merge this PR first, then rebase #3093 onto `main`).

internal/store/builder.go
```diff
@@ -731,6 +746,28 @@ func (b *Builder) startReflector(
 	}
 }
 
+// WaitForStoresSync blocks until every reflector started by the latest Build() has
+// completed its initial list, or until ctx/timeout elapses.
+func (b *Builder) WaitForStoresSync(ctx context.Context, timeout time.Duration) bool {
```
bhope, line 751: Should this distinguish the two scenarios - reflector never synced because its stop channel closed vs reflector is still trying? Right now both look identical and both wait out the full timeout.

bhope:
Heads up - as written this doesn't fully close the race it's meant to fix. The discovery-side migration lands in #3093 so let's change the PR description from fixes to relates.

Left an inline as well.

internal/store/builder.go
harishkashyap569: Good question. Right now we only look at `LastSyncResourceVersion`, so a reflector whose stop channel already closed looks the same as one that's still listing — both sit there until `--store-sync-timeout`. If it never synced we keep the previous writers and retry, so scrapes shouldn't go empty, but we can burn the full timeout on a dead reflector.

client-go doesn't expose "was stopped". Distinguishing it would mean tracking the stop ch we pass to `Run` and treating closed + empty `LastSync` as will-never-sync. I can add that here if you want, otherwise I'd rather leave it as a follow-up so this stays scoped to the swap-after-sync behavior.

pkg/metricshandler/metrics_handler.go
```diff
@@ -69,35 +87,171 @@ func New(opts *options.Options, kubeClient kubernetes.Interface, storeBuilder ks
...
+	m.mtx.RLock()
+	defer m.mtx.RUnlock()
+	return m.writersInstalled
+}
+
+// ConfigureStore applies storeBuilder configuration under mtx, then rebuilds
+// writers. Runtime callers that mutate the shared builder (for example CR
+// discovery) must use this instead of calling builder With* methods directly,
+// so configuration cannot race with Build(). Returns an error without rebuilding
+// when configuration fails.
```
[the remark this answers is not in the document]
harishkashyap569: Same race Brian flagged — discovery still mutates `storeBuilder` directly in this PR. `ConfigureStore` is the API for that; the discovery-side switch is in #3093 so we can land the sync-before-swap change first.

internal/store/builder_sync_test.go
```diff
@@ -0,0 +1,51 @@
...
+func TestWaitForStoresSync_Timeout(t *testing.T) {
+	b := NewBuilder()
+	store := cache.NewStore(cache.MetaNamespaceKeyFunc)
+	lw := &cache.ListWatch{
+		ListFunc: func(_ metav1.ListOptions) (runtime.Object, error) {
+			return &v1.PodList{}, nil
+		},
+	}
+	reflector := cache.NewReflectorWithOptions(lw, &v1.Pod{}, store, cache.ReflectorOptions{})
+	b.reflectors = []*cache.Reflector{reflector}
```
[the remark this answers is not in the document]
harishkashyap569: Makes sense — I'll take `reflectorsMu` in that test so it matches what `WaitForStoresSync` does.

harishkashyap569:
@bhope Yeah that's fair — this PR only waits for store sync before swapping writers. Discovery still calls `With*` on the shared builder, so the race isn't fully closed here. The `ConfigureStore` migration is in #3093. I'll update the description to relate to that instead of implying this fully fixes it.

PR #3058, opened 2026-08-05
fix: fix issues with wrong labels being applied to endpointslices metric

odinuge:

This fixes the ordering between terminating and hostname labels for endpointslice metrics, where they are swapped causing kube-state-metrics to emit strange metrics. This also ensures we don't overrwrite serving with terminating, and that terminating actually gets a value.

This means it fixes these three labels that have been broken
- serving, was set to the 'terminating' value
- terminating, was set to the 'hostname' value
- hostname, was set to the terminating value that was always an empty string

Fixes: 75e7841 ("endpointslices: Expose empty labels")

Fixes #

## Summary by CodeRabbit

* **Bug Fixes**
  * Corrected endpoint metrics to accurately report serving and terminating conditions.
  * Restored the correct hostname and label ordering in endpoint metrics.
  * Ensured endpoint metrics remain accurate when topology hints are present.

### DIFF
--- a/internal/store/endpointslice.go
+++ b/internal/store/endpointslice.go
@@ -153,7 +153,7 @@ func createEndpointsSliceEndpoints() generator.FamilyGenerator {
 				}
 
 				if ep.Conditions.Terminating != nil {
-					serving = strconv.FormatBool(*ep.Conditions.Terminating)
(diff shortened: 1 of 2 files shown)


mrueg:
thanks for the fix!
/lgtm

PR #3059, opened 2026-08-09
fix(pod): do not emit nil metrics for unparseable pod IPs

mrueg:

`createPodIPFamilyGenerator` sized the metric slice with the number of pod IPs and assigned by index, but skipped entries whose IP failed to parse. Every skipped IP therefore left a `nil *metric.Metric` in the family, which `wrapPodFunc` then dereferences while merging the default labels — so a malformed `status.podIPs` entry crashes metric generation instead of producing one fewer series.

This appends the parsed entries instead, so skipped IPs leave no hole.

While here, the address is resolved with `net/netip` rather than the deprecated `ParseIPSloppy` helper from `k8s.io/utils/net`, and v4-in-v6 addresses are unmapped so they keep reporting `ip_family="4"` (covered by a new test case). Note that `netip.ParseAddr` is stricter than `ParseIPSloppy`: an address with leading zeros (`010.1.1.1`) is now skipped rather than reported. Kubelet-assigned IPs never use that form.

The label keys are hoisted into a package-level variable, matching `descPodLabelsDefaultLabels`, so they are not reallocated on every pod event.

**How does this change affect the cardinality of KSM:** does not change cardinality, except that a pod with an unparseable IP no longer crashes.

Split out of #2916.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved pod IP metric handling for invalid and IPv4-mapped addresses.
  * Invalid pod IPs are now excluded from metrics instead of producing empty entries.
  * IPv4-mapped IPv6 addresses are correctly reported as IPv4.

### DIFF
--- a/internal/store/pod.go
+++ b/internal/store/pod.go
@@ -18,11 +18,11 @@ package store
 
 import (
 	"context"
+	"net/netip"
 	"strconv"
 
 	basemetrics "k8s.io/component-base/metrics"
 	"k8s.io/component-helpers/storage/ephemeral"
-	"k8s.io/utils/net"
 
 	"k8s.io/kube-state-metrics/v2/pkg/constant"
 	"k8s.io/kube-state-metrics/v2/pkg/metric"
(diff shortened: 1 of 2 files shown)


bhope (approved):
Thanks for the fix!
/lgtm

PR #3060, opened 2026-08-09
perf: reduce allocations in metric generation and header sanitization

mrueg:

Three independent allocation reductions on the per-event and per-scrape paths, each measured with `benchstat` (6 runs).

**1. `SanitizeHeaders`: precompute per-store headers** — headers were re-parsed and rewritten on every scrape. They are now parsed once in `NewMetricsStore`, for both exposition formats.

```
SanitizeHeaders/text-format_unique_headers     16.03m -> 4.55m    -71.60%
                                           B/op 10.82Mi -> 1.56Mi -85.60%
                                     allocs/op 100265  -> 10      -99.99%
```

`escapeString` also gained a fast path that skips the `strings.Replacer` when nothing needs escaping (`MetricWrite`: -7 to -11% time, same allocs).

**2. `mapToPrometheusLabels`** — pre-sizes the sorted-keys slice and only allocates the conflict map when two Kubernetes keys actually sanitize to the same Prometheus label name (the uncommon case). Runs per object whenever a labels/annotations allowlist is configured:

```
labels-8    allocs/op  84 -> 75   -10.71%    B/op 2.41Ki -> 2.22Ki  -8.16%
labels-32   allocs/op 333 -> 291  -12.61%    B/op 13.0Ki -> 8.87Ki -31.95%   sec/op -38.83%
```

A `BenchmarkMapToPrometheusLabels` covering both the plain and the conflict path is included.

**3. `MetricsStore.Add`: one buffer per object, not per family** — every family allocated its own `bytes.Buffer`, which then grew geometrically as metrics were written into it (~59 buffers per pod). All families of an object now render into a single buffer sized up front from `Family.SizeHint()`, and each family gets a view into it. Empty families no longer pay `bytes.Buffer.Grow`s minimum allocation, and trailing slack is trimmed so the retained buffer is not larger than the per-family slices it replaces:

```
BenchmarkPodStoreAdd  sec/op 38.37µ -> 30.36µ  -20.89%
                        B/op 34.07Ki -> 25.00Ki -26.62%
                   allocs/op    380 -> 282      -25.79%
```

`Family.AppendBytes` and `Family.SizeHint` are additive; `FamilyInterface` is unchanged and anything not implementing them falls back to `ByteSlice()`. The scrape path (`WriteAll`) is unaffected.

**How does this change affect the cardinality of KSM:** does not change cardinality.

Split out of #2916.

## Summary by CodeRabbit

* **Performance**
  * Improved metric rendering with reusable buffers and size estimation, reducing allocations during output generation.
  * Optimized label conversion, string escaping, and metric-store updates.
  * Streamlined header processing and reuse for faster metrics responses.
* **Compatibility**
  * Preserved existing metric output and escaping behavior.
  * Continued consistent normalization of OpenMetrics metadata and label names.
* **Testing**
  * Added performance benchmarks covering metric rendering, label conversion, and metric-store updates.

### DIFF
--- a/internal/store/utils.go
+++ b/internal/store/utils.go
@@ -86,7 +86,7 @@ func mapToPrometheusLabels(labels map[string]string, prefix string) ([]string, [
 	labelKeys := make([]string, 0, len(labels))
 	labelValues := make([]string, 0, len(labels))
 
-	sortedKeys := make([]string, 0)
+	sortedKeys := make([]string, 0, len(labels))
 	for key := range labels {
 		sortedKeys = append(sortedKeys, key)
 	}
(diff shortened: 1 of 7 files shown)


bhope (commented):
These are reasonable performance optimizations and the changes look good. Thanks for the PR.

/lgtm
/hold

Feel free to unhold.

PR #3061, opened 2026-08-10
fix(metrics_store): write headers when only a later store has metrics

mrueg:

`WriteAll` emits each `# HELP`/`# TYPE` header from inside a `Range` over `m.stores[0]`, so the header is skipped whenever the first store holds no objects. The metric families underneath are written from *every* store:

```go
if help != "" {
    m.stores[0].metrics.Range(func(...) bool { io.WriteString(w, help); return false })
}
for _, s := range m.stores {           // <- every store
    s.metrics.Range(... w.Write(metricFamilies[i]) ...)
}
```

`buildStores` creates one store per namespace, so a deployment scoped to several namespaces (`--namespaces=a,b`) emits bare metric lines whenever the first namespace happens to hold no objects of that type. The type metadata is silently lost, and it comes back as soon as that namespace gains an object, so the exposition flaps with cluster contents rather than staying stable.

Before, with an empty first store:

```
kube_service_info_1{namespace="b",uid="b1"} 1
```

After:

```
# HELP kube_service_info_1 Info 1 about services
# TYPE kube_service_info_1 gauge
kube_service_info_1{namespace="b",uid="b1"} 1
```

This PR determines once, across all stores, whether there is anything to describe, and keeps the existing behaviour of writing no headers at all when every store is empty (covered by `TestWriteAllWithEmptyStores`). As a side effect it replaces a `Range` per header with a single pass.

Added `TestWriteAllWithEmptyFirstStore`, which fails on `main` with exactly the output above and passes with the fix.

This is long-standing rather than a recent regression — the pattern predates the current header-sanitization code.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed metrics output when the first store is empty but later stores contain data.
  * Metric help headers are now emitted correctly for all available metric families.
  * Preserved proper newline handling and error reporting during metrics output.

### DIFF
--- a/pkg/metrics_store/metrics_writer.go
+++ b/pkg/metrics_store/metrics_writer.go
@@ -80,28 +80,41 @@ func (m MetricsWriter) WriteAll(w io.Writer) error {
 		return nil
 	}
 
+	// Headers describe the families written below, so they are emitted only when
+	// there is at least one object to describe. That has to consider every store,
+	// not just the first: there is one store per namespace, so the first can be
+	// empty while a later one holds objects whose families are still written.
+	// The answer is the same for every header, so determine it once.
+	hasMetrics := false
+	for _, s := range m.stores {
+		s.metrics.Range(func(_ interface{}, _ interface{}) bool {
+			hasMetrics = true
+			return false
+		})
+		if hasMetrics {
+			break
+		}
+	}
+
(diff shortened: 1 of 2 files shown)


bhope (commented):
Thanks for the fix, one of the cases for multi-NS scope :) 

/lgtm

PR #3064, opened 2026-08-10
fix(store): never emit the same label name twice

mrueg:

`mapToPrometheusLabels` resolves colliding sanitized label names by renaming them to `<name>_conflictN`, but it never checks whether the generated name is itself already in use. A key that sanitizes directly onto that name collides with it:

```
{"A.b.conflict1": "v0", "a-b": "v1", "a.b": "v2"}     // all valid Kubernetes label keys
=> keys [label_a_b_conflict1  label_a_b_conflict1  label_a_b_conflict2]
```

`A.b.conflict1` sanitizes to `label_a_b_conflict1` and sorts first; the later `a-b`/`a.b` collision then renames the second entry onto that same name. `pkg/metric/metric.go` writes label keys verbatim with no dedup, so the exposed sample carries `label_a_b_conflict1` twice.

**Prometheus rejects a sample with a duplicate label name, and it fails the entire scrape rather than the individual series** — so one object carrying such a label pair suppresses every metric from the target, not just its own. The keys are reachable through `--metric-labels-allowlist` (or the annotations equivalent) for any label a tenant controls on their own namespaces, deployments, pods, and so on.

This PR tracks the names already handed out and advances the conflict suffix past any that are taken, so the output is `_conflict1`/`_conflict2`/`_conflict3` and every emitted name is unique.

Existing conflict behaviour is unchanged — the two current conflict test cases pass untouched. Added a case for the collision above; it fails on `main` with the duplicate shown and passes here.

Incidentally this replaces the linear rescan of `labelKeys` with a map lookup, so conflict resolution is no longer O(n²) in the number of labels on an object.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed Prometheus label handling when label names conflict with generated suffixes.
  * Ensured all labels remain unique while preserving their associated values.

* **Tests**
  * Added coverage for conflicting label names and suffix progression.

### DIFF
--- a/internal/store/utils.go
+++ b/internal/store/utils.go
@@ -92,51 +92,52 @@ func mapToPrometheusLabels(labels map[string]string, prefix string) ([]string, [
 	}
 	slices.Sort(sortedKeys)
 
-	// conflictDesc holds some metadata for resolving potential label conflicts
-	type conflictDesc struct {
-		// the number of conflicting label keys we saw so far
-		count int
+	// used maps an emitted label name to its offset in labelKeys, so that the
+	// first time a name collides the entry already emitted under it can be
+	// renamed. conflicts counts the suffixes handed out per sanitized name.
+	// Prometheus rejects a sample carrying the same label name twice, so a
+	// generated "_conflictN" name must be checked against used as well: it can
+	// collide with a name some other key sanitized to.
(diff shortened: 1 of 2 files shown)


bhope:
/lgtm 

Thanks for the PR!

PR #3066, opened 2026-08-10
fix(customresourcestate): do not panic on a metric that fails to compile

mrueg:

`compileCommon` returns a nil `*compiledCommon` alongside its error, but all three metric types assign through `cc` before checking `err`:

```go
cc, err := compileCommon(m.Gauge.MetricMeta)
cc.t = metric.Gauge                                  // <- nil deref when err != nil
if err != nil {
	return nil, fmt.Errorf("each.gauge: %w", err)
}
```

So any `path` or `labelsFromPath` that fails to compile panics instead of being reported. A plausible config typo is enough — a list lookup written without its `key=value`:

```yaml
metrics:
- name: bar
  each:
    type: Gauge
    gauge: {path: [status, conditions, "[Ready]"]}   # meant [type=Ready]
```

`newCompiledMetric` is reached from the factory generator running on the discovery poll goroutine, so nothing recovers and the process dies. Worse, it only fires once a *matching CRD is discovered*, so it presents as a random crash some time after startup rather than as the config error it is. All three types (`Gauge`, `Info`, `StateSet`) are affected.

This PR checks the error before touching `cc` at all three sites.

**Also in here** (same file, one line): `compilePath`'s list-index op returned `fmt.Errorf("list index out of range: %s", part)` as the resolved *value* rather than signalling non-resolution. Every other unresolvable path yields nil and is skipped, so an out-of-range index in the final path position surfaced as a label value:

```
idx="list index out of range: 5"
```

and through a gauge path as the misleading `expected number but was list index out of range: 5`. It now returns nil like its neighbours. Only a trailing index was affected — with a following path element the error object was consumed and turned into nil anyway, which is why this went unnoticed.

Both are covered by new tests that fail on `main` (the first with the nil-pointer panic, the second with the bogus label value) and pass here.

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid list indexes are now treated as missing values instead of causing errors.
  * Metric configuration failures now return clear errors without causing crashes.
  * Prevented invalid metric definitions from attempting to write through unavailable results.

* **Tests**
  * Added coverage for invalid list indexes and metric compilation failures.

### DIFF
--- a/pkg/customresourcestate/registry_factory.go
+++ b/pkg/customresourcestate/registry_factory.go
@@ -160,10 +160,12 @@ func newCompiledMetric(m Metric) (compiledMetric, error) {
 			return nil, errors.New("expected each.gauge to not be nil")
 		}
 		cc, err := compileCommon(m.Gauge.MetricMeta)
-		cc.t = metric.Gauge
 		if err != nil {
+			// compileCommon returns a nil *compiledCommon alongside its error,
+			// so nothing may be assigned through cc before this check.
 			return nil, fmt.Errorf("each.gauge: %w", err)
 		}
+		cc.t = metric.Gauge
 		valueFromPath, err := compilePath(m.Gauge.ValueFrom)
 		if err != nil {
 			return nil, fmt.Errorf("each.gauge.valueFrom: %w", err)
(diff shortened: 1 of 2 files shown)


bhope:
/lgtm

Thanks for the PR!

PR #3067, opened 2026-08-10
fix(metric_generator): escape reserved characters in help text

mrueg:

`generateHeader` writes the help string verbatim into the HELP line:

```go
header.WriteString("# HELP ")
header.WriteString(g.Name)
header.WriteByte(' ')
header.WriteString(g.Help)      // <- unescaped
header.WriteByte('\n')
```

The Prometheus text format reserves `\` and newline in HELP — upstream `expfmt` escapes exactly those two — but KSM escapes label values only. The consequences are not local to the family:

| help text | result |
| --- | --- |
| `matches \d+ digits` | `text format parsing error in line 1: invalid escape sequence '\d'` |
| `ends with \` | `invalid escape sequence '\<newline>'` |
| `line one\nline two` | `text format parsing error in line 2: expected float as value, got "two"` |

A scrape that fails to parse is discarded in full, so one bad help string suppresses every metric the process exposes. A newline additionally breaks `parseHeaderStatic`, which locates the TYPE line by scanning to the first `\n`; with a multi-line HELP the OpenMetrics-only `info`/`stateset` suffix is then never rewritten to `gauge` for `text/plain`.

The built-in help strings contain neither character, so this only bites users of the CustomResourceState feature — where `help:` is user-supplied YAML that reaches this function with no validation or escaping anywhere on the path (`registry_factory.go` → `famGen` → `NewFamilyGeneratorWithStability`). A backslash while describing a path or a regex is an easy thing to write.

This PR escapes both characters on both write paths (the STABLE branch prefixes the stability level and needs the same treatment).

`pkg/metric_generator` had **no test files at all**; this adds the first, covering plain text, backslash, trailing backslash, newline, and the STABLE path. They fail on `main` and pass here. Built-in help strings are unaffected — I checked none contain a backslash, and the `internal/store` and `pkg/metrics_store` suites, which assert exposition output extensively, are unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved Prometheus metric headers by safely escaping backslashes, newlines, and carriage returns in help text.
  * Preserved correct formatting for both stable and non-stable metrics, including headers that contain trailing backslashes.

### DIFF
--- a/pkg/metric_generator/generator.go
+++ b/pkg/metric_generator/generator.go
@@ -76,15 +76,20 @@ func (g *FamilyGenerator) Generate(obj interface{}) *metric.Family {
 	return family
 }
 
+// escapeHelp escapes the characters the exposition format reserves in a HELP
+// line. A newline would end the line early and a backslash would start an
+// escape sequence, either of which makes the whole exposition unparseable.
+var escapeHelp = strings.NewReplacer(`\`, `\\`, "\n", `\n`)
+
 func (g *FamilyGenerator) generateHeader() string {
 	header := strings.Builder{}
 	header.WriteString("# HELP ")
 	header.WriteString(g.Name)
 	header.WriteByte(' ')
 	if g.StabilityLevel == basemetrics.STABLE {
-		fmt.Fprintf(&header, "[%v] %v", g.StabilityLevel, g.Help)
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

Thanks for the fix!

PR #3068, opened 2026-08-10
fix: three fixes in the server request and shutdown paths

mrueg:

Three small, independent fixes in the serving path. Grouped because each is a couple of lines; happy to split if preferred.

### 1. Gzip negotiation wraps the response writer once per token

`pkg/metricshandler/metrics_handler.go` — the loop over `Accept-Encoding` has no `break`, unlike `promhttp.decorateWriter` which the comment says it was taken from:

```go
for _, part := range parts {
	if part == "gzip" || strings.HasPrefix(part, "gzip;") {
		writer = gzip.NewWriter(writer)      // <- once per matching token
		resHeader.Set("Content-Encoding", "gzip")
	}
}
```

Only the outermost writer is closed at the end of the handler, so with a duplicated token the inner `gzip.Writer` is never flushed or finalised and its trailer never reaches the response. The client gets `Content-Encoding: gzip` and a **zero-length body**, failing with `gzip: unexpected EOF`.

Requires `--enable-gzip-encoding` plus a client or intermediary that emits the token twice — Prometheus itself does not — so the blast radius is small, but the fix is a single `break`.

Covered by a new test over `gzip`, `gzip, gzip` and `gzip;q=1.0, gzip`; the duplicate cases fail on `main` with `unexpected EOF` and the single-token case is unaffected either way.

### 2. The shutdown grace period was always zero

`pkg/app/server.go` — both server interrupt handlers did:

```go
ctxShutDown, cancel := context.WithTimeout(ctx, 3*time.Second)
```

`ctx` is precisely the context whose cancellation triggered the shutdown, so the derived context is already `Done` and `Shutdown` closes the listeners and returns `ctx.Err()` immediately, without draining in-flight scrapes. Derived from `context.Background()` now, so the intended three seconds actually apply.

### 3. The `/healthz` and `/livez` delegation had no deadline

`handleClusterDelegationForProber` issued its upstream request with `context.Background()`, and the shared `rest.Config` sets no client `Timeout`. Against an API server that accepts the connection but never responds — packets dropped rather than refused — each probe pins a handler goroutine and a connection for the TCP keepalive duration, minutes after the kubelet has given up. The server's `WriteTimeout` closes the client connection but does not unblock the goroutine. It now uses the request context, so it is cancelled when the prober disconnects.

Fixes 2 and 3 are not unit-tested: both need a real shutdown sequence or a blackholed API server to observe, and neither seemed worth the harness. They are visible by inspection.

### DIFF
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -379,7 +379,10 @@ func RunKubeStateMetrics(ctx context.Context, opts *options.Options) error {
 			klog.InfoS("Started kube-state-metrics self metrics server", "telemetryAddress", telemetryListenAddress)
 			return web.ListenAndServe(&telemetryServer, &telemetryFlags, sLogger)
 		}, func(error) {
-			ctxShutDown, cancel := context.WithTimeout(ctx, 3*time.Second)
+			// Not derived from ctx: the interrupt handler runs because ctx was
+			// cancelled, so a child of it is already done and Shutdown would
+			// return immediately instead of draining in-flight scrapes.
+			ctxShutDown, cancel := context.WithTimeout(context.Background(), 3*time.Second)
 			defer cancel()
 			telemetryServer.Shutdown(ctxShutDown)
 		})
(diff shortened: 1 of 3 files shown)


bhope (commented):
/lgtm

All three fixes make sense. Thanks!

PR #3070, opened 2026-08-10
fix: serialize config reloads so one does not strand the other

mrueg:

`RunKubeStateMetricsWrapper` watches up to three files — the options config, the custom resource state config and the kubeconfig — and each gets its own `viper` instance. `WatchConfig` starts a goroutine per instance, so every `OnConfigChange` callback runs on its own goroutine. All three closed over the same `ctx` and `cancel` and wrote them with no synchronisation:

```go
cancel()
<-time.After(3 * time.Second)
ctx, cancel = context.WithCancel(context.Background())   // shared, unsynchronised
go KSMRunOrDie(ctx)
```

Two of them firing together is not an edge case: the options config and the custom resource state config are normally keys in the **same** ConfigMap, and kubelet updates the mounted files together, so one edit fires both watchers.

When that happens the second callback overwrites `cancel` before the first has used it. The instance the lost `cancel` belonged to is never stopped, so two instances end up contending for `:8080`; the loser's bind error propagates to `KSMRunOrDie`, which calls `klog.FlushAndExit(1)`. A routine config reload takes the process down.

The three callbacks were byte-identical, so this collapses them into a single `restart` function guarded by a mutex — reloads now queue instead of racing, and the duplication goes away with it.

Not unit-tested: the reload path has no test harness today, and exercising it means driving real `fsnotify` events against three watchers with controlled timing. `go test -race` would not have caught this for the same reason.

### DIFF
--- a/internal/wrapper.go
+++ b/internal/wrapper.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"os"
 	"path/filepath"
+	"sync"
 	"time"
 
 	"github.com/fsnotify/fsnotify"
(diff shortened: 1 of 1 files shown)


bhope (commented):
/lgtm
/hold

@mrueg  restart serializes but doesn't merge concurrent triggers, so one `ConfigMap` edit touching both configs still causes two sequential full restarts. Might be worth doing a follow up on?

Feel free to unhold

PR #3071, opened 2026-08-10
fix(hpa): skip a status metric whose source is not set

mrueg:

`createHPAStatusTargetMetric` switches on `MetricStatus.Type` and dereferences the matching source pointer with no nil check:

```go
switch m.Type {
case autoscaling.ObjectMetricSourceType:
	metricName = m.Object.Metric.Name        // <- m.Object may be nil
	currentMetric = m.Object.Current
...
```

The **spec** counterpart is safe because `validateMetricSpec` requires the field named by `Type` to be populated. The **status** counterpart is not: `ValidateHorizontalPodAutoscalerStatusUpdate` validates only `currentReplicas`, `desiredReplicas` and the conditions, and `autoscalerStatusStrategy.ValidateUpdate` is the only validation applied to the `/status` subresource. `status.currentMetrics` is entirely unchecked, so this is accepted:

```
PATCH hpa/status  {"status":{"currentMetrics":[{"type":"Resource"}]}}
```

Generating metrics for that object panics with a nil pointer dereference on the reflector goroutine, and nothing in `pkg/` or `internal/` recovers, so the process dies. Writing an HPA's status requires privileges the HPA controller normally holds — the controller itself always writes a well-formed status — so this is a robustness gap rather than an unprivileged DoS, but the fix is cheap and mirrors what the switch already does for an unrecognised type.

Added `TestHPAStatusTargetMetricWithNilSource` covering all five source types; each panics on `main` and passes here.

### DIFF
--- a/internal/store/horizontalpodautoscaler.go
+++ b/internal/store/horizontalpodautoscaler.go
@@ -266,21 +266,40 @@ func createHPAStatusTargetMetric() generator.FamilyGenerator {
 				// The variable maps the type of metric to the corresponding value
 				metricMap := make(map[metricTargetType]float64)
 
+				// Unlike spec.metrics, status.currentMetrics is not validated:
+				// ValidateHorizontalPodAutoscalerStatusUpdate checks only the
+				// replica counts and the conditions, so the source matching Type
+				// may be absent. Skip such an entry rather than dereferencing it.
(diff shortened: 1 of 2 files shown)


bhope:
/lgtm

Thanks for the PR!

PR #3073, opened 2026-08-10
fix(metricshandler): scope the scrape lock to reading the writers

mrueg:

`ServeHTTP` holds `mtx.RLock` for the entire response:

```go
func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
	m.mtx.RLock()
	defer m.mtx.RUnlock()
```

That covers negotiation, gzip wrapping, header sanitisation and the write itself — i.e. as long as the slowest client takes to read the body. `BuildWriters` and `ConfigureSharding` need the write lock, so a single slow scraper delays a re-shard or a config rebuild for its whole duration. Worse, Go's `sync.RWMutex` blocks *new* readers once a writer is waiting, so one slow client plus one re-shard stalls every scrape queued behind them.

The lock is only needed to read `m.metricsWriters`. Writers are replaced wholesale by `BuildWriters` rather than mutated in place, so snapshotting the slice header under the lock and releasing it immediately gives the response a self-consistent view even if a rebuild lands mid-write: the old writers stay readable, they simply stop being updated.

No dedicated test — reproducing the stall needs a slow client racing a rebuild, which is timing-dependent and would be flaky. `go test -race` on the package is clean, and `TestResourceFiltering` in `pkg/app` continues to pass, which exercises `ServeHTTP` end to end.

**Previously in this PR, now reverted:** I had also changed the resource filters to compare on length rather than against nil, on the theory that `?resources=` (empty value) returning no metrics was an undocumented asymmetry. It is not — `TestResourceFiltering` already pins `?resources=,,,,` to return nothing and `?exclude_resources=,,,,` to return everything, which is coherent: `resources` is an allowlist, so an empty one allows nothing, and `exclude_resources` is a denylist, so an empty one denies nothing. CI caught it. Only the lock scoping remains.

## Summary by CodeRabbit

* **Performance**
  * Improved metrics response handling so configuration updates can continue while responses are being generated.
  * Ensured each metrics response uses a consistent set of configured writers.

### DIFF
--- a/pkg/metricshandler/metrics_handler.go
+++ b/pkg/metricshandler/metrics_handler.go
@@ -192,8 +192,16 @@ func (m *MetricsHandler) Run(ctx context.Context) error {
 // ServeHTTP implements the http.Handler interface. It writes all generated metrics to the response body.
 // Note that all operations defined within this procedure are performed at every request.
 func (m *MetricsHandler) ServeHTTP(w http.ResponseWriter, r *http.Request) {
+	// Snapshot the writers instead of holding the lock for the whole response.
+	// BuildWriters and ConfigureSharding need the write lock, so a slow client
+	// would otherwise delay a re-shard -- and since a waiting writer blocks new
+	// readers, every scrape queued behind it too. Writers are replaced wholesale
+	// rather than mutated, so a snapshot stays readable and self-consistent even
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

PR #3074, opened 2026-08-10
fix(discovery): read the GVK cache under the lock

mrueg:

`ResolveGVKToGVKPs` reads and range-iterates `r.Map` with no lock held — at four separate points, one per resolution path. Every other access to that map goes through `SafeRead`/`SafeWrite`, including the CRD informer's add, update and delete handlers, which write it from the informer goroutine.

`ResolveGVKToGVKPs` runs on the discovery poll goroutine, via the factory generator in `pkg/customresourcestate/config.go`. So any CRD add, update or delete that lands while a rebuild is resolving configured resources is a concurrent map read and write. That is a **fatal runtime throw** (`fatal error: concurrent map read and map write`), not a panic anything can recover from — the process dies. Installing an operator bundle with several CRDs, or ordinary CRD status churn, is enough.

**Why the obvious fix does not work:** wrapping the body in `SafeRead` deadlocks. The same function records and clears the "missing GVK" warning through `markMissingGVKWarned`/`clearMissingGVKWarning`, which take the write lock themselves, and `sync.RWMutex` is not reentrant — so the function would block against a lock it already holds.

Instead the body takes the lock once and uses lock-free `...Locked` variants of the warning helpers. One acquisition covers both the map reads and the warning bookkeeping, which is also what makes the read-then-mark sequence atomic rather than a check-then-act across two acquisitions. The `klog` call is deliberately left outside the critical section.

`clearMissingGVKWarning`'s locking wrapper became dead once the caller moved to the locked variant, so it is removed rather than kept behind a lint suppression; `markMissingGVKWarned` keeps its wrapper because the existing dedup test calls it.

### Testing

`TestResolveGVKToGVKPsIsRaceFree` drives the informer's append/remove against all four resolution paths concurrently. Against the current code it reports `WARNING: DATA RACE`; with this change it is clean. CI already runs unit tests with `-race`, so this stays covered.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved concurrent discovery resolution to prevent race conditions.
  * Stabilized exact and wildcard matching when mappings are updated simultaneously.
  * Improved handling and logging of missing-resource warnings during concurrent operations.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -144,74 +144,86 @@ func (r *CRDiscoverer) ResolveGVKToGVKPs(gvk schema.GroupVersionKind) (resolvedG
 	}
 	hasVersion := v != "" && v != "*"
 	hasKind := k != "" && k != "*"
-	// No need to resolve, return.
-	if hasVersion && hasKind {
-		for _, el := range r.Map[g][v] {
-			if el.Kind == k {
-				r.clearMissingGVKWarning(gvk)
-				return []groupVersionKindPlural{
-					{
-						GroupVersionKind: schema.GroupVersionKind{
-							Group:   g,
-							Version: v,
-							Kind:    k,
-						},
-						Plural: el.Plural,
-					},
-				}, nil
-			}
-		}
-		if r.markMissingGVKWarned(gvk) {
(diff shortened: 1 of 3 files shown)


bhope (approved):
/lgtm

Thanks for the PR

PR #3075, opened 2026-08-10
fix(store): guard the builder state shared with discovery

mrueg:

`availableStores` is package state, seeded with the built-in resources, and `WithCustomResourceStoreFactories` adds entries to it:

```go
if _, ok := availableStores[gvrString]; ok { ... }
availableStores[gvrString] = func(b *Builder) []cache.Store { ... }
```

Custom resource discovery calls that from its poll goroutine with no lock held (`internal/discovery/discovery.go`, before `BuildWriters`), while `Build` and `BuildStores` read the same map from whichever goroutine drove the rebuild — the autosharding informer, for instance, reacting to a StatefulSet replica change.

A concurrent map read and write is a **fatal runtime throw**, not a panic anything can recover from. So a CRD add, update or delete landing alongside a re-shard can kill the process. The window is short and no test drives both paths, which is why `-race` in CI has never seen it.

This puts the map behind a `sync.RWMutex`, reached through `lookupStore` and `registerStore` so no access can bypass it, and guards the `range` in `availableResources` too. `resourceExists` goes through `lookupStore`.

`b.enabledResources` has exactly the same pair of writers and readers — `WithEnabledResources` appends and sorts it from the discovery goroutine, `Build` ranges over it — so it gets a lock as well. `Build` takes a **copy** rather than holding the read lock for the whole build: the store constructors it invokes mutate the builder as they run (`buildStores` swaps `b.namespaces` and restores it via defer), so holding a read lock across them would mean a write under a read lock.

### Scope

The two remaining fields discovery writes, `customResourceClients` and `buildCustomResourceStoresFunc`, are whole-value assignments read later during the build. They are still races in the strict sense, but they are single-word writes rather than map mutation, so they cannot produce the fatal throw this PR is about. Fixing them properly means making `Build` stop mutating the builder mid-flight, which is a larger refactor and better kept separate.

### Testing

`TestBuilderSharedStateIsRaceFree` drives factory registration and resource enablement against the rebuild's reads. Against the current code the race detector reports three data races; with this change it is clean. CI already runs unit tests with `-race`.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved reliability when store configuration and enabled resources are accessed or updated concurrently.
  * Prevented inconsistent resource and store lookups during simultaneous operations.
  * Custom resource replacements now report their registration status accurately.

* **Tests**
  * Added concurrency coverage to verify race-free behavior across repeated parallel operations.

### DIFF
--- a/internal/store/builder.go
+++ b/internal/store/builder.go
@@ -23,6 +23,7 @@ import (
 	"slices"
 	"strconv"
 	"strings"
+	"sync"
 	"time"
 
 	"github.com/prometheus/client_golang/prometheus"
@@ -80,11 +81,15 @@ type Builder struct {
 	// namespaceFilter is inside fieldSelectorFilter
 	fieldSelectorFilter string
 	namespaces          options.NamespaceList
-	enabledResources    []string
-	totalShards         int
-	shard               int32
-	useAPIServerCache   bool
-	objectLimit         int64
+	// enabledResourcesMu guards enabledResources, which WithEnabledResources
+	// appends to from the custom resource discovery goroutine while Build reads
+	// it from the goroutine driving the rebuild.
+	enabledResourcesMu sync.RWMutex
+	enabledResources   []string
+	totalShards        int
+	shard              int32
+	useAPIServerCache  bool
+	objectLimit        int64
 
 	GetGVKStopChan func(gvk string) chan struct{}
 }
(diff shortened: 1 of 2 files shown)


bhope:
/lgtm

Thanks for the fix!

PR #3076, opened 2026-08-10
fix(endpointslice): build each hint's labels independently

mrueg:

`kube_endpointslice_endpoints_hints` builds one metric per zone by appending onto label slices shared across the loop:

```go
labelKeys = append(labelKeys, "address")
labelValues = append(labelValues, ep.Addresses[0])

for _, zone := range ep.Hints.ForZones {
    m = append(m, &metric.Metric{
        LabelKeys:   append(labelKeys, "for_zone"),
        LabelValues: append(labelValues, zone.Name),
        Value:       1,
    })
}
```

Every iteration appends to the *same* base slice. That is correct today only by accident of allocation: the base is built with a single `append` onto a nil slice, which yields `cap == len == 1`, so each `append` in the loop is forced to reallocate. Give the base any spare capacity — a second base label, or someone switching to `make([]string, 0, n)` — and all zones share one backing array and report the last value:

```
for_zone values: got [zone-c zone-c zone-c], want [zone-a zone-b zone-c]
```

That is the output of the new test against the same code with a `make([]string, 0, 4)` base, so the failure mode is one edit away rather than hypothetical.

Building the two slices per zone removes the hazard and is shorter than the append dance it replaces.

**Also in here:** the same block indexed `ep.Addresses[0]` on the strength of a comment quoting the API docs ("This must contain at least one address"). Validation does enforce that today, so it is not reachable — but it is one relaxation away from a panic on the reflector goroutine, which nothing recovers, so it now skips such an endpoint.

Two tests: one asserting each zone keeps its own value, one asserting an endpoint with no addresses is skipped rather than panicking. Existing `kube_endpointslice_endpoints_hints` output is unchanged.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved EndpointSlice topology-hint metrics by skipping endpoints without addresses or routing hints.
  * Ensured zone-specific metrics retain the correct labels and avoid incorrect data sharing.
  * Prevented errors when processing endpoints that have no addresses.

### DIFF
--- a/internal/store/endpointslice.go
+++ b/internal/store/endpointslice.go
@@ -102,26 +102,32 @@ func createEndpointsSliceHints() generator.FamilyGenerator {
 			for _, ep := range e.Endpoints {
 				// Hint is populated when the endpoint is configured to be zone aware and preferentially route requests to its local zone.
 				// If there is no hint, skip this metric
-				if ep.Hints != nil && len(ep.Hints.ForZones) > 0 {
-					var (
-						labelKeys,
-						labelValues []string
-					)
+				if ep.Hints == nil || len(ep.Hints.ForZones) == 0 {
+					continue
+				}
 
-					// Per Docs.
-					// This must contain at least one address but no more than
-					// 100. These are all assumed to be fungible and clients may choose to only
-					// use the first element. Refer to: https://issue.k8s.io/106267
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

Thanks for the fix!

PR #3077, opened 2026-08-10
fix(customresourcestate): do not write GVK labels into the shared config

mrueg:

`compile` takes `Resource` by value, but `CommonLabels` is a map, so the copy shares it with the configured resource and with every other copy made from it — and wildcard resolution produces one copy per discovered GVK:

```go
if resource.CommonLabels == nil {
	resource.CommonLabels = map[string]string{}
}
resource.CommonLabels[customResourceState+"_group"] = resource.GroupVersionKind.Group
resource.CommonLabels[customResourceState+"_version"] = resource.GroupVersionKind.Version
resource.CommonLabels[customResourceState+"_kind"] = resource.GroupVersionKind.Kind
```

Those three writes go into a map every copy points at, and into the user's configuration. Compiling `myteam.io/v1 Foo` and then `myteam.io/v2 Bar` from one configured resource leaves the caller's map as:

```
expected: map[string]string{"team":"myteam"}
actual:   map[string]string{"customresource_group":"myteam.io",
                            "customresource_kind":"Bar",
                            "customresource_version":"v2",
                            "team":"myteam"}
```

**To be clear about severity:** the emitted metrics are correct today. Each `compile` overwrites all three keys before `Merge` copies them into the family, and factories are built one at a time, so no family ends up with another GVK's labels. This is a hazard rather than a live bug — but it is one concurrent or partially-overwriting compile away from being one, and mutating configuration the caller owns is wrong regardless.

Building a new map removes the sharing. The test asserts both that each resolution keeps its own GVK labels and that the configured map is left as the user wrote it; the second assertion fails on `main` with the output above.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where resources resolved from wildcards could share or overwrite labels.
  * Common labels now remain unchanged when resource-specific labels are added.
  * Resources with different group, version, and kind values retain independent labels, improving label accuracy and consistency.

* **Tests**
  * Added regression coverage for label isolation and preservation of configured common labels.

### DIFF
--- a/pkg/customresourcestate/registry_factory_test.go
+++ b/pkg/customresourcestate/registry_factory_test.go
@@ -594,3 +594,37 @@ func mustCompilePath(t *testing.T, path ...string) valuePath {
 	}
 	return out
 }
+
+// Wildcard resolution produces one Resource copy per discovered GVK, all sharing
+// the CommonLabels map of the configured resource. Writing the GVK labels into
+// that shared map would have each copy overwrite the last one's values, and
+// would leak them back into the caller's configuration.
+func Test_compile_doesNotShareCommonLabels(t *testing.T) {
+	configured := map[string]string{"team": "myteam"}
+	base := Resource{
+		Labels: Labels{CommonLabels: configured},
+		Metrics: []Generator{{
+			Name: "foo",
+			Each: Metric{Type: metric.Gauge, Gauge: &MetricGauge{MetricMeta: MetricMeta{Path: []string{"status"}}}},
+		}},
+	}
+
+	v1 := base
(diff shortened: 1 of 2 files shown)


pkg/customresourcestate/registry_factory_test.go
```diff
@@ -594,3 +594,37 @@ func mustCompilePath(t *testing.T, path ...string) valuePath {
 	}
 	return out
 }
+
+// Wildcard resolution produces one Resource copy per discovered GVK, all sharing
```
bhope: We probably may not need them in the tests. WDYT?

bhope (commented):
/lgtm
/hold

Feel free to unhold.

PR #3078, opened 2026-08-10
fix(discovery): check the CRD fields instead of asserting them

mrueg:

`extractGVKPs` reads the CRD spec with bare type assertions:

```go
objSpec := u.Object["spec"].(map[string]interface{})
g := objSpec["group"].(string)
k := objSpec["names"].(map[string]interface{})["kind"].(string)
p := objSpec["names"].(map[string]interface{})["plural"].(string)
for _, version := range objSpec["versions"].([]interface{}) {
	v := version.(map[string]interface{})["name"].(string)
```

Seven assertions, any of which panics on an object shaped differently than expected. Every field is required on `apiextensions.k8s.io/v1`, so a CRD the API server accepted does have them — I could not construct a realistic response that trips this, and I am **not** claiming a reachable bug.

The reason to fix it anyway is the blast radius. `extractGVKPs` runs from the CRD informer's event handlers, on the informer goroutine, where nothing recovers: a failed assertion is not a skipped object, it is the exporter exiting. The cost of being wrong about the shape of one object should be that object being skipped, not the process dying — the same reasoning that already applies to the `DeletedFinalStateUnknown` and `*unstructured.Unstructured` checks immediately above these lines.

Each field is now checked, with the failure logged and scoped: an unusable spec skips the whole CRD, a malformed single version skips just that version and keeps the rest.

Six new cases in `TestExtractGVKPs` cover a missing spec, a non-object spec, missing group, names and versions, and a versions list mixing a non-object, a nameless entry and a valid one. They panic on `main` (`interface conversion: interface {} is nil, not map[string]interface {}`) and pass here.

## Summary by CodeRabbit

- **Bug Fixes**
  - Improved handling of malformed or incomplete custom resource definitions.
  - Invalid definitions and version entries are now safely skipped instead of causing failures.
  - Valid versions continue to be processed as expected.

- **Tests**
  - Added coverage for missing fields, incorrect data types, and malformed version entries.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -95,12 +95,22 @@ func extractGVKPs(obj interface{}) []groupVersionKindPlural {
 			klog.ErrorS(nil, "CRD version has no name", "crd", u.GetName())
 			continue
 		}
-		// Versions that are not served by the API server cannot be listed or watched,
-		// so any reflector started for them would fail indefinitely. `served` is a
-		// required field on apiextensions.k8s.io/v1, treat it as served if absent.
-		if served, ok := versionSpec["served"].(bool); ok && !served {
-			klog.V(1).InfoS("skipping CRD version that is not served by the API server", "group", g, "version", v, "kind", k)
-			continue
+		// Versions that are not served by the API server cannot be listed or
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

PR #3079, opened 2026-08-10
fix(proc): start the reaper only once

mrueg:

`StartReaper` spawns a goroutine that runs for the lifetime of the process and has no stop path:

```go
func StartReaper() {
	if os.Getpid() == 1 {
		go func() {
			sigs := make(chan os.Signal, 1)
			signal.Notify(sigs, syscall.SIGCHLD)
			for { ... }        // never returns
		}()
	}
}
```

It is called from `RunKubeStateMetrics` (`pkg/app/server.go:275`), and `internal/wrapper.go` restarts `RunKubeStateMetrics` on **every config reload**. So each reload leaves another reaper behind, blocked on its own signal channel — which the `signal` package holds a reference to, so it is never collected. Reproduced with the same goroutine body:

```
after 5 calls: goroutines grew by 5
```

It only applies when the exporter runs as pid 1, which is the ordinary case for a container without an init system. And duplicate reapers are otherwise harmless — `Wait4` uses `WNOHANG`, so they race to reap and the losers simply find no children — which is exactly why this leaks quietly instead of breaking something visible. The cost is a goroutine and a channel per reload, so it matters for deployments whose configuration changes regularly.

Guarded with a `sync.Once`. The pid lookup and the goroutine launch are indirected through package variables so the guard can be tested without the test binary having to run as pid 1 or registering a real `SIGCHLD` handler in the test process — the latter matters, because a live handler calling `Wait4(-1, ...)` in a test binary could reap children belonging to other tests.

Two tests: repeated calls launch exactly one reaper (5 launches on `main`, 1 here), and nothing is launched when the process is not pid 1. Both are `//go:build linux`, matching the file they cover.

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved child-process cleanup when running as the system’s initial process.
  * Prevented duplicate reaper processes, signal handlers, and background tasks when initialization is requested repeatedly.
  * Ensured child-process reaping continues reliably without blocking.

* **Tests**
  * Added coverage for initialization behavior under different process IDs.
  * Added verification that repeated initialization starts only one reaper.

### DIFF
--- a/pkg/util/proc/reaper.go
+++ b/pkg/util/proc/reaper.go
@@ -21,33 +21,54 @@ package proc
 import (
 	"os"
 	"os/signal"
+	"sync"
 	"syscall"
 
 	"k8s.io/klog/v2"
 )
 
+var (
+	reaperOnce = new(sync.Once)
+
+	// Indirected so the guard can be tested without the test binary running as
+	// pid 1 or registering a real SIGCHLD handler.
+	getpid = os.Getpid
+	launch = func() { go reapChildren() }
+)
+
 // StartReaper starts a goroutine to reap processes if called from a process
 // that has pid 1.
+//
+// It is safe to call more than once: the reaper runs for the lifetime of the
+// process and has no stop path, and RunKubeStateMetrics -- which calls this --
+// is restarted by the config file watchers on every reload. Without the guard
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

PR #3080, opened 2026-08-10
fix(util): rebuild the clients from the current kubeconfig

mrueg:

`CreateKubeClient` returns a memoized client and ignores its arguments entirely:

```go
func CreateKubeClient(apiserver string, kubeconfig string) (clientset.Interface, error) {
	if currentKubeClient != nil {
		return currentKubeClient, nil
	}
	...
	if config == nil {
		config, err = clientcmd.BuildConfigFromFlags(apiserver, kubeconfig)
```

and the `*rest.Config` behind it is built once into a package variable that `CreateCustomResourceClients` and `CreateDiscoveryClient` then reuse.

`internal/wrapper.go` watches the kubeconfig and restarts KSM when it changes, precisely so that new credentials take effect. But the restart calls `CreateKubeClient` again and gets the client built from the **superseded** file, while `pkg/app/server.go:186` builds its own fresh `rest.Config` for the CRD discoverer and the auth filter. So after a certificate or token rotation the two halves of the process are running on different credentials, and the store client keeps failing until the pod is actually restarted — which is the very thing the watcher exists to avoid.

`tests/e2e/hot-reload-kubeconfig_test.go` only asserts that port 8080 comes back up, so it does not catch this.

**Why not just key the cache on the arguments:** a rotation rewrites the kubeconfig and leaves its path alone, so `(apiserver, kubeconfig)` is unchanged across exactly the event we need to react to. Any cache keyed on the arguments would still hand back the old credentials. The memoization has to go.

Each call now builds from the file as it stands. The shared settings move into a `buildConfig` helper, so:

- the custom resource and discovery clients keep getting the user agent and content types that previously only `CreateKubeClient` applied — today they inherit them by side effect, and only if `CreateKubeClient` happened to run first;
- `CreateDiscoveryClient` and `CreateCustomResourceClients` stop depending on call ordering for their configuration.

The cost is re-reading the kubeconfig per call. `CreateKubeClient` runs once per KSM run; `CreateCustomResourceClients` runs per discovery rebuild, where it is negligible against the store reconstruction that follows.

### Testing

`pkg/util` had no test file. Two added, both failing on `main`:

- `TestBuildConfigReflectsCurrentKubeconfig` — rewrites the same path and asserts the second config carries the new host and token. On `main`: `second host: got "https://first.example.com", want "https://second.example.com"`.
- `TestCreateKubeClientRebuildsAfterRotation` — points the kubeconfig at one `httptest` server, rewrites it to a second, and asserts each server sees exactly one `/version` request. On `main` the second sees none.

## Summary by CodeRabbit

- **Bug Fixes**
  - Kubernetes and discovery connections now use the latest configuration whenever they are created.
  - Updated kubeconfig files, including rotated credentials and changed endpoints, are recognized automatically.
  - Improved reliability when configuration changes while the application is running.
  - Reduced the risk of stale connection details being used after configuration updates.
  - Added coverage for configuration refresh scenarios to help ensure consistent behavior.

### DIFF
--- a/pkg/util/utils.go
+++ b/pkg/util/utils.go
@@ -23,7 +23,6 @@ import (
 
 	"k8s.io/apimachinery/pkg/api/meta"
 	"k8s.io/apimachinery/pkg/runtime/schema"
-	"k8s.io/client-go/discovery"
 	clientset "k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/rest"
 	"k8s.io/client-go/tools/clientcmd"
@@ -105,15 +104,6 @@ func CreateCustomResourceClients(apiserver string, kubeconfig string, factories 
 		customResourceClients[gvrString] = customResourceClient
 	}
 	return customResourceClients, nil
-}
-
-// CreateDiscoveryClient creates a Kubernetes discovery client.
-func CreateDiscoveryClient(apiserver string, kubeconfig string) (*discovery.DiscoveryClient, error) {
-	config, err := buildConfig(apiserver, kubeconfig)
-	if err != nil {
-		return nil, err
-	}
-	return discovery.NewDiscoveryClientForConfig(config)
 }
 
 // GVRFromType returns the GroupVersionResource for a given type.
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

Thanks for the fix!

PR #3083, opened 2026-08-11
fix(deployment): do not panic on an unset rollingUpdate value

mrueg:
Both rolling-update generators guard the outer pointers and then re-raise any
error from the conversion as a panic:

```go
if d.Spec.Strategy.RollingUpdate == nil || d.Spec.Replicas == nil {
	return &metric.Family{}
}

maxUnavailable, err := intstr.GetScaledValueFromIntOrPercent(
	d.Spec.Strategy.RollingUpdate.MaxUnavailable, int(*d.Spec.Replicas), false)
if err != nil {
	panic(err)
}
```

The guard does not cover what the helper actually rejects. It returns
`errors.New("nil value for IntOrString")` when the *inner* `MaxUnavailable` /
`MaxSurge` pointer is nil, and a parse error when a percentage string is
malformed. Neither is implied by `RollingUpdate != nil`:

```yaml
spec:
  strategy:
    type: RollingUpdate
    rollingUpdate:
      maxUnavailable: null
```

Defaulting fills these in for objects that go through the API server, but
nothing in kube-state-metrics depends on that, and the blast radius is the
whole process rather than the one Deployment: a panic on the reflector
goroutine takes down every metric for every resource.

Return an empty family instead, which is what the same generators already do
when the outer pointers are nil, and what `kube_cronjob_next_schedule_time`
does for an unparseable schedule.

Test added: a Deployment with `RollingUpdate: &v1.RollingUpdateDeployment{}`
panics with `nil value for IntOrString` before this change.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Prevented deployment metrics collection from crashing when rolling-update settings are missing, nil, malformed, or invalid.
  * Deployment replica metrics continue to be reported when rolling-update values cannot be converted.
  * Rolling-update metric families remain available even when they contain no sample values.

* **Tests**
  * Added coverage for unset, non-percentage, and unparsable rolling-update limits.

### DIFF
--- a/internal/store/deployment.go
+++ b/internal/store/deployment.go
@@ -311,9 +311,13 @@ func deploymentMetricFamilies(allowAnnotationsList, allowLabelsList []string) []
 					return &metric.Family{}
 				}
 
+				// GetScaledValueFromIntOrPercent also fails on a nil
+				// MaxUnavailable and on a percentage that does not parse, neither
+				// of which the guard above covers. A single such Deployment must
+				// not crash the exporter, so skip its metric instead.
 				maxUnavailable, err := intstr.GetScaledValueFromIntOrPercent(d.Spec.Strategy.RollingUpdate.MaxUnavailable, int(*d.Spec.Replicas), false)
 				if err != nil {
-					panic(err)
+					return &metric.Family{}
 				}
 
 				return &metric.Family{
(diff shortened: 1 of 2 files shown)


bhope (commented):
/lgtm

Thanks for your PR!

PR #3085, opened 2026-08-11
fix(watch): make --object-limit bound the total, not the page

mrueg:
`List` sets the limit as a request page size and truncates the page it gets
back, but leaves the response's continue token alone:

```go
if i.limit > 0 {
	options.Limit = i.limit
	...
}
res, err := i.lw.List(options)
...
if int64(len(list)) > i.limit {
	meta.SetList(res, list[0:i.limit])       // <- Continue still set
}
```

`cache.Reflector` does not call `List` once. It lists through
`pager.ListPager`, which re-invokes `List` with the previous response's
`Continue` token and concatenates the pages, stopping only when the token is
empty. Truncating a page therefore caps nothing -- it just makes the pager ask
for the next one, and the store still ends up holding every object.

The flag documents itself as "The total number of objects to list per
resource" (docs/developer/cli-arguments.md), so on a 50k-pod cluster
`--object-limit=100` is expected to hold 100 pods and instead holds 50000.

Paging is on for exactly the lists that matter here: the reflector disables it
only when a specific `resourceVersion` is requested, so both the initial list
and every relist after a `410 Gone` page.

Clear `Continue` (and `RemainingItemCount`, which would then describe a list
that no longer exists) once the response has been truncated. A short page that
still carries a token also terminates the list -- overshooting the cap is the
one thing this flag exists to prevent.

Also:
  - `meta.ExtractList` ran on every list even with no limit configured,
    allocating a `[]runtime.Object` over the whole collection for nothing. It
    is now only reached when a limit is set.
  - the error from `meta.SetList` was discarded.
  - `kube_state_metrics_list_objects` now reports the number of objects
    actually returned rather than the limit.

Adds pkg/watch/watch_test.go -- the package had no tests.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Enforced configured limits on listed resources and prevented unnecessary pagination.
  * Correctly removed continuation and remaining-item metadata when results are truncated.
  * Improved handling of list responses, errors, request limits, and API-server caching.
  * Ensured metrics reflect the actual number of returned objects.

* **Tests**
  * Added coverage for list limits, pagination behavior, caching, and response metadata.

### DIFF
--- a/pkg/watch/watch_test.go
+++ b/pkg/watch/watch_test.go
@@ -1,0 +1,188 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
+
+Licensed under the Apache License, Version 2.0 (the "License");
+you may not use this file except in compliance with the License.
+You may obtain a copy of the License at
+
+    http://www.apache.org/licenses/LICENSE-2.0
+
+Unless required by applicable law or agreed to in writing, software
+distributed under the License is distributed on an "AS IS" BASIS,
+WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+See the License for the specific language governing permissions and
+limitations under the License.
+*/
+
+package watch
+
+import (
+	"fmt"
+	"testing"
+
+	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
+	"k8s.io/apimachinery/pkg/api/meta"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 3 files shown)


pkg/watch/watch_test.go
```diff
@@ -0,0 +1,188 @@
+/*
+Copyright 2026 The Kubernetes Authors All rights reserved.
```
bhope, line 2: Should we match k/k pattern here where all new test files do not mention the year?

mrueg: we could do that in a follow up if we want for all test files, if needed.

PR #3087, opened 2026-08-11
fix(options): validate the options before starting

mrueg:
`main` calls `Validate` after `Parse` returns:

```go
if err := opts.Parse(); err != nil {
	klog.FlushAndExit(klog.ExitFlushTimeout, 1)
}
if err := opts.Validate(); err != nil {      // <- unreachable
	klog.ErrorS(err, "Validating options error")
	klog.FlushAndExit(klog.ExitFlushTimeout, 1)
}
```

`Parse` is `o.cmd.Execute()`, which runs `cmd.Run`, which is
`internal.RunKubeStateMetricsWrapper`. That function ends in `select {}` and
never returns, so `Parse` only returns when the process is already shutting
down — and on the error path it exits before reaching the next statement
either way. `Validate` has therefore never run in any real invocation.

What that lets through today:

    --node=n --resources=deployments   accepted, then silently collects
                                       nothing (the field selector does not
                                       apply to deployments)

and, once the early `return nil` in `Validate` is removed (#3062):

    --auto-gomemlimit-ratio=7          GOMEMLIMIT set to 7x available memory
    --object-limit=-1                  accepted

Move the call into the command's `PreRunE`, which cobra runs after flag
parsing and before `Run`, and which propagates its error out through
`Execute`. A rejected option now aborts before any client, informer or
listener is created.

`SilenceUsage` is set on the root command so a one-line validation failure
is not buried under the full flag usage dump.

Note this is orthogonal to #3062: that PR fixes `Validate` returning early
when `--node` is empty, this one makes `Validate` run at all. The test uses
the `--node` check because it is the only one reachable before #3062 lands.

---

## Summary by CodeRabbit

* **Bug Fixes**
  * Invalid command-line options are now rejected before execution begins.
  * Validation errors no longer trigger full command usage output.
  * Parsing failures continue to be logged and terminate the process appropriately.

* **Tests**
  * Added coverage confirming invalid options prevent command execution while valid options proceed normally.

### DIFF
--- a/main.go
+++ b/main.go
@@ -31,11 +31,11 @@ func main() {
 		internal.RunKubeStateMetricsWrapper(opts)
 	}
 	opts.AddFlags(cmd)
+	// Parse blocks for the lifetime of the process: Execute runs cmd.Run, which
+	// only returns once kube-state-metrics is shutting down. Option validation
+	// therefore happens in the command's PreRunE, not here.
 	if err := opts.Parse(); err != nil {
-		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
-	}
-	if err := opts.Validate(); err != nil {
-		klog.ErrorS(err, "Validating options error")
+		klog.ErrorS(err, "Running kube-state-metrics failed")
 		klog.FlushAndExit(klog.ExitFlushTimeout, 1)
 	}
 }
--- a/pkg/options/autoload.go
+++ b/pkg/options/autoload.go
@@ -140,4 +140,7 @@ var InitCommand = &cobra.Command{
 	Short: "Add-on agent to generate and expose cluster-level metrics.",
 	Long:  "kube-state-metrics is a simple service that listens to the Kubernetes API server and generates metrics about the state of the objects.",
 	Args:  cobra.NoArgs,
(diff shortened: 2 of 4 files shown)


mrueg:
Correct, and deliberately out of scope here — that early `return nil` is what #3062 removes:

```go
 func (o *Options) Validate() error {
 	shardableResource := "pods"
-	if o.Node == "" {
-		return nil
-	}
+	if o.Node != "" {
+		...
```

The two PRs fix different halves of the same problem and neither is sufficient alone:

- **#3062** makes `Validate` check the right things — but `Validate` is never called, so it changes nothing at runtime.
- **This PR** makes `Validate` run — but it still returns early without `--node`, so `--auto-gomemlimit-ratio` and `--object-limit` are still unchecked.

Only with both does `--auto-gomemlimit-ratio=7` actually get rejected. They are textually independent (this one touches `main.go` / `AddFlags` / `autoload.go`, #3062 touches the body of `Validate`), so they can merge in either order.

The regression test here uses the `--node` resource check precisely because it is the only branch reachable before #3062 lands. #3062 already carries the ratio, object-limit, NaN and non-node-resource cases.

PR #3088, opened 2026-08-12
feat: add kube_pod_status_disruption_reason metric

### LINKED ISSUE #2774: [Feature Request] Support detailed disruption reason detection via new metric kube_pod_status_disruption_reason

bhope:

`kube_pod_status_reason` only matches `pod.status.reason`, condition reasons, and terminated container reasons against a fixed list. Pods evicted through the Eviction API (`kubectl drain`, PDB-aware evictions, autoscalers like Karpenter) set a `DisruptionTarget` condition with reason `EvictionByEvictionAPI` instead, which is not in that list, so the metric stays 0 for the most common disruption path.

Adds `kube_pod_status_disruption_reason`, scoped to the `DisruptionTarget` condition, covering `PreemptionByScheduler`, `DeletionByTaintManager`, `EvictionByEvictionAPI`, `DeletionByPodGC`, and `TerminationByKubelet`. The existing `kube_pod_status_reason` is unchanged for backward compatibility.

Unlike `kube_pod_status_reason`, this metric only emits a series when a reason is actually set, instead of one row per known reason with 0/1 values. Most pods are never disrupted, so the dense form would waste series for no benefit. The tradeoff: `reason == 0` queries and joins against other per-pod metrics do not work for non-disrupted pods, since no series exists for them.

**How does this change affect the cardinality of KSM:** increases, by at most 1 series per disrupted pod, 0 for pods with no disruption.

Fixes #2774

## Summary by CodeRabbit

* **New Features**
  * Added an experimental pod disruption reason metric.
  * Reports `DisruptionTarget` condition reasons with pod, namespace, reason, and UID labels.
  * Maps unrecognized reasons to `Other`; missing series do not indicate a false reason.
  * Remains separate from the existing pod status reason metric.

* **Documentation**
  * Documented the metric, labels, supported conditions, and interpretation of missing series.

* **Tests**
  * Added coverage for recognized, missing, unknown, and unrelated disruption conditions.


internal/store/pod.go
```diff
@@ -37,9 +38,10 @@ import (
 )
 
 var (
-	descPodLabelsDefaultLabels = []string{"namespace", "pod", "uid"}
-	podStatusReasons           = []string{"Evicted", "NodeAffinity", "NodeLost", "PreemptionByScheduler", "SchedulingGated", "Shutdown", "TerminationByKubelet", "UnexpectedAdmissionError"}
-	descPodIPsLabelKeys        = []string{"ip", "ip_family"}
+	descPodLabelsDefaultLabels    = []string{"namespace", "pod", "uid"}
+	podStatusReasons              = []string{"Evicted", "NodeAffinity", "NodeLost", "PreemptionByScheduler", "SchedulingGated", "Shutdown", "TerminationByKubelet", "UnexpectedAdmissionError"}
+	podDisruptionConditionReasons = []string{"PreemptionByScheduler", "DeletionByTaintManager", "EvictionByEvictionAPI", "DeletionByPodGC", "TerminationByKubelet"}
```
mrueg: k8s.io/api/core/v1 exports two of the five:

v1.PodReasonPreemptionByScheduler  // "PreemptionByScheduler"
v1.PodReasonTerminationByKubelet   // "TerminationByKubelet"

can we use these?

internal/store/pod.go
```diff
@@ -1736,6 +1739,34 @@ func getPodStatusReasonValue(p *v1.Pod, reason string) float64 {
...
+		"kube_pod_status_disruption_reason",
+		"The pod disruption condition reason",
+		metric.Gauge,
+		basemetrics.ALPHA,
+		"",
+		wrapPodFunc(func(p *v1.Pod) *metric.Family {
+			ms := []*metric.Metric{}
+
+			for _, cond := range p.Status.Conditions {
+				if cond.Type != v1.DisruptionTarget || !slices.Contains(podDisruptionConditionReasons, cond.Reason) {
```
mrueg: I wonder if it makes sense to set the reason to "Unknown", "Untracked in Kube-State-Metrics", "Other" or similar?

bhope: @mrueg That's a good suggestion. However, we follow similar (silent drop) pattern in `kube_pod_status_reason` metric. Also a place-holder-unknown may not provide much information to the user, but still produce additional metric per pod. That's why I am leaning toward the silent drop mechanism. What do you think?

internal/store/pod.go
bhope: Good catch. Fixed!

internal/store/pod.go
mrueg: "Other" would provide a signal that it has detected something that is not allowlisted and made the user wonder what that is. I'm not sure if we otherwise want to actively figure out new status reasons ourselves all the time?

bhope: Makes sense. I will add it as "Other" for now. We will retrofit this into other metric to follow the suit separately. Thanks for the suggestion.

bhope: Updated.

mrueg:
/lgtm
thanks for your contribution!
/hold

feel free to unhold

PR #3089, opened 2026-08-12
feat: reduce kube_pod_status_reason cardinality

### LINKED ISSUE #3004: kube_pod_status_reason: reduce cardinality by not enumerating all known reasons per pod
Stop enumerating every known reason per pod for metric `kube_pod_status_reason`. Only emit a row when the pod actually has a reason set:

`kube_pod_status_reason{reason="OutOfcpu"} 1   // emitted only when true`

This decouples the metric's cardinality from how many reasons kube-state-metrics knows about and removes the permanent all-zero rows for every pod that has never hit any of these reasons.

**Why is this needed:**
`kube_pod_status_reason` currently emits one time series per pod for every entry in the hardcoded `podStatusReasons` list (internal/store/pod.go), with value 0 or 1 depending on whether that reason actually applies to the pod.

**Additional context:**
This is a breaking change to the metric shape for any consumer relying on the explicit zero rows. May need a deprecation/migration note in CHANGELOG and docs.

bhope:

`kube_pod_status_reason` emitted one row per entry in the hardcoded `podStatusReasons` list for every pod, with value 0 or 1, regardless of how many reasons kube-state-metrics knows about or whether the pod ever hit any of them.

Emits a row only for the reason that is actually set. This decouples cardinality from the size of the known-reasons list and removes the permanent all-zero rows for pods that never hit any of them.

This is a breaking change for consumers relying on the explicit zero rows: `reason == 0` no longer matches, since no series exists for a false reason. The two documented example queries relying on this (non-standard "Terminating" pod state and the `PodsBlockedInTerminatingState` alert) are updated to use `unless` instead of `== 0`.

**How does this change affect the cardinality of KSM:** decreases, from one row per known reason per pod (currently 8) to at most one row per pod, 0 for pods with no reason set.

Fixes #3004

## Bug Fixes

* Pod status-reason metrics now report only the currently active reason.
* Unknown non-empty reasons are grouped under `Other`.
* Unrelated condition and container-termination reasons no longer produce status-reason metrics.
* Improved detection and alerting for Pods in the `Terminating` state.

## Documentation

* Clarified that missing reason series do not indicate false values.
* Updated PromQL examples to reflect the revised metric behavior.


internal/store/pod.go
```diff
@@ -1704,12 +1704,14 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
...
+					continue
+				}
+				ms = append(ms, &metric.Metric{
 					LabelKeys:   []string{"reason"},
 					LabelValues: []string{reason},
-					Value:       getPodStatusReasonValue(p, reason),
-				}
-				ms = append(ms, m)
+					Value:       1,
+				})
```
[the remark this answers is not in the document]
bhope: This is  metric's pre-existing design, not introduced by this PR.

internal/store/pod.go
```diff
@@ -1704,12 +1704,14 @@ func createPodStatusReasonFamilyGenerator() generator.FamilyGenerator {
 			ms := []*metric.Metric{}
 
 			for _, reason := range podStatusReasons {
-				m := &metric.Metric{
+				if getPodStatusReasonValue(p, reason) != 1 {
```
mrueg: do we want to set this to "Other" as well, to match behaviour?

bhope: Unlike `disruption_reason`, this one also scans every condition + container-terminated reason, most of which are normal. So "Other" would fire on nearly every pod. Could scope it to just `p.Status.Reason` instead, which is much rarer. Will update with this fix.

bhope: Updated with the narrowed scope, let me know what you think. Thanks.

mrueg: Should we turn the getPodStatusReasonValue into `hasPodStatusReason(p, reason) bool` since it's not used anywhere else?

CatherineF-dev:
This change can be highlighted in https://github.com/kubernetes/kube-state-metrics/blob/main/CHANGELOG.md in next release.

/lgtm
/hold
 feel free to unhold

internal/store/pod.go
bhope: Thats a good suggestion - updated to return bool. Thanks

bhope:
Hi @CatherineF-dev @mrueg Can I get another look? Addressed the comments. Thanks.

PR #3091, opened 2026-08-13
feat: Add metric for configuration errors (`kube_state_metrics_config_errors_total`)

### LINKED ISSUE #2472: Add metric(s) for KSM errors
Seconding Damien, I'm not for panicking the binary since we expect a [reload](https://github.com/kubernetes/kube-state-metrics/blob/f8aa7d9bb9d8e29876e19f4859391a54a7e61d63/internal/wrapper.go#L61-L62) on configuration change, and then a log (or metric) indicating an issue with the given configuration. Crashing the binary goes against the controller-esque behavior and allows anyone (with access) to crash a running instance simply by supplying a bad config.

> That said I can definitely see a pain point here and maybe we should add a new metrics to report any kind of misconfiguration so that it would be more visible to cluster admin when kube-state-metrics is misconfigured.

+1 all the way. I'll add one.

_Originally posted by @rexagod in https://github.com/kubernetes/kube-state-metrics/issues/2434#issuecomment-2285671557_

Mujib-Ahasan:
This adds `kube_state_metrics_config_errors_total` to track errors encountered while processing kube-state-metrics configuration. The metric distinguishes between regular configuration and Custom Resource State configuration.

Fixes #2472 

## Summary by CodeRabbit

* **New Features**
  * Added metrics for configuration processing errors.
  * Added separate tracking for general configuration and custom resource configuration failures.
  * Metrics now record failures during configuration loading, creation, and refresh.

### DIFF
--- a/pkg/app/server.go
+++ b/pkg/app/server.go
@@ -55,6 +55,7 @@ import (
 	"k8s.io/kube-state-metrics/v2/internal/discovery"
 	"k8s.io/kube-state-metrics/v2/internal/store"
 	"k8s.io/kube-state-metrics/v2/pkg/allowdenylist"
+	"k8s.io/kube-state-metrics/v2/pkg/customresource"
 	"k8s.io/kube-state-metrics/v2/pkg/customresourcestate"
 	generator "k8s.io/kube-state-metrics/v2/pkg/metric_generator"
 	"k8s.io/kube-state-metrics/v2/pkg/metricshandler"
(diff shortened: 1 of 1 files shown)


Mujib-Ahasan:
There is a problem I noticed (from copilot comments) from handling intial configuration error is that we currently hit those error path before the telemetry server start so even if we Inc() `kube_state_metrics_config_errors_total` the metric isn't actually scrapeable. and I think this is also true for `kube_state_metrics_last_config_reload_successful`. Any thoughts?? 

cc @CatherineF-dev @mrueg

PR #3093, opened 2026-08-14
fix(discovery): avoid no-op CRD updates forcing writer rebuilds

harishkashyap569:
## Summary

Second PR in a two-part split (see #3055 for store-sync-before-swap). **Merge #3055 first**, then merge this PR.

- Serialize `storeBuilder` updates through `MetricsHandler.ConfigureStore` so discovery cannot race async rebuilds
- Skip no-op CRD add/update/delete events that do not change the discoverable GVK/plural set
- Track `cacheRevision` so delete/re-add cycles with the same GVR still trigger rebuilds
- Preserve `WasUpdated` on factory/client/`WithEnabledResources` errors for retry on the next tick
- Run all writer rebuilds asynchronously (including the first CRS build) so discovery reconciliation does not block in `WaitForStoresSync`

Until #3055 lands, this PR's diff includes both changes. After #3055 merges, the diff should shrink to the two discovery commits below. A rebase onto `main` is only needed if GitHub shows a stale/duplicate diff (e.g. after a squash merge) or if there are merge conflicts.

Review the discovery commits:
- `fix(discovery): avoid no-op CRD updates forcing writer rebuilds`
- `fix(metricshandler): run all writer rebuilds asynchronously`

## Test plan

- [x] `go test ./internal/discovery/... ./pkg/metricshandler/...`
- [ ] Merge after #3055
- [ ] Internal cluster validation of full stack

## Summary by CodeRabbit

- **New Features**
  - Added configurable store synchronization timeout through `--store-sync-timeout` and YAML settings, defaulting to two minutes.
  - Metrics rebuilds now wait for stores to synchronize before switching writers, with retries and fallback to the previous working metrics.
  - Readiness checks remain unavailable until metrics initialization is complete.
  - Added support for updating enabled custom resources without affecting built-in resources.

- **Bug Fixes**
  - Prevented unnecessary rebuilds when resource discovery has not changed.
  - Improved handling of discovery updates, removals, and synchronization failures.

### DIFF
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -344,14 +344,20 @@ func (r *CRDiscoverer) PollForCacheUpdates(
 		}
 		// Apply builder config under the same lock Build() uses, then rebuild.
 		// Mutating the shared builder directly would race async rebuilds.
+		var configErr error
 		m.ConfigureStore(ctx, func(b ksmtypes.BuilderInterface) {
 			b.WithCustomResourceClients(discoveredCustomResourceClients)
 			b.WithCustomResourceStoreFactories(customFactories...)
 			if err := b.WithEnabledResources(enabledCustomResources); err != nil {
 				klog.ErrorS(err, "failed to update custom resource stores")
+				configErr = err
 			}
 			b.WithGenerateCustomResourceStoresFunc(b.DefaultGenerateCustomResourceStoresFunc())
 		})
+		if configErr != nil {
+			// Preserve WasUpdated so the next tick retries.
+			return false
+		}
 		lastAppliedEnabledKey = enabledKey
 		lastAppliedRevision = observedRevision
 		// Clear only the revision we applied. If discovery changed while clients
--- a/internal/discovery/discovery.go
+++ b/internal/discovery/discovery.go
@@ -345,9 +345,13 @@ func (r *CRDiscoverer) PollForCacheUpdates(
 		// Apply builder config under the same lock Build() uses, then rebuild.
 		// Mutating the shared builder directly would race async rebuilds.
 		if err := m.ConfigureStore(ctx, func(b ksmtypes.BuilderInterface) error {
+			replacer, ok := b.(ksmtypes.CustomResourceReplacer)
+			if !ok {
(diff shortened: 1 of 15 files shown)


harishkashyap569:
Depends on #3055 — please merge #3055 first, then I'll rebase this branch so the diff is discovery-only.

harishkashyap569:
Updated the PR description: merge #3055 first, then merge this PR. Rebase is only needed if the diff stays stale after #3055 lands (e.g. squash merge) or if GitHub reports conflicts.

internal/discovery/discovery.go
```diff
@@ -307,28 +323,54 @@ func (r *CRDiscoverer) PollForCacheUpdates(
...
+		// Apply builder config under the same lock Build() uses, then rebuild.
+		// Mutating the shared builder directly would race async rebuilds.
+		if err := m.ConfigureStore(ctx, func(b ksmtypes.BuilderInterface) error {
+			replacer, ok := b.(ksmtypes.CustomResourceReplacer)
+			if !ok {
+				return fmt.Errorf("builder does not support replacing custom resources")
+			}
+			b.WithCustomResourceClients(discoveredCustomResourceClients)
+			b.WithCustomResourceStoreFactories(customFactories...)
+			if err := replacer.ReplaceEnabledCustomResources(enabledCustomResources); err != nil {
```
bhope, line 354: `ReplaceEnabledCustomResources` now replaces `enabledResources` each tick (instead of appending), so a poll landing mid CRD update/delete race can now briefly drop a resource's metrics which is different from the old behavior. Is this intentional?

harishkashyap569: Yeah, that difference is intentional. `WithEnabledResources` only appends, so a CR that dropped out of discovery stayed in `enabledResources` forever. `ReplaceEnabledCustomResources` makes the CR set match what discovery currently sees (built-ins are left alone).

A tick that races a CRD delete/re-add can drop that CR from the next generation. We still serve the previous writers until the new ones sync, so `/metrics` shouldn't go blank mid-rebuild. After a successful swap that CR is gone until the next tick; `cacheRevision` is there so a delete/re-add with the same GVR still rebuilds rather than getting skipped as a no-op.

